Issue #8, opened 2019-05-30
Add the new kinds decided in Java


bogdandrutu:
See some discussion about this:
https://github.com/open-telemetry/opentelemetry-specification/issues/43

Here is where this is defined:
https://github.com/open-telemetry/opentelemetry-proto/blob/master/src/opentelemetry/proto/trace/v1/trace.proto#L107

PR #9, opened 2019-05-31
Add new types of SpanKind.

songy23:
Fixes https://github.com/open-telemetry/opentelemetry-proto/issues/8.

### DIFF
--- a/src/opentelemetry/proto/trace/v1/trace.proto
+++ b/src/opentelemetry/proto/trace/v1/trace.proto
@@ -115,6 +115,19 @@ message Span {
     // Indicates that the span covers the client-side wrapper around an RPC or
     // other remote request.
     CLIENT = 2;
+
+    // Indicates that the span is used internally. Default value.
+    INTERNAL = 3;
+
+    // Indicates that the span describes producer sending a message to a broker.
+    // Unlike client and  server, there is no direct critical path latency relationship
+    // between producer and consumer spans.
+    PRODUCER = 4;
+
+    // Indicates that the span describes consumer recieving a message from a broker.
+    // Unlike client and  server, there is no direct critical path latency relationship
+    // between producer and consumer spans.
+    CONSUMER = 5;
   }
 
   // Distinguishes between spans generated in a particular context. For example,
--- a/src/opentelemetry/proto/trace/v1/trace.proto
+++ b/src/opentelemetry/proto/trace/v1/trace.proto
@@ -106,6 +106,7 @@ message Span {
   // in addition to a parent/child relationship.
   enum SpanKind {
     // Unspecified. Do NOT use as default.
+    // Implementations MAY assume SpanKind to be INTERNAL when receiving UNSPECIFIED.
     SPAN_KIND_UNSPECIFIED = 0;
 
     // Indicates that the span is used internally. Default value.
--- a/src/opentelemetry/proto/trace/v1/trace.proto
+++ b/src/opentelemetry/proto/trace/v1/trace.proto
@@ -105,19 +105,19 @@ message Span {
   // Type of span. Can be used to specify additional relationships between spans
   // in addition to a parent/child relationship.
   enum SpanKind {
-    // Unspecified.
+    // Unspecified. Do NOT use as default.
     SPAN_KIND_UNSPECIFIED = 0;
+
+    // Indicates that the span is used internally. Default value.
+    INTERNAL = 1;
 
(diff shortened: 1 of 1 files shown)


src/opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -115,6 +115,19 @@ message Span {
     // Indicates that the span covers the client-side wrapper around an RPC or
     // other remote request.
     CLIENT = 2;
+
+    // Indicates that the span is used internally. Default value.
```
bogdandrutu: Can we respect the order from Java?

src/opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -115,6 +115,19 @@ message Span {
...
+    INTERNAL = 3;
```
SergeyKanzhelev: there is no more `SPAN_KIND_UNSPECIFIED `. It was replaced with `INTERNAL`

bogdandrutu: proto3 convention to have UNSPECIFIED for the 0 value. https://developers.google.com/protocol-buffers/docs/style#top_of_page look for enums.

SergeyKanzhelev: Semantic of unspecified is `INTERNAL`. So than get rid of explicit `INTERNAL`? Perhaps rename to `SPAN_KIND_INTERNAL_OR_UNSPECIFIED`?

src/opentelemetry/proto/trace/v1/trace.proto
songy23: Updated. I assume we haven't released proto so it's safe to just change the order.

src/opentelemetry/proto/trace/v1/trace.proto
songy23: For now I left it as-is. I think this is more of a syntax restriction of proto. Happy to change it if we have a different consensus here.

SergeyKanzhelev: It will be easier if enum in language will match exactly the enum values in proto. Thus I was suggesting to make `INTERNAL` and `UNSPECIFIED` the same value. Also it eliminates question on how to process `UNSPECIFIED`.

For instance, if service recieves `UNSPECIFIED` - should it preserve the value or free to switch to `INTERNAL`. If need to preserve - Go SDK will need to expose the `UNSPECIFIED` as an option for creating `SpanData`

songy23: SG, merged these two in https://github.com/open-telemetry/opentelemetry-proto/pull/9/commits/2e24a8bf3b023f3ea61eb63d7dc340633a85d8fd.

songy23:
PTAL.

src/opentelemetry/proto/trace/v1/trace.proto
bogdandrutu: I think UNSPECIFIED is proto3 specific because in proto3 they removed the notion of has value. I would suggest that we keep the UNSPECIFIED to be consistent with the proto3 recommendations.

About changing to INTERNAL, that is probably a valid behavior, alternative we can simply reject the request if UNSPECIFIED by saying that this field is required.

SergeyKanzhelev: I'm not a proto expert. Why the proposed change is bad? What's the reason to keep them separate if semantically we want internal to be exactly when `UNSPECIFIED` is used for? (In code it will be the same. If you forgot to tell us what you want it to be - it will be `INTERNAL`)

bogdandrutu: If we ever want to distinguish between not set and default value this will be impossible with the current approach, and as the language suggested this value should be considered not set not a default value.

SergeyKanzhelev: if you feel strongly about it - I'm fine. Perhaps put a comment saying that implementation can assume `INTERNAL` when received `UNDEFINED`.

songy23: Reverted 2e24a8b and added a note in https://github.com/open-telemetry/opentelemetry-proto/pull/9/commits/03c06742b895249dd52b4d9ba2955446543ec427.

songy23:
Thanks for the review, please help me merge this PR since I'm not authorized to do so :)

PR #14, opened 2019-05-31
Remove same_process_as_parent_span.

### LINKED ISSUE #7: Remove or Update the name for same_process_as_parent_span
See comments for what is intended to be used, and decide if it is needed:

https://github.com/open-telemetry/opentelemetry-proto/blob/master/src/opentelemetry/proto/trace/v1/trace.proto#L245

songy23:
Fixes #7.

### DIFF
--- a/src/opentelemetry/proto/trace/v1/trace.proto
+++ b/src/opentelemetry/proto/trace/v1/trace.proto
@@ -32,7 +32,7 @@ option java_outer_classname = "TraceProto";
 // multiple root spans, or none at all. Spans do not need to be
 // contiguous - there may be gaps or overlaps between spans in a trace.
 //
-// The next id is 16.
+// The next id is 15.
 message Span {
   // A unique identifier for a trace. All spans from the same trace share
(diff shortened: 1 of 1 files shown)


src/opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -237,16 +237,9 @@ message Span {
...
-  // trace crosses a process boundary. True when the parent_span belongs
-  // to the same process as the current span. This flag is most commonly
-  // used to indicate the need to adjust time as clocks in different
-  // processes may not be synchronized.
-  google.protobuf.BoolValue same_process_as_parent_span = 14;
-
   // An optional number of child spans that were generated while this span
   // was active. If set, allows an implementation to detect missing child spans.
-  google.protobuf.UInt32Value child_span_count = 15;
+  google.protobuf.UInt32Value child_span_count = 14;
```
bogdandrutu: At the top of the Span proto message there is a message with the next id, please update.

songy23: Good catch, fixed.

Issue #13, opened 2019-05-31
Set up continuous integration.


songy23:
As an example see https://github.com/census-instrumentation/opencensus-proto/blob/master/.travis.yml.

PR #15, opened 2019-05-31
Setup Travis.

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-proto/issues/13

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+sudo: required
+dist: trusty
+
+matrix:
+  include:
+  - language: java
+    jdk: oraclejdk8
+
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,20 @@
+sudo: required
+dist: trusty
```
bai: Nitpick: `trusty` support ended [2 months ago (01 Apr 2019)](https://wiki.ubuntu.com/Releases). It might make sense to use `xenial` (16.04) instead.

bogdandrutu: Happy to accept a PR :)

PR #16, opened 2019-05-31
Ubuntu 14.04 is EOLd, use 16.04 instead

bai:
This PR switches Travis CI from Ubuntu 14.04 Trusty to Ubuntu 16.04 Xenial. Ubuntu 14.04 support [ended 2 months ago](https://endoflife.date/ubuntu).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 sudo: required
-dist: trusty
+dist: xenial
 
 matrix:
   include:
   - language: java
-    jdk: oraclejdk8
+    jdk: openjdk8
 
 script:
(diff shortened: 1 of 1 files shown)


bogdandrutu:
Restarted the build to confirm if this is a transitive error, otherwise please fix :)

bai:
Sorry for premature CODEOWNERS ping. I pushed an update, please take a look.

OracleJDK is not included in Ubuntu after 14.04 so I included OpenJDK instead, which should be the same™ apart from Flight Recorder and all those things. Let me know what you think 👀

bogdandrutu:
Happy to be more "open" :)

PR #18, opened 2019-06-03
Ignore VS Code settings in gitignore.

songy23:


.gitignore
```diff
@@ -12,6 +12,9 @@ out/
 
 # VS Code
 .vscode
```
SergeyKanzhelev: I found it useful to exchange some vscode settings in the project. Like markdown line length or so. unfortunately some settings only available as vscode setting, not editorconfig

PR #21, opened 2019-07-18
Capture Wire Protocol design goals

tigrannajaryan:
This is a follow up from Spec SIG meeting on Jul 18, 2019, recapping
the topics that I mentioned in the meeting.

The document will help us design the right wire protocol.

### DIFF
--- a/specification/proto-design-goals.md
+++ b/specification/proto-design-goals.md
@@ -1,0 +1,19 @@
+# Design Goals for OpenTelemetry Wire Protocol
+
+We want to design a telemetry data exchange protocol that has the following characteristics:
+
+- Be suitable for use between all of the following node types: instrumented applications, telemetry backends, local agents, stand-alone collectors/forwarders.
+
+- Have high reliability of data delivery and clear visibility when the data cannot be delivered.
+
+- Have low CPU usage for serialization and deserialization.
+
+- Impose minimal pressure on memory manager, including pass-through scenarios, where deserialized data is short-lived and must be serialized as-is shortly after and where such short-lived data is created and discarded at high frequency (think telemetry data forwarders).
+
+- Support ability to efficiently modify deserialized data and serialize again to pass further. This is related but slightly different from the previous requirement.
+
+- Ensure high throughput (within the available bandwidth) in high latency networks (e.g. scenarios where telemetry source and the backend are separated by high latency network).
+
+- Allow backpressure signalling.
+
+- Be load-balancer friendly (do not hinder re-balancing).


specification/proto-design-goals.md
```diff
@@ -0,0 +1,19 @@
+# Design Goals for OpenTelemetry Wire Protocol
+
+We want to design a telemetry data exchange protocol that has the following characteristics:
+
+- Be suitable for use between all of the following node types: instrumented applications, telemetry backends, local agents, stand-alone collectors/forwarders.
+
+- Have high reliability of data delivery and clear visibility when the data cannot be delivered.
```
ccaraman: for "data cannot be delivered" - is this the only type of errors?

tigrannajaryan: This is one of the problems that we encountered in production particularly with OpenCensus protocol. This requirement can be refined further if there is an evidence of other common error types.

specification/proto-design-goals.md
```diff
@@ -0,0 +1,19 @@
+# Design Goals for OpenTelemetry Wire Protocol
```
songy23: Should backwards compatibility be part of the goals?

tigrannajaryan: Do you mean backwards compatibility with any existing protocol? I think that should be a non-goal.

If you mean backwards compatibility within its own revisions then I think it goes without saying. That should be a goal of any protocol that can be deployed on a variety of devices at different times by different people. We can call this out specifically if needed.

songy23: I mean within its own versions. Agreed it may already be implied and not need to point it specifically.

specification/proto-design-goals.md
```diff
@@ -0,0 +1,19 @@
...
+
+We want to design a telemetry data exchange protocol that has the following characteristics:
+
+- Be suitable for use between all of the following node types: instrumented applications, telemetry backends, local agents, stand-alone collectors/forwarders.
+
+- Have high reliability of data delivery and clear visibility when the data cannot be delivered.
+
+- Have low CPU usage for serialization and deserialization.
```
pjanotti: nit: I prefer `resource` instead of `CPU` since we are at high-level.

tigrannajaryan: I think it is worth listing CPU and Memory separately. They are related but somewhat distinct with possibly different solutions. Especially the memory, I wanted to mention it in a separate paragraph so that we realize the specific pattern of allocating and using memory can be quite unfriendly to GCs.

specification/proto-design-goals.md
```diff
@@ -0,0 +1,19 @@
+# Design Goals for OpenTelemetry Wire Protocol
+
+We want to design a telemetry data exchange protocol that has the following characteristics:
```
yurishkuro: is this intended to be a single protocol for different types of telemetry?

tigrannajaryan: My past experience is with traces and logs (but not metrics). I think the answer should be yes, one protocol for all types, unless we think other telemetry types have significantly different requirements for the protocol. Thoughts are welcome.

specification/proto-design-goals.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Impose minimal pressure on memory manager, including pass-through scenarios, where deserialized data is short-lived and must be serialized as-is shortly after and where such short-lived data is created and discarded at high frequency (think telemetry data forwarders).
+
+- Support ability to efficiently modify deserialized data and serialize again to pass further. This is related but slightly different from the previous requirement.
+
+- Ensure high throughput (within the available bandwidth) in high latency networks (e.g. scenarios where telemetry source and the backend are separated by high latency network).
+
+- Allow backpressure signalling.
+
+- Be load-balancer friendly (do not hinder re-balancing).
```
tedsuo: At what granularity are we load balancing? Per span?

tigrannajaryan: Typically per batch of spans. L7 load balancers trigger rebalancing in specific situations. For gRPC we found that it typically happens when gRPC stream is opened. By controlling what you include in one stream you can control the frequency of rebalancing.
Doing it per span is likely going to be prohibitively expensive.

specification/proto-design-goals.md
reyang: How do we think about de-duplicating the data? (e.g. in case of transmission failures, where sender has no idea whether the backend accepted the data or not)

tigrannajaryan: I'll have this in the RFC. In short, I suggest to aim for reliable delivery at the cost of possible duplicates. Details in the RFC.

reyang: Thank you, I'll look into the RFC.

specification/proto-design-goals.md
tedsuo: I think this makes sense... there _is_ a possible issue with the fact that metrics systems do not consume traces, and there are perhaps emerging standards for wire protocols in that space. How do we play well with prometheus, etc? Or is that out of scope? AKA we will always be doing some data conversion/processing in a gateway node between the telemetry network and the metrics backend.

tigrannajaryan:
Reviewers, if there are no objections about these goals, let's merge this.

tigrannajaryan:
Reviewers, I need more comments or one more approval to merge this.

tigrannajaryan:
Approvers, please merge.

PR #22, opened 2019-07-22
Add OpenTelemetry Protocol requirements

tigrannajaryan:
This follows from the design goals and will feed the protocol RFC.

### DIFF
--- a/ specification/proto-requirements.md
+++ b/ specification/proto-requirements.md
@@ -1,0 +1,93 @@
+# OpenTelemetry Protocol Requirements
+
+This document will drive OpenTelemetry Protocol design and RFC.
+
+## Goals
+
+See the goals of OpenTelemetry Protocol design [here](proto-design-goals.md).
+
+## Vocabulary
+
+There are 2 parties involved in telemetry data exchange. In this document the party that is the source of telemetry data is called the Client, the party that is the destination of telemetry data is called the Server.
+
+Examples of a Client are instrumented applications or sending side of telemetry collectors, examples of Servers are telemetry backends or receiving side of telemetry collectors (so a Collector is typically both a Client and a Server depending on which side you look from).
+
+## Known Issues with Existing Protocols
+
+Our experience with OpenCensus and other protocols has been that many of them have one or more of the following drawbacks:
+
+- High CPU consumption for serialization and especially deserialization of received telemetry data.
+- High and frequent CPU consumption by Garbage Collector.
+- Lack of delivery guarantees for certain protocols (e.g. stream-based gRPC OpenCensus protocol) which makes troubleshooting of telemetry pipelines difficult.
+- Not aware / not cooperating with load balancers resulting in potentially large imbalances in horizontally scaled backends.
+- Support either traces or metrics but not both.
+
+Our goal is to avoid or mitigate these known issues in the new protocol.
+
+## Requirements
+
+The following are OpenTelemetry protocol requirements.
+
+#### Supported Node Types
+
+The protocol must be suitable for use between all of the following node types: instrumented applications, telemetry backends, telemetry agents running as local daemons, stand-alone collector/forwarder services.
+
+#### Supported Data Types
+
+The protocol must support traces and metrics as data types.
+
+#### Reliability of Delivery
+
+The protocol must ensure reliable data delivery and clear visibility when the data cannot delivered. This should be achieved by sending data acknowledgements from the Server to the Client. 
+
+Note that acknowledgements alone are not sufficient to guarantee that: a) no data will be lost and b) no data will be duplicated. Acknowledgements can help to guarantee a) but not b). Guaranteeing both at the same is difficult. Because it is usually preferable for telemetry data to be duplicated than to lose it, we choose to guarantee that there are no data losses while potentially allowing duplicate data.
+
+Duplicates can typically happen in edge cases (e.g. on reconnections, network interruptions, etc) when the client has no way of knowing if last sent data was delivered. In these cases the client will usually choose to re-send the data to guarantee the delivery which in turn may result in duplicate data on the server side.
+
+_To avoid having duplicates the client and the server could each track sent and delivered items using uniquely identifying ids. For this approach to be possible the client should be able to request the id of the last received data from the server and be able to resume sending from that item. Such request may be impossible to support on the server efficiently as it requires the server to have a consistent view of its data which may not be possible to do in a partitioned state, or it requires sticky sessions/connections which can be detrimental to load balancing and availability of servers._
+
+For this reason we have slightly relaxed requirements and consider duplicate data acceptable in rare cases.
+
+#### Throughput
+
+The protocol must ensure high throughput in high latency networks when the client and the server are not in the same data center.
+
+This requirement may rule out half-duplex protocols. The throughput of half-duplex protocols is highly dependent on network roundtrip time and request size. To achieve good throughput request sizes may be too large to be practical.
+
+#### Compression
+
+The protocol must achieve high compression ratios for telemetry data. The protocol design must consider batching of telemetry data and grouping of similar data (both can help to achieve better compression using common compression algorithms).
+
+#### Encryption
+
+Industry standard encryption (e.g. TLS/HTTPS) must be supported.
+
+#### Backpressure Signalling and Throttling
+
+The protocol must allow backpressure signalling.
+
+If the server is unable to keep up with the pace of data it receives from the client then it must be able to signal that fact to the client. The client may then throttle itself to avoid overwhelming the server.
+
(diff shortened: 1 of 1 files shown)


 specification/proto-requirements.md
```diff
@@ -0,0 +1,93 @@
...
+
+The protocol must be suitable for use between all of the following node types: instrumented applications, telemetry backends, telemetry agents running as local daemons, stand-alone collector/forwarder services.
+
+#### Supported Data Types
+
+The protocol must support traces and metrics as data types.
+
+#### Reliability of Delivery
+
+The protocol must ensure reliable data delivery and clear visibility when the data cannot delivered. This should be achieved by sending data acknowledgements from the Server to the Client.
```
songy23:
```suggestion
The protocol must ensure reliable data delivery and clear visibility when the data cannot be delivered. This should be achieved by sending data acknowledgements from the Server to the Client. 
```

tigrannajaryan:
>Suggestion: add a section about crash recovery. (It's fine that we don't attempt to recover anything but I think it's worth documenting here.)

@songy23 can you expand on this? I think crash recovery is orthogonal to the protocol, perhaps I am missing something.

 specification/proto-requirements.md
tigrannajaryan: Fixed.

songy23:
> can you expand on this? I think crash recovery is orthogonal to the protocol, perhaps I am missing something.

I mean when client/server of this protocol crashed, the behavior on undelivered telemetry data should be specified (whether dropping or storing/recovering). This could be an implementation detail but I think it's worth mentioning.

Bringing this up because there're many asks about crash recovery previously in the OpenCensus protocol.

 specification/proto-requirements.md
```diff
@@ -0,0 +1,93 @@
...
+
+Duplicates can typically happen in edge cases (e.g. on reconnections, network interruptions, etc) when the client has no way of knowing if last sent data was delivered. In these cases the client will usually choose to re-send the data to guarantee the delivery which in turn may result in duplicate data on the server side.
+
+_To avoid having duplicates the client and the server could each track sent and delivered items using uniquely identifying ids. For this approach to be possible the client should be able to request the id of the last received data from the server and be able to resume sending from that item. Such request may be impossible to support on the server efficiently as it requires the server to have a consistent view of its data which may not be possible to do in a partitioned state, or it requires sticky sessions/connections which can be detrimental to load balancing and availability of servers._
```
reyang: I agree with `To avoid having duplicates the client and the server could each track sent and delivered items using uniquely identifying ids`. I disagree with `For this approach to be possible the client should be able to request the id of the last received data from the server and be able to resume sending from that item`, which is not a must-have requirement.
De-dupe could happen at different places in the telemetry pipeline, either at agent, ingestion service, or the final storage. The client could keep sending the data (using a reasonable retry policy) until it received a success response.

 specification/proto-requirements.md
```diff
@@ -0,0 +1,93 @@
...
+
+#### Supported Data Types
+
+The protocol must support traces and metrics as data types.
+
+#### Reliability of Delivery
+
+The protocol must ensure reliable data delivery and clear visibility when the data cannot be delivered. This should be achieved by sending data acknowledgements from the Server to the Client. 
+
+Note that acknowledgements alone are not sufficient to guarantee that: a) no data will be lost and b) no data will be duplicated. Acknowledgements can help to guarantee a) but not b). Guaranteeing both at the same is difficult. Because it is usually preferable for telemetry data to be duplicated than to lose it, we choose to guarantee that there are no data losses while potentially allowing duplicate data.
```
reyang: Even for a), it is extremely difficult when we look at the entire pipeline.

For example SDK sends data to agent, agent sends data to Stackdriver. If the agent takes incoming data and store them temporarily in-memory (or even on local disk), data loss could happen when agent got unexpected restart (or the machine which agent is running on could get stolen). If the agent is supposed to return a success to the SDK **only if** the data got delivered to Stackdriver (or in another words, the agent could say "I've done my part, now it is Stackdriver's problem"), the agent **has** to actually deliver it to Stackdriver and receive a success acknowledgement. Eventually this system will become a fully distributed transaction system which uses something like two-phase-commit and becomes super slow and fragile.

 specification/proto-requirements.md
tigrannajaryan: >De-dupe could happen at different places in the telemetry pipeline, either at agent, ingestion service, or the final storage. The client could keep sending the data (using a reasonable retry policy) until it received a success response.

I think what you say does not contradict to what this proposal says. The client will send the data again if it is not sure it was received. If the server wants to dedupe then that's server's responsibility, the client is not going to do anything special about it. 

The deduping you refer to that can happen at different places in the pipeline is orthogonal to the protocol and is not a protocol concern if protocol is not going to do anything about it.

Trace and Span Ids are already part of telemetry data. We don't need to add anything in the protocol to make deduplication possible. One can say that Ids are _already_ tracked, it is just not tracked at the protocol level, so from protocol's perspective we really do not care about the ids and are not going to do anything about ids or about preventing duplicates.

 specification/proto-requirements.md
tigrannajaryan: The protocol is concerned with one pair of client/server only. What happens after the server acknowledges the receipt is not a protocol concern. The server may store the data persistently or may discard it, may forward it, etc. It is not a protocol concern. You are correct that unless the protocol requires end-to-end acknowledgement delivery it cannot guarantee delivery end-to-end. However, it is not a goal for the protocol at all. It can be a design goal for a pipeline engineer. We are not designing a pipeline here, we only design telemetry data delivery between a pair of nodes.

This is identical to how TCP guarantees delivery of data but it is not concerned with what the receiver of said data does with the data afterwards. On the "protocol" abstraction level we are not concerned with that.

 specification/proto-requirements.md
reyang: I think we need two changes to the proposal:
1. Loose the statement `For this approach to be possible the client should be able to request the id of the last received data from the server and be able to resume sending from that item.` since it is one option but not a must-have thing.
2. Need some de-dupe id (for traces we can just reuse trace-id/span-id pair, for metrics/logs we need something else).

 specification/proto-requirements.md
reyang: I see where you're coming from, and that sounds totally reasonable.

My worry is that in monitoring space people might misinterpret "no data loss" into different things (e.g. the end-to-end guarantee), it will be great to clarify it.

 specification/proto-requirements.md
tigrannajaryan: OK, I'll remove 1 and will add a sentence explicitly mentioning that id tracking and deduping can be done at the layer above protocol layer.

 specification/proto-requirements.md
tigrannajaryan: Makes sense, I will add a clarification.

tigrannajaryan: Clarification added.

 specification/proto-requirements.md
tigrannajaryan: Fixed.

 specification/proto-requirements.md
tigrannajaryan: Actually, I need to think a bit more about this. Thanks for raising the issue.

tigrannajaryan:
>I mean when client/server of this protocol crashed, the behavior on undelivered telemetry data should be specified (whether dropping or storing/recovering). This could be an implementation detail but I think it's worth mentioning.

>Bringing this up because there're many asks about crash recovery previously in the OpenCensus protocol.

@songy23 typically crash resilience is achieved by implementing persistent storage of received messages on the server (receiver) side and making sure received data is safely persisted and fsynced before the acknowledgement of that particular data is sent to the client. This does not put additional requirements on the protocol that I can think of.

The way I have implemented this in the past for a log collection and forwarding agent was by storing received log messages in a persistent queue implemented on top of a very simple sqlite database. The persistent queue needs an ID value to find and remove delivered items when acknowledgement is received from the destination however that ID can be completely internal and is independent from any ID that may or may not be present on the protocol level. This allows to decouple the crash resiliency / persistent queue implementation from the exchange protocol. 

Sometimes you _can_ reuse the IDs assigned at the protocol level for persistent queue implementation. However often this is not possible because the particular protocol or the data that it delivers has no notion of IDs. For this reason I think we should stay away from leaking protocol-level IDs (even if we have them) into the implementation of persistent queue. There are other reasons too: particularly that protocol-level IDs are typically not guaranteed to be unique across different connection sessions which makes them useless for persistent use cases.

We can implement similar persistent queue functionality for OpenTelemetry Service and I do not think we need special support in the protocol for this.

tigrannajaryan:
As discussed in today's Data Format meeting I moved the requirements document to opentelemetry-specification repo 

Closing this, please continue review / discussion here https://github.com/open-telemetry/opentelemetry-specification/pull/193

PR #26, opened 2019-08-21
Remove gradle and keep only proto files

### LINKED ISSUE #25: Remove gradle build and keep only .proto files
In Java client we would like to consume this library as a GIT submodule and run the build directly from there. The model classes will be published to maven central as part of the client build (as a separate module).

I think this approach will work for all clients. This way we don't have to configure publishing to central registries in multiple places and we also assure a client use the same versions of proto dependencies (e.g. java client builds Jaeger proto and otel proto)

cc) @bogdandrutu  @tigrannajaryan @c24t @reyang @mayurkale22 @vmarchaud @fbogsany @SergeyKanzhelev @lmolkova @tsloughter 

Here is an example of using proto files in Java https://github.com/open-telemetry/opentelemetry-java/pull/504 - at the moment it uses https://github.com/pavolloffay/opentelemetry-proto

pavolloffay:
Resolves #25


.travis.yml
```diff
@@ -1,20 +1,9 @@
...
-
-before_cache:
-  - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
-
-cache:
-  directories:
-    - $HOME/.gradle
-    - $HOME/.gradle/caches/
-    - $HOME/.gradle/wrapper/
+  - ./java_build.sh
```
pavolloffay: I have included this to test the compilation

specification/proto-design-goals.md
```diff
@@ -1,19 +0,0 @@
-# Design Goals for OpenTelemetry Wire Protocol
```
pavolloffay: This was already moved to RFC repo

agent/metrics/v1/metrics_service.proto
```diff
@@ -16,9 +16,9 @@ syntax = "proto3";
 
 package opentelemetry.proto.agent.metrics.v1;
 
-import "opentelemetry/proto/agent/common/v1/common.proto";
-import "opentelemetry/proto/metrics/v1/metrics.proto";
-import "opentelemetry/proto/resource/v1/resource.proto";
+import "agent/common/v1/common.proto";
+import "metrics/v1/metrics.proto";
+import "resource/v1/resource.proto";
```
bogdandrutu: If these files will be built with bazel then imported in C++ or other proto files will be strange to have them included as "metrics/..." it is very generic path. 

What is the motivation to move the files?

pavolloffay: No reason. At least java files use correct package name `io.opentelemetry` I am not sure what the build for other languages produces. 

I have reverted the path.

specification/proto-design-goals.md
bogdandrutu: Can we not do this in the current PR and open a different one with this? I want @tigrannajaryan to comment and others.

bogdandrutu:
@pavolloffay ping

pavolloffay:
sorry I was stuck on other things, I am going to update the PR now.

pavolloffay:
@bogdandrutu I have updated the PR

PR #29, opened 2019-10-24
Refine Metrics Protobuf definitions

tigrannajaryan:
This change applies the refinement approach that is already performed on Traces
Protobuf definitions as part of https://github.com/open-telemetry/oteps/pull/59 and
which proved to yield significant performance improvements.

Notable changes are:
- Replace google.protobuf.Timestamp by int64 time in unix epoch nanoseconds.
- Eliminate unnecessary messages, move the fields into the containing message.
- Replace oneof by a set of fields.

Simple benchmark in Go demonstrates the following improvement compared to OpenCensus
Metrics encoding and decoding:

```
BenchmarkEncode/OpenCensus/Metrics-8         	       2	 645252504 ns/op
BenchmarkEncode/OTLP/Metrics-8               	       4	 288457433 ns/op
BenchmarkDecode/OpenCensus/Metrics-8                   1	1154650804 ns/op
BenchmarkDecode/OTLP/Metrics-8                         3	 475370617 ns/op
```

Encoding is about 2.2 times faster, decoding is about 2.4 times faster.

Benchmarks encode and decode 500 batches of 2 metrics: one int64 Gauge with 5 time series
and one Histogram of doubles with 1 time series and single bucket. Each time series for
both metrics contains 5 data points. Both metrics have 2 labels.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -16,8 +16,6 @@ syntax = "proto3";
 
 package opentelemetry.proto.metrics.v1;
 
-import "google/protobuf/timestamp.proto";
-import "google/protobuf/wrappers.proto";
 import "opentelemetry/proto/resource/v1/resource.proto";
 
 option java_multiple_files = true;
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -110,87 +117,104 @@ message TimeSeries {
...
+  // [bucket_bounds[i], +infinity) for i == N-1
+  // The values in bucket_bounds array must be strictly increasing and > 0.
   // Don't change bucket boundaries within a TimeSeries if your backend doesn't
   // support this.
-  BucketOptions bucket_options = 3;
+  // If bucket_bounds field is unspecified then bucket bounds are not explicitly defined.
+  // TODO: If OpenMetrics decides to support (a, b] intervals we should add
+  // support for these by defining a boolean value which decides what type of
+  // intervals to use.
+  repeated double bucket_bounds = 3;
```
bogdandrutu: We may want to support the other types of the BucketOptions like Linear. So I don't think this optimization can be done this way.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -199,67 +223,62 @@ message HistogramValue {
...
-    }
+    // The observation (sampling) time of the above value.
+    int64 exemplar_timestamp_unixnano = 3;
 
-    // Exemplars are example points that may be used to annotate aggregated
-    // Histogram values.
-    Exemplar exemplar = 2;
+    // timestamp_unspecified must be set to true if timestamp_unixnano value is unspecified.
+    // In that case timestamp_unixnano value will be decided by the backend.
+    bool exemplar_timestamp_unspecified = 4;
```
bogdandrutu: we can consider 0 as not set. or a negative value.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -199,67 +223,62 @@ message HistogramValue {
...
+
+    // exemplar_attachments are contextual information about the example value.
+    repeated StringKeyValuePair exemplar_attachments = 5;
```
bogdandrutu: Add a comment that these need unique keys?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -199,67 +223,62 @@ message HistogramValue {
...
 
-      // Contextual information about the example value.
-      map<string, string> attachments = 3;
-    }
+    // The observation (sampling) time of the above value.
+    int64 exemplar_timestamp_unixnano = 3;
 
-    // Exemplars are example points that may be used to annotate aggregated
-    // Histogram values.
-    Exemplar exemplar = 2;
```
bogdandrutu: I am not sure if this is worth it. Exemplars will not be used for all the histograms, and the current change increases the memory usage for all the histograms to avoid one extra allocation. Can we split this in a separate PR and discuss?

tigrannajaryan:
I am closing this PR. Will create separate smaller PRs.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: I'm not sure this exemplar stuff is useful enough as specified. I'd like to support more than one exemplar. I'd like exemplars to carry a sample weight too.

tigrannajaryan: @jmacd I am going to create a separate PR, we can discuss there.

PR #30, opened 2019-10-25
[WIP] Performance optimize Point protobuf message

tigrannajaryan:
Changes in Point message:

- Replace google.protobuf.Timestamp by int64 time in unix epoch nanoseconds.
- Replace oneof by a set of fields.

This is part 1 one of overall changes that yield about 2x performance improvement.

Benchmarking
============

Simple Go benchmark of *all* changes that I plan to propose demonstrates the following
improvement compared to OpenCensus Metrics encoding and decoding:

```
BenchmarkEncode/OpenCensus/Metrics-8    8	 660200437 ns/op	80432000 B/op	 1787000 allocs/op
BenchmarkEncode/OTLP/Metrics-8         19	 296717207 ns/op	49152000 B/op	    1000 allocs/op
BenchmarkDecode/OpenCensus/Metrics-8    5	1040309846 ns/op	458263993 B/op	11877000 allocs/op
BenchmarkDecode/OTLP/Metrics-8         12	 501670950 ns/op	346696049 B/op	 7124000 allocs/op
```

Encoding and decoding is about 2 times faster with new format. Memory usage is also
significantly smaller, number of allocations during decoding is roughly half of OpenCensus
which reflects the fact that many messages were eliminated and embedded inside others.

Benchmarks encode and decode 1000 batches of 2 metrics: in each batch 50 int64 Gauges with
5 time series and 50 Histograms of doubles with 1 time series and single bucket. Each time
series for both metrics contains 5 data points. Both metrics have 2 labels.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -122,32 +122,51 @@ message LabelValue {
   bool has_value = 2;
 }
 
-// A timestamped measurement.
+// Point is a timestamped measurement.
 message Point {
-  // Must be present for counter/cumulative metrics. The time when the
-  // cumulative value was reset to zero. The cumulative value is over the time
-  // interval (start_timestamp, timestamp]. If not specified, the backend can
-  // use the previous recorded value.
-  google.protobuf.Timestamp start_timestamp = 1;
+  // start_time_unixnano is the time when the cumulative value was reset to zero.
+  // Must be present for counter/cumulative metrics. The cumulative value is over the time
+  // interval (start_time_unixnano, timestamp_unixnano].
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
+  int64 start_time_unixnano = 1;
 
-  // The moment when this point was recorded.
-  // If not specified, the timestamp will be decided by the backend.
-  google.protobuf.Timestamp timestamp = 2;
+  // timestamp_unixnano is the moment when this point was recorded.
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
+  int64 timestamp_unixnano = 2;
 
-  // The actual point value.
-  oneof value {
-    // A 64-bit integer.
-    int64 int64_value = 3;
+  // start_time_unspecified must be set to true if start_time_unixnano value is unspecified.
+  // In that case the backend can use the previous recorded value for start_time_unixnano.
+  bool start_time_unspecified = 3;
 
-    // A 64-bit double-precision floating-point number.
-    double double_value = 4;
+  // timestamp_unspecified must be set to true if timestamp_unixnano value is unspecified.
+  // In that case timestamp_unixnano value will be decided by the backend.
+  bool timestamp_unspecified = 4;
 
-    // A histogram value.
-    HistogramValue histogram_value = 5;
+  // ValueType is the enumeration of possible types that Point's value can have.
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -122,32 +122,51 @@ message LabelValue {
...
-    int64 int64_value = 3;
+  // timestamp_unixnano is the moment when this point was recorded.
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
+  int64 timestamp_unixnano = 3;
 
-    // A 64-bit double-precision floating-point number.
-    double double_value = 4;
+  // timestamp_unspecified must be set to true if timestamp_unixnano value is unspecified.
+  // In that case timestamp_unixnano value will be decided by the backend.
+  bool timestamp_unspecified = 4;
```
bogdandrutu: What about we use 0 as default/unspecified and remove the bool?

tigrannajaryan: I am not a big fan of magic values and 0 is technically a valid timestamp although probably unrealistic in the context of timeseries.
Using a separate bool field doesn't add any bytesize in the encoded form when timestamp is present and is a very small increase for in-memory representation.

P.S. This reminded me that I should probably rearrange the fields to ensure they are better packed in memory in the struct based on their alignment requirements. I'll have a look at that.

tigrannajaryan: Yeah, that indeed makes a difference. Rearranging bools to be next to each other saves 8 bytes from unsafe.Sizeof(Point), from 96 to 88 bytes. This doesn't make any differrence on the total allocated memory during test, because of heap allocator granularity (the nearest class sizes are 80 and 96 bytes, so both get 96 bytes).
I think it is still worth it since if more fields are added in the future we will have room for growth for the current block size of 96.
So I am going to change the order of the fields to this:
```protobuf
  int64 start_time_unixnano = 1;
  int64 timestamp_unixnano = 2;
  bool start_time_unspecified = 3;
  bool timestamp_unspecified = 4;
```
This is slightly less natural but I think is fine and worth it.

bogdandrutu: I am still confused why 0 is not a good default/unspecified value. Sorry :(

tigrannajaryan: 0 nanoseconds from unix epoch start is technically a valid timestamp, it is exactly 00:00:00 UTC on 1 January 1970.

I agree, this is probably nonsense for a timeseries. No-one is going to send timestamps for year 1970 for an existing observability system.

I am almost ready to agree with you if you think that this is really not an issue. I like the fact that we can save space and remove the bool field.

Anyone else has any supporting arguments that make 0 a safe special value, which means timestamp is unknown?

tigrannajaryan:
Please do not merge this yet even if it gets approved. I am exploring an alternate representation for Point, which may be even better.

jmacd:
I'm willing to go with what you feel is best, regarding the 0 vs boolean question.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Let's go without the boolean for the moment. If we get feedback from others we can address this. Maybe just create an issue to track this discussion.

tigrannajaryan:
I am closing this one and will open a series of separate PRs with incremental improvements.

PR #31, opened 2019-10-29
Add Python build script also usable on Windows

Oberon00:
This PR adds a Python script to call protoc. The script is useful for local development on Windows, but could also be used to entirely replace the bash script.

### DIFF
--- a/genproto.py
+++ b/genproto.py
@@ -7,26 +7,36 @@ import subprocess
 import argparse
 import time
 
+
 def rmtree_retry(dpath):
     if not path.exists(dpath):
         return
     try:
         shutil.rmtree(dpath)
     except OSError:
-        time.sleep(1) # For antivirus (?)
+        time.sleep(1)  # For antivirus (?)
         shutil.rmtree(dpath)
+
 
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev:
If you are adding a script to repo - it is a good practice to add a note in CONTRIBUTING how to use it to simplify local development. This will put more context on why it is needed.

bogdandrutu:
@Oberon00 we said that we will use git submodules, so I would suggest that this script to be moved where you plan to write the opentelemetry exporter and use git submodules. What do you think?

tigrannajaryan:
> @Oberon00 we said that we will use git submodules, so I would suggest that this script to be moved where you plan to write the opentelemetry exporter and use git submodules. What do you think?

I agree, we should publish only .proto file. It would be nice to have a CI that runs `protoc` on the files in this repo for validation purposes, but that doesn't require Windows support.

SergeyKanzhelev:
@Oberon00 I haven't looked into the script. If you believe it will be useful for many people - reopen the PR add a note to CONTRIBUTING on how to use it. Otherwise closing the PR.

PR #34, opened 2019-10-30
Split timeseries by data types in Metrics Protobuf definitions

tigrannajaryan:
Previously we had Point message which was a value of oneof the data types. This is
unnecessary flexibility because points in the same timeseries cannot be of different
data type. This also costed performance.

Now we have separate timeseries message definitions for each data type and the
timeseries used is defined by the oneof entry in the Metric message.

This change is stacked on top of https://github.com/open-telemetry/opentelemetry-proto/pull/33

Simple benchmark in Go demonstrates the following improvement of encoding and decoding
compared to the baseline state:

```
===== Encoded sizes
Baseline/MetricOne              24200 bytes  [1.000], gziped 1813 bytes  [1.000]
Proposed/MetricOne              19150 bytes  [1.264], gziped 1586 bytes  [1.143]

Baseline/MetricSeries           56022 bytes  [1.000], gziped 6691 bytes  [1.000]
Proposed/MetricSeries           43115 bytes  [1.299], gziped 6328 bytes  [1.057]

goos: darwin
goarch: amd64
BenchmarkEncode/Baseline/MetricOne-8         	      26	 212691770 ns/op	27376010 B/op	  351000 allocs/op
BenchmarkEncode/Proposed/MetricOne-8         	      50	 122404576 ns/op	20880001 B/op	   51000 allocs/op

BenchmarkEncode/Baseline/MetricSeries-8      	       8	 653270997 ns/op	71344004 B/op	 1751000 allocs/op
BenchmarkEncode/Proposed/MetricSeries-8      	      19	 313939544 ns/op	51152000 B/op	  251000 allocs/op

BenchmarkDecode/Baseline/MetricOne-8         	      14	 383343126 ns/op	186296037 B/op	 5274000 allocs/op
BenchmarkDecode/Proposed/MetricOne-8         	      24	 250832929 ns/op	156696050 B/op	 4274000 allocs/op

BenchmarkDecode/Baseline/MetricSeries-8      	       5	1031024029 ns/op	455096041 B/op	12174000 allocs/op
BenchmarkDecode/Proposed/MetricSeries-8      	      10	 567579261 ns/op	339096060 B/op	 8374000 allocs/op
```

It is 30-50% faster and is 20-25% smaller on the wire and in memory.

Benchmarks encode and decode 500 batches of 2 metrics: one int64 Gauge with 5 time series
and one Histogram of doubles with 1 time series and single bucket. Each time series for
both metrics contains either 1 data point (MetricOne) or 5 data points (MetricSeries).
Both metrics have 2 labels.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -28,13 +28,19 @@ message Metric {
   // The descriptor of the Metric.
   MetricDescriptor metric_descriptor = 1;
 
-  // One or more timeseries for a single metric, where each timeseries has
-  // one or more points.
-  repeated TimeSeries timeseries = 2;
-
   // The resource for the metric. If unset, it may be set to a default value
   // provided for a sequence of messages in an RPC stream.
-  opentelemetry.proto.resource.v1.Resource resource = 3;
+  opentelemetry.proto.resource.v1.Resource resource = 2;
+
+  // data is a list of one or more TimeSeries for a single metric, where each timeseries has
+  // one or more points. Only one of the following fields is used for the data, depending on
+  // the type of the metric defined by MetricDescriptor.type field.
+  oneof data {
+    Int64TimeSeriesList int64_data = 3;
+    DoubleTimeSeriesList double_data = 4;
+    HistogramTimeSeriesList histogram_data = 5;
+    SummaryTimeSeriesList summary_data = 6;
+  }
 }
 
 // Defines a metric type and its schema.
@@ -100,17 +106,72 @@ message MetricDescriptor {
   repeated string label_keys = 5;
 }
 
-// A collection of data points that describes the time-varying values
-// of a metric.
-message TimeSeries {
+// Int64TimeSeriesList is a list of timeseries of int64 values.
+message Int64TimeSeriesList {
+  repeated Int64TimeSeries list = 1;
+}
+
+// DoubleTimeSeriesList is a list of timeseries of double values.
+message DoubleTimeSeriesList {
+  repeated DoubleTimeSeries list = 1;
+}
+
+// HistogramTimeSeriesList is a list of timeseries of Histogram.
+message HistogramTimeSeriesList {
+  repeated HistogramTimeSeries list = 1;
+}
+
+// SummaryTimeSeriesList is a list of timeseries of Summary.
+message SummaryTimeSeriesList {
+  repeated SummaryTimeSeries list = 1;
+}
+
+// Int64TimeSeries is a list of data points that describes the time-varying values
+// of a int64 metric.
+message Int64TimeSeries {
   // The set of label values that uniquely identify this timeseries. Applies to
   // all points. The order of label values must match that of label keys in the
   // metric descriptor.
   repeated LabelValue label_values = 1;
 
-  // The data points of this timeseries. Point.value type MUST match the
-  // MetricDescriptor.type.
-  repeated Point points = 2;
+  // The data points of this timeseries.
+  repeated Int64Value points = 2;
+}
+
+// DoubleTimeSeries is a list of data points that describes the time-varying values
+// of a double metric.
+message DoubleTimeSeries {
+  // The set of label values that uniquely identify this timeseries. Applies to
+  // all points. The order of label values must match that of label keys in the
+  // metric descriptor.
+  repeated LabelValue label_values = 1;
+
+  // The data points of this timeseries.
+  repeated DoubleValue points = 2;
+}
+
+// HistogramTimeSeries is a list of data points that describes the time-varying values
+// of a Histogram.
+message HistogramTimeSeries {
+  // The set of label values that uniquely identify this timeseries. Applies to
+  // all points. The order of label values must match that of label keys in the
+  // metric descriptor.
+  repeated LabelValue label_values = 1;
+
+  // The data points of this timeseries.
+  repeated HistogramValue points = 2;
+}
+
+// SummaryTimeSeries is a list of data points that describes the time-varying values
+// of a Summary metric.
+message SummaryTimeSeries {
+  // The set of label values that uniquely identify this timeseries. Applies to
+  // all points. The order of label values must match that of label keys in the
+  // metric descriptor.
+  repeated LabelValue label_values = 1;
+
+  // The data points of this timeseries.
+  repeated SummaryValue points = 2;
 }
 
 message LabelValue {
@@ -121,8 +182,8 @@ message LabelValue {
   bool has_value = 2;
 }
 
-// A timestamped measurement.
-message Point {
+// Int64Value is a timestamped measurement of int64 value.
+message Int64Value {
   // start_time_unixnano is the time when the cumulative value was reset to zero.
   // This is used for Counter type only. For Gauge the value is not specified and
   // defaults to 0.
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -100,17 +106,72 @@ message MetricDescriptor {
   repeated string label_keys = 5;
 }
 
-// A collection of data points that describes the time-varying values
-// of a metric.
-message TimeSeries {
+// Int64TimeSeriesList is a list of timeseries of int64 values.
+message Int64TimeSeriesList {
```
bogdandrutu: Can we get rid of the "List" messages? and have oneof { repeated xyz }?

tigrannajaryan: No, unfortunately that's not valid in Protobuf :-(

bogdandrutu: But we can get rid of the entire oneof then. The type of the data is known from the metric descriptor. If [gauge|couter]_long then is Int64, etc.

tigrannajaryan: I am not sure I understand how your propose the message structure to be. Do you mean get rid of oneof and just list all fields (flatten as we did for Span AttributeValue)? Like this:
```protobuf
// Defines a Metric which has one or more timeseries.
message Metric {
  // The descriptor of the Metric.
  MetricDescriptor metric_descriptor = 1;

  // The resource for the metric. If unset, it may be set to a default value
  // provided for a sequence of messages in an RPC stream.
  opentelemetry.proto.resource.v1.Resource resource = 2;

  // data is a list of one or more TimeSeries for a single metric, where each timeseries has
  // one or more points. Only one of the following fields is used for the data, depending on
  // the type of the metric defined by MetricDescriptor.type field.
  repeated Int64TimeSeries int64_data = 3;
  repeated DoubleTimeSeries double_data = 4;
  repeated HistogramTimeSeries histogram_data = 5;
  repeated SummaryTimeSeries summary_data = 6;
}
```
This will work.

bogdandrutu: Yes that is what I think will be better. Not sure if the usage of keyward data is right but the structure looks like what I thought

tigrannajaryan: Yeah, I think this is even better (and will use less CPU). It will mean very small increase of in-memory size but simplies the message structure and removes a lot of message definitions. Let me change.

java_build.sh
```diff
@@ -17,4 +17,4 @@ protoc --java_out=${OUTDIR} opentelemetry/proto/resource/v1/resource.proto \
 OPENAPI_OUTDIR="./build/gen-openapi"
 
 mkdir -p $OPENAPI_OUTDIR
-protoc --plugin=protoc-gen-swagger=/usr/bin/protoc-gen-swagger --swagger_out=logtostderr=true,grpc_api_configuration=opentelemetry/proto/agent/trace/v1/trace_service_http.yaml:$OPENAPI_OUTDIR opentelemetry/proto/agent/trace/v1/trace_service.proto
+#protoc --plugin=protoc-gen-swagger=/usr/bin/protoc-gen-swagger --swagger_out=logtostderr=true,grpc_api_configuration=opentelemetry/proto/agent/trace/v1/trace_service_http.yaml:$OPENAPI_OUTDIR opentelemetry/proto/agent/trace/v1/trace_service.proto
```
SergeyKanzhelev: curious about this change...

tigrannajaryan: Ahh, slipped when committing, reverting.

tigrannajaryan: Fixed.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -121,8 +160,8 @@ message LabelValue {
   bool has_value = 2;
 }
 
-// A timestamped measurement.
-message Point {
+// Int64Value is a timestamped measurement of int64 value.
+message Int64Value {
   // start_time_unixnano is the time when the cumulative value was reset to zero.
   // This is used for Counter type only. For Gauge the value is not specified and
```
SergeyKanzhelev: will it help to have different data types for gauge and counter then? May be another place to save on some perf and memory

bogdandrutu: On the wire there is no memory saved because a default fixed64 is not sent (0 value). Only internal memory may decrease, I think it is a small RAM saving.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -28,13 +28,17 @@ message Metric {
...
   // provided for a sequence of messages in an RPC stream.
-  opentelemetry.proto.resource.v1.Resource resource = 3;
+  opentelemetry.proto.resource.v1.Resource resource = 2;
+
+  // data is a list of one or more TimeSeries for a single metric, where each timeseries has
+  // one or more points. Only one of the following fields is used for the data, depending on
+  // the type of the metric defined by MetricDescriptor.type field.
+  repeated Int64TimeSeries int64_timeseries = 3;
+  repeated DoubleTimeSeries double_timeseries = 4;
+  repeated HistogramTimeSeries histogram_timeseries = 5;
```
SergeyKanzhelev: should we call it doubleHistogram? In case we will decide to have int64 histogram in future... sorry if it was already discussed before

SergeyKanzhelev: this comments is not super relevant for this change =)

opentelemetry/proto/metrics/v1/metrics.proto
tigrannajaryan: Yeah, I had that version initially (with separate messages for guage and counter), we discussed with with @bogdandrutu and decided that small memory saving is not worth the added number of messages that the user need to choose correctly from. IMO it is the right tradeoff, the simplicity of the proto and small number of messages is important.

opentelemetry/proto/metrics/v1/metrics.proto
tigrannajaryan: That's a good question. We can add int64 histogram later, but you are right that it would be nice to get the name for this histogram right, so that in the future we don't end up with assymetric naming like `Histogram` and `Int64Histogram`. If there is a feeling int64 histogram may be needed in the future, it may be worth renaming Histogram to `DoubleHistogram` now (or perhaps `HistogramOfDoubles` as a better name? Neither name is very nice to me).

bogdandrutu: let's evaluate that later.

SergeyKanzhelev: https://github.com/open-telemetry/opentelemetry-proto/issues/35

tigrannajaryan: @SergeyKanzhelev thanks for creating the issue. Can we merge this PR and if we decide to rename the message we can do in a follow-up PR?

SergeyKanzhelev: I was waiting for a day for more comments... Ill merge by EOD ok?

tigrannajaryan: Sounds good, thanks.

PR #36, opened 2019-10-31
Optimize LabelValues in metric proto

tigrannajaryan:
Optimize LabelValues in metric proto

LabelValues is now a string array.

Benchmarking results are below (Baseline is current `master`, Proposed is after this commit).
This change reduces CPU usage up to 40% for one-data-point timeseries encoding and reduces
memory consumption by about 25%.

```
===== Encoded sizes
Baseline/Metric/Histogram       13569 bytes  [1.000], gziped  774 bytes  [1.000]
Proposed/Metric/Histogram       13159 bytes  [1.031], gziped  781 bytes  [0.991]

Baseline/Metric/MixOne          48530 bytes  [1.000], gziped 1671 bytes  [1.000]
Proposed/Metric/MixOne          45720 bytes  [1.061], gziped 1677 bytes  [0.996]

Baseline/Metric/MixSeries       97867 bytes  [1.000], gziped 6620 bytes  [1.000]
Proposed/Metric/MixSeries       95067 bytes  [1.029], gziped 6587 bytes  [1.005]

goos: darwin
goarch: amd64
BenchmarkEncode/Baseline/Metric/Int64-8        	      34	 156113692 ns/op
BenchmarkEncode/Proposed/Metric/Int64-8        	      67	  91936291 ns/op

BenchmarkEncode/Baseline/Metric/Summary-8      	     120	  50891750 ns/op
BenchmarkEncode/Proposed/Metric/Summary-8      	     141	  42172658 ns/op

BenchmarkEncode/Baseline/Metric/Histogram-8    	      91	  64816532 ns/op
BenchmarkEncode/Proposed/Metric/Histogram-8    	     100	  56837230 ns/op

BenchmarkEncode/Baseline/Metric/HistogramSeries-8         	      37	 159643499 ns/op
BenchmarkEncode/Proposed/Metric/HistogramSeries-8         	      40	 149983727 ns/op

BenchmarkEncode/Baseline/Metric/Mix-8                     	      20	 287582195 ns/op
BenchmarkEncode/Proposed/Metric/Mix-8                     	      31	 196378683 ns/op

BenchmarkEncode/Baseline/Metric/MixSeries-8               	       8	 681122815 ns/op
BenchmarkEncode/Proposed/Metric/MixSeries-8               	      10	 540018386 ns/op

BenchmarkDecode/Baseline/Metric/Int64-8                   	      18	 319038450 ns/op	206696040 B/op	 5724000 allocs/op
BenchmarkDecode/Proposed/Metric/Int64-8                   	      22	 265034276 ns/op	154696039 B/op	 4724000 allocs/op

BenchmarkDecode/Baseline/Metric/Summary-8                 	      52	 117972148 ns/op	79496034 B/op	 2024000 allocs/op
BenchmarkDecode/Proposed/Metric/Summary-8                 	      60	 104489249 ns/op	69096035 B/op	 1824000 allocs/op

BenchmarkDecode/Baseline/Metric/Histogram-8               	      36	 161440606 ns/op	104296028 B/op	 2624000 allocs/op
BenchmarkDecode/Proposed/Metric/Histogram-8               	      40	 147001580 ns/op	93896032 B/op	 2424000 allocs/op

BenchmarkDecode/Baseline/Metric/HistogramSeries-8         	      16	 345726946 ns/op	233896053 B/op	 5324000 allocs/op
BenchmarkDecode/Proposed/Metric/HistogramSeries-8         	      16	 333473715 ns/op	223496045 B/op	 5124000 allocs/op

BenchmarkDecode/Baseline/Metric/Mix-8                     	       9	 611361046 ns/op	391240035 B/op	10326000 allocs/op
BenchmarkDecode/Proposed/Metric/Mix-8                     	      10	 531026674 ns/op	318440033 B/op	 8926000 allocs/op

BenchmarkDecode/Baseline/Metric/MixSeries-8               	       5	1179081837 ns/op	776840057 B/op	18026000 allocs/op
BenchmarkDecode/Proposed/Metric/MixSeries-8               	       5	1077597483 ns/op	704040035 B/op	16626000 allocs/op
```

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -160,20 +160,24 @@ message MetricDescriptor {
 message Int64TimeSeries {
   // The set of label values that uniquely identify this timeseries. Applies to
   // all points. The order of label values must match that of label keys in the
-  // metric descriptor.
-  repeated LabelValue label_values = 1;
+  // metric descriptor. Each element in this array is the value of the key defined at the
+  // corresponding index of MetricDescriptor.label_keys array. The number of elements
+  // in this array and MetricDescriptor.label_keys array must be the same.
+  repeated string label_values = 1;
 
   // The data points of this timeseries.
   repeated Int64Value points = 2;
 }
 
-// eDoubleTimeSeries is a list of data points that describes the time-varying values
+// DoubleTimeSeries is a list of data points that describes the time-varying values
 // of a double metric.
 message DoubleTimeSeries {
   // The set of label values that uniquely identify this timeseries. Applies to
   // all points. The order of label values must match that of label keys in the
-  // metric descriptor.
-  repeated LabelValue label_values = 1;
+  // metric descriptor. Each element in this array is the value of the key defined at the
+  // corresponding index of MetricDescriptor.label_keys array. The number of elements
+  // in this array and MetricDescriptor.label_keys array must be the same.
+  repeated string label_values = 1;
 
   // The data points of this timeseries.
   repeated DoubleValue points = 2;
@@ -184,8 +188,10 @@ message DoubleTimeSeries {
 message HistogramTimeSeries {
   // The set of label values that uniquely identify this timeseries. Applies to
   // all points. The order of label values must match that of label keys in the
-  // metric descriptor.
-  repeated LabelValue label_values = 1;
+  // metric descriptor. Each element in this array is the value of the key defined at the
+  // corresponding index of MetricDescriptor.label_keys array. The number of elements
+  // in this array and MetricDescriptor.label_keys array must be the same.
+  repeated string label_values = 1;
 
   // The data points of this timeseries.
   repeated HistogramValue points = 2;
@@ -227,20 +233,13 @@ message HistogramTimeSeries {
 message SummaryTimeSeries {
   // The set of label values that uniquely identify this timeseries. Applies to
   // all points. The order of label values must match that of label keys in the
-  // metric descriptor.
-  repeated LabelValue label_values = 1;
+  // metric descriptor. Each element in this array is the value of the key defined at the
+  // corresponding index of MetricDescriptor.label_keys array. The number of elements
+  // in this array and MetricDescriptor.label_keys array must be the same.
+  repeated string label_values = 1;
 
   // The data points of this timeseries.
   repeated SummaryValue points = 2;
-}
-
-// Label value or indication that the value is missing
-message LabelValue {
-  // The value for the label.
-  string value = 1;
-  // If false the value field is ignored and considered not set.
-  // This is used to differentiate a missing label from an empty string.
-  bool has_value = 2;
 }
 
 // Int64Value is a timestamped measurement of int64 value.


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -146,20 +176,12 @@ message SummaryTimeSeries {
...
   // The data points of this timeseries.
   repeated SummaryValue points = 2;
 }
 
-message LabelValue {
-  // The value for the label.
-  string value = 1;
-  // If false the value field is ignored and considered not set.
-  // This is used to differentiate a missing label from an empty string.
-  bool has_value = 2;
```
SergeyKanzhelev: do we have to differentiate missing value from empty string at a first place? How important is it?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -104,13 +104,43 @@ message MetricDescriptor {
...
 }
 
+// LabelValues is a list of label string values.
+message LabelValues {
+  // Values of labels. Each element in this array is the value of the key defined at the
+  // corresponding index of MetricDescriptor.label_keys array. The number of elements
+  // in these 2 arrays must be the same. Note that values array must include empty
+  // strings and must also contain an empty string value for labels for which the
+  // value was originally unspecified.
+  repeated string values = 1;
```
SergeyKanzhelev: I'd say aggregation of metrics would be challenging as there will be a need to track a separate array of unspecified values. Is it a concern?

Looking at API - we may end up with various label set quite often as LabelSet is not pre-created when metric is registered. @jmacd do you think it will be a common case of missing labels?

tigrannajaryan: I don't think that's a problem. When exporting we will need to Go over internal metric representation and convert in-memory LabelSet to LabelValues proto message. Should not be an issue to build the message. unspecified values don't really make this materially more difficult. It will be a straightforward array iteration and appending of indexes.
Alternatively this can be done once per LabelSet and cached.

opentelemetry/proto/metrics/v1/metrics.proto
tigrannajaryan: This is explicitly required in the API: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-metrics-user.md#missing-label-keys

>When the SDK interprets a LabelSet in the context of grouping aggregated values for an exporter, and where there are keys that are missing, the SDK is required to consider these values explicitly unspecified, a distinct value type of the exported data model.

Another way this could be modeled in `Metric` data structure is to require every metric timeseries that has a different set of labels with defined values to be stored in a separate instance of `Metric` which would point to `MetricDescriptor` containing label_keys only for defined labels. Hard to tell which approach would be more efficient, it all depends on statistical composition of data.

For now in this change I tried not to change existing functionality but simply modeled it in a slightly more efficient way.

opentelemetry/proto/metrics/v1/metrics.proto
SergeyKanzhelev: API declares metric and then on every measurement recording you can pass whatever LabelSet you want. If aggregation logic will attempt to merge different LabelSet's - there potentially be many missing dimensions quite often. If LabelSet will not be merged and data will be aggregated separately - we do not need a missing labels functionality

jmacd: I'm not sure what it means to "merge" label sets, as nothing like that was included in the specification. There has been _speculation_ that people will demand more of label sets, such as a way to extend them (e.g., `labelSet.Add(oneMoreLabel)`, or `labelSet.Combine(more, label, sets)`).

LabelSets contain arbitrary labels and are not bound to an instrument. Since instruments may be aggregated along different dimensions, it's certainly possible to have unspecified values. I think it's prudent to preserve this information -- otherwise there will be ambiguity.

In the currently open WIP PR adding a stdout exporter in Go (https://github.com/open-telemetry/opentelemetry-go/pull/265), I left a TODO about this: https://github.com/open-telemetry/opentelemetry-go/blob/63d79a833e3312b89b2de9fcfbe744561fe70563/sdk/metric/batcher/stateful/stateful.go#L76

SergeyKanzhelev:
> We declared that no, the value is explicitly unspecified if the LabelSet/Handle did not declare it.

From consumption standpoint - what's the difference of non-specified and empty value? Would one want to create a stacked chart where unspecified label and empty label are two different values? If empty is that important for this scenario - perhaps one can report it with the actual word "EMPTY"?

If there are some notes I can read thru to understand it - please let me know

jmacd:
If we specified that empty strings were invalid values, then we could get away with using the empty string to encode an unspecified value. I have mixed feelings here, because it's so much easier to use an empty string than to distinguish unspecified values. But then some users, who have legitimate empty string values will be forced to use "EMPTY", which is also a valid value, and then they're left with ambiguity.

tigrannajaryan:
I feel empty/unspecified string is a discussion that likely needs to happen in the realm of Specification. If there is a decision not to support unspecified or empty strings we can then modify the proto accordingly. For now the proto and this PR mirrors what the Spec requires.

tigrannajaryan:
@SergeyKanzhelev @jmacd can we move forward with this change? It does not introduce new concepts, but merely follows what spec requires today.

I created a separate issue to clarify the need for "unspecified" lablels here: https://github.com/open-telemetry/opentelemetry-specification/issues/345

If the conclusion eventually is that we don't need the unspecified labels I will happily modify this proto to remove them in a future PR. For now I believe this PR is good as is.

jmacd:
In hindsight, I'm not sure we should have written that into the specification. It was related to the concept of _required_ aggregation keys, which was softened in the final spec to _recommended_ aggregation keys. If a system is expecting precisely a fixed number of dimensions and those dimensions are not included in the export format, they are effectively unspecified. 

The reason we discussed explicitly unspecified values was to declare that the implementation would not be required to scan through the context for distributed correlation labels to satisfy the missing values. 

@bogdandrutu I'd be happy to strike this from the specification.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -155,25 +155,55 @@ message MetricDescriptor {
...
+
+  // Some metric sources allow setting the value of a label to be equal not just to
+  // empty string but to an unspecified value, or in other words the value of a particular
+  // label defined in label_keys array may be missing in a particular timeseries.
+  // In order to support this concept and because ProtoBuf cannot distinguish empty string
+  // from unspecified string we store indexes of unspecified labels in "values" array in
+  // this field. This field contains only indexes of unspecified labels.
+  // This field should not contain duplicate numbers, numbers less than zero or
+  // numbers greater than or equal to the length of values array.
+  repeated int32 unspecified_value_indexes = 2;
```
SergeyKanzhelev: should the array be ordered to simplify merging?

tigrannajaryan:
> I'm fine with this change, but declaring empty strings and unspecified values to be the same would be a better optimization and easier to understand

@SergeyKanzhelev I agree and if spec decides to do that I will make the corresponding change here.

jmacd:
@SergeyKanzhelev I didn't say empty string--that I would reject. I said leaving out a key:value entirely (i.e., no key, no value) is effectively saying unspecified.

opentelemetry/proto/metrics/v1/metrics.proto
tigrannajaryan: I am trying to think of a usecase when such merging is needed for the proto data and fail to find one. This data typically either gets generated in client libraries based on in-memory representation of the metric as-is or is received by backends and stored in a way that is likely decoupled from this proto representation completely. So, basically this array is a likely a write-once, read-once, immutable data.

Even if there is an obscure use case this array is likely to be mostly empty or contain at most a few elements. Merging them is going to be fast performance-wise even if it is unsorted.

tigrannajaryan:
> @SergeyKanzhelev I didn't say empty string--that I would reject. I said leaving out a key:value entirely (i.e., no key, no value) is effectively saying unspecified.

And that's exactly possible to do now, by omitting the key from `MetricDescriptor.label_keys`. The only downside to that is that when exporting metrics you will have to create distinct instances of `MetricDescriptor` for each combination of specified labels (as opposed to have a single `MetricDescriptor` for all instances of one metric that have different label values). It is doable and my gut feeling it is not a big deal since it is going to be rare.

jmacd:
I could imagine a solution where instead of having label values be a repeated strings, it would be repeated integers referring to a dictionary of string values. This would support compression -- where the same string is used more than once, and it would allow us to have an explicit value like -1 to indicate unspecified.

tigrannajaryan:
How common is it to have labels with repeated values?

Also compression-wise if the concern is about wire size the proto most likely is going to be gziped and that compresses repetitive strings quite efficiently. Performance-wise this elimination of redundant labels could probably be a win, but for performance I am looking into an approach that looks promising, so I'd leave that aside for a moment.

Compressing the way you suggested is possible for other entities as well but so far I avoided going in that direction because i think it makes working with data slightly more complicated since you need to use an additional indirection (the way you suggested). It may be worth exploring, I can do some experimentation later if time permits.

jmacd:
I am starting to agree that we should allow empty strings. Let's do that and see what trouble ensues. Should we specify that the empty string is not a valid attribute/label value?

SergeyKanzhelev:
@jmacd giving I'm not very deep into metrics - I like the idea to try equating empty strings with unspecified. If there will be legitimate use cases for an opposite - we can always add an indexes array with the indexes saying which empty strings are empty and which are unspecified. Similar to what's proposed in this PR.

SergeyKanzhelev:
Again, pardon my ignorance... I'm trying to get involved, but have too much on my plate now.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: OpenMetrics does not make this separation. Not 100% sure if it is needed or not to separate between empty string and not present.

bogdandrutu:
Hard to follow this discussion. Can we summarize what is the current status? I personally prefer to first decide if we need to support unspecified or we can do what OpenMetrics does and treat unspecified as empty.

tigrannajaryan:
> Hard to follow this discussion. Can we summarize what is the current status? I personally prefer to first decide if we need to support unspecified or we can do what OpenMetrics does and treat unspecified as empty.

@bogdandrutu to summarize: I believe we can move forward and merge this PR since it is a straight improvement over current state and does not change any semantics, it is simply a performance optimization without functional changes.

We can separately discuss if we want to eliminate "unspecified label" concept as part of https://github.com/open-telemetry/opentelemetry-specification/issues/345 that I created for that discussion.

Depending on the decision resulting from that discussion I will make any necessary changes in this proto repository.

SergeyKanzhelev:
@tigrannajaryan I think there is a difference - unspecified values on API and unspecified values on schema (it is my take). It seems like there is little objection to have "empty mean unspecified" on schema. So this PR can be improved further

tigrannajaryan:
@SergeyKanzhelev @jmacd I removed the ability to have unspecified value from this PR. Please have another look.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -155,25 +155,33 @@ message MetricDescriptor {
   repeated string label_keys = 5;
 }
 
+// LabelValues is a list of label string values.
+message LabelValues {
```
bogdandrutu: Do you need this extra message anymore? You can directly use repeated string

SergeyKanzhelev:
@jmacd I'm merging this just to move forward. If you feel that we need to revisit this decision - please file an issue.

jmacd:
LGTM

PR #37, opened 2019-11-02
Optimize HistogramTimeseries and HistogramValue for speed

tigrannajaryan:
Optimize HistogramTimeseries and HistogramValue for speed

This change removes unnecessary nesting from BucketOptions and moves them
from HistogramValue to HistogramTimeseries. The later change allows multiple
points in the timeseries to use the same bucket options and significantly
improves performance in that case.

Benchmark in Go demonstrates the following improvement of encoding and decoding
compared to the baseline state:

```
===== Encoded sizes
Baseline/Metric/Histogram       13769 bytes  [1.000], gziped  769 bytes  [1.000]
Proposed/Metric/Histogram       13569 bytes  [1.015], gziped  774 bytes  [0.994]

Baseline/Metric/MixSeries       109867 bytes  [1.000], gziped 7006 bytes  [1.000]
Proposed/Metric/MixSeries       97867 bytes  [1.123], gziped 6620 bytes  [1.058]

BenchmarkEncode/Baseline/Metric/HistogramOne-8 	      61	  82956741 ns/op
BenchmarkEncode/Proposed/Metric/HistogramOne-8 	      93	  67504571 ns/op

BenchmarkEncode/Baseline/Metric/HistogramSeries-8         	      20	 278817032 ns/op
BenchmarkEncode/Proposed/Metric/HistogramSeries-8         	      36	 167044602 ns/op

BenchmarkDecode/Baseline/Metric/HistogramOne-8            	      31	 193549762 ns/op	106696051 B/op	 2824000 allocs/op
BenchmarkDecode/Proposed/Metric/HistogramOne-8            	      36	 166249091 ns/op	104296031 B/op	 2624000 allocs/op

BenchmarkDecode/Baseline/Metric/HistogramSeries-8         	       9	 592067126 ns/op	287496035 B/op	 7524000 allocs/op
BenchmarkDecode/Proposed/Metric/HistogramSeries-8         	      15	 374523053 ns/op	233896049 B/op	 5324000 allocs/op
```

For single histogram value timeseries it is about 16-18% faster, for timeseries composed
of 5 histogram values it is about 36-40% faster.

https://github.com/tigrannajaryan/exp-otelproto/blob/4c8771fac0543fd8e8d0892a16dcfeb1be43c3aa/encodings/encoding_test.go

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -183,6 +183,7 @@ message SummaryTimeSeries {
   repeated SummaryValue points = 2;
 }
 
+// Label value or indication that the value is missing
 message LabelValue {
   // The value for the label.
   string value = 1;
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -314,6 +314,7 @@ message HistogramValue {
 }
 
 // StringKeyValuePair is a pair of key/value strings.
+// TODO: consider unifying this with Resource and Span Attribute key/value pairs.
 message StringKeyValuePair {
   string key = 1;
   string value = 2;
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -138,6 +138,41 @@ message HistogramTimeSeries {
 
   // The data points of this timeseries.
   repeated HistogramValue points = 2;
+
+  // ExplicitBounds specifies buckets with explicitly defined bounds for values.
+  message ExplicitBounds {
+    // The bucket boundaries are described by "bounds" field.
+    //
+    // This defines size(bounds) + 1 (= N) buckets. The boundaries for bucket
+    // at index i are:
+    //
+    // [0, bounds[i]) for i == 0
+    // [bounds[i-1], bounds[i]) for 0 < i < N-1
+    // [bounds[i], +infinity) for i == N-1
+    // The values in bounds array must be strictly increasing and > 0.
+    //
+    // Note: only [a, b) intervals are currently supported for each bucket. If we decides
+    // to also support (a, b] intervals we should add support for these by defining a boolean
+    // value which decides what type of intervals to use.
+    repeated double bounds = 1;
+  }
+
+  // A histogram may optionally contain the distribution of the values in the population.
+  // In that case "bucket_options" field below and "buckets" field in HistogramValue
+  // both must be defined. Otherwise both fields must be omitted in which case the
+  // distribution of values in the histogram is unknown and only the total count and sum
+  // are known.
+  //
+  // bucket_options apply to all points in this HistogramTimeSeries. To define
+  // different bucket_options for different points create separate instances of
+  // HistogramTimeSeries each with its own bucket_options.
+  oneof bucket_options {
+    // explicit_bounds is the only supported option currently.
+    ExplicitBounds explicit_bounds = 3;
+
+    // Other options such linear buckets or exponential buckets may be added below
+    // in the future.
+  }
 }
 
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -226,79 +258,69 @@ message HistogramValue {
   // compact encoding on the wire.
   // If the value of 0 occurs for the first data point in the timeseries it means that
   // the timestamp is unspecified. In that case the timestamp may be decided by the backend.
+  // Note: this field is always unspecified and ignored if MetricDescriptor.type==GAUGE_HISTOGRAM.
   sfixed64 start_time_unixnano = 1;
```
bogdandrutu: Nit: I am not sure if I know of an usecase where this is not the same for all points, just to document this.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -226,79 +258,69 @@ message HistogramValue {
...
+  // number of elements in bucket_bounds array.
+  //
+  // Note: if HistogramTimeSeries.bucket_options defines bucket bounds then this field
+  // must also be present and number of elements in this field must be equal to the
+  // number of buckets defined by bucket_options.
+  repeated Bucket buckets = 5;
+}
+
+// StringKeyValuePair is a pair of key/value strings.
+message StringKeyValuePair {
```
bogdandrutu: Add a todo to look to unify this with label/resource etc.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -138,6 +138,41 @@ message HistogramTimeSeries {
...
+  // A histogram may optionally contain the distribution of the values in the population.
+  // In that case "bucket_options" field below and "buckets" field in HistogramValue
+  // both must be defined. Otherwise both fields must be omitted in which case the
+  // distribution of values in the histogram is unknown and only the total count and sum
+  // are known.
+  //
+  // bucket_options apply to all points in this HistogramTimeSeries. To define
+  // different bucket_options for different points create separate instances of
+  // HistogramTimeSeries each with its own bucket_options.
+  oneof bucket_options {
```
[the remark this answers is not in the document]
tigrannajaryan: Yes, that's a good idea unless we think we may have a large number of different bucket options (which would result in memoery size increase). I think we are likely only going to add Linear and Exponential. In that case oneof can be eliminated and we will save CPU time. Let me do a benchmark and post results.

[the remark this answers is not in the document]
tigrannajaryan: Benchmarked without oneof and 3 bucket options (with only one set). Memory consumption is virtually identical as with oneof but we gain a few more percent of CPU speed.

With single bucket option (onyl explicit_bounds, like we have now) the oneof version is even worse.

I will eliminate oneof since it is a clear improvement.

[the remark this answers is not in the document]
tigrannajaryan: I removed oneof and updated PR description with latest benchmark results.

opentelemetry/proto/metrics/v1/metrics.proto

opentelemetry/proto/metrics/v1/metrics.proto
tigrannajaryan: Good to know. If that is the case then perhaps this can be moved to HistogramTimeSeries. The same change can be done for all other value types, so let's consider it in a separate PR. I added an issue https://github.com/open-telemetry/opentelemetry-proto/issues/38

bogdandrutu: FYI: We need some more investigations not just benchmarks for that change.

SergeyKanzhelev:
Was active for a business day and non-controversial... will try to merge now.

*I will now close/re-open so CI status will get updated*

PR #40, opened 2019-11-06
Eliminate Snapshot message from SummaryValue

tigrannajaryan:
This change removes unnecessary Snapshot message and moves percentile
values to SummaryValue. Also removed count and sum from Snapshot
because there are no known use-cases for these.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -402,33 +402,17 @@ message SummaryValue {
   // This field must be unset if the sum is not available.
   google.protobuf.DoubleValue sum = 4;
 
-  // The values in this message can be reset at arbitrary unknown times, with
-  // the requirement that all of them are reset at the same time.
-  message Snapshot {
-    // The number of values in the snapshot. Optional since some systems don't
-    // expose this.
-    google.protobuf.Int64Value count = 1;
+  // Represents the value at a given percentile of a distribution.
+  message ValueAtPercentile {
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -395,40 +394,27 @@ message SummaryValue {
...
-      double percentile = 1;
-
-      // The value at the given percentile of a distribution.
-      double value = 2;
-    }
+  double sum = 4;
+
+  // count_and_sum_unspecified must be set to true if count and sum values are unknown or
+  // unspecified.
+  bool count_and_sum_unspecified = 5;
```
bogdandrutu: Please split this into 2:
1. Removes the Snapshot -> makes it closer to openmetrics
2. Remove usage of the wrapped values.

tigrannajaryan: Done. This is part 1.

bogdandrutu:
Please resync

tigrannajaryan:
> Please resync

Done.

bogdandrutu:
I will wait for one more approve before merging.

bogdandrutu:
Please mention that this is compatible with OM

tigrannajaryan:
> Please mention that this is compatible with OM

OpenMetrics proto exists only as a PR. It is not merged into their master branch. Is it useful to refer to a PR that may not end up being merged? By the time it is merged it may be changed again and our claim to be compatible with OM may be wrong.

PR #41, opened 2019-11-07
Replace attributes map by AttributeKeyValue pair list

tigrannajaryan:
This is the first of commits that implements RFC0059 [1]. I will submit changes introduced
in the RFC in separate PRs for easier review.

This change is already benchmarked as part of the RFC, so I will not submit individual benchmarks
with each PR.

Note: this change has been thoroughly reviewed, commented and discussed as part of the RFC
process. This commit merely implements part of the RFC.

[1] - https://github.com/open-telemetry/oteps/blob/master/text/0059-otlp-trace-data-format.md

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -159,27 +159,20 @@ message Span {
   // This field is required.
   google.protobuf.Timestamp end_time = 9;
 
-  // A set of attributes, each with a key and a value.
-  message Attributes {
-    // The set of attributes. The value can be a string, an integer, a double
-    // or the Boolean values `true` or `false`. Note, global attributes like
-    // server name can be set as tags using resource API. Examples of attributes:
-    //
-    //     "/http/user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36"
-    //     "/http/server_latency": 300
-    //     "abc.com/myattribute": true
-    //     "abc.com/score": 10.239
-    map<string, AttributeValue> attribute_map = 1;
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -244,16 +245,42 @@ message Span {
...
   // An optional number of child spans that were generated while this span
   // was active. If set, allows an implementation to detect missing child spans.
-  google.protobuf.UInt32Value child_span_count = 14;
+  google.protobuf.UInt32Value child_span_count = 15;
+}
+
+// AttributeKeyValue is a key-value pair that is used to store Span attributes, Link
+// attributes, etc.
+message AttributeKeyValue {
+  // ValueType is the enumeration of possible types that value can have.
```
SergeyKanzhelev: I like that STRING is zero. Perhaps mention that if omitted - `string_value` will be used

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -244,16 +245,42 @@ message Span {
...
+  enum ValueType {
+    STRING  = 0;
+    INT64   = 1;
```
SergeyKanzhelev: nit: Do we need the 64 specifier? Before it was simply int.

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: That's the default behavior of protobufs. It is not possible to distinguish omitted from 0 value on the receiving side and the sender automatically omits zero values when encoding, so you don't really need to do anything special about this, it will just work automatically like that (at least in languages which auto-initialize structs to zero values, like Go).
Do you think it is worth a comment?

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: I agree. Changed to simply `INT`.

PR #42, opened 2019-11-07
Use unary gRPC for export requests

tigrannajaryan:
This implements RFC0035 requirement to use unary gRPC:
https://github.com/open-telemetry/oteps/blob/master/text/0035-opentelemetry-protocol.md#otlp-over-grpc


yurishkuro:
This doesn't seem controversial given that OTEP was approved, so merging.

PR #45, opened 2019-11-08
Change timestamp data type to sfixed64 nanoseconds since Unix epoch

tigrannajaryan:
The change to int64 is proposed by RFC0059 [1]. In this commit there is a small deviation from
the RFC: here we use sfixed64 instead of int64 to make it consistent with changes already
done for Metrics proto [2]. Experiments proved sfixed64 to be more suitable for timestamps.

[1] https://github.com/open-telemetry/oteps/blob/master/text/0059-otlp-trace-data-format.md
[2] https://github.com/open-telemetry/opentelemetry-proto/pull/33

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -16,7 +16,6 @@ syntax = "proto3";
 
 package opentelemetry.proto.trace.v1;
 
-import "google/protobuf/timestamp.proto";
 import "google/protobuf/wrappers.proto";
 import "opentelemetry/proto/resource/v1/resource.proto";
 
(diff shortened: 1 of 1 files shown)


yurishkuro:
> Experiments proved sfixed64 to be more suitable for timestamps.

more suitable than `fixed64` too? signed value seems odd for the epoch-style timestamp.

tigrannajaryan:
> > Experiments proved sfixed64 to be more suitable for timestamps.
> 
> more suitable than `fixed64` too? signed value seems odd for the epoch-style timestamp.

`fixed64` should be slightly better (double the range in the future), but we already use `sfixed64` in metrics. Fro consistency it is best to use the same type for timestamp everywhere. I can update in all places (not sure if we want it in this PR or separate).

yurishkuro:
If there's no perf benefit of sfixed64, I would go for fixed64, which is unsigned and makes more sense for epoch-style timestamps. Otherwise we'll be answering this question over and over.

bogdandrutu:
I agree we can change metrics too to use fixed64

SergeyKanzhelev:
merging this. @tigrannajaryan can you please send separate PR for switch to unsigned everywhere

PR #47, opened 2019-11-11
Implement Result Code as required by OTLP

tigrannajaryan:
This change implements OTLP requirement of Result Code. OTLP describes
how it is used in the export responses:
https://github.com/open-telemetry/oteps/blob/master/text/0035-opentelemetry-protocol.md#result-code

### DIFF
--- a/opentelemetry/proto/agent/common/v1/common.proto
+++ b/opentelemetry/proto/agent/common/v1/common.proto
@@ -95,3 +95,20 @@ message ServiceInfo {
 
   // TODO(songya): add more fields as needed.
 }
+
+// ExportResultCode is the result code returned in the export responses.
+enum ExportResultCode {
+  // Telemetry data is successfully processed by the server.
+  Success = 0;
+
+  // processing of telemetry data failed. The client MUST NOT retry
+  // sending the same telemetry data. The telemetry data MUST be dropped.
+  // This for example can happen when the request contains bad data and
+  // cannot be deserialized or otherwise processed by the server.
+  FailedNotRetryable = 1;
+
+  // Processing of telemetry data failed. The client SHOULD record the
+  // error and may retry exporting the same data after some time. This
+  // for example can happen when the server is overloaded.
+  FailedRetryable = 2;
+}
--- a/opentelemetry/proto/agent/metrics/v1/metrics_service.proto
+++ b/opentelemetry/proto/agent/metrics/v1/metrics_service.proto
@@ -51,4 +51,6 @@ message ExportMetricsServiceRequest {
 }
 
 message ExportMetricsServiceResponse {
+  // result_code indicates the result of processing the export request.
+  opentelemetry.proto.agent.common.v1.ExportResultCode result_code = 1;
 }
--- a/opentelemetry/proto/agent/trace/v1/trace_service.proto
+++ b/opentelemetry/proto/agent/trace/v1/trace_service.proto
@@ -80,4 +80,6 @@ message ExportTraceServiceRequest {
 }
 
 message ExportTraceServiceResponse {
+  // result_code indicates the result of processing the export request.
+  opentelemetry.proto.agent.common.v1.ExportResultCode result_code = 1;
 }


opentelemetry/proto/agent/common/v1/common.proto
```diff
@@ -95,3 +95,20 @@ message ServiceInfo {
...
+// ExportResultCode is the result code returned in the export responses.
+enum ExportResultCode {
+  // Telemetry data is successfully processed by the server.
+  Success = 0;
+
+  // processing of telemetry data failed. The client MUST NOT retry
+  // sending the same telemetry data. The telemetry data MUST be dropped.
+  // This for example can happen when the request contains bad data and
+  // cannot be deserialized or otherwise processed by the server.
+  FailedNoneRetryable = 1;
```
yurishkuro: did you mean NotRetryable?

tigrannajaryan: This typo keep sneaking in. Fixing.

opentelemetry/proto/agent/common/v1/common.proto
```diff
@@ -95,3 +95,20 @@ message ServiceInfo {
 
   // TODO(songya): add more fields as needed.
 }
+
+// ExportResultCode is the result code returned in the export responses.
+enum ExportResultCode {
```
bogdandrutu: I remember you told me that this is controlled by the status we return from grpc not by the message.

tigrannajaryan: grpc status is used for throttling purposes: https://github.com/open-telemetry/oteps/blob/master/text/0035-opentelemetry-protocol.md#throttling

ResultCode is different and is an explicit field in the response.

The reason I am using grpc code for throttling is that there is already clearly defined grpc code for this (Unavailable). As opposed to that there is no grpc code that allows to distinguish between FailedRetryable and FailedNotRetryable cases and I don't want to artificially model them using other grpc codes.

bogdandrutu: But from the Status you can determine if a retry is possible or not correct?

tigrannajaryan: Is there a clearly defined condition / set of status codes that allow client to make a decision about retrying or not retrying based just the grpc status code? If there is such a way then we can eliminate ResultCode.

At the moment I am not sure how a client should make such decision by looking at the returned gRPC code. For example if gRPC Status is `Unknown` is it retryable or no?

Perhaps instead of ResultCode we list here the status codes which must be interpreted by the client as Retryable and then make a recommendation for server to return specific errors codes to Retryable and NonRetryable cases? That could work.

bogdandrutu: FYI: So you always get a status, but I think you never get a response + an error status (if I remember correctly).

So the logic I would use retry if:
* If the status contains the RetryInfo is a retryable 
* For some status codes without a RetryInfo (deadline exceeded, need to determine the full list) you may do retry.

tigrannajaryan: >FYI: So you always get a status, but I think you never get a response + an error status (if I remember correctly).

Yes, that is correct. A response only comes when status is success. Any other status is received without a response.

>If the status contains the RetryInfo is a retryable

So you are saying RetryInfo with RetryDelay=0 would be equivalent to ResultCode=FailedRetryable. The same RetryInfo with RetryDelay>0 would be throttling indication (also retryable obviously, but after a delay).

>For some status codes without a RetryInfo (deadline exceeded, need to determine the full list) you may do retry.

Make sense. Also, we need to decide which status code to return to indicate NotRetryable. Let me think though and submit an update.

tigrannajaryan: I created https://github.com/open-telemetry/oteps/pull/65
If it gets accepted I will discard this PR.

tigrannajaryan:
Reviewers, please have a look at https://github.com/open-telemetry/oteps/pull/65 instead. This uses a different approach that Bogdan proposed and which I think is better.

SergeyKanzhelev:
> Reviewers, please have a look at open-telemetry/oteps#65 instead. This uses a different approach that Bogdan proposed and which I think is better.

I'm assuming this have to be closed? I'm closing, please re-open if needed

PR #48, opened 2019-11-11
Move StringKeyValue and AttributeKeyValue to a common.proto

tigrannajaryan:
Moved StringKeyValue and AttributeKeyValue into a common.proto file so that
they can be used from other proto files if necessary.

### DIFF
--- a/opentelemetry/proto/common/v1/common.proto
+++ b/opentelemetry/proto/common/v1/common.proto
@@ -1,0 +1,54 @@
+// Copyright 2019, OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package opentelemetry.proto.common.v1;
+
+option java_multiple_files = true;
+option java_package = "io.opentelemetry.proto.common.v1";
+option java_outer_classname = "CommonProto";
+
+// AttributeKeyValue is a key-value pair that is used to store Span attributes, Link
+// attributes, etc.
+message AttributeKeyValue {
+  // ValueType is the enumeration of possible types that value can have.
+  enum ValueType {
+    STRING  = 0;
+    INT   = 1;
+    DOUBLE  = 2;
+    BOOL    = 3;
+  };
+
+  // key part of the key-value pair.
+  string key = 1;
+
+  // type of the value.
+  ValueType type = 2;
+
+  // Only one of the following fields is supposed to contain data (determined by `type` field).
+  // This is deliberately not using Protobuf `oneof` for performance reasons (verified by benchmarks).
+
+  string string_value = 3;
+  int64 int_value = 4;
+  double double_value = 5;
+  bool bool_value = 6;
+}
+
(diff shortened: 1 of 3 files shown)


opentelemetry/proto/resource/v1/resource.proto
```diff
@@ -16,12 +16,20 @@ syntax = "proto3";
...
 option java_outer_classname = "ResourceProto";
 
 // Resource information.
 message Resource {
-  // Set of labels that describe the resource.
-  map<string,string> labels = 1;
+  // labels is a collection of attributes that describe the resource. See OpenTelemetry
+  // specification semantic conventions for standardized label names:
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-resource-semantic-conventions.md
+  repeated opentelemetry.proto.common.v1.AttributeKeyValue labels = 1;
```
yurishkuro: https://github.com/open-telemetry/opentelemetry-proto/issues/50

opentelemetry/proto/resource/v1/resource.proto
```diff
@@ -16,12 +16,20 @@ syntax = "proto3";
...
+
+  // dropped_labels_count is the number of dropped labels. If the value is 0, then
+  // no labels were dropped.
+  int32 dropped_labels_count = 2;
```
bogdandrutu: I don't think we have yet define the capability to drop labels from resource.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -343,7 +344,7 @@ message HistogramValue {
 
       // exemplar_attachments are contextual information about the example value.
       // Keys in this list must be unique.
-      repeated StringKeyValuePair attachments = 3;
+      repeated opentelemetry.proto.common.v1.AttributeKeyValue attachments = 3;
```
bogdandrutu: this changes the data model, do we have an agreement that we want to support labels that are not string? This increases the size of the messages (in-memory) without a real benefit (unless you prove me that there are some important systems that support multiple types for labels).

opentelemetry/proto/resource/v1/resource.proto
tigrannajaryan: Intermediary nodes can drop labels to save bandwidth, etc. It does not necessarily have to be part of OpenTelemetry API or SDK.

SergeyKanzhelev:
@bogdandrutu can you please add a review status "request changes" if you don't agree with the change or need clarification. It will simplify maintainence as the status will be seen from the list of PRs

opentelemetry/proto/metrics/v1/metrics.proto
tigrannajaryan: My understanding is that this can be any set of attributes that you want to attach to a measurement that can be later explored in the backend UI. I think non-string values can be useful here, just like they are for e.g. Span attributes. It seems to be the same use case, providing additional information to understand what operation is represented (and exemplar is a representation of an operation, isn't it?).

opentelemetry/proto/common/v1/common.proto
```diff
@@ -0,0 +1,47 @@
...
+  string key = 1;
+
+  // type of the value.
+  ValueType type = 2;
+
+  // Only one of the following fields is supposed to contain data (determined by `type` field).
+  // This is deliberately not using Protobuf `oneof` for performance reasons (verified by benchmarks).
+
+  string string_value = 3;
+  int64 int_value = 4;
```
tylerbenson: (caveat: I'm not a proto expert) Any significant downside to using a `long` type instead of int?

tigrannajaryan: ProtoBufs don't have a `long` type: https://developers.google.com/protocol-buffers/docs/proto3#scalar (or you can call int64 a long depending on how you look at it).

tylerbenson: oh, sorry I missed the int64 already defined there.  carry on!

tigrannajaryan:
@bogdandrutu done, PTAL.

PR #49, opened 2019-11-11
Remove unnecessary wrapping of int32

tigrannajaryan:
Implement change approved in https://github.com/open-telemetry/oteps/blob/master/text/0059-otlp-trace-data-format.md

Minor deviation from the approved change: we use uint32 instead of int32
since child count cannot be a negative number and unsigned number is
semantically a better fit.

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -245,9 +245,9 @@ message Span {
   // Status.Ok (code = 0).
   Status status = 14;
 
-  // An optional number of child spans that were generated while this span
-  // was active. If set, allows an implementation to detect missing child spans.
-  google.protobuf.UInt32Value child_span_count = 15;
+  // child_span_count is an optional number of local child spans that were generated while this
+  // span was active. If set, allows an implementation to detect missing child spans.
+  uint32 child_span_count = 15;
 }
 
 // AttributeKeyValue is a key-value pair that is used to store Span attributes, Link


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -245,9 +245,9 @@ message Span {
   // Status.Ok (code = 0).
   Status status = 14;
 
-  // An optional number of child spans that were generated while this span
-  // was active. If set, allows an implementation to detect missing child spans.
-  google.protobuf.UInt32Value child_span_count = 15;
+  // child_span_count is an optional number of local child spans that were generated while this
+  // span was active. If set, allows an implementation to detect missing child spans.
+  uint32 child_span_count = 15;
```
bogdandrutu: How do we know if 0 means it was real 0 or not set?

tigrannajaryan: It will not be possible. We had this discussion on this field here https://github.com/open-telemetry/oteps/pull/59/files#r338174095

If you believe the distinction between 0 and "not set" is important then this change will not work. Can you think of a use case where it matters?

Oberon00: I know that at least with protobuf v2 and Java, for each field a `hasMyField` method was generated you could use to check if it was received.

bogdandrutu: @Oberon00 not true for v3 :(

tigrannajaryan: @Oberon00 protobuf v3 removed the distinction: https://developers.google.com/protocol-buffers/docs/proto3#default
>Note that for scalar message fields, once a message is parsed there's no way of telling whether a field was explicitly set to the default value (for example whether a boolean was set to false) or just not set at all: you should bear this in mind when defining your message types.

bogdandrutu: This field is used to determine completion of a trace. If it is 0 the system cannot determine if this is a real leaf or the value is not set.

lizthegrey: it's only _local_ leaves though, there's no way of knowing about remote leaves... so I'm not sure how this helps with knowing if the trace is finished...

bogdandrutu: Another option is to reserve MAX_INT for undefined.

lizthegrey: oh, I like that one.

jmacd: I'd suggest -1 is a better value for undefined. However, I think @lizthegrey is right--perhaps this should be named "internal_child_count" (for `SpanKind==INTERNAL`) or "local_child_count".

It would also make sense, to me, to have a count of the number of times a context is extracted for a remote child, but I'm not sure it would be reliable way to test for completion since the child span could fail or be disregarded in some way.

Oberon00: I think we should really think about this remote/local child problem. Discussion had already started in the original OTEP https://github.com/open-telemetry/oteps/pull/59/files#r336737075.

tigrannajaryan:
Dropping this PR for now until the discussion about child counts is resolved. I created a new issue to discuss it here: https://github.com/open-telemetry/opentelemetry-specification/issues/355

PR #51, opened 2019-11-12
Remove wrapping of int and double value from SummaryValue message

tigrannajaryan:
This micro-optimization reduces CPU and memory usage.

```
goos: darwin
goarch: amd64
BenchmarkEncode/Baseline/Metric/Summary-8      	     120	  50042473 ns/op
BenchmarkEncode/Proposed/Metric/Summary-8      	     136	  43423088 ns/op

BenchmarkEncode/Baseline/Metric/Mix-8                     	      28	 206595979 ns/op
BenchmarkEncode/Proposed/Metric/Mix-8                     	      30	 199870715 ns/op

BenchmarkEncode/Baseline/Metric/MixSeries-8               	       9	 592705353 ns/op
BenchmarkEncode/Proposed/Metric/MixSeries-8               	      10	 539199628 ns/op

BenchmarkDecode/Baseline/Metric/Summary-8                 	      51	 119929790 ns/op	78696041 B/op	 2024000 allocs/op
BenchmarkDecode/Proposed/Metric/Summary-8                 	      56	 104306363 ns/op	69096032 B/op	 1824000 allocs/op

BenchmarkDecode/Baseline/Metric/Mix-8                     	      10	 545856843 ns/op	328040062 B/op	 9126000 allocs/op
BenchmarkDecode/Proposed/Metric/Mix-8                     	      10	 526862286 ns/op	318440044 B/op	 8926000 allocs/op

BenchmarkDecode/Baseline/Metric/MixSeries-8               	       5	1161078696 ns/op	752040067 B/op	17626000 allocs/op
BenchmarkDecode/Proposed/Metric/MixSeries-8               	       5	1075300086 ns/op	704040064 B/op	16626000 allocs/op
```

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -16,7 +16,6 @@ syntax = "proto3";
 
 package opentelemetry.proto.metrics.v1;
 
-import "google/protobuf/wrappers.proto";
 import "opentelemetry/proto/common/v1/common.proto";
 import "opentelemetry/proto/resource/v1/resource.proto";
 
@@ -389,12 +388,11 @@ message SummaryValue {
 
   // The total number of recorded values since start_time. Optional since
   // some systems don't expose this.
-  google.protobuf.Int64Value count = 3;
+  fixed64 count = 3;
 
   // The total sum of recorded values since start_time. Optional since some
   // systems don't expose this. If count is zero then this field must be zero.
-  // This field must be unset if the sum is not available.
-  google.protobuf.DoubleValue sum = 4;
+  double sum = 4;
 
   // Represents the value at a given percentile of a distribution.
   message ValueAtPercentile {


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -394,12 +393,15 @@ message SummaryValue {
...
 
   // The total sum of recorded values since start_time. Optional since some
   // systems don't expose this. If count is zero then this field must be zero.
-  // This field must be unset if the sum is not available.
-  google.protobuf.DoubleValue sum = 4;
+  double sum = 4;
+
+  // count_and_sum_unspecified must be set to true if count and sum values are unknown or
+  // unspecified.
+  bool count_and_sum_unspecified = 5;
```
bogdandrutu: I remember I saw cases when one was define but not the other.

tigrannajaryan: OK, I'll separate the fields.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -394,12 +393,17 @@ message SummaryValue {
...
+  uint64 count = 3;
 
   // The total sum of recorded values since start_time. Optional since some
   // systems don't expose this. If count is zero then this field must be zero.
-  // This field must be unset if the sum is not available.
-  google.protobuf.DoubleValue sum = 4;
+  double sum = 4;
+
+  // count_unspecified must be set to true if count value is unknown or unspecified.
+  bool count_unspecified = 5;
```
yurishkuro: curious, isn't `(count == 0) <==> (count_unspecified == true)`? Same for sum.

tigrannajaryan: I think those are different assertions:
- `count == 0` means "no observations were made"
- `count_unspecified == true` mean "we don't know how many observations were made"

I don't know how _important_ is the distinction though. In both cases you cannot calculate the average for example.

Also, alternatively the second assertion can be modeled differently. We could assume that:
- if `count==0 && len(percentile_values)==0` then it means "no observations were made"
- if `count==0 && len(percentile_values)>0` then it means "we don't know how many observations were made, but we have data about their distribution"

Could this be good enough and we don't really need to have a distinct condition for unspecified count?

bogdandrutu: I am fine with removing the bools for the moment. If we really need them we can add them back later.

tigrannajaryan: @bogdandrutu I like that. Let me do it.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -389,12 +388,11 @@ message SummaryValue {
 
   // The total number of recorded values since start_time. Optional since
   // some systems don't expose this.
-  google.protobuf.Int64Value count = 3;
+  uint64 count = 3;
```
bogdandrutu: fixed64?

tigrannajaryan: Given that most values are small that may increase wire size (not a big deal). I do not think we will gain anything from fixed64 in this case.
I would say it doesn't matter, can change if you have an opinion about it.

bogdandrutu: This is the total number of samples observed from the beginning of the process. This will easily grow.

jmacd: I still want to support sampled metrics, in which case the count could be approximate (I'd suggest `double`). Maybe it would be appropriate to use a separate field to represent approximate counts, though.

tigrannajaryan: I think we can use `double` universally. It is going to be a negligible performance hit to convert from int if needed.

But, I am not sure I understand the need for fractional count. How do you sample a metric or more precisely why would you need to lose the information about the real count (I am guessing by multiplying `count` and `sum` by some sampling coefficient)?

tigrannajaryan: I created a separate issue to discuss the need for fractional counts: https://github.com/open-telemetry/opentelemetry-proto/issues/59
Let's merge this one as is for now.

tigrannajaryan: @bogdandrutu done.

tigrannajaryan: Wait, wrong place :-(

tigrannajaryan: @bogdandrutu OK, it is fixed64 now.

jmacd: I don't see why fixed64 makes sense here.  We're talking about a Summary, which implies it is infrequently reported, so probably CPU usage is not a concern.  An int64 is a smaller encoding for values up to 2^49 or so.

Anyway it's not a big deal.

tigrannajaryan: I believe the difference is negligible. int64 is slightly slower to encode/decode and slightly smaller uncompressed but (IIRC) often compresses worse than fixed64 by zlib (I didn't dig into why - could entire because of my test data). The combined difference is below my threshold for optimizations (and below my measurement error for CPU time).
I am fine with either approach.

jmacd: That's a good point--about compression. Thanks.

PR #52, opened 2019-11-12
Use fixed64 fields instead of []byte for trace and span ids

tigrannajaryan:
This reduces CPU and memory usage, as well as wire size.

```
===== Encoded sizes
Baseline/Trace/Attribs          18791 bytes  [1.000], gziped 4027 bytes  [1.000]
Proposed/Trace/Attribs          17591 bytes  [1.068], gziped 3932 bytes  [1.024]

Baseline/Trace/Events           14089 bytes  [1.000], gziped 1730 bytes  [1.000]
Proposed/Trace/Events           12489 bytes  [1.128], gziped 1616 bytes  [1.071]

goos: darwin
goarch: amd64
BenchmarkEncode/Baseline/Trace/Attribs-8      	      86	  63775703 ns/op
BenchmarkEncode/Proposed/Trace/Attribs-8      	      99	  60474051 ns/op

BenchmarkEncode/Baseline/Trace/Events-8       	      74	  78882425 ns/op
BenchmarkEncode/Proposed/Trace/Events-8       	      81	  75893047 ns/op

BenchmarkDecode/Baseline/Trace/Attribs-8      	      66	 102103922 ns/op	82618718 B/op	 1416925 allocs/op
BenchmarkDecode/Proposed/Trace/Attribs-8      	      70	  91743212 ns/op	76888944 B/op	 1216925 allocs/op

BenchmarkDecode/Baseline/Trace/Events-8       	      45	 134349730 ns/op	103504028 B/op	 1924000 allocs/op
BenchmarkDecode/Proposed/Trace/Events-8       	      54	 118756653 ns/op	97704014 B/op	 1724000 allocs/op
```

`Trace/Attribs` is 1000 batches, each containing 100 spans, each span containing 3 attributes (2 int64, 1 string) and 0 TimedEvents.

`Trace/Events` is 1000 batches, each containing 100 spans, each span containing 0 attributes and 3 TimedEvents, each TimedEvent containing 1 int64 attribute.

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -24,49 +24,64 @@ option java_multiple_files = true;
 option java_package = "io.opentelemetry.proto.trace.v1";
 option java_outer_classname = "TraceProto";
 
-// A span represents a single operation within a trace. Spans can be
+// Span represents a single operation within a trace. Spans can be
 // nested to form a trace tree. Spans may also be linked to other spans
-// from the same or different trace. And form graphs. Often, a trace
+// from the same or different trace and form graphs. Often, a trace
 // contains a root span that describes the end-to-end latency, and one
 // or more subspans for its sub-operations. A trace can also contain
 // multiple root spans, or none at all. Spans do not need to be
 // contiguous - there may be gaps or overlaps between spans in a trace.
 //
-// The next id is 17.
+// The next field id is 18.
 message Span {
-  // A unique identifier for a trace. All spans from the same trace share
-  // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes
-  // is considered invalid.
+  // trace_id is the unique identifier of a trace. All spans from the same trace share
+  // the same `trace_id`.
   //
-  // This field is semantically required. Receiver should generate new
-  // random trace_id if empty or invalid trace_id was received.
+  // This field is semantically required. If an invalid trace ID was received:
+  // - The receiver MAY reject the invalid data and respond with the appropriate error
+  //   code to the sender.
+  // - The receiver MAY accept the invalid data and attempt to correct it.
   //
-  // This field is required.
-  bytes trace_id = 1;
+  // The ID is a 16-byte value represented using two fixed64 fields. An ID with
+  // value of 0 for all bytes is considered invalid.
+  // To convert this pair of fixed64 fields to an ID represented as a 16-byte array put
+  // trace_id_high in big-endian order in the first 8 bytes and trace_id_low in the
+  // last 8 bytes. Go example code:
+  //    var traceIDBytes [16]byte
+  // 	binary.BigEndian.PutUint64(traceIDBytes[:], trace_id_high)
+  //	binary.BigEndian.PutUint64(traceIDBytes[8:], trace_id_low)
+  fixed64 trace_id_high = 1;
+  fixed64 trace_id_low = 2;
 
-  // A unique identifier for a span within a trace, assigned when the span
-  // is created. The ID is an 8-byte array. An ID with all zeroes is considered
-  // invalid.
+  // span_id is a unique identifier for a span within a trace, assigned when the span
+  // is created.
   //
-  // This field is semantically required. Receiver should generate new
-  // random span_id if empty or invalid span_id was received.
+  // This field is semantically required. If an invalid span ID was received:
+  // - The receiver MAY reject the invalid data and respond with the appropriate error
+  //   code to the sender.
+  // - The receiver MAY accept the invalid data and attempt to correct it.
   //
-  // This field is required.
-  bytes span_id = 2;
+  // The ID is an 8-byte value represented as a fixed64 field. An ID with value of 0
+  // for all bytes is considered invalid.
+  // To convert this field to an ID represented as an 8-byte array put span_id in big-endian
+  // order in the byte array. Go example code:
+  //    var spanIDBytes [8]byte
+  // 	binary.BigEndian.PutUint64(traceIDBytes[:], span_id)
+  fixed64 span_id = 3;
 
   // tracestate conveys information about request position in multiple distributed tracing graphs.
   // It is a tracestate in w3c-trace-context format: https://www.w3.org/TR/trace-context/#tracestate-header
   // See also https://github.com/w3c/distributed-tracing for more details about this field.
-  string tracestate = 3;
+  string tracestate = 4;
 
-  // The `span_id` of this span's parent span. If this is a root span, then this
-  // field must be empty. The ID is an 8-byte array.
-  bytes parent_span_id = 4;
+  // parent_span_id is the `span_id` of this span's parent span. If this is a root span, then this
+  // field must be 0. See span_id for ID value byte order explanation.
+  fixed64 parent_span_id = 5;
 
   // An optional resource that is associated with this span. If not set, this span
   // should be part of a ResourceSpans that does include the resource information, unless
   // resource information is unknown.
-  opentelemetry.proto.resource.v1.Resource resource = 5;
+  opentelemetry.proto.resource.v1.Resource resource = 6;
 
   // A description of the span's operation.
   //
@@ -81,7 +96,7 @@ message Span {
   // receiver to fix the empty span name.
   //
   // This field is required.
-  string name = 6;
+  string name = 7;
 
   // Type of span. Can be used to specify additional relationships between spans
   // in addition to a parent/child relationship.
@@ -115,7 +130,7 @@ message Span {
   // Distinguishes between spans generated in a particular context. For example,
   // two spans with the same name may be distinguished using `CLIENT` (caller)
   // and `SERVER` (callee) to identify queueing latency associated with the span.
-  SpanKind kind = 7;
+  SpanKind kind = 8;
 
   // start_time_unixnano is the start time of the span. On the client side, this is the time
   // kept by the local machine where the span execution starts. On the server side, this
@@ -123,7 +138,7 @@ message Span {
   // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
   //
   // This field is semantically required and it is expected that end_time >= start_time.
-  fixed64 start_time_unixnano = 8;
+  fixed64 start_time_unixnano = 9;
 
   // end_time_unixnano is the end time of the span. On the client side, this is the time
   // kept by the local machine where the span execution ends. On the server side, this
@@ -131,7 +146,7 @@ message Span {
   // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
   //
   // This field is semantically required and it is expected that end_time >= start_time.
-  fixed64 end_time_unixnano = 9;
+  fixed64 end_time_unixnano = 10;
 
   // attributes is a collection of key/value pairs. The value can be a string,
   // an integer, a double or the Boolean values `true` or `false`. Note, global attributes
@@ -141,12 +156,12 @@ message Span {
   //     "/http/server_latency": 300
   //     "abc.com/myattribute": true
   //     "abc.com/score": 10.239
-  repeated opentelemetry.proto.common.v1.AttributeKeyValue attributes = 10;
+  repeated opentelemetry.proto.common.v1.AttributeKeyValue attributes = 11;
 
   // dropped_attributes_count is the number of attributes that were discarded. Attributes
   // can be discarded because their keys are too long or because there are too many
   // attributes. If this value is 0, then no attributes were dropped.
-  int32 dropped_attributes_count = 11;
+  int32 dropped_attributes_count = 12;
 
   // Event is a time-stamped annotation of the span, consisting of user-supplied
   // text description and key-value pairs.
@@ -166,33 +181,35 @@ message Span {
   }
 
   // events is a collection of Event items.
-  repeated Event events = 12;
+  repeated Event events = 13;
 
   // dropped_events_count is the number of dropped events. If the value is 0, then no
   // events were dropped.
-  uint32 dropped_events_count = 13;
+  uint32 dropped_events_count = 14;
 
   // A pointer from the current span to another span in the same trace or in a
   // different trace. For example, this can be used in batching operations,
   // where a single batch handler processes multiple requests from different
   // traces or when the handler receives a request from a different project.
   message Link {
-    // A unique identifier of a trace that this linked span is part of. The ID is a
-    // 16-byte array.
-    bytes trace_id = 1;
+    // trace_id is a unique identifier of a trace that this linked span is part of.
+    // See Span.trace_id for trace ID value byte order explanation.
+    fixed64 trace_id_high = 1;
+    fixed64 trace_id_low = 2;
 
-    // A unique identifier for the linked span. The ID is an 8-byte array.
-    bytes span_id = 2;
+    // span_id is a unique identifier for the linked span.
+    // See Span.span_id for span ID value byte order explanation.
+    fixed64 span_id = 3;
 
     // The tracestate associated with the link.
-    string tracestate = 3;
+    string tracestate = 4;
 
     // attributes is a collection of attribute key/value pairs on the link.
-    repeated opentelemetry.proto.common.v1.AttributeKeyValue attributes = 4;
+    repeated opentelemetry.proto.common.v1.AttributeKeyValue attributes = 5;
 
     // dropped_attributes_count is the number of dropped attributes. If the value is 0,
     // then no attributes were dropped.
-    int32 dropped_attributes_count = 5;
+    int32 dropped_attributes_count = 6;
   }
 
   // A collection of links, which are references from this span to a span
@@ -207,16 +224,16 @@ message Span {
   }
 
   // The included links.
-  Links links = 14;
+  Links links = 15;
 
   // An optional final status for this span. Semantically when Status
   // wasn't set it is means span ended without errors and assume
   // Status.Ok (code = 0).
-  Status status = 15;
+  Status status = 16;
 
   // An optional number of child spans that were generated while this span
   // was active. If set, allows an implementation to detect missing child spans.
-  google.protobuf.UInt32Value child_span_count = 16;
+  google.protobuf.UInt32Value child_span_count = 17;
 }
 
 // The Status type defines a logical error model that is suitable for different


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -23,35 +23,37 @@ option java_multiple_files = true;
...
+  // fixed64 fields. An ID with value of 0 is considered invalid.
   //
-  // This field is semantically required. Receiver should generate new
-  // random trace_id if empty or invalid trace_id was received.
+  // This field is semantically required. If 0 or invalid trace_id was received:
+  // - The receiver MAY reject the invalid data and respond with the appropriate error
+  //   code to the sender.
+  // - The receiver MAY accept the invalid data and attempt to correct it.
+  // Low 64 bits of ID.
+  fixed64 trace_id_low = 1;
```
bogdandrutu: Does this change make a real difference? Once we do this we need to deal with endianness which I am worried that we are going to have mistakes.

tigrannajaryan: Benchmarks show 5-10% less CPU and memory usage. I think it is worth it.
I can add a comment explaining how to convert a 16 byte IDs to low/high 64 bits and back to avoid any confusion.

bogdandrutu: Sorry I cannot remember all the scenarios for the benchmarks. Can you update description, need to understand if this 5% is in what case.

tigrannajaryan: `Trace/Attribs` is 1000 batches, each containing 100 spans, each span containing 3 attributes (2 int64, 1 string) and 0 TimedEvents.

`Trace/Events` is 1000 batches, each containing 100 spans, each span containing 0 attributes and 3 TimedEvents, each TimedEvent containing 1 int64 attribute.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -23,35 +23,37 @@ option java_multiple_files = true;
...
 //
-// The next id is 15.
+// The next field id is 17.
 message Span {
-  // A unique identifier for a trace. All spans from the same trace share
-  // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes
-  // is considered invalid.
+  // trace_id is the unique identifier of a trace. All spans from the same trace share
+  // the same `trace_id`. The ID is a 128 bit (16 byte) value represented using two
+  // fixed64 fields. An ID with value of 0 is considered invalid.
```
bogdandrutu: Not sure I understand. In OpenTelemetry the trace id is defined as a 16 bytes array. So you need to explain how that will be converted to unit128 or directly to the fixed64.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -23,35 +23,37 @@ option java_multiple_files = true;
...
-  // This field is semantically required. Receiver should generate new
-  // random span_id if empty or invalid span_id was received.
-  //
-  // This field is required.
-  bytes span_id = 2;
+  // This field is semantically required. If 0 or invalid span_id was received:
+  // - The receiver MAY reject the invalid data and respond with the appropriate error
+  //   code to the sender.
+  // - The receiver MAY accept the invalid data and attempt to correct it.
+  fixed64 span_id = 3;
```
bogdandrutu: Same as trace id. This is defined by the w3c standard that we use as a 8 bytes array. Not clear how that gets converted to a fixed64.

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: I'll add a clarification.

tigrannajaryan: I added a clarification, please have a look.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -23,35 +23,48 @@ option java_multiple_files = true;
...
+  //   code to the sender.
+  // - The receiver MAY accept the invalid data and attempt to correct it.
   //
-  // This field is required.
-  bytes trace_id = 1;
-
-  // A unique identifier for a span within a trace, assigned when the span
-  // is created. The ID is an 8-byte array. An ID with all zeroes is considered
-  // invalid.
+  // The ID is a 128 bit (16 byte) value represented using two fixed64 fields.
```
bogdandrutu: The id is not a 128 bit

tigrannajaryan: I am not sure I understand what the objection is. Do you prefer that it just says `ID is a 16 bytes value represented using two fixed64 fields`?

bogdandrutu: You say that it is 128 bit then compare it with 0 which means you assume a number representation of the 16 bytes array. Which it does not have.

tigrannajaryan: Re-worded. Please have another look.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -23,35 +23,50 @@ option java_multiple_files = true;
...
+  //   code to the sender.
+  // - The receiver MAY accept the invalid data and attempt to correct it.
   //
-  // This field is required.
-  bytes trace_id = 1;
-
-  // A unique identifier for a span within a trace, assigned when the span
-  // is created. The ID is an 8-byte array. An ID with all zeroes is considered
-  // invalid.
+  // The ID is a 16 bytes value represented using two fixed64 fields. An ID with
```
bogdandrutu: `The ID is a 16 bytes value **is** represented using two fixed64 fields.`

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -23,35 +23,50 @@ option java_multiple_files = true;
...
 message Span {
-  // A unique identifier for a trace. All spans from the same trace share
-  // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes
-  // is considered invalid.
+  // trace_id is the unique identifier of a trace. All spans from the same trace share
+  // the same `trace_id`.
   //
-  // This field is semantically required. Receiver should generate new
-  // random trace_id if empty or invalid trace_id was received.
+  // This field is semantically required. If 0 or invalid trace_id was received:
```
bogdandrutu: I think you are duplicated the comment about 0 being invalid.

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: I think what you suggest is not syntactically correct English. I may be wrong though, I am not a native speaker :-)

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: Reworded, check again please.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -23,35 +23,50 @@ option java_multiple_files = true;
...
 message Span {
-  // A unique identifier for a trace. All spans from the same trace share
-  // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes
-  // is considered invalid.
+  // trace_id is the unique identifier of a trace. All spans from the same trace share
+  // the same `trace_id`.
   //
-  // This field is semantically required. Receiver should generate new
-  // random trace_id if empty or invalid trace_id was received.
+  // This field is semantically required. If invalid trace ID was received:
```
bogdandrutu: If "an" invalid?

opentelemetry/proto/trace/v1/trace.proto
bogdandrutu: I would defer to a native speaker :)

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -23,35 +23,50 @@ option java_multiple_files = true;
...
-  // This field is semantically required. Receiver should generate new
-  // random span_id if empty or invalid span_id was received.
+  // This field is semantically required. If invalid span ID was received:
+  // - The receiver MAY reject the invalid data and respond with the appropriate error
+  //   code to the sender.
+  // - The receiver MAY accept the invalid data and attempt to correct it.
   //
-  // This field is required.
-  bytes span_id = 2;
+  // The ID is an 8 bytes value represented as a fixed64 field. An ID with value of 0
```
bogdandrutu: 0 for all bytes is considered invalid.

tigrannajaryan: Fixed.

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: Fixed.

opentelemetry/proto/trace/v1/trace.proto
jmacd: Language looks good. I'd write "a 16-byte value", no big deal :)
This is making regret that the Go tracer switched from two uint64s to [16]byte.

tigrannajaryan: Changed to "a 16-byte value", thanks.

>This is making regret that the Go tracer switched from two uint64s to [16]byte.

Is it too late to switch back? Would be probably nice to use something like `struct { hi,lo uint64 }`.

tigrannajaryan:
Can we merge this?

opentelemetry/proto/trace/v1/trace.proto
jmacd: It's not. That's what we had before. :)

tigrannajaryan: @jmacd I am going to resolve this, leaving it up to you if you want to consider changing TraceID data type in Go lib.

tigrannajaryan:
Can we merge this?

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -24,35 +24,50 @@ option java_multiple_files = true;
...
   //
-  // This field is required.
-  bytes trace_id = 1;
-
-  // A unique identifier for a span within a trace, assigned when the span
-  // is created. The ID is an 8-byte array. An ID with all zeroes is considered
-  // invalid.
+  // The ID is a 16-byte value represented using two fixed64 fields. An ID with
+  // value of 0 for all bytes is considered invalid.
+  // To convert this pair of fixed64 fields to an ID represented as a 16-byte array put
```
SergeyKanzhelev: is little-endian used in other systems like Jaeger and Zipkin?

yurishkuro: Jaeger uses big endian, evidently: https://github.com/jaegertracing/jaeger/blob/master/model/ids.go#L81

bogdandrutu: I would prefer to be the consistent with Jaeger :)

yurishkuro: according to wikipedia:

> big-endianness is the dominant ordering in networking protocols (IP, TCP, UDP)

bogdandrutu: And performance wise even though Intel is little endian the transformation is an operation that is super fast (no transistors involved) change a wire reverse.

tigrannajaryan: ProtoBuf uses little-endian encoding. Also most (?) server CPUs today use little endian in-memory representation. This means on little-endian CPUs ProtoBuf does straight copying of fixed64 fields when serializing. Requiring big-endian for this field will require extra endianness conversion on little-endian CPUs.

tigrannajaryan: So while I agree it would be nice to be consistent with Jaeger it will also mean a performance hit. I can benchmark and see if it is a significant performance hit.

tigrannajaryan: There is no measurable performance difference so I am changing to use big endian byte order so that we are consistent with Jaeger.

tigrannajaryan: @SergeyKanzhelev I changed to big endian, explanation above. Please unblock and merge.

SergeyKanzhelev:
@tigrannajaryan last change was less than 24 hours before. However it seems that the main concern was resolved. I'm merging to keep it going, please raise an issue if you disagree with the change

Flarna:
One question regarding the CPU usage improvement: Which code parts were included in the test?
Only the encoding/decoding of protobuf itself or also the conversion of strings from/to int64?
In special Javascript requires quite some parsing of a traceid string to get two int64.

tigrannajaryan:
Only encoding/decoding in Go was benchmarked. I do not know much about ProtoBuf support in JS, how do int64s work given that JS's doesn't support them natively? int64's are used throughout the proto, not just for IDs.

Flarna:
I played around with [protobufjs](https://www.npmjs.com/package/protobufjs). It uses [long](https://www.npmjs.com/package/long) to represent 64 bit numbers which are in fact an object containing two native numbers for high/low 32 bits + a boolean for signed/unsigned.

The whole path to get the traceid (represented as a 32 char hex string) in `SpanContext` to the object finally serialized to a proto message is to first parse the high and low 16 hex chars into two `Long` objects. These four  32 bit `numbers` are then encoded into the final `Buffer`.

This looks like quite some extra work compared to only covert from hex string to a buffer.

Is there a similar conversion needed in go? Or is the trace-id stored as two int64 in SpanContext already? If a conversion is needed it would be interesting how overal performance is effected if the conversion is included in a benchmark.

tigrannajaryan:
Go SDK will likely use native int64 (@jmacd can tell whether they decided to do it or no), so there is no conversion at all.

I understand that it may be extra code for JS. It would be also useful to understand how big the performance impact is compared to hex-to-bin.

If we find this highly problematic for JS it is not too late to revert this change.

Flarna:
I did a fast benchmark to compared putting the traceId of 100 SpanContext objects into a `ExportTraceServiceRequest` message and encode it. I have omitted all other fields in this test.
The Cpu time needed to convert and encode
two fixed64: 193us
bytes: 106us
Size of resulting message:
two fixed64: 2003 bytes
bytes:       2103 bytes

jmacd:
The Go client would prefer to use the native types to avoid a conversion, but I doubt there's a significant difference between using native uint64s vs an array of [...]byte. This sounds like a trade-off between certain other clients and the server performance.

tigrannajaryan:
>I doubt there's a significant difference between using native uint64s vs an array of [...]byte

In Go there is a visible difference that benchmarks clearly show (benchmarks results are in this PR description).

That said, I do not think we need to favor Go over JS, especially given that this hits JS performance seriously and gains for Go are quite modest. I suggest that we revert this particular change.

jmacd:
Looking back at the original data, I'd want to see the actual code to make a conclusion. I wouldn't expect a difference in the number of allocations for encoding the two schemes, but maybe I haven't thought this through enough. I would agree that if the issue is significantly worse in JS, we should just skip this. The Go repository can revert to two int64s for traceID.

tigrannajaryan:
>Looking back at the original data, I'd want to see the actual code to make a conclusion.

The benchmark to compare encodings is here https://github.com/tigrannajaryan/exp-otelproto/blob/master/encodings/encoding_test.go (the specific bytes vs int64 comparison is no longer in the codebase, but can be easily added back to Baseline/Proposed encodings that are part of the benchmark).

>I wouldn't expect a difference in the number of allocations for encoding the two schemes

`bytes` Protobuf type is represented as a byte slice in Go which is an extra allocation. `int64` Protobuf type is represented as native int64 field which is not an extra allocation.

yurishkuro:
in Jaeger's Go code we use a custom type TraceID (via gogo annotations) which is a pair of unit64's with a custom marshaller from a byte array, so we don't have an extra allocation. The proto field remains as `bytes`.

tigrannajaryan:
Good point. Not sure if Goprotobuf also allows custom marshalers like Gogoproto does and whether it is possible to declare the protobuf schema in a way that makes possible to use either of these libs.

yurishkuro:
Is there a particular reason why you need to use the official goprotobuf?  gogo claims to be much more efficient, in part due to support for custom types and custom marshallers.

A proto file can be always stripped from gogo annotations, or you can keep them (as long as gogo files are on the import path) and not specify the gogo plugin, in which case they are ignored.

Flarna:
If `traceId` gets reverted I would vote also to do the same for `spanId`. I find it confusing to use numbers for something which is actually a byte sequence which in turn requires specification of endieness.

Other `fixed64` like `start_time_unixnano` are numbers therefore a good match.

Anyhow, this is no hard requirement and I don't think the performance difference will be that high to get a real issue.

jmacd:
@tigrannajaryan (of course, thanks for the explanation)

We use gogo at LightStep, mainly for the `gogoproto.nullable` annotation, but custom marshalers is also a solution to the performance problem here.

yurishkuro:
from Jaeger:
```
message Span {
  bytes trace_id = 1 [
    (gogoproto.nullable) = false,
    (gogoproto.customtype) = "TraceID",
    (gogoproto.customname) = "TraceID"
  ];
  bytes span_id = 2 [
    (gogoproto.nullable) = false,
    (gogoproto.customtype) = "SpanID",
    (gogoproto.customname) = "SpanID"
  ];
}
```

tigrannajaryan:
>Is there a particular reason why you need to use the official goprotobuf

We (Omnition) use gogoproto internally. However some users are able to use goprotobuf only. If I remember correctly @bogdandrutu mentioned this was a limitation for them, he may be able to tell more.

>A proto file can be always stripped from gogo annotations, or you can keep them (as long as gogo files are on the import path) and not specify the gogo plugin, in which case they are ignored.

So this will mean that they will be compatible on the wire but incompatible on the code level. Which may be fine for some applications and not OK for some others. It is indeed a problem for Collector, where we want to stay with Protobuf schema declarations that end up producing the same code interface using Goprotobug and Gogoproto, although Gogoproto's implementation is faster. This way we are able to re-compile existing code that is originally built against Goprotobuf and replace it during build by Gogoproto without touching any code and still have significant performance gains. This is precisely how it works today with OpenCensus Protobuf schema (the official Collector build uses Goprotobuf, but Omnition has a custom build that uses Gogoproto).

I think this ability to switch Protobuf implementation without touching the code that uses it is an important property of a schema definition that I would want to have in OTLP as well.

tigrannajaryan:
Created a PR to revert this change: https://github.com/open-telemetry/opentelemetry-proto/pull/82

PR #56, opened 2019-11-14
Eliminate unnecessary wrapping of TimedEvent

tigrannajaryan:
This change improves CPU and RAM usage by 15-20%:

```
BenchmarkEncode/Baseline/Trace/Events-8       	      62	  90245433 ns/op
BenchmarkEncode/Proposed/Trace/Events-8       	      79	  77156010 ns/op

BenchmarkDecode/Baseline/Trace/Events-8       	      33	 167595535 ns/op	119504035 B/op	 2324000 allocs/op
BenchmarkDecode/Proposed/Trace/Events-8       	      45	 131056988 ns/op	103504032 B/op	 1924000 allocs/op
```

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -32,7 +32,7 @@ option java_outer_classname = "TraceProto";
 // multiple root spans, or none at all. Spans do not need to be
 // contiguous - there may be gaps or overlaps between spans in a trace.
 //
-// The next id is 15.
+// The next id is 17.
 message Span {
   // A unique identifier for a trace. All spans from the same trace share
   // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -239,16 +226,16 @@ message Span {
...
   // An optional final status for this span. Semantically when Status
   // wasn't set it is means span ended without errors and assume
   // Status.Ok (code = 0).
-  Status status = 14;
+  Status status = 15;
 
   // An optional number of child spans that were generated while this span
   // was active. If set, allows an implementation to detect missing child spans.
-  google.protobuf.UInt32Value child_span_count = 15;
+  google.protobuf.UInt32Value child_span_count = 16;
```
bogdandrutu: Update the next index at the top of the file?

tigrannajaryan: Done.

bogdandrutu:
If we do all this 15-20% improvements do we get to 0 cost at one point? :))

tigrannajaryan:
> If we do all this 15-20% improvements do we get to 0 cost at one point? :))

Almost :-)

We will be 3-5x faster than OpenCensus for traces and 2-3x faster for metrics (and similar reduction in RAM usage). I am looking for more ways, not sure if eventually it will be 0 cost :-)

PR #62, opened 2019-11-22
Remove unneeded data that is now conveyed via generic Resource attrib…

tigrannajaryan:
…utes

The PR https://github.com/open-telemetry/opentelemetry-specification/pull/363/files
defines semantic conventions for library, host and service. There is no longer
need to hard code this information as protocol fields.

### DIFF
--- a/opentelemetry/proto/agent/common/v1/common.proto
+++ b/opentelemetry/proto/agent/common/v1/common.proto
@@ -33,12 +33,6 @@ message Node {
   // Identifier that uniquely identifies a process within a VM/container.
   ProcessIdentifier identifier = 1;
 
-  // Information on the OpenTelemetry Library that initiates the stream.
-  LibraryInfo library_info = 2;
-
-  // Additional information on service.
-  ServiceInfo service_info = 3;
-
   // Additional attributes.
   map<string, string> attributes = 4;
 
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/agent/common/v1/common.proto
```diff
@@ -49,49 +43,9 @@ message Node {
 // Identifier that uniquely identifies a process within a VM/container.
 message ProcessIdentifier {
```
bogdandrutu: I don't think this is needed as well.

opentelemetry/proto/agent/common/v1/common.proto
```diff
@@ -33,12 +33,6 @@ message Node {
   // Identifier that uniquely identifies a process within a VM/container.
```
bogdandrutu: Can we remove all the Node?

tigrannajaryan: Yes. I will do in separate PR.

opentelemetry/proto/agent/common/v1/common.proto
tigrannajaryan: Correct, was planning to do in separate PR.

bogdandrutu:
Please rebase

PR #63, opened 2019-11-22
Remove config capability

tigrannajaryan:
This existed also in OpenCensus but as far as I know was never
used in real world. Removing now as unneeded functionality.
If there is a future evidence that this is needed we can add back.

### DIFF
--- a/java_build.sh
+++ b/java_build.sh
@@ -7,7 +7,6 @@ rm -rf $OUTDIR
 mkdir -p $OUTDIR
 
 protoc --java_out=${OUTDIR} opentelemetry/proto/resource/v1/resource.proto \
-    && protoc --java_out=${OUTDIR} opentelemetry/proto/trace/v1/trace_config.proto \
     && protoc --proto_path=./ --java_out=${OUTDIR} opentelemetry/proto/trace/v1/trace.proto \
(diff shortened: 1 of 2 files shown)


opentelemetry/proto/trace/v1/trace_config.proto
```diff
@@ -1,77 +0,0 @@
...
-
-package opentelemetry.proto.trace.v1;
-
-option java_multiple_files = true;
-option java_package = "io.opentelemetry.proto.trace.v1";
-option java_outer_classname = "TraceConfigProto";
-
-// Global configuration of the trace service. All fields must be specified, or
-// the default (zero) values will be used for each type.
-message TraceConfig {
```
bogdandrutu: Please don't remove this for the moment. We still have a config capability for the library. I know it is manual for the moment but we can use this as reference to what can be configured.

tigrannajaryan: Do you suggest to remove `rpc Config` from `TraceService` but keep trace_config.proto file for future work?

bogdandrutu: Yes

tigrannajaryan: Done

PR #72, opened 2019-11-25
Refine child span count field

tigrannajaryan:
The field was previously defined ambiguously and it was not clear
whether it was the count of local or all children (including remotely
generated).

This change makes it clear that this is only local children. A counter
of remote children may be added in the future if there is an understanding
about how to obtain this number (open discussion is here:
https://github.com/open-telemetry/opentelemetry-specification/issues/355).

The field type is also changed to a more efficient sfixed32 with
special-casing -1 as "value not known".

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -16,7 +16,6 @@ syntax = "proto3";
 
 package opentelemetry.proto.trace.v1;
 
-import "google/protobuf/wrappers.proto";
 import "opentelemetry/proto/common/v1/common.proto";
 
 option java_multiple_files = true;
@@ -220,9 +219,10 @@ message Span {
   // Status.Ok (code = 0).
   Status status = 16;
 
-  // An optional number of child spans that were generated while this span
-  // was active. If set, allows an implementation to detect missing child spans.
-  google.protobuf.UInt32Value child_span_count = 17;
+  // An optional number of local child spans that were generated while this span
+  // was active. Value of -1 indicates that the number of local child spans is unknown.
+  // If local_child_span_count>=0, allows an implementation to detect missing child spans.
+  sfixed32 local_child_span_count = 17;
 }
 
 // The Status type defines a logical error model that is suitable for different
--- a/opentelemetry/proto/common/v1/common.proto
+++ b/opentelemetry/proto/common/v1/common.proto
@@ -26,7 +26,7 @@ message AttributeKeyValue {
   // ValueType is the enumeration of possible types that value can have.
   enum ValueType {
     STRING  = 0;
-    INT   = 1;
+    INT     = 1;
     DOUBLE  = 2;
     BOOL    = 3;
   };


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,9 +219,10 @@ message Span {
   // Status.Ok (code = 0).
   Status status = 16;
 
-  // An optional number of child spans that were generated while this span
-  // was active. If set, allows an implementation to detect missing child spans.
-  google.protobuf.UInt32Value child_span_count = 17;
+  // An optional number of local child spans that were generated while this span
+  // was active. Value of -1 indicates that the number of local child spans is unknown.
+  // If local_child_span_count>=0, allows an implementation to detect missing child spans.
```
jmacd: nit: You might say if local_child_span_count>0 it allows detecting missing children. The value 0 positively states that there were no local children.

yurishkuro: I would prefer to treat 0 as “unknown”. It has the same practical result that it can’t be used for validation, but it works better with an optional field. 

Also, it may be worth emphasizing that this only counts in-process children.

tigrannajaryan: What if we do know that there is 0 local children? If we treat 0 as unknown we have no way to convey this fact.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,9 +219,10 @@ message Span {
...
+  sfixed32 local_child_span_count = 17;
```
SergeyKanzhelev: Comment on the Span ` The next field id is 18.` have to be updated

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,9 +219,10 @@ message Span {
   // Status.Ok (code = 0).
   Status status = 16;
 
-  // An optional number of child spans that were generated while this span
-  // was active. If set, allows an implementation to detect missing child spans.
-  google.protobuf.UInt32Value child_span_count = 17;
+  // An optional number of local child spans that were generated while this span
```
SergeyKanzhelev: Can you add a small note on how consumer can use this information. I.e. semantics of this field beyond just description. Or link to spec describing it?

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: I think the comment is correct. It means the next _available_. If it is unclear I can fix it.

SergeyKanzhelev: Oh, I totally missed that this is not a new property. Sorry for confusion

opentelemetry/proto/trace/v1/trace.proto
SergeyKanzhelev: @yurishkuro I'm merging. Please file an issue if you feel strongly about `0 means unknown`. I understand how it may be confusing - if one simply forgot to set this property to any value - it will mean `0` children which may be factually incorrect

yurishkuro: @tigrannajaryan we don't need to know, because there is no practical difference. Validation is only useful when this value > 0.

SergeyKanzhelev: https://github.com/open-telemetry/opentelemetry-proto/issues/76

Issue #54, opened 2019-11-14
Eliminate multi data point TimeSeries


tigrannajaryan:
Metric proto currently allows to represent multiple data points for each metric, i.e. a timeseries with several values of the same metric, one value per timestamp.

However, reading the [metric SDK proposal](https://github.com/open-telemetry/opentelemetry-specification/pull/347) tells me that normally each collection will emit only a single data point per metric, the last aggregate state of the metric. There may be custom exporters which accumulate data points and emit them all in one batch but it seems to be a possible exceptional case and not a norm. Typically an exporter will simply send the last aggregate value.

Another use case where we could possibly have multiple data points per metric is in Collector, where we could in theory batch points from the same metric into one timeseries. However, in practice I do not think we will do it since it is a very costly operation that requires maintaining maps of metrics.

Given that a single data point per metric is the most common case I suggest that we optimize the proto for that case. Currently the proto definition is the following (only int64 is shown for simplicity):

```proto
message Metric {
  MetricDescriptor metric_descriptor = 1;
  Resource resource = 2;
  repeated Int64TimeSeries int64_timeseries = 3;
}
message Int64TimeSeries {
  repeated string label_values = 1;
  repeated Int64Value points = 2;
}
message Int64Value {
  fixed64 start_time_unixnano = 1;
  fixed64 timestamp_unixnano = 2;
  sfixed64 value = 3;
}
```

I propose to change this to the following:

```proto
message Metric {
  MetricDescriptor metric_descriptor = 1;
  Resource resource = 2;
  repeated Int64Point int64_points = 3;
}
message Int64Point {
  repeated string label_values = 1;
  fixed64 start_time_unixnano = 2;
  fixed64 timestamp_unixnano = 3;
  sfixed64 value = 4;
}
```

Preliminary benchmarking shows that for the case when we have 1 int64 data point per metric this change reduces CPU and RAM usage by 25% for encoding and decoding.

Note: this schema still allows representing multiple data points per metric if neccessary, however at the cost of repeating label values.

tigrannajaryan:
@jmacd @bogdandrutu I need your opinion please.

bogdandrutu:
>Another use case where we could possibly have multiple data points per metric is in Collector, where we could in theory batch points from the same metric into one timeseries. However, in practice I do not think we will do it since it is a very costly operation that requires maintaining maps of metrics.

We are working on a processor that does this, in order to reduce cardinality for the labels (some users need this).

[...]

PR #77, opened 2019-11-27
Eliminate multi-data point timeseries

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-proto/issues/54

After a discussion we came to conclusion that there is no good use case
for having multiple data points. All possible ways to combine data points
that we imagined are either very expensive or make no semantic sense.

This commit eliminates timeseries that can have multiple data points and replaces
them with single data points.

If in the future we discover a use case for multiple data points we have 2
possible paths to extend the proto:

- Add a corresponding points repeated field in the Int64DataPoint.
- Introduce a different message type that is capable of storing multiple data
  points and assign it a different type in the MetricDescriptor.Type enum.

There are probably other ways as well.


tigrannajaryan:
> Can you split the removing of the generated code in a separate PR?

@bogdandrutu I do not understand, can you clarify? I did not remove any code, just re-generated it to match proto changes.

bogdandrutu:
@tigrannajaryan I think I understand what happens. You actually regenerated the files. This is so hard to review. I think in the future will be a big mess if we do re-generate the files in the PR. Maybe we should just configure the CI to regenerate everything after each PR merged to master.

tigrannajaryan:

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -46,46 +45,46 @@ option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/metric
...
-  // one or more points. Only one of the following fields is used for the data, depending on
-  // the type of the metric defined by MetricDescriptor.type field.
-  repeated Int64TimeSeries int64_timeseries = 2;
-  repeated DoubleTimeSeries double_timeseries = 3;
-  repeated HistogramTimeSeries histogram_timeseries = 4;
-  repeated SummaryTimeSeries summary_timeseries = 5;
+  // Data is a list of one or more DataPoints for a single metric. Only one of the
+  // following fields is used for the data, depending on the type of the metric defined
+  // by MetricDescriptor.type field.
+  repeated Int64DataPoint int64_datapoints = 2;
```
yurishkuro: if I am reading this right, it means for every measurement point recorded we need to send all the labels, even though a time series is defined as data points with identical labels, which is the most prevalent use case (especially when using handle/bound instruments).

tigrannajaryan: @yurishkuro You are reading it right. See discussion on https://github.com/open-telemetry/opentelemetry-proto/issues/54

Can you describe a use case which you believe is prevalent where there is a need to transmit multiple data points with the same set of label values in one request? Or are saying that we could avoid transmitting the labels at all if they didn't change (that would be some sort of streaming or history-aware protocol, which this is not).

If there is indeed a good use-case for multiple data points this PR should not be merged.

bogdandrutu: @yurishkuro not for every measurement. At the bound level we will aggregate for a period of time (e.g. 10 sec) then export one point that represents the current aggregated value at that bound.

tigrannajaryan: Hmm, right, what Bogdan said. I believe SDK is going to do collection cycles periodically and only export one data point per labelset per cycle.

yurishkuro: I remember we had a discussion some time ago about raw measures vs. instruments, with one example someone mentioned was that even a counter can be treated as a raw measure if we treat each delta as an independent event. Are we discounting the possibility of ever exporting these raw measures without any aggregation on the client side?

In any case, this is not a blocking issue for me.

cc @robskillington

yurishkuro: PS I do agree with the proposal in #54 that for such special case of multi-valued metrics we could have a different API / data structure.

tigrannajaryan: >Are we discounting the possibility of ever exporting these raw measures without any aggregation on the client side?

It is still possible using this schema, though it will be less efficient (label values will be transmitted for every data point). Also, you are right, we have the option to have a different request type / data structure that is more optimal for that case.

jmacd:
Just to add a little more depth, I can imagine two scenarios calling for different formats, ideally.  You might wish to export a list of timestamped values where each value has its own distinct timestamp.  In that case, the original protocol would do. You might wish to elide the timestamp for a short time period by recording the values and a single timestamp, in which case a `packed` protobuf encoding would be desirable.  Either of these could be added in the future, so I'm happy with this PR as it stands.

tigrannajaryan:
@open-telemetry/specs-approvers Can I get more reviews on this, please?

tigrannajaryan:
Thanks @yurishkuro

PR #87, opened 2020-01-06
Rename agent to collector

### LINKED ISSUE #67: Rename `agent` directory to `collector`?

SergeyKanzhelev:
Fixes #67


opentelemetry/proto/collector/metrics/v1/metrics_service.proto
```diff
@@ -14,18 +14,18 @@
...
 option java_multiple_files = true;
-option java_package = "io.opentelemetry.proto.agent.metrics.v1";
+option java_package = "io.opentelemetry.proto.collector.metrics.v1";
 option java_outer_classname = "MetricsServiceProto";
-option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/agent/metrics/v1";
+option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/collector/metrics/v1";
 
 // Service that can be used to push metrics between one Application
-// instrumented with OpenTelemetry and an agent, or between an agent and a
+// instrumented with OpenTelemetry and an collector, or between an collector and a
```
reyang:
```suggestion
// instrumented with OpenTelemetry and a collector, or between a collector and a
```

yurishkuro:
please describe motivation here or in the ticket

SergeyKanzhelev:
@yurishkuro 

This change is mostly driven by the fact that OpenTelemetry Collector used to be called Agent and recognized by this name. But recently we discovered many people confused by it and the decision was to only call it collector. Thus this change.

yurishkuro:
would it still be usable as host agent / daemon set? ie accessible via loopback.

SergeyKanzhelev:
Yes, definitely. This change is simply to make a terminology clearer since the schema is used in collector that can run as "agent". But it is still a collector. The confusion is coming from situations when agent is used in a form of JavaAgent thing

tigrannajaryan:
> would it still be usable as host agent / daemon set? ie accessible via loopback.

@yurishkuro Yes, Collector is intended to be run either as a standalone service (i.e. a "Collector") or as a local Agent.

This change simply normalizes the terminology across different repos.

SergeyKanzhelev:
@yurishkuro please comment if you still have concerns. I will merge it later today

yurishkuro:
No concerns.

PR #88, opened 2020-01-06
osx install script

SergeyKanzhelev:
Is there better way? simply adopted linux script =)


bogdandrutu:
If anyone knows a better way we are happy to accept suggestions :)

PR #90, opened 2020-01-06
omitted field means unknown

### LINKED ISSUE #76: Zero means unknown for child span count
I think it is important to discuss [this](https://github.com/open-telemetry/opentelemetry-proto/pull/72#discussion_r351121524). Copying some notes from the thread. Note, that the main problem is that semantics of the field and how it will be used is not clearly defined.

Let's treat 0 as “unknown”. It has the same practical result that it can’t be used for validation, but it works better with an optional field.

If one simply forgot to set this property to any value - it will mean 0 children which may be factually incorrect.

Depending on semantics of the field and how it should be used there may be no practical difference. Validation is only useful when this value > 0.

SergeyKanzhelev:
Fixes #76 

To re-iterate on reasoning: we need to chose between the potential problem when omitted (by mistake) field will be treated as zero children and the ability to distinguish zero children and unknown cases. Also having default as zero should eliminate the need to send `-1` with every span for the SDK that doesn't count children. 

I'm not sure what's the all use cases for this field, so if there are any that require this differentiation - happy to close the PR.


SergeyKanzhelev:
@open-telemetry/specs-approvers any other comments? 

I was also thinking - semantics of this field and intended use is not clear. If nobody plans to use it now - should we just remove it?

SergeyKanzhelev:
this will be superseeded by another PR to remove the field for now

PR #94, opened 2020-01-14
Add InstrumentationLibrary and plugin into the collector

bogdandrutu:


bogdandrutu:
The Library info are the data used for the "Named Tracer/Meter" which are associated with every Span/Instrument created by these instances. I picked the name library info instead of "NamedXYZ".

tigrannajaryan:
Why does it need to be a separate data structure and not simply part of Resource message? We already have semantic conventions that allow to differentiate library attributes from the rest, including library name and version: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-resource-semantic-conventions.md#library

bogdandrutu:
Based on my understanding the library name inside the Resource is the "instrumentation" library (99% of the time opentelemetry), this is the "instrumented" library (e.g. mongo, etc.).

tigrannajaryan:
> Based on my understanding the library name inside the Resource is the "instrumentation" library (99% of the time opentelemetry), this is the "instrumented" library (e.g. mongo, etc.).

You are, right, my mistake. However, the original question still holds: would it be better if we defined semantic conventions for "instrumented" library or you think this is information is fundamental property of a span, will always be present and is better represented as a hard-coded Protobuf field?

bogdandrutu:
Do you mean semantic conventions inside the Resource? I think this does not belong to the resource (which describes the application) and is a sub-component of the application described by the resource.

tigrannajaryan:
> Do you mean semantic conventions inside the Resource? I think this does not belong to the resource (which describes the application) and is a sub-component of the application described by the resource.

Yes, I was thinking library info could be added to Resource and one request could carry multiple ResourceSpans/ResourceMetrics messages, each carrying data from a particular Resource/Library combination.

However, it seems like it logically is separate, so making it a separate message is the right approach (what you did).

opentelemetry/proto/collector/trace/v1/trace_service.proto
```diff
@@ -37,23 +38,33 @@ service TraceService {
...
   // in that case this array will contain multiple elements.
   repeated ResourceSpans resource_spans = 1;
 }
 
-// A collection of spans from a Resource.
+// A collection of LibrarySpans from a Resource.
 message ResourceSpans {
-  // A list of Spans that originate from a resource.
+  // A list of LibrarySpans that originate from a resource.
   repeated opentelemetry.proto.trace.v1.Span spans = 1;
```
Oberon00:
```suggestion
   repeated opentelemetry.proto.trace.v1.LibrarySpan spans = 1;
```

opentelemetry/proto/common/v1/common.proto
```diff
@@ -53,3 +53,10 @@ message StringKeyValue {
   string key = 1;
   string value = 2;
 }
+
+// LibraryInfo is a message representing the instrumented library informations such as the fully
```
Oberon00: Should this be
```suggestion
// LibraryInfo is a message representing the instrumenting library informations such as the fully
```
and represent the information that comes from the "named tracer"?

Otherwise, where would this information come from?

Oberon00:
Just calling out that we currently do not have any API nor conventions to report the "instrument**ed**" library, just the "instrument**ing**" (aka reporting) library.

I did not know that "instrument**ation**" is yet another distinct thing, I thought that to be synonymous to "instrument**ing**" library.

opentelemetry/proto/collector/metrics/v1/metrics_service.proto
```diff
@@ -34,25 +35,33 @@ service MetricsService {
...
 
-  // The resource for the metrics in this message that do not have an explicit
-  // Metric.resource field set. If neither this field nor Metric.resource are set then no
-  // resource info is known.
+  // The resource for the metrics in this message.
+  // If this field is not set then no resource info is known.
   opentelemetry.proto.resource.v1.Resource resource = 2;
 }
 
+// A collection of Metrics from a LibraryInfo.
```
tigrannajaryan: I suggest moving `LibraryMetrics` to `proto/metrics/v1/metrics.proto`. This file contains declarations which are specific gRPC service declaration. 
If in the future we implement OTLP over any other transport it will need to use `LibraryMetrics`.

By the same logic `ResourceMetrics` also belongs to `proto/metrics/v1/metrics.proto`, so I think that needs to be moved too. This file will contain `MetricsService`, `ExportMetricsServiceRequest` and `ExportMetricsServiceResponse` messages which are gRPC-specific.

Same for the corresponding span messages.

bogdandrutu: Looks like you read the issue https://github.com/open-telemetry/opentelemetry-proto/issues/95 :) Will do that in a separate PR.

opentelemetry/proto/common/v1/common.proto
bogdandrutu: It can be both instrumenting and instrumented. Think if a library has a dependency on otel and instruments itself (not via a plugin).

SergeyKanzhelev:
I have the same qq as @tigrannajaryan . We already have this: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-resource-semantic-conventions.md#library 

BTW, I do believe storing library info in resource has some side effects. Change like this may be needed, I want to make sure this is what you are trying to solve

opentelemetry/proto/common/v1/common.proto
```diff
@@ -53,3 +53,10 @@ message StringKeyValue {
...
+// qualified name and version. 
+message LibraryInfo {
```
SergeyKanzhelev: language like here: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-resource-semantic-conventions.md#library ?

Also if we will want to distinguish library that instrumented code vs. SDK version that uploaded telemetry. Will SDK version be part of LibraryInfo or be part of a resource attributes?

SergeyKanzhelev:
I have the same qq as @tigrannajaryan . We already have this: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-resource-semantic-conventions.md#library 

BTW, I do believe storing library info in resource has some side effects. Change like this may be needed, I want to make sure this is what you are trying to solve.

opentelemetry/proto/common/v1/common.proto
tigrannajaryan: I expect that SDK version is part of `Resource` and follows `Library` section of semantic conventions doc that you linked to. Perhaps for clarity we need to rename the section to `Telemetry Library`. There is always one SDK per `Resource`.

There can be many _Instrumenting libraries_ (and corresponding Named Tracer/Meter) per `Resource` and this is what I think this PR suggests to represent differently.

opentelemetry/proto/common/v1/common.proto
Oberon00: A library *can* be both. But that is a special case and in all cases where it's different using the "instrument**ed**" library is wrong.

bogdandrutu:
@SergeyKanzhelev I think you got some answers, can you clarify what did we miss to explain?

opentelemetry/proto/common/v1/common.proto
SergeyKanzhelev: sorry, it was two comments in one. Splitting

SergeyKanzhelev: Spec https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-resource-semantic-conventions.md#library ? also defines a language. Should it be defined here as well?

opentelemetry/proto/collector/trace/v1/trace_service.proto
```diff
@@ -37,23 +38,33 @@ service TraceService {
...
   repeated ResourceSpans resource_spans = 1;
 }
 
-// A collection of spans from a Resource.
+// A collection of LibrarySpans from a Resource.
 message ResourceSpans {
-  // A list of Spans that originate from a resource.
-  repeated opentelemetry.proto.trace.v1.Span spans = 1;
+  // A list of LibrarySpans that originate from a resource.
+  repeated LibrarySpans library_spans = 1;
```
SergeyKanzhelev: This pattern helps with the specific problem - single process often host many libraries. This pattern can be generalized further and insted of single outing the Library info - we can have enclosed resource objects at multiple levels. So other patterns like springboot multi-app hosting in a single process or collector appending host information to different applications can also be optimized in a similar way. What do you think?

Oberon00: That's something that is that should be discussed in the spec or even an OTEP. It sounds useful but also potentially complicated to implement.

bogdandrutu: https://github.com/open-telemetry/oteps/pull/84/files

SergeyKanzhelev: Will this proposal change the implementation? Should it be renamed to component spans?

bogdandrutu:
Updated to match the current names proposed in the https://github.com/open-telemetry/oteps/pull/84

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -24,14 +24,24 @@ option java_package = "io.opentelemetry.proto.metrics.v1";
...
-  // A list of metrics that originate from a resource.
-  repeated Metric metrics = 1;
-
   // The resource for the metrics in this message.
   // If this field is not set then no resource info is known.
-  opentelemetry.proto.resource.v1.Resource resource = 2;
+  opentelemetry.proto.resource.v1.Resource resource = 1;
+
+  // A list of metrics that originate from a resource.
+  repeated InstrumentationLibraryMetrics instrumentation_library_metrics = 2;
```
tigrannajaryan:
```suggestion
  repeated InstrumentationLibraryMetrics library_metrics = 2;
```
I don't see much point in the long name. It is strongly-typed anyway.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -24,14 +24,24 @@ option java_package = "io.opentelemetry.proto.metrics.v1";
...
+}
+
+// A collection of Metrics from a InstrumentationLibrary.
+message InstrumentationLibraryMetrics {
+  // The instrumentation library information for the metrics in this message.
+  // If this field is not set then no library info is known.
+  opentelemetry.proto.common.v1.InstrumentationLibrary instrumentation_library = 1;
```
tigrannajaryan:
```suggestion
  opentelemetry.proto.common.v1.InstrumentationLibrary library = 1;
```
Same here, I think short name is easier to use in the code and is more readable.

opentelemetry/proto/common/v1/common.proto
bogdandrutu: Does the discussion from today sig meeting + https://github.com/open-telemetry/oteps/pull/84 + https://github.com/open-telemetry/opentelemetry-specification/pull/494 clarifies.

opentelemetry/proto/collector/trace/v1/trace_service.proto
bogdandrutu: The "component" changed, so we stay with library.

opentelemetry/proto/common/v1/common.proto
Oberon00: @SergeyKanzhelev As discussed today, the linked resource convention is unrelated, so "language" does not apply here (yet).

opentelemetry/proto/common/v1/common.proto
```diff
@@ -53,3 +53,10 @@ message StringKeyValue {
   string key = 1;
   string value = 2;
 }
+
+// InstrumentationLibrary is a message representing the instrumentation library informations
+// such as the fully qualified name and version. 
+message InstrumentationLibrary {
+  string name = 1;
+  string version = 2;
```
Oberon00: Me might consider using a structured way to represent the version in the future to split up the ` semver:1.2.3` into two parts.

tigrannajaryan: Do note that more fields and structure usually means slower performance and more memory allocations with ProtoBufs.
A string with well-defined conventions appears to be the right balance here.

opentelemetry/proto/common/v1/common.proto
```diff
@@ -53,3 +53,10 @@ message StringKeyValue {
   string key = 1;
   string value = 2;
 }
+
+// InstrumentationLibrary is a message representing the instrumentation library informations
```
arminru:
```suggestion
// InstrumentationLibrary is a message representing the instrumentation library information
```

You might want to reference the glossary at this point since people tend to confuse those terms.
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/glossary.md#instrumenting-library

opentelemetry/proto/common/v1/common.proto
arminru: Please specify the format or add a link to the spec.
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#obtaining-a-tracer is a bit vague on the version format, though.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -24,14 +24,24 @@ option java_package = "io.opentelemetry.proto.metrics.v1";
...
   // The resource for the metrics in this message.
   // If this field is not set then no resource info is known.
-  opentelemetry.proto.resource.v1.Resource resource = 2;
+  opentelemetry.proto.resource.v1.Resource resource = 1;
+
+  // A list of metrics that originate from a resource.
+  repeated InstrumentationLibraryMetrics library_metrics = 2;
+}
+
+// A collection of Metrics from a InstrumentationLibrary.
```
arminru:
```suggestion
// A collection of Metrics produced by an InstrumentationLibrary.
```
Otherwise it sounds like this was about "meta-"metrics regarding the instrumentation itself, not the instrumented application.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -24,13 +24,23 @@ option java_package = "io.opentelemetry.proto.trace.v1";
...
   // The resource for the spans in this message.
   // If this field is not set then no resource info is known.
   opentelemetry.proto.resource.v1.Resource resource = 1;
 
-  // A list of Spans that originate from a resource.
+  // A list of InstrumentationLibrarySpans that originate from a resource.
+  repeated InstrumentationLibrarySpans instrumentation_library_spans = 2;
+}
+
+// A collection of Spans from a InstrumentationLibrary.
```
arminru:
```suggestion
// A collection of Spans produced by an InstrumentationLibrary.
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -24,13 +24,23 @@ option java_package = "io.opentelemetry.proto.trace.v1";
...
+message InstrumentationLibrarySpans {
+  // The instrumentation library information for the metrics in this message.
```
arminru:
```suggestion
  // The instrumentation library information for the spans in this message.
```

Issue #95, opened 2020-01-14
Move Resource[Spans|Metrics] to the main proto files


bogdandrutu:
Currently this messages are defined in the collector directory but they should probably live directly in the [metrics|trace].proto files because they define the structure of the data produce by the library.

Same comment applies to the Library[Spans|Metrics] if https://github.com/open-telemetry/opentelemetry-proto/pull/94 gets merged.

PR #96, opened 2020-01-31
Move Resource[Span|Metrics] out of collector

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-proto/issues/95

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -17,11 +17,22 @@ syntax = "proto3";
 package opentelemetry.proto.trace.v1;
 
 import "opentelemetry/proto/common/v1/common.proto";
+import "opentelemetry/proto/resource/v1/resource.proto";
 
 option java_multiple_files = true;
 option java_package = "io.opentelemetry.proto.trace.v1";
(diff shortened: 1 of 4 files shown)


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -17,12 +17,23 @@ syntax = "proto3";
...
 import "opentelemetry/proto/common/v1/common.proto";
+import "opentelemetry/proto/resource/v1/resource.proto";
 
 option java_multiple_files = true;
 option java_package = "io.opentelemetry.proto.trace.v1";
 option java_outer_classname = "TraceProto";
 option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/trace/v1";
 
+// A collection of spans from a Resource.
+message ResourceSpans {
```
SergeyKanzhelev: I'm curious if order matters here? Would it be better for a very long collection of spans to read resource first and them stream thru the collection? It may be a dumb question as I was never tried to optimize anything proto-related

bogdandrutu: Actually even logically the proposed order makes more sense, don't know if a real optimization but makes sense

Issue #79, opened 2019-12-02
Remove generated code from repository


Oberon00:
The generated go code should not be checked in into the repository. See also https://github.com/open-telemetry/opentelemetry-proto/pull/77#issuecomment-559260610

Usually you should generate the files only during the build and maybe publish an artifact with them but never check them into source control (the same way you don't check in binary executables). I see #74 stated reasons to generate the code in this repository, and I think having the build files here is fine. But having the generated code here seems wrong.

Oberon00:
@tigrannajaryan What were the reasons to include the generated code anyway?

tigrannajaryan:
@Oberon00 So that the code can be consumed by other repos that depend on it. This was the approach used by [OpenCensus proto](https://github.com/census-instrumentation/opencensus-proto) and I went with it.

I am not opposed to removing generated code from this repo provided that there is a clear way to consume the proto from dependent projects.

Oberon00:
Coming from more of a Java background, I would have expected the build in the repository to publish some artifacts to some external repository instead of committing the generated files to version control.

Oberon00:
Yet another possibility would be to invoke the build script from the client projects. If it's more than a single line of client build script to do so, tune the proto build so that it becomes just a single line 😄

jmacd:
We could use Docker with a prebuilt protoc for this task.

tigrannajaryan:
Docker is also good option. We use this https://hub.docker.com/r/znly/protoc for our internal builds.

tigrannajaryan:
@open-telemetry/specs-approvers does anybody know how to generate Go ProtoBuf code correctly from a different repo?

Generated Go files import each other and import statements include full URL as specified in `go_package` option in .proto files. This is currently set to this repository's URL. The net result is that the generated files cannot be compiled because they end up importing non-existing packages.

What is the right way to do this? I was not able to find a `protoc` option to override import paths. Is one expected to manually alter the import paths after generation (or `go_package` option before generation)?

ansungchn:
> @open 원격 분석 / 사양 승인자는 다른 리포지토리에서 Go ProtoBuf 코드를 올바르게 생성하는 방법을 알고 있습니까?
> 
[...]

PR #99, opened 2020-02-05
Regenerate Go files to match latest changes

tigrannajaryan:
Go files are stale. They need to be regenerated any time .proto files
are changed. I executed `make gen-go`.

This needs to happen until https://github.com/open-telemetry/opentelemetry-proto/issues/79 is implemented.


tigrannajaryan:
@open-telemetry/specs-approvers this is a non-change from protocol perspective, it simply ensures Go files match .proto files. Please review/approve.

tigrannajaryan:
@yurishkuro thanks!

PR #100, opened 2020-02-06
Rename Event.description to name to match spec.

Oberon00:
See https://github.com/open-telemetry/opentelemetry-specification/blob/f2af73c0fb71732bc08a77ae106e4ae14154ecaf/specification/api-tracing.md#add-events

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -160,8 +160,10 @@ message Span {
     // time_unixnano is the time the event occurred.
     fixed64 time_unixnano = 1;
 
-    // description is a user-supplied text.
-    string description = 2;
+    // name of the event.
+    //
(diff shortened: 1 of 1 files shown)


tigrannajaryan:
Please re-generate go files. `make gen-go` will do it (and you may need to run one of `install-proto*` before that if this the first time).

yurishkuro:
In Jaeger we have a CI step that fails the build if rerunning proto generation produces a different results from what's checked in.

Oberon00:
@tigrannajaryan I don't have an environment ready where I want to run some global install script. If someone else could push the make gen-go changes, that would be nice. Otherwise it will take some time.

tigrannajaryan:
@Oberon00 here https://github.com/open-telemetry/opentelemetry-proto/pull/101
Please close this one.

Oberon00:
Closing in favor of #101

PR #102, opened 2020-02-11
Remove separation between label keys and values, add constant labels

bogdandrutu:
This change is required in order to be able to transform from OpenMetrics into our protocol, also allows us to export (using otel protocol) data from libraries that do not predefined the label keys.


tigrannajaryan:
@bogdandrutu can you please clarify how OpenMetrics would not fit the previous proto? If I understand the change correctly the old and new proto are logically equivalent, the difference is in what data composition you optimize for. The old format was more efficient if you had multiple data points with the same label keys but different values, the new format is more efficient if data points have different label keys or have the same key/values. Is there any data that says the later data composition is more likely?

We had previously discussed the topic of single vs multiple data points and I think the conclusion was that most likely there will always be a single data point in each metric (because that's how metrics are likely emitted). Has this understand changed since we discussed last time? If not then it seem that this change will not make a difference from efficiency perspective.

That said, the new format has the advantage of having better modeled key/value pair list (I never liked the parallel key/value arrays placed in different messages), so if all else is equal I would prefer this new proto.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -162,20 +162,16 @@ message MetricDescriptor {
   }
   Type type = 4;
 
-  // The label keys associated with the metric descriptor. Keys in this list must
-  // be unique.
-  repeated string label_keys = 5;
+  // The set of constant label associated with the metric descriptor. Labels in this list
+  // apply to all data points.
+  repeated opentelemetry.proto.common.v1.StringKeyValue constant_labels = 5;
```
jmacd: How are constant labels different from process-wide resources?

bogdandrutu: Good question, I can think of examples like this:
* "queue_provider=kafka" - this metric is for kafka.

They can be provided when the instrument is created or if we have a View definition that allows to do this:
`Aggregate instrument X as MinMaxCountSum where correlation_context.bar == "baz"` then `bar == baz` will become a constant label.

tigrannajaryan:
@bogdandrutu can you please tell what's the status of this? Are you waiting for more feedback? We need the metrics part of proto to be declared "Beta" as well so that we can start implementation in the Collector.

tigrannajaryan:
@open-telemetry/specs-approvers do we consider API and data model for traces and metrics to be at maturity level "Beta" already? This is necessary for us to declare proto maturity "Beta" as well.

SergeyKanzhelev:
Readme said it's in Alpha. I haven't seen anybody claiming otherwise. I'd suggest we merge this change.

opentelemetry/proto/metrics/v1/metrics.proto
tigrannajaryan: Is listing labels here semantically equivalent to listing the same labels and values all data points and is merely a storage optimization or they are semantically different? I am asking because the comment here and the comment describing the meaning of `labels` field in the data point messages are not the same. This one does not mention say anything about whether this labels identify the timeseries or no.

bogdandrutu:
Going to merge this. Thanks everyone for review.

PR #103, opened 2020-02-12
Declare Trace, Resource and Common as Beta

tigrannajaryan:
This adds the maturity levels for individual parts of the protocol
according to recently agreed project-wide definition of maturity levels.

After merging this change we will declare Trace (Span), Resource and
Common parts of the protocol as Beta, which means no more than once in
3 months changes are allowed after that.

We need this declaration of maturity so that we can begin implementation
of OTLP in the Collector.

One known pending change that prevents us from declaring the components
as Stable is https://github.com/open-telemetry/opentelemetry-proto/pull/94

We assume that there will be no other changes within the next 3-months timeframe.
The assumption seems reasonable since there are no other known shortcomings
in this part of the protocol that may require more modifications.

Note that metrics part is considered Alpha since it is still under active
development.


tigrannajaryan:
@open-telemetry/specs-approvers please review. This blocks Collector progress.

tigrannajaryan:
@open-telemetry/specs-approvers please have a look and if you are OK with the proposal to declare parts of the protocol "Beta" please approve. Please do note that it means we do not expect major or frequent changes to data model of traces and spans in the API/SDK anymore.

README.md
```diff
@@ -4,4 +4,19 @@ The proto files can be consumed as GIT submodule or copied over and built direct
...
+-------------------------|----------|
+collector/metrics/*      | Alpha    |
+collector/trace/*        | Beta     |
+common/*                 | Beta     |
+metrics/*                | Alpha    |
+resource/*               | Beta     |
+trace/trace.proto        | Beta     |
+trace/trace_config.proto | Alpha    |
+
+(See [maturity-matrix.yaml](https://github.com/open-telemetry/community/blob/47813530864b9fe5a5146f466a58bd2bb94edc72/maturity-matrix.yaml#L57)
```
SergeyKanzhelev: should this be described in a yaml file though?

tigrannajaryan: I wanted this to be visible and easily readable. Happy to update the yaml file in the community repo too, which currently does not currently have an entry for the protocol (I can add).

PR #108, opened 2020-02-24
Allow SummaryValue percentile of 0.0

jkwatson:
A percentile of 0.0 is useful for representing the minimum value in a Summary.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -391,7 +391,7 @@ message SummaryValue {
   // Represents the value at a given percentile of a distribution.
   message ValueAtPercentile {
     // The percentile of a distribution. Must be in the interval
-    // (0.0, 100.0].
+    // [0.0, 100.0].
     double percentile = 1;
 
     // The value at the given percentile of a distribution.


SergeyKanzhelev:
I don't think this is the right terminology. At least description doesn't match the PR. From stackoverflow:

```
A bracket means that end of the range is inclusive -- it includes the element listed. A parenthesis means that end is exclusive and doesn't contain the listed element. So for [first1, last1), the range starts with first1 (and includes it), but ends just before last1.

Assuming integers:

- (0, 5) = 1, 2, 3, 4
- (0, 5] = 1, 2, 3, 4, 5
- [0, 5) = 0, 1, 2, 3, 4
- [0, 5] = 0, 1, 2, 3, 4, 5
```

So how it is currently written it includes `0.0` and `100.0`.

jkwatson:
@SergeyKanzhelev it's currently written as `(0.0, 100.0]` which means that 0.0 is excluded. This PR is to include it.

jmacd:
FWIW I gave a more specific definition for quantile here: https://github.com/open-telemetry/opentelemetry-specification/blob/2b75442b05fd968f197422dc18124338a955f3a2/specification/sdk-metric.md#quantile-definition

bogdandrutu:
@SergeyKanzhelev the intent is to include 0.0 - min value and 100 - max value.

SergeyKanzhelev:
Oh, I am totally blind =). Of course. Sorry

PR #110, opened 2020-02-27
Generate HTTP/JSON -> gRPC gateway boilerplate in Go

nilebox:
As part of Collector receiver implementation (https://github.com/open-telemetry/opentelemetry-collector/issues/574), we need gRPC gateway, following OpenCensus receiver as an example: [opencensusreceiver/opencensus.go#L234-L242](https://github.com/open-telemetry/opentelemetry-collector/blob/d6147389a6cd0ba192e1ba0fad2f26bc3f4c655b/receiver/opencensusreceiver/opencensus.go#L234-L242).

There is already a YAML config for gRPC gateway, but `gen-go` build target is missing invocations of `protoc-gen-grpc-gateway`.


yurishkuro:
One thing that upsets me about proto-json mapping is that `bytes` fields like traceID are encoded in JSON using base64. This makes them look in JSON very different from the hex representation used in Trace-Context. And even if there is a way to swap the marshaling using gogoproto annotations, it's not compatible with other languages.

nilebox:
@yurishkuro are you suggesting some specific change? As you mentioned, custom types are not cross-language compatible, so we can't do that (https://github.com/grpc-ecosystem/grpc-gateway/issues/754). I suppose we may switch from `bytes` to `string` but that would make binary protocol less efficient?

yurishkuro:
I don't have a better suggestion. This is the main reason why we are still using custom JSON for Jaeger's /api/* instead of the one generated from proto.

install-proto-osx.sh
```diff
@@ -9,11 +9,15 @@ wget https://github.com/grpc-ecosystem/grpc-gateway/releases/download/v1.9.6/pro
     && chmod +x protoc-gen-swagger-v1.9.6-darwin-x86_64 \
     && sudo ln -s -f ${DIR}/protoc-gen-swagger-v1.9.6-darwin-x86_64 /usr/local/bin/protoc-gen-swagger
 
+wget https://github.com/grpc-ecosystem/grpc-gateway/releases/download/v1.9.6/protoc-gen-grpc-gateway-v1.9.6-darwin-x86_64 \
```
nilebox: The latest release is `v1.13.0`, but I wanted to keep it consistent with `protoc-gen-swagger` above.
Could bump both if there is no objection.

makefile
```diff
@@ -20,6 +20,8 @@ ci: gen-java gen-swagger
 .PHONY: gen-go
 gen-go:
 	$(foreach file,$(PROTO_FILES),$(call exec-command,protoc --go_out=plugins=grpc:$(GOPATH)/src $(file)))
+	protoc --grpc-gateway_out=logtostderr=true,grpc_api_configuration=opentelemetry/proto/collector/trace/v1/trace_service_http.yaml:$(GOPATH)/src opentelemetry/proto/collector/trace/v1/trace_service.proto
+	protoc --grpc-gateway_out=logtostderr=true,grpc_api_configuration=opentelemetry/proto/collector/trace/v1/trace_service_http.yaml:$(GOPATH)/src opentelemetry/proto/collector/metrics/v1/metrics_service.proto
```
nilebox: This is not a typo, `trace_service_http.yaml` contains mappings for both tracing and metrics for some reason.

gen/go/collector/metrics/v1/metrics_service.pb.go
```diff
@@ -132,11 +132,11 @@ var fileDescriptor_75fb6015e6e64798 = []byte{
 
 // Reference imports to suppress errors if they are not otherwise used.
 var _ context.Context
-var _ grpc.ClientConn
+var _ grpc.ClientConnInterface
 
 // This is a compile-time assertion to ensure that this generated file
 // is compatible with the grpc package it is being compiled against.
-const _ = grpc.SupportPackageIsVersion4
+const _ = grpc.SupportPackageIsVersion6
```
nilebox: Not sure why the gRPC version went from 4 to 6. Possibly a requirement of gRPC gateway?

nilebox: Fixed, generated using correct version of gen tools.

makefile
nilebox: Fixed, moved metrics part of the config to appropriate dir.

opentelemetry/proto/collector/trace/v1/trace_service_http.yaml
```diff
@@ -5,8 +5,5 @@ config_version: 3
 http:
  rules:
  - selector: opentelemetry.proto.collector.trace.v1.TraceService.Export
-   post: /v1/trace
```
nilebox: Not sure why `/v1/trace1` was mapped to MetricsService?
The code was moved from `opentelemetry-java` (https://github.com/open-telemetry/opentelemetry-java/issues/235), so it's hard to track the history of this change.

install-proto-osx.sh
tigrannajaryan: I think it is fine to bump both.

SergeyKanzhelev:
@nilebox to the @yurishkuro's feedback - can you post an example of JSON?

If this JSON will be used universally across the project we probably will want the same as an stdout output and similar.

(not blocking this PR, just want to use it as a step to drive consistency).

nilebox:
> can you post an example of JSON?

For example, OpenTelemetry's JSON `traceId: "7uGbfsPBsXM="` corresponds to `trace-id` `eee19b7ec3c1b173` in W3C Trace-Context, while having the same binary representation internally.
See below why.

---

@SergeyKanzhelev see JSON of OpenCensus as an example: https://github.com/open-telemetry/opentelemetry-collector/blob/6340d3c0477fcc8334c614a564e46207ebe72c46/receiver/opencensusreceiver/opencensus_test.go#L79-L80

Both OpenCensus and OpenTelemetry declare `trace_id` and `span_id` as byte array (`bytes`), which gRPC translates to JSON string via `base64` encoding.

As @yurishkuro points out, Trace-Context uses a different hex (`base16`) encoding, you can find example of `trace-id` header here: https://www.w3.org/TR/trace-context/#examples-of-http-traceparent-headers

---

> we probably will want the same as an stdout output and similar

We can use `base64` encoding in stdout and everywhere else with text representation, this is not a problem. It will still be different from Trace-Context though (which is probably fine).

install-proto-osx.sh
nilebox: Will do separately to see if there any side effects of that.

nilebox:
Is anything blocking this PR from being merged?

tigrannajaryan:
This is blocking OTLP receiver implementation in Collector.

nilebox:
@owais @jmacd Can you please merge this PR? I don't have permissions to do that myself.

install-proto-osx.sh
bogdandrutu: Please consider to move this changes to bump protoc version in a separate PR.

bogdandrutu:
@nilebox can you do the split to bump the protoc in a separate PR?

install-proto-osx.sh
nilebox: I didn't change version, I added installation of the new `protoc-gen-grpc-gateway` binary that is required for generating gRPC gateway boilerplate, matching the version of existing `protoc-gen-swagger`.

nilebox:
> can you do the split to bump the protoc in a separate PR?

@bogdandrutu as I replied in the inline comment, I didn't bump `protoc`, but just added a new binary that is required for this PR.

tigrannajaryan:
@open-telemetry/specs-approvers we need more reviews on this.

bogdandrutu:
@yurishkuro are you ok with merging this as it is, since this is not necessary the official opentelemetry protocol over HTTP, is just an extension to our gRPC protocol? Also I would open an issue to discuss what to do in the future about this issue: 1. change trace/span id to hex? 2. define a different json schema for otlp over HTTP? etc.

yurishkuro:
no objections to merge

PR #112, opened 2020-02-27
[WIP] Validate in CI that generated Go files are in sync with protos

nilebox:
1. Currently in CI `gen-go` doesn't get executed at all.
2. There is no validation that generated Go files are up to date.

This change adds `validate-go` target executed as part of `make ci`.

### DIFF
--- a/gen/go/metrics/v1/metrics.pb.go
+++ b/gen/go/metrics/v1/metrics.pb.go
@@ -1077,7 +1077,7 @@ func (m *SummaryValue) GetPercentileValues() []*SummaryValue_ValueAtPercentile {
 // Represents the value at a given percentile of a distribution.
 type SummaryValue_ValueAtPercentile struct {
 	// The percentile of a distribution. Must be in the interval
-	// (0.0, 100.0].
+	// [0.0, 100.0].
 	Percentile float64 `protobuf:"fixed64,1,opt,name=percentile,proto3" json:"percentile,omitempty"`
 	// The value at the given percentile of a distribution.
 	Value                float64  `protobuf:"fixed64,2,opt,name=value,proto3" json:"value,omitempty"`
--- a/makefile
+++ b/makefile
@@ -33,8 +33,8 @@ validate-go: gen-go
 # Generate ProtoBuf implementation for Go.
 .PHONY: gen-go
 gen-go:
-	$(foreach file,$(PROTO_FILES),$(call exec-command,protoc --go_out=plugins=grpc:$(GOPATH)/src $(file)))
 	rm -rf ./$(GENDIR)/go
+	$(foreach file,$(PROTO_FILES),$(call exec-command,protoc --go_out=plugins=grpc:$(GOPATH)/src $(file)))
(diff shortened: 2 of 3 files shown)


gen/go/metrics/v1/metrics.pb.go
```diff
@@ -1077,7 +1077,7 @@ func (m *SummaryValue) GetPercentileValues() []*SummaryValue_ValueAtPercentile {
 // Represents the value at a given percentile of a distribution.
 type SummaryValue_ValueAtPercentile struct {
 	// The percentile of a distribution. Must be in the interval
-	// (0.0, 100.0].
+	// [0.0, 100.0].
```
nilebox: This is a real (though harmless) example that currently in master generated Go files are out of sync.

makefile
```diff
@@ -12,9 +12,23 @@ $(1)
...
+then \
+	echo "Git tree is not clean. Did you forget to commit some files?" ;\
+	git --no-pager status ;\
+fi
+endef
+
 # CI build
 .PHONY: ci
-ci: gen-java gen-swagger
+ci: validate-go gen-java gen-swagger
```
nilebox: Order matters: we need to run `validate-go` before generating Java and OpenAPI definitions, otherwise the git directory will be dirty before the check.

makefile
```diff
@@ -12,15 +12,35 @@ $(1)
 
 endef
 
+# Function to make sure that the git directory is clean.
+define check-git-status
+if [[ -n $$(git --no-pager status -s 2> /dev/null) ]] ;\
+then \
+	echo "Git tree is not clean. Did you forget to commit some files?" ;\
```
tigrannajaryan: This is also printed when generated files are not in sync, right? Perhaps mention that in the error message.

makefile
```diff
@@ -12,15 +12,35 @@ $(1)
...
+gen-go-ci:
+	rm -rf ./$(GENDIR)/go
+	$(foreach file,$(PROTO_FILES),$(call exec-command,protoc --go_out=plugins=grpc:$(GOPATH)/src $(file)))
 
 # Generate ProtoBuf implementation for Go.
 .PHONY: gen-go
 gen-go:
-	$(foreach file,$(PROTO_FILES),$(call exec-command,protoc --go_out=plugins=grpc:$(GOPATH)/src $(file)))
 	rm -rf ./$(GENDIR)/go
+	$(foreach file,$(PROTO_FILES),$(call exec-command,protoc --go_out=plugins=grpc:$(GOPATH)/src $(file)))
```
tigrannajaryan: This seems identical to lines 36-37 (I cannot see a difference if there is). If so why not `gen-go: gen-go-ci` or some other way to reuse it?

nilebox: The difference between `gen-go` and `gen-go-ci` is that apparently in Travis we check out sources into GOPATH, so `cp` fails to copy from directory to itself.

Agreed, need to deduplicate this.

nilebox:
I will send a new PR which doesn't rely on `git status`, since Travis CI handles it poorly due to checking out sources with "detached HEAD".

PR #113, opened 2020-03-02
(2) Run gen-go target in CI

nilebox:
This makes sure that `make gen-go` succeeds.
A few changes to make `gen-go` runnable in CI, and add it to `make ci` target.

- [x] Merge #115 first for changes in `gen/go` to disappear.


makefile
```diff
@@ -14,14 +16,18 @@ endef
 
 # CI build
 .PHONY: ci
-ci: gen-java gen-swagger
+ci: gen-go gen-java gen-swagger
```
nilebox: Change 1: Run `gen-go` as part of `make ci` in Travis CI.

makefile
```diff
@@ -14,14 +16,18 @@ endef
...
 
 # Generate ProtoBuf implementation for Go.
 .PHONY: gen-go
 gen-go:
-	$(foreach file,$(PROTO_FILES),$(call exec-command,protoc --go_out=plugins=grpc:$(GOPATH)/src $(file)))
 	rm -rf ./$(GENDIR)/go
-	cp -R $(GOPATH)/src/github.com/open-telemetry/opentelemetry-proto/$(GENDIR)/go ./gen/
```
nilebox: Change 2: This line fails if the source is checked out into `$(GOPATH)/src/github.com/open-telemetry/opentelemetry-proto`, which is actually a canonical path for Go, and it's being used by Travis CI.

So I made this `cp` conditional: we only need to copy generated files if we are in the custom directory.

makefile
```diff
@@ -14,14 +16,18 @@ endef
...
 # CI build
 .PHONY: ci
-ci: gen-java gen-swagger
+ci: gen-go gen-java gen-swagger
 
 # Generate ProtoBuf implementation for Go.
 .PHONY: gen-go
 gen-go:
-	$(foreach file,$(PROTO_FILES),$(call exec-command,protoc --go_out=plugins=grpc:$(GOPATH)/src $(file)))
 	rm -rf ./$(GENDIR)/go
```
nilebox: Also moved the `rm` command to the top, otherwise we delete Go files we just generated, when the source is checked out into GOPATH (see below).

PR #114, opened 2020-03-02
(3) Verify in CI that gen/go files are in sync with protos

nilebox:
The script is simple:
- Calculate the checksum of `gen/go` directory
- Run `make gen-go`
- Calculate the checksum of `gen/go` directory again
- If the checksum has changed, the `gen/go` directory is out of sync

NOTE: This is part 3 of changes. If all 3 PRs look good, we can just merge this one to avoid extra git rebasing between merges.

- [x] Merge #115 first for changes in `gen/go` to disappear.
- [x] Merge #113 with changes in the makefile


makefile
```diff
@@ -1,5 +1,7 @@
+GOPATH_DIR := $(GOPATH)/src/github.com/open-telemetry/opentelemetry-proto
```
nilebox: Changes in the makefile are from #113, please ignore.

verify-go.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+set -e
+
+gen_checksum()
+{
+    find ./gen -type f -exec md5sum {} \; | sort -k 2 | md5sum
```
nilebox: This is more portable than `git status` which didn't work with detached HEAD in Travis CI.

verify-go.sh
```diff
@@ -0,0 +1,27 @@
...
+}
+
+print_checksum_details()
```
nilebox: This prints checksum for each file separately, might be useful for debugging to see which specific files are out of sync.
If this is to noisy, I may remove this.

verify-go.sh
bogdandrutu: Can you please add this as a comment in the file?

nilebox: Done, added comments to functions.

verify-go.sh
```diff
@@ -0,0 +1,27 @@
...
+{
+    find ./gen -type f -exec md5sum {} \; | sort -k 2
+}
+
+print_checksum_details
+CHECKSUM_BEFORE=$(gen_checksum)
+echo "checksum before: ${CHECKSUM_BEFORE}"
+make gen-go
```
nilebox: Now we are running `gen-go` twice in CI - here and in https://github.com/open-telemetry/opentelemetry-proto/blob/f09a72cf8c520c6974cea8cb2f7907e34fd81fe9/makefile#L19

The one inside `ci` can be removed if you think that's worth it.
OTOH, if it doesn't add to much extra time, it might still be useful for `make ci` to generate all files at once, e.g. when testing on local machine.

PR #116, opened 2020-03-05
Flatten data point definitions

bogdandrutu:
Historical decisions:
We decided to have values separated from the data points because we may add a repeated values in the future.

Proposal:
Flatten the current protos (avoid unnecessary extra messages/allocations). If in the future we need to optimize (not duplicate the labels) we can have a Timeseries message { labels, repeated DataPoints} where labels in the DataPoints are not present (or redefine the Timeseries completely). Not sure we need that, so better to keep the current definition clean.


bogdandrutu:
@open-telemetry/specs-approvers please review :)

PR #119, opened 2020-03-06
Flatten bucket options for explicit bounds

bogdandrutu:
* Also it generates the new go files.


bogdandrutu:
@tigrannajaryan in the current design let's assume we add ExponentialBuckets definition, still that will not be part of the ExplicitBucket message. So we will have a:

oneof {  // the oneof can be explicit or implicit by convention.
  repeated double explicit_buckets;
  ExponentialBuckets exponential_buckets
}

We don't need to have the ExplicitBucket message, or maybe I miss something.

bogdandrutu:
@tigrannajaryan in other words having a different message for ExplicitBuckets only gives us the capability to add more properties for this type of defining buckets, which I fail to see how we can enhance that in the future.

PR #120, opened 2020-03-10
Use trace_state instead of tracestate

bogdandrutu:
This is to confirm with almost all the languages that use `TraceState` and not `Tracestate`.


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -67,7 +67,7 @@ message Span {
   // tracestate conveys information about request position in multiple distributed tracing graphs.
   // It is a tracestate in w3c-trace-context format: https://www.w3.org/TR/trace-context/#tracestate-header
   // See also https://github.com/w3c/distributed-tracing for more details about this field.
-  string tracestate = 3;
+  string trace_state = 3;
```
SergeyKanzhelev: I'm confused. We intentionally made it a single word in W3C to avoid choosing the separator between words. Can we revert this?

Oberon00: --> #130

PR #122, opened 2020-03-10
Add MinMaxSumCount data type

MrAlias:
The specification includes this as a default aggregation type, however, the proto has no way to transfer this data natively. This adds a new data type and datapoints to be used in this transfer.


bogdandrutu:
We decided to use Summary with 0 percentile min and 1 percentile max

jmacd:
(https://gitter.im/open-telemetry/metrics-specs?at=5e5050594880f07ed1ee45c9)

MrAlias:
@jmacd & @bogdandrutu: further down in the conversation that @jmacd linked both @c24t and @jkwatson continued to discuss the fact that the OpenTelemetry proto definition did not match the OpenTelemetry specified aggregations, and there was desire there to have them match.

It seems like this proto definition is a loud echo (or a direct renaming) of the OpenCensus proto. As such, it seems like this kind of hold over is going to provide future confusion and friction. Parts of the OpenTelemetry project will be bound by a previous specification's design choices.

The solution @bogdandrutu proposed indeed _can_ be used to solve this transport mismatch, but _should_ it? I think for the same reasons we decided to include more than just `Measurement` instruments in the metric API we should include message types that match OpenTelemetry specified aggregations. It makes the project a cohesive set of ideas instead of a patch work one.

Are there other reasons besides an opposition to changing what was inherited that are leading you to oppose this inclusion? I'm interested to understand both of your reasons on this issue, and guessing I've overlooked something.

bogdandrutu:
Your comment makes me think that we actually misdefined what is the difference between `Aggregation` and `Aggregator`.

The `Aggregation` represents the result of an `Instrument` + `Aggregator`, and is defined in a generic way so that different combinations of `Instruments` and `Aggregators` can produce the same `Aggregation`. As an example `DDSketch` and `MinMaxSumCount` (or any other algorithm that allows to calculate quantiles) aggregators applied to a Measure instrument will both produce summary aggregation (with different level of confidence and .

Another reason is that the cardinality of the combinations `Instruments` + `Aggregators` will be very high and make it impossible to define an object for every `Aggregation`.

I would be happy to add a new Aggregation in the proto, if that gives us a clear benefit, but for this specific request I do not see what is the advantage to know that this is `MinMaxSumCount` vs `Summary`.

 > It seems like this proto definition is a loud echo (or a direct renaming) of the OpenCensus proto. As such, it seems like this kind of hold over is going to provide future confusion and friction. Parts of the OpenTelemetry project will be bound by a previous specification's design choices.

We are trying to also stay compatible with OpenMetrics protocol which does not define a MinMaxCountSum representation.

jkwatson:
What if we add an optional `minimum` to the Summary? The thing that bothers me the most is that a percentile value of 0.0 is mathematical nonsense. We're shoehorning the minimum into that percentile slot because we don't have a proper place for it, rather than it being a good idea.

bogdandrutu:
@jkwatson that is an interesting reason and I would I would like to document a bit about that. I personally didn't know that is the case, will investigate more. But hope that my explanation makes sense why we don't need an aggregations for every possible aggregators.

jkwatson:
To quote wikipedia:
```
A percentile (or a centile) is a measure used in statistics indicating the value below which a given percentage of observations in a group of observations falls. For example, the 20th percentile is the value (or score) below which 20% of the observations may be found. Similarly, 80% of the observations are found above the 20th percentile.
```

so, there is, by definition, nothing in the "0th" percentile. :)

MrAlias:
I've realized I probably should have opened an issue prior to this to have the discussion there. Going to close in favor of #125 where I hope we can come to a solution to better support OpenTelemetry aggregations.

Issue #125, opened 2020-03-13
Metrics Proto Does Not Support the OpenTelemetry Specification


MrAlias:
As language SIGs include support to export metrics using the OTLP it is being discovered that the native export format they are exporting from do not fit nicely or at all into the OTLP structure. This issue is intended to track these friction points, proposed changes, and action items to resolution.

Ideally the OTLP would natively support the output of all combination of `Instrument`s and `Aggregator`s defined in the OpenTelemetry Specification. Currently, this is not the case.

### [First Class Support For Min and Max Values of a MinMaxSumCount Aggregation](https://github.com/open-telemetry/opentelemetry-proto/pull/122)

There is no metric kind for a `MinMaxSumCount` aggregator that outputs the minimum, maximum, sum, and count of events observed.

It has been suggested that a `Summary` metric kind can be used to transport these values. The `Summary` kind does have fields for the count and sum, however, it does not have dedicated fields for the maximum nor minimum.

The suggested work around here is to send the maximum as the 100th percentile and the minimum as the 0th percentile. Both of which are not obvious and the latter is mathematically incorrect (the 0th percentile is the value which 0% of events occured, where as the minimum is the minimal value where at least 1 event occured).

### [HistogramValue.Bucket.Exemplar is inadequate](https://github.com/open-telemetry/opentelemetry-proto/issues/81)

As outlined in the linked ticket:

1. It's not possible to provide exemplars for non-histograms
2. It's not possible to provide more than one exemplar per bucket
3. It requires converting certain data types to a string, which will require additional specification

### Instrumentation Instead of Aggregation

The OTLP metric kinds are centered around instruments not output from Aggregation:

```
    ...
    GAUGE_INT64 = 1;
    GAUGE_DOUBLE = 2;
    GAUGE_HISTOGRAM = 3;
    COUNTER_INT64 = 4;
    COUNTER_DOUBLE = 5;
    CUMULATIVE_HISTOGRAM = 6;
    SUMMARY = 7;
```

The incongruence of the instruments these kinds were modeled after and the actual instruments of the OpenTelemetry Specification is secondary to the fact that they are modeled after instruments in the first place.

If the goal of the OTLP is to transport the output of `Instrument` -> `Aggregator`, it should be modeled after the output of the `Aggregator`s, aggregations. And while, yes, the `Histogram` is one of these aggregations, the included histogram kinds are conflated with instrument qualifiers (`Gauge`, `Cumulative`).

[...]

PR #127, opened 2020-03-13
Update SummaryDataPoint percentile comment

MrAlias:
Include information on the way we are fitting the `MinMaxSumCount` aggregation into the `Summary` metric kind.

Relates to #125


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -340,6 +340,18 @@ message SummaryDataPoint {
   double sum = 5;
 
   // Represents the value at a given percentile of a distribution.
+  //
+  // To support the Min and Max values of a MinMaxSumCount aggregation the
+  // following conventions are used:
```
bogdandrutu: I would not make a direct reference to `MinMaxSumCount`:

To record Min and Max values following conventions are used:
  //  - The 100th percentile is equivalent to the maximum value observed.
  //  - The 0th percentile is equivalent to the minimum value observed.

Also I couldn't find a good source to confirm that 0/100 are not mathematically correct (I found some sources saying that they may not be generally accepted), but I would not put a strong phrase here without a good proof.

MrAlias: I'm not entirely sure why we wouldn't include a reference to the `MinMaxSumCount` aggregation as OpenTelemetry implementations will expect guidance from the OTLP as to what should be done with this native aggregation. It seems like I've missed the target audience of who will be using this OTLP. Who do you envision using this?

As for the mathematical correctness, from the linked ticket in the comment:

> the 0th percentile is the value which 0% of events occured, where as the minimum is the minimal value where at least 1 event occured

It is based on the definition of a percentile that the incorrectness is determined.

bogdandrutu: Where is the source of the comment in the issue?

What I found so far:
* https://www.quora.com/How-do-you-calculate-the-0th-and-the-100th-percentile-of-a-group-of-set-of-numbers
* Even on the wikipedia there are mentions about 0th and 100th "Note that in theory the 0th percentile falls at negative infinity and the 100th percentile at positive infinity, although in many practical applications, such as test results, natural lower and/or upper limits are enforced.".

So I am not sure where the sentence "0th percentile is mathematically incorrect" comes from.

bogdandrutu: @MrAlias I don't care that you include a reference to that, but what about DDSketch do we include a reference to that as well? Or any other algorithm that can produce percentiles?

MrAlias: > I don't care that you include a reference to that, but what about DDSketch do we include a reference to that as well? Or any other algorithm that can produce percentiles?

Ah! Makes sense. I'll remove it.

With regard to the source of the comment in the issue, it was primarily based my prior work with percentiles, but the Wikipedia article you linked also defines them the same way:

> A percentile (or a centile) is a measure used in statistics indicating the value below which a given percentage of observations in a group of observations falls. For example, the 20th percentile is the value (or score) below which 20% of the observations may be found. Similarly, 80% of the observations are found above the 20th percentile.

From that it follow that the 0th percentile is the value below which 0% of the observations may be found, meaning zero events.

Again, I was pulling the definition of a minimum from past work, but as [defined by Wikipedia](https://en.wikipedia.org/wiki/Sample_maximum_and_minimum):

> In statistics, the sample maximum and sample minimum, also called the largest observation and smallest observation, are the values of the greatest and least elements of a sample.

From this definition it follows that _at least_ one event had this value.

MrAlias: I also just realized it might not be so much the correctness of the statement as it is the strength of it which is the main reason you wanted me to take it out. Sorry, I think I missed that on the first read through. Updated to the suggested language.

bogdandrutu:
Please rebase so I can merge.

Issue #128, opened 2020-03-16
Take beta declaration serious or remove it.


Oberon00:
https://github.com/open-telemetry/opentelemetry-proto/pull/103 declared large parts of the protocols as beta, but purely cosmetic yet breaking changes like https://github.com/open-telemetry/opentelemetry-proto/pull/120 were still incorporated. What's the definition of beta, this repository is following?
(I think our official definition is a bit over the top https://github.com/open-telemetry/community/blob/master/maturity-matrix.yaml#L43-L49)

tigrannajaryan:
I agree, we should take beta declaration seriously. Note that we are still allowed to make braking changes but very infrequently. According to API maturity level [definitions](https://github.com/open-telemetry/community/blob/master/maturity-matrix.yaml):
>"beta" maturity APIs should not introduce backwards-incompatible changes
 more than once every three months; and when those changes are introduced, the
 authors will make a best effort to provide compatibility bridges.

Given that there are no existing OTLP implementations yet arguably we are still OK with making changes, but I agree, we should all begin to respect the definition of "beta" or agree that it is too early and remove the label. The later will have ramifications: work that depends on protocol being "beta"-stable will be halted.

SergeyKanzhelev:
should we revert the #120 than or keep it? I also filed an issue that the change may not be great for interoperability

Oberon00:
Good question. Actually, I think a (beta) release should be tagged (currently there are no tags at all https://github.com/open-telemetry/opentelemetry-proto/tags), otherwise the beta declaration is either meaningless or every commit counts as release.

SergeyKanzhelev:
Totally agree. Tag would be very helpful. @bogdandrutu revert or keep the change?

tigrannajaryan:
I suggest we keep #120. It is done and already used by dependent repos, reverting is worse.

Oberon00:
So the action item here is to tag a release?

Oberon00:
Since a release was tagged now, are we fine with keeping the beta-"promise" in the README?

CC @bogdandrutu

bogdandrutu:
What does "beta promise" means?

Oberon00:
@bogdandrutu See @tigrannajaryan's comment: https://github.com/open-telemetry/opentelemetry-proto/issues/128#issuecomment-599686709

tigrannajaryan:
I think we recently made a sufficient efforts to be serious about maturity declarations. I suggest that we close this issue.

PR #132, opened 2020-03-20
Fix consistency for timestamp vs time and use unix_nano

bogdandrutu:
This is a small breaking change, but not for the on-wire format so binaries build with an older version will be able to talk to the collector still. Only when updating the protos will cause breaking.


Oberon00:
I liked _unixnano better.

bogdandrutu:
@Oberon00  unixnano is not a word :( all the IDEs that I use complain about that.

SergeyKanzhelev:
should we do it after tagging the beta (https://github.com/open-telemetry/opentelemetry-proto/issues/128) or after?

bogdandrutu:
I will do it immediately after this. An official beta release

SergeyKanzhelev:
@Oberon00 do you still have concerns? If not, let's wrap up the beta

Oberon00:
I am a bit concerned about the renamings in tracing/ but I think I'm not particularly affected, so for me it's OK to tag a beta with these changes.

SergeyKanzhelev:
thank you for confirming

Issue #81, opened 2019-12-10
HistogramValue.Bucket.Exemplar is inadequate


jmacd:
The HistogramValue Bucket supports **one** exemplar per bucket, with exemplar values being restricted to **string**-valued attributes. This is particularly restrictive, for a number of reasons:

1. It's not possible to provide exemplars for non-histograms
2. It's not possible to provide more than one exemplar per bucket
3. It requires converting certain data types to a string, which will require additional specification

I would remove this support entirely until a full set of requirements can be drafted. (1) It's meaningful and reasonable to support encoding exemplars for any kind of metric instrument. (2) It's certainly useful to provide more than one exemplar per instrument/value-range, as a means of conveying a sample distribution, (3) There is a common request (e.g., https://github.com/open-telemetry/opentelemetry-specification/issues/381, https://kccncna19.sched.com/event/UaXX/deep-linking-metrics-and-traces-with-opentelemetry-openmetrics-and-m3-rob-skillington-chronosphere) to include trace context in these exemplars, and forcing them to be encoded as string-attribute lists is onerous (and inefficient)--probably trace context should be specialized.

PR #137, opened 2020-04-15
Proposal for metric proto refactor

MrAlias:
Update metric descriptor to contain information on the data qualities.
This includes:

 - The meaning of the time interval measurements were made over is now described by the `interval` value.
 - The monotonic nature of the data is captured if known.
 - If the data values are restricted to a subset of numbers.

The `SummaryDataPoint` and `HistogramDataPoint` were merged into a unified `Distribution` message that contains statistics about the observed population of values.

Adds support for multiple histogram bucket definitions and adds a linear bucket definition message.

Includes an explicit minimum and maximum for a `Distribution` (#122).

Removes the exemplar code (#81). This can be added back in as a more generalized case now given the new `Data` structure of the Metrics. But, it has been left for a separate PR.

Unifies the common parts of the previous `*DataPoints` into a unified `Data` message.

This is not complete:

- [ ] It likely needs better names.
- [ ] it certainly needs documentation work.
- [ ] it absolutely needs feedback on the approach.

I wanted to get this posted sooner rather than later to not let perfection hinder progress.

Related to #125
Fixes to #81


MrAlias:
I've marked this as a draft as this is still pretty rough and I expect to review and iterate on this. However, I would still appreciate anyone who has participated on #125 to take a look (@tigrannajaryan, @c24t , @jkwatson , @jmacd ) and provide any thoughts you have.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -118,255 +145,122 @@ message MetricDescriptor {
...
-  // A gauge is an instantaneous measurement of a value.
-  //
-  // A counter/cumulative measurement is a value accumulated over a time
-  // interval. In a time series, cumulative measurements should have the same
-  // start time, increasing values, until an event resets the cumulative value
-  // to zero and sets a new start time for the subsequent points.
-  enum Type {
-    // Do not use this default value.
+  // Interval describes how metric values are reported over time.
+  enum Interval {
```
jkwatson, line 149: I don't love this name, since INSTANTANEOUS is kind of "not an interval". Is `TemporalQuality` too verbose?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -118,255 +145,122 @@ message MetricDescriptor {
...
-
-  // time_unix_nano is the moment when this value was recorded.
-  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
-  fixed64 time_unix_nano = 3;
-
-  // value itself.
-  int64 value = 4;
-}
+  // Subset describes the subset of numbers metric values belong to if any.
+  enum Subset {
```
jkwatson, line 190: What about `NumericQuality` to go along with the proposed `TemporalQuality` ?

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: I originally was going to go with `Temporality` or similar `Temporal*` but was worried about the definition being confused with the [a secular possession definition](https://www.merriam-webster.com/dictionary/temporality):

> 1a: civil or political as distinguished from spiritual or ecclesiastical power or authority
> 1b: an ecclesiastical property or revenue —often used in plural

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -87,23 +87,50 @@ message InstrumentationLibraryMetrics {
...
+    fixed64 start_time_unix_nano = 2;
+
+    // time_unix_nano is the time when this data point was recorded. It MUST
+    // be represented as the UNIX Epoch time in nanoseconds (nanoseconds since
+    // 00:00:00 UTC on 1 January 1970). If zero, it is treated as unspecified
+    // and may be set by the receiver.
+    fixed64 time_unix_nano = 3;
+
+    // Value is the measured value.
+    oneof Value {
```
c24t, line 128: This means `metric.data` might have a mix of value types right? Vs. https://github.com/open-telemetry/opentelemetry-proto/pull/137/files#diff-ad849336b0a70e8a18346421908f69edL103-L106.

jmacd, line 128: `oneof` has a high cost, unfortunately. Either the recently-released (new) Go protobuf library can help or improve on this, otherwise I suggest we just remove it and document that only one of the three types should be present, defined accordingly in the metric descriptor.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -118,255 +145,122 @@ message MetricDescriptor {
...
-  // To record Min and Max values following conventions are used:
-  // - The 100th percentile is equivalent to the maximum value observed.
-  // - The 0th percentile is equivalent to the minimum value observed.
-  //
-  // See the following issue for more context:
-  // https://github.com/open-telemetry/opentelemetry-proto/issues/125
-  message ValueAtPercentile {
+// Distribution is a data point in a timeseries containing statistics for
+// an observed population of values.
+message Distribution {
```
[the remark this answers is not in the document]
c24t: Nevermind, I see you've got "documentation work" in the PR description.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: I don't like `Interval`, it reminds me that sometimes statisticians distinguish the kind of a number as being an "interval" or "ratio" measurement, a distinction that is close to the debates we're having in OTEP 93.

`Temporality` sounds good to me; I personally will have no confusion over the ecclesiastical usage. `TemporalQuality` is good too. :)

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -87,23 +87,50 @@ message InstrumentationLibraryMetrics {
...
-  repeated Int64DataPoint int64_data_points = 2;
-  repeated DoubleDataPoint double_data_points = 3;
-  repeated HistogramDataPoint histogram_data_points = 4;
-  repeated SummaryDataPoint summary_data_points = 5;
+  repeated Data data = 2;
+
+  message Data {
+    // The set of labels that uniquely identify this timeseries.
+    //
+    // These labels take precedence over any labels with the same key defined
```
jmacd, line 99: May we treat a duplicate label key found in the Data vs. the Descriptor as an error?

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: I've been trying to use the word "Kind" to indicate something that belongs to a certain category, and "Type" when it refers to a certain value. Maybe "TemporalKind"?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -87,23 +87,50 @@ message InstrumentationLibraryMetrics {
...
+    //
+    // If this data point is a part of a timeseries for an DELTA, or
+    // CUMULATIVE metric, this value represents the instant before the
+    // measuring time interval began (i.e. for time t to be in the measuring
+    // interval than start_time_unix_nano < t <= time_unix_nano).
+    //
+    // This time MUST be reset so as to represent a new measurement lifetime
+    // if this data point is part of a timeseries for a CUMULATIVE metric and
+    // the metric value is reset to zero.
+    fixed64 start_time_unix_nano = 2;
```
jmacd, line 119: I've read and re-read this explanation, and I think you're saying what I think we want, but we should clarify this, especially for cumulative values. I found a helpful explanation in Stackdriver's documentation:

https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries#point

> The time interval to which the data point applies. For GAUGE metrics, the start time is optional, but if it is supplied, it must equal the end time. For DELTA metrics, the start and end time should specify a non-zero interval, with subsequent points specifying contiguous and non-overlapping intervals. For CUMULATIVE metrics, the start and end time should specify a non-zero interval, with subsequent points specifying the same start time and increasing end times, until an event resets the cumulative value to zero and sets a new start time for the following points.

For DELTA instruments: can we recommend that start and end times equal the start and end time of the interval? If so, in a simple configuration where all metrics are collected on the same interval, all Instantaneous and Delta measurements can infer their end (Instantaneous) or start/end (Delta) time from the report and be skipped in the Data field.  At least I think we should have an option to leave out timestamps when they can be inferred from the report.

We have to pay particular attention to how cumulatives are reported.  The start-time of a cumulative should remain constant as long as the cumulative is not reset, as stated in the Stackdriver docs. It means that the SDK will be responsible for remembering cumulative values that are actively being reported.  When some number of intervals pass without reporting that cumulative, the SDK may forget it, as follows.

1. Remember the end timestamp fo the last interval that reported the metric
2. Reset **all** values for the instrument, retain only the memory of the end timestamp
3. Any new cumulative value for the instrument (i.e., distinct label set) will be reported as having a start time equal to the last timestamp where the instrument was reset.

Item (2) ensures that we can aggregate cumulative values for a single metric within the process, since all values will have the same reset time.  This does not ensure that multiple instruments can be aggregated in the process: an error condition would arise as we try to aggregate multiple cumulative values with different star times.  The same condition may arise in a distributed setting, where if multiple processes report the same metric with different cumulative start times, aggregation is not possible.  I suggest we leave it at that, there is a known condition where cumulatives cannot be aggregated except when they have the same start time.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -87,23 +87,50 @@ message InstrumentationLibraryMetrics {
...
+    // time in nanoseconds (nanoseconds since 00:00:00 UTC on 1 January 1970).
+    // If zero, it is treated as unspecified and may be set by the receiver.
+    //
+    // If this data point is a part of a timeseries for an INSTANTANEOUS
+    // metric, this value is ignored.
+    //
+    // If this data point is a part of a timeseries for an DELTA, or
+    // CUMULATIVE metric, this value represents the instant before the
+    // measuring time interval began (i.e. for time t to be in the measuring
+    // interval than start_time_unix_nano < t <= time_unix_nano).
```
jmacd, line 114: This `start < t <= time` bothers me.  Why not `start <= t < time`? I state this because it's possible for an event to happen concurrently with the start of a time window and be recorded in that window, but it's not possible for an event to happen concurrently with the end of a window and be included in the window.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: I'm debating whether the descriptor should indicate the expected Data field, or the Data struct should do so itself. I can see it both ways. It's simpler if the descriptor identifies a constant field to be expected. I've also described how I believe that we should be able to report single-value distributions using a single value, which means that an observer reporting a distribution should report a scalar value when there's no aggregation, but should report a distribution value when there is aggregation.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -118,255 +145,122 @@ message MetricDescriptor {
...
-  Type type = 4;
+  Interval interval = 4;
 
-  // The set of labels associated with the metric descriptor. Labels in this list apply to
-  // all data points.
-  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 5;
-}
+  // monotonic is true if the metric values are entirely non-increasing, or
+  // entirely non-decreasing.
+  bool monotonic = 5;
```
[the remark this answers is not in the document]
jmacd: It's easier to describe if we encode the concrete types as an enum, but less forward-compatible if we add more and more instruments. It's harder to describe if we include refinements individually. (I'm still thinking about this...)

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: I'm onboard abandoning `Interval`. I'm leaning towards `Temporality`, but not entirely opposed to `TemporalQuality`. The `*Quality` part feels a bit redundant (especially if going with the suggestion to make these all "`refinement`" flags) but could be convinced otherwise.

@jkwatson : thoughts?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -118,255 +145,122 @@ message MetricDescriptor {
...
-
-      // exemplar_attachments are contextual information about the example value.
-      // Keys in this list must be unique.
-      repeated opentelemetry.proto.common.v1.StringKeyValue attachments = 3;
-    }
-
-    // exemplar is an optional representative value of the bucket.
-    Exemplar exemplar = 2;
+    // NONPOSITIVE is the set of numbers less than or equal to zero.
+    NONPOSITIVE  = 2;
```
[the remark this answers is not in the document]
MrAlias: Could be used if measuring your enjoyment of a global pandemic :trollface:.

Yeah, it was just added for completeness. Do you think we should just remove it?

opentelemetry/proto/metrics/v1/metrics.proto
jkwatson: I'm fine with `Temporality`.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: the way how "time" is implemented in the current systems cannot guarantee a happens-before relationship so I would not bother to justify what should happen if an event happens at that same moment - unless there is another way to synchronize these events time cannot guarantee the ordering of the events.

> I state this because it's possible for an event to happen concurrently with the start of a time window and be recorded in that window, but it's not possible for an event to happen concurrently with the end of a window and be included in the window.

If an event happens at the same time (if you read `time.Now` exactly when the event happens) with closing the time we close the window (if you read `time.Now`) there is no guarantee if the event will be included in the current window or the next window unless there are other mechanisms to ensure the happens before relationship, so none of the option is ideal.

The idea behind using this definition of the (start, end] is that when using an Observer the value is "added" at the end time so it feels kind of more natural this way - again none of the versions can be 100% correct, so we also look at what others do like OpenMetrics/Stackdriver those being the system which have start time.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -118,255 +145,122 @@ message MetricDescriptor {
...
+
+// LinearBounds define evenly spaced bucket bounds.
+//
+// This defines number + 2 (= N) buckets. The boundaries for buckets at
+// index i are:
+//
+// (-infinity, minimum) for i == 0
+// [minimum + ((i-1)*width), minimum + (i*width)) for 0 < i < N-1
+// [minimum + (number*width), +infinity) for i == N-1
+message LinearBounds {
```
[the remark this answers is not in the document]
MrAlias: Yeah, I was mostly including more than one to gauge peoples opinion on the methodology of how to add these bounds. You are correct where this came from.

Sounds like there is some concern of using the `oneof` here, so might need to adjust how these are presented. But, I'll plan to include an exponential bucketing definition.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -118,255 +145,122 @@ message MetricDescriptor {
...
+//
+// This defines size(bounds) + 1 (= N) buckets. The boundaries for bucket at
+// index i are:
+//
+//   (-infinity, bounds[i]) for i == 0
+//   [bounds[i-1], bounds[i]) for 0 < i < N-1
+//   [bounds[i], +infinity) for i == N-1
+message ExplicitBounds {
+  // The values in bounds array must be monotonically increasing.
+  repeated double bounds = 3;
```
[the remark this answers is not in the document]
MrAlias: That's a good point. I'll see if I can't include typing in the next iteration of the bounds.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: I agree and we did especially change it, so look also to the history of PRs for more discussion.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -118,255 +145,122 @@ message MetricDescriptor {
...
+
+  message BucketBounds {
+    oneof BoundKinds {
+      ExplicitBounds explicit_bounds = 1;
+      LinearBounds linear_bounds = 2;
+    }
+  }
+  BucketBounds bucket_bounds = 6;
+
+  message Percentile {
```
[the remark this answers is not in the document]
MrAlias: Yeah, I was leaving it up to the user to decide on what to include. I think it would be fine if someone wanted to include all values here, with the percentiles being like you said, so that it will "avoid having to compute quantiles lossily from histogram counts".

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: Yeah, it currently does. I think (for a variety of reasons) I'm going to go back to a format like before.

I think it is still possible in the old format for the sender to send a mix of data, but with the distinct types it is likely easier to ignore this error.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -118,255 +145,122 @@ message MetricDescriptor {
...
-    SUMMARY = 7;
+    // INSTANTANEOUS is a metric whose values are measured at a particular
+    // instant.
+    //
+    // If an INSTANTANEOUS metric is reported over any time interval the
+    // values are representative of distinct measurements within the
+    // interval.
+    INSTANTANEOUS = 1;
+
+    // DELTA is a metric whose values are measured and successively added
```
bogdandrutu, line 161: s/added/accumulated or aggregated? added implies Sum.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -118,255 +145,122 @@ message MetricDescriptor {
...
+    // from the previous interval.
+    //
+    // Each reported value for a DELTA metric over a time interval is
+    // self-contained, the values do not depend on values in other time
+    // intervals. This is unlike a CUMMULATIVE metric where values are sums
+    // of all measured events in the time interval and all previously
+    // measured values of the metric.
+    //
+    // DELTA metrics are not guaranteed to be monotonic.
+    DELTA = 2;
```
bogdandrutu, line 172: I've never seen a system where Deltas are not monotonic, are we trying to do too much here and overgeneralize?

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
MrAlias: This is a great question. Don't want to leave it unaddressed, but I'm still contemplating the answer to this one.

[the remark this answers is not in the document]
bogdandrutu: @tigrannajaryan I think protobuf go should do this optimization internally. If this is the case probably worth filling a bug to the protobuf library they should pack the members in the best way to minimize memory usage.

I would not try right now to optimize on that.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: I think you're correct and we should treat this as an error.

Do you think we should specify a behavior for when this error occurs? Like the receiver will disregard? Or maybe, the behavior is undetermined and up to the receiver to handle?

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu, line 187: This is not enough for systems like Prometheus/Stackdriver/(probably others) which expect only monotonic increasing. How can an exporter for Prometheus/Stackdriver know if it is monotonic increasing or decreasing?

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
bogdandrutu: This can also be added later. Can you please split this in a separate PR.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu, line 210: I am not sure I understand why are these merged. Also can be done in a separate PR.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -44,7 +44,7 @@ message InstrumentationLibraryMetrics {
   repeated Metric metrics = 2;
 }
 
-// Defines a Metric which has one or more timeseries.
+// Metric represents timeseries produced by a telemetry system.
```
yurishkuro, line 47:
```suggestion
// Metric represents a timeseries produced by a telemetry system.
```
Did you mean a singular or plural form of timeseries?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -87,23 +87,50 @@ message InstrumentationLibraryMetrics {
...
+
+    // start_time_unix_nano is the beginning of a time interval over which
+    // this data point was recorded. It MUST be represented as the UNIX Epoch
+    // time in nanoseconds (nanoseconds since 00:00:00 UTC on 1 January 1970).
+    // If zero, it is treated as unspecified and may be set by the receiver.
+    //
+    // If this data point is a part of a timeseries for an INSTANTANEOUS
+    // metric, this value is ignored.
+    //
+    // If this data point is a part of a timeseries for an DELTA, or
```
yurishkuro, line 111:
```suggestion
    // If this data point is a part of a timeseries for a DELTA, or
```

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: this and the subset value define the concept you are talking about.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: delta is the sum of all observed events. If you are returning a delta value for a quantity that can both increase or decrease in the observation time this would not be monotonic.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -118,255 +145,122 @@ message MetricDescriptor {
...
+    // measured values of the metric.
+    //
+    // DELTA metrics are not guaranteed to be monotonic.
+    DELTA = 2;
+
+    // CUMULATIVE is a metric whose values are measured and successively
+    // added to the lifetime total of the metric.
+    //
+    // Each reported value for a CUMULATIVE metric is the sum of all
+    // measurements up to and including that one.
```
yurishkuro, line 178: I don't think this sentence is accurate, as it reads almost like a math sequence `V_n = Sum_{i=1..n-1} V_i`

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -118,255 +145,122 @@ message MetricDescriptor {
...
-  fixed64 start_time_unix_nano = 2;
-
-  // time_unix_nano is the moment when this value was recorded.
-  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
-  fixed64 time_unix_nano = 3;
-
-  // value itself.
-  int64 value = 4;
-}
+  // Subset describes the subset of numbers metric values belong to if any.
```
yurishkuro, line 189: "subset" is an odd name, I think you're referring to "domain" of values

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: Singular. Thanks.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -118,255 +145,122 @@ message MetricDescriptor {
...
+
+  message Percentile {
     // The percentile of a distribution. Must be in the interval
     // [0.0, 100.0].
     double percentile = 1;
 
     // The value at the given percentile of a distribution.
     double value = 2;
   }
+  repeated Percentile percentiles = 7;
```
yurishkuro, line 238: Should this really be a first-class citizen? Since percentiles cannot be aggregated, supporting it directly sounds like encouraging bad practices.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: That formula is mostly correct based on my understanding of what we are trying to convey. Only thing I would add is that the latest measurement is included in the sum. I.E.

```
Value_n = Sum_{i=1...n} Measurement_i
```

Do you have a different way of understanding what a CUMULATIVE metric is, or maybe a better way of stating this?

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: +1 on `Domain`.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: I agree. It is influenced by both the previous proto definition that had a `Summary` type to support this, and both the Prometheus' `Summary` and part of Stackdriver's `Distribution`.

I whole hardily agree these lead users to make nonsense aggregations, but was keeping it around to answer the opposite question of why did you remove them.

If there is a consensus for not including these I'm 100% onboard to dropping.

MrAlias: To add to this, in my experience, I have seem limited use of Prometheus `Summary`s for this exact reason. It is much more flexible to send the distribution data as a histogram and calculate a close approximation of an arbitrary percentile when needed (not when measured).

I think if you wanted to optimization of sending percentiles, you could send them as Int64 or Double data on your own.

opentelemetry/proto/metrics/v1/metrics.proto
yurishkuro: Does CUMULATIVE only applies to sums?

Your formula is still confusing (note that mine didn't have observations _at all_, it was purely a deterministic math sequence), because `Value` refers to reported values, which can have multiple measurements in-between. E.g. if we're reporting request count every second, but in the API we're calling `counter.Inc(1)` for every request, then I expect `Measurement_i == 1`, while `Value_j` are cumulative counts.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: @jmacd : That's a good point. I'm not sure either. I can see the inclusion of the type in the `Data` itself allowing for simpler extension of a metric instrument (likely going to be really useful when there are multiple views on the same instrument), but also how it might be easier to reason about distinct metric descriptor that fully defines the data.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: Sorry, just re-read. :expressionless: 

I see the omission you are talking about. I'll be sure to distinquish monotonic increasing/decreasing.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: I pushed a change to put this info in the descriptor. If you come to a different conclusion @jmacd, let me know.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: > Does CUMULATIVE only applies to sums?

Good point. No, I don't think it does in theory. Likely in practice it does, but I'll change the sentence to be more general.

I don't want to let the discussion of `Sum`s distract from the underlying question you have about the sentence. To that end maybe we can use your example to make sure we share a common understanding of what a CUMULATIVE value is. To expand on it:

1. The system starts.
1. A request comes in, we call `counter.Inc(1)`.
1. Another request comes in, we call `counter.Inc(1)`.
1. Another request comes in, we call `counter.Inc(1)`.
1. The 1 second collection cycle ends. A metric is created for the export pipeline with a value of `3`.
1. A request comes in, we call `counter.Inc(1)`.
1. Another request comes in, we call `counter.Inc(1)`.
1. The 1 second collection cycle ends. A metric is created for the export pipeline with a value of `5` (`2` from this collection cycle, `3` from the previous totaling `5` (assuming a sum aggregation)).
1. A request comes in, we call `counter.Inc(1)`.
1. The 1 second collection cycle ends. A metric is created for the export pipeline with a value of `6` 

The value for the metric in this example is said to be CUMULATIVE. Does this sound correct to you?

If so is there a way you think we can rephrase the sentence to better describe this behavior?

If not, can you provide an revision of the above example that matches what you are thinking?

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: @c24t : When I looked into implementing the `refinement` approach I wasn't sure how to restrict  mutually exclusive refinements using the flag system. Using the enums like the current approach builds sets restrictions using syntax, but I'm not sure how to achieve the same otherwise (i.e. you could set both `monotonic_increasing` and `monotonic_decreasing`).

Do you have a way in mind to prevent this with this kind of implementation? Want to make sure I didn't miss something (because I like the uniformity of this approach).

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
tigrannajaryan: @bogdandrutu in memory representation of a ProtoBuf message is a Go `struct`. If I remember correctly the order of the fields in the struct matches their order in the ProtoBuf declaration. I do not think ProtoBuf generator reorders fields in order to reduce wasted memory due to alignment requirements.
I do not think it should be considered a ProtoBuf generator bug. I would expect that the struct fields preserve the order otherwise readability of the code will be impacted.
This is lower priority issue but at least I would check the memory usage to see how much we are wasting. Perhaps it is negligible.

opentelemetry/proto/metrics/v1/metrics.proto
yurishkuro: If you want to generalize from sum, then it's an aggregation of all measurements from some fixed point in time up to the last reporting snapshot. And I think it's fine to provide a small example as well.

But I am still curious what aggregations other than a sum this could be used for.

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
MrAlias: Removed this based on this: https://github.com/open-telemetry/opentelemetry-proto/pull/137#discussion_r410489326

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: > If you want to generalize from sum, then it's an aggregation of all measurements from some fixed point in time up to the last reporting snapshot. And I think it's fine to provide a small example as well.

:+1: 

> But I am still curious what aggregations other than a sum this could be used for.

Currently there are not any alternate aggregations (nor do I anticipate there ever being). I have seen monitoring systems in the past measure cumulative averages and rates. Both of which are pretty useless IMO.

The thing I don't want to do is imply is that there is a predefined aggregation that should be assumed if the the `Temporality` is cumulative. That would incorrectly couple the concepts.

opentelemetry/proto/metrics/v1/metrics.proto
c24t: AFAIK there's no way to to prevent invalid bitsets in the proto itself, I assume it'd be up to the collector to return some sensible error.

One benefit of using flags instead of enums is that we can add refinements without updating the proto package. (Whether this is a benefit or a detriment maybe depends on your use case: it would be very bad news if the client and the collector interpreted this field differently!) But this is only true if the proto doesn't know anything about refinement semantics.

I don't have a strong opinion on this vs. enums, just pointing out that there's a tradeoff here between safety and development speed.

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
c24t: I think if we're going to put the refinements (instead of e.g. the instrument names) in the proto then it makes sense to start small, and only include the refinements for the instruments we expect to support. Assuming it's easier to add these later than remove them.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: > we also look at what others do like OpenMetrics/Stackdriver

@bogdandrutu : where is this defined for OpenMetrics?

MrAlias: FWIW, Stackdriver is an inclusive interval: https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeInterval

E.g. `[start, end]`

MrAlias: Looking at a token example of a check-pointing process:

```
=========================
||     Checkpointer    ||
=========================


-------------------------
| Start: t_0            |
-------------------------
       |                            -----------------
       |  <-------------------------| measurement_0 |
       |                            -----------------
-------------------------
| Checkpoint 1          |
| --------------------- |           -----------------
| Begin Checkpoint: t_1 |  <--------| measurement_1 |
|                       |           -----------------
| Processing...         |
|                       |           -----------------
|                       |  <--------| measurement_2 |
|                       |           -----------------
-------------------------
       |                            -----------------
       |  <-------------------------| measurement_3 |
       |                            -----------------
-------------------------
| Checkpoint 2          |
| --------------------- |
| Begin Checkpoint: t_2 |
|                       |
| Processing...         |
|                       |
-------------------------
       |
       ...
```

If `start <= t < time`:

- Checkpoint 1 (`t_0 <= t < t_1`) includes:
  - measurement_0
- Checkpoint 2 (`t_1 <= t < t_2`) includes:
  - measurement_1, measurement_2, measurement_3

If `start < t <= time`:

- Checkpoint 1 (`t_0 < t <= t_1`) includes:
  - measurement_0, measurement_1
- Checkpoint 2 (`t_1 < t <= t_2`) includes:
  - measurement_2, measurement_3

(side note: with this interval definition, start is not included in any interval. This is not idea, but likely minor enough to not be a real problem in the real world)

@jmacd : From my understanding of you comment, you are saying that the `t_0 < t <= t_1` definition of the interval means that while Checkpoint 1 should contain both measurement_0 and measurement_1, it is not possible to do in practice. Mainly because the checkpoint process is no longer receiving measurements when `t_1` is defined by the process, right?

bogdandrutu: https://github.com/OpenObservability/OpenMetrics/pull/126/files#diff-d1792782ae179ab49b581de250c5def2R103 

I talked to one of the StackDriver authors and they explained to me that they actually don't want to have inclusive interval because that may cause duplicate "report" (they also said that in practice nobody does implement that way - if an event happens at exactly same time as the endtime of the interval it will not be included in both deltas, the decision to be in the current delta or the next delta happens using other synchronization mechanisms that ensure a happens before relationship and ensures events are counted only once).

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: I think what I said is that I've never seen a system (backend: Prometheus, Stackdriver, SignalFx, etc.) that support delta for non-monotonic values. I did not say that I don't understand what you wrote, just an observation and something to think about if we need to support this if nobody needs this.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: > https://github.com/OpenObservability/OpenMetrics/pull/126/files#diff-d1792782ae179ab49b581de250c5def2R103
> 
> I talked to one of the StackDriver authors and they explained to me that they actually don't want to have inclusive interval because that may cause duplicate "report" (they also said that in practice nobody does implement that way - if an event happens at exactly same time as the endtime of the interval it will not be included in both deltas, the decision to be in the current delta or the next delta happens using other synchronization mechanisms that ensure a happens before relationship and ensures events are counted only once).

Thanks for the link and info. We should not consider a revision of an open PR and undocumented behavior given in a conversation to be prior art. Furthermore, I think the point @jmacd made is a technical one and I think we should engage the point on that level.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: I can speak for New Relic, in that it supports non-monotonic delta. I'm a bit surprised to hear Stackdriver no longer supports this as I have in the past sent data of this type to it.

I'm wondering if I might have been misinterpreted here though. The goal was to make sure that it is clear this type is not coupled to the monotonic refinement. This goal seems prudent with [the impending inclusion of an instrument that produces just this kind of data](https://github.com/open-telemetry/oteps/pull/98/files#diff-95b3f96b8945175022788618354f953eR25).

bogdandrutu: I think I am fine with having this, it was mostly something that I want others to think about and see if it is useful. It seems that this is useful so we can have this :)

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: @MrAlias I state back to my comment, using time to determine a happens-before relationship is impossible (or very hard to implement unless we are going to implement something like true-time). So I am not sure I understand your example, and how would it be possible for an event to happen before we start the checkpointer.

MrAlias: @bogdandrutu: Hmm, it seems like we might be talking past each other. I'm wondering if this is stemming from the differences in concurrency design in the implementations we work in.

Maybe this is something we can talk about in the upcoming Metric SIG meeting? I'll add it to the agenda if you can make the meeting.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -118,144 +152,121 @@ message MetricDescriptor {
...
+  // Monotonic describes the metric values to be entirely non-increasing,
+  // or entirely non-decreasing.
+  enum Monotonic {
+    // UNKNOWN is the default, and means it the monotonic nature of the
+    // metric values is known.
+    UNKNOWN = 0;
+
+    // NONINCREASING means all the successive metric values decrease or
+    // remain constant.
+    NONINCREASING = 1;
```
bogdandrutu, line 229: Do we need to support this mode? It seems that we can simply ask users to use the "-value" when recording and we have to deal with only one type.

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
bogdandrutu: I would also vote for removing things that we are not sure about usefulness.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -87,23 +87,57 @@ message InstrumentationLibraryMetrics {
...
-  // Data is a list of one or more DataPoints for a single metric. Only one of the
-  // following fields is used for the data, depending on the type of the metric defined
-  // by MetricDescriptor.type field.
-  repeated Int64DataPoint int64_data_points = 2;
-  repeated DoubleDataPoint double_data_points = 3;
-  repeated HistogramDataPoint histogram_data_points = 4;
-  repeated SummaryDataPoint summary_data_points = 5;
+  repeated Data data = 2;
+
+  message Data {
```
bogdandrutu, line 96: Having one message Data with multiple value types is less efficient than having multiple data points, because the generated class/struct in languages like go, java, c++ will be larger (will include all value types). So I suggest we stick with the previous definition of data points and deal with duplicate definition for labels, start/end timestamp.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: Thanks for the suggestions, they are much appreciated.

I've updated the time interval handling. PTAL and let me know if you have any other suggestions.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: Updated, PTAL.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: removed.

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
MrAlias: removed.

MrAlias:
Worked with @bogdandrutu on a plan forward by breaking up the changes included here into smaller PRs.

I'm going to close this to avoid confusion. If there are further general questions/suggestion about the approach feel free to continue commenting here.

Thank you everyone who provided feedback, it was greatly appreciated.

PR #139, opened 2020-04-24
Update install-proto.sh

MrAlias:
- Save downloaded binaries and libraries to a persistent location. This preserves the setup across restarts.
- Unify install location to default to `/usr/local`.
- Add argument to override install location.
- Scope variables to avoid malicious variable overrides to be executed with elevated permissions.
- Conditionally check if `sudo` is needed instead of just defaulting.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,6 +12,13 @@ for general information about the project.
 Install manually or use:
 
 - `install-proto.sh` (if on Linux).
+    - Requires: `bash`, `wget`, `unzip`, `go`, and `sudo`.
+    - The default install directory is `/usr/local` for all downloaded binaries and libraries.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -12,6 +12,13 @@ for general information about the project.
 Install manually or use:
 
 - `install-proto.sh` (if on Linux).
+    - Requires: `bash`, `wget`, `unzip`, `go`, and `sudo`.
+    - The default install directory is `/usr/local` for all downloaded binaries and libraries.
+      This location can be overridden by passing a alternate location to the script.
```
SergeyKanzhelev, line 17:
```suggestion
      This location can be overridden by passing an alternate location to the script.
```

SergeyKanzhelev:
if there will be no more feedback I can merge later today. Seems not controversial.

Issue #141, opened 2020-04-25
Dual placement of Labels makes certain translations expensive


tigrannajaryan:
### Problem

Metric proto currently allows Labels in 2 places: in the MetricDescriptor and in each DataPoint.

The semantics is that Labels in MetricDescriptor apply to all DataPoints. This requires complicated and expensive merging of Labels maps (which possibly is made even more expensive since they are stored as lists, requiring either intermediary data structures or O(nm) merging) when performing translations from OTLP to other formats that have no equivalent notion of having label keys per data point and per group of data points (e.g. OpenCensus).

### Approach 1

I suggest that we eliminate one or the other and have Labels in one place only.

### Approach 2

Alternatively, we can require that Labels in MetricDescriptor and in DataPoints cannot overlap, i.e. there do not exists any entries that have the same key in both places. This way the Labels in MetricDescriptor allow to avoid repeating common labels for all data points. In this case merging is possible by plain concatenation and there is no need to expensive merging that ensures uniqueness of the keys.

This approach is probably less desirable since it makes impossible to translate series of data points from OTLP to OpenCensus TimeSeries because each DataPoint can have a different set of keys (not possible in OpenCensus).

However, this would still be an improvement over current state of things if we really want to keep dual Labels placement.

tigrannajaryan:
@bogdandrutu what do you think?

bogdandrutu:
> This approach is probably less desirable since it makes impossible to translate series of data points from OTLP to OpenCensus TimeSeries because each DataPoint can have a different set of keys (not possible in OpenCensus).

This is already the case unfortunately (I tried my best to fight for having this as a requirement that a metric always produces the same set of label keys, @jmacd pushed to not have this requirement). So in open-telemetry we don't know the set of label-keys that we produce and the default aggregation for all the instruments will produce unbounded number of label-keys. I know this is not easily compatible with OpenCensus but I don't think we can do anything at this moment, unless we add the constraint to predefined the keys. There was a tentative way to solve this by using the "recommended label keys", but that was not enforced as requirement to produce aggregation only using these keys, so the problem was not resolved.

> I suggest that we eliminate one or the other and have LabelKeys in one place only.

We can remove only the labels from the MetriDescriptor if we want to remove any of them.

[...]

PR #144, opened 2020-04-27
Remove common labels from MetricDescriptor

tigrannajaryan:
Resolves: https://github.com/open-telemetry/opentelemetry-proto/issues/141

Metric proto previously allowed Labels in 2 places: in the MetricDescriptor and in each DataPoint.

The semantics is that Labels in MetricDescriptor apply to all DataPoints.
This requires complicated and expensive merging of Labels maps (which
possibly is made even more expensive since they are stored as lists,
requiring either intermediary data structures or O(nm) merging) when performing
translations from OTLP to other formats that have no equivalent notion of
having label keys per data point and per group of data points (e.g. OpenCensus).

This commit removes Labels from MetricDescriptor. My understanding is that
there is currently no client library that emits MetricDescriptor with labels
so this is not used and can be safely removed.

This will simplify and make faster translation code in the Collector.

If this is needed we can think carefully about adding it back after there
is a clear justification and tradeoffs are well understood.


bogdandrutu:
I agree that this was a premature optimization and we can always bring it back later.

PR #145, opened 2020-04-28
Add Monotonic Types

MrAlias:
Add the following types:

* `MONOTONIC_INT64`
* `MONOTONIC_DOUBLE`

These type values are still transported in the same DataPoints as their non-monotonic counterpoints, but add the significance to their values as being monotonic.

This is a partial replacement of #137 
Relates to #125


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -227,9 +227,39 @@ message MetricDescriptor {
...
+    // this Monotonic value.
+  }
+
+  // monotonic describes the Monotonic refinement of values this metric has.
+  Monotonic monotonic = 6;
+
   // The set of labels associated with the metric descriptor. Labels in this list apply to
   // all data points.
-  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 6;
+  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 7;
```
bogdandrutu, line 262: hmm... this was removed? did you make a merge mistake in the previous PR?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -227,9 +227,39 @@ message MetricDescriptor {
...
+    // remain constant.
+    NONDECREASING = 1;
+
+    // NONINCREASING is currently omitted due to the lack of a known
+    // use-case. If you have a fitting use-case please open an issue to add
+    // this Monotonic value.
+  }
+
+  // monotonic describes the Monotonic refinement of values this metric has.
+  Monotonic monotonic = 6;
```
bogdandrutu, line 258: Based on a comment from gitter:
> I think we can have only `Domain`, and we don't need "monotonicity". For example if domain refers to the raw data not to the aggregated data, then by having a non negative values domain implies nondecreasing monotonicity. The advantage being that we can apply this to data in "distributions" and not only to int64/double.

Currently the "monotonic" property is hard to correlate with Histogram/Summary (or even the newly proposed Distribution in that closed PR).

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: Oh shoot! You are correct, I did. I can fix it here, or in a separate PR if you want.

bogdandrutu: Sent https://github.com/open-telemetry/opentelemetry-proto/pull/146.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: The problem with only including `Domain` and not `Monotonic` is that `Domain` does not imply `Monotonic`. For example, if data for a metric continuously switches back-and-forth between say `0` and `1`, it is in the `Domain` of "greater than or equal to zero", but it is not monotonic. If you calculate the derivative on this data to generate its rate thinking it is monotonic it will result in large artifacts due to the discontinuous nature.

Conversely, it is also true that `Monotonic` does not imply `Domain`. So I don't think it is a choice of overlapping concepts to be made here.

I agree that the monotonic nature of Histogram/Summary data is not intuitively defined. But I don't think you improve the situation by not including this information. Take for instance when you want to translate into Prometheus metrics, without including a description of the monotonic nature of the data you would not be able to translate into any data into Prometheus Histograms or Counters.

I think instead, I can add better descriptions of how this monotonic refinement relates to the exposition formats. How does that sound?

bogdandrutu: I do agree that if we cannot make sure that it implies "monotonicity" we should not apply that proposal.

Based on this example I think I see something wrong in my thinking, for me int64/double types meant SumInt64/SumDouble, so you cannot get 0 -> 1 -> 0 unless you have a "-1" somewhere:
> For example, if data for a metric continuously switches back-and-forth between say 0 and 1, it is in the Domain of "greater than or equal to zero", but it is not monotonic. If you calculate the derivative on this data to generate its rate thinking it is monotonic it will result in large artifacts due to the discontinuous nature.

bogdandrutu: Is that the difference? You think for int64/double that they may not be a sum?

bogdandrutu: Just a random thought: Maybe another way to think about this is domain + temporality determines monotonicity, in your example of 0->1 transitions that will be an instantaneous metric, so will always be a gauge in Prometheus. If temporality != Instantaneous then we can determine the monotonicity?

tigrannajaryan:
With 3 different enums we now use 3x32=96bits to encode 6 bits worth of info. This can have significant impact on memory consumption in Collector which may hold many thousands or even millions of metrics in queues.

Should we use a single field with bitmasks instead? We can keep the enums and just use a single uint32 with lots of reserved fields for future use.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: Yes, I think that is the difference.

bogdandrutu:
@tigrannajaryan I see that as an optimization, which I think we can do it as a next step after we finalize the data format and all the properties we need.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: What do you think about my latest comment (https://github.com/open-telemetry/opentelemetry-proto/pull/145#discussion_r417000445), is that a correct assumption?

MrAlias: > Just a random thought: Maybe another way to think about this is domain + temporality determines monotonicity, in your example of 0->1 transitions that will be an instantaneous metric, so will always be a gauge in Prometheus. If temporality != Instantaneous then we can determine the monotonicity?

Yeah, I'm currently thinking this exact same thing through. I worry there might be a looseness with terms though.

If the metric is CUMULATIVE is there really a guarantee that the data is monotonic? Based on the [new UpDownSumObserver](https://github.com/open-telemetry/oteps/pull/98/files?short_path=95b3f96#diff-95b3f96b8945175022788618354f953e) I think that is going to be possible.

bogdandrutu: After I wrote that random thought I now have the impression that monotonicity does not apply to any instantaneous metric, correct? :)) Probably I need to go have a drink or sleep :)))

bogdandrutu: > If the metric is CUMULATIVE is there really a guarantee that the data is monotonic? Based on the new UpDownSumObserver I think there is going to be a situation where that is not the case.

No, what I am trying to say is: if the metric is cumulative or delta and domain is not-negative then is monotonic. if the metric is instantaneous is not monotonic.

MrAlias: > Probably I need to go have a drink or sleep :)

Same. :smiley: 

> ... if the metric is cumulative or delta and domain is not-negative then is monotonic.

I think that is true if domain describes the domain of the measurement an instrument can make. The way I'm using these refinements is to describe the quality of the aggregation values.

It is true that if you only ever put non-negative values through a sum aggregator the resulting aggregation is monotonic. However, if you instead say that the resulting aggregation values are non-negative you cannot say anything about the monotonicity of the that aggregation.

Are referring to different parts of the pipeline when we are using these terms to describe data?

bogdandrutu: Yes I think I had that in the initial comment "For example if domain refers to the raw data not to the aggregated data". That was the idea :)

bogdandrutu: Now that we agree on that, do you think actually referring to the raw data domain is better? Argument being that we can determine interesting properties for Distributions as well as for sum values, for example in a Distribution we can determine if the sum is monotonic (and that percentiles are positive and never going to have negative buckets, etc).

Alternative we can go with your approach and extend the comment that monotonic applies to: int64/double delta or cumulative and to sum in the distribution if delta or cumulative. I think monotonicity should not apply to instantaneous aggregations.

MrAlias: > Now that we agree on that, do you think actually referring to the raw data domain is better? Argument being that we can determine interesting properties for Distributions as well as for sum values, for example in a Distribution we can determine if the sum member is monotonic (and that percentiles are positive and never going to have negative buckets, etc).
> 
> Alternative we can go with your approach and extend the comment that monotonic applies to: int64/double delta or cumulative and to sum in the distribution if delta or cumulative. I think monotonicity should not apply to instantaneous aggregations.

I'm not 100% sure what the best approach here is.

It seemed to me like describing the data contained here would be the most useful to the user of this proto instead of making them infer things from understanding the OpenTelemetry processing pipeline. Additionally, it means that changes in the OpenTelemetry instruments/processors/aggregators are agnostically supported here. This would match the Type property as that describes the way the data is stored here.

That being said, it seems like the Temporality quality is already describing how the raw data is combined in the OpenTelemetry pipeline, so there is precedence in both ways.

I'm interested what @jmacd and @c24t think about this. @c24t had raised a similar question on the initial #137.

tigrannajaryan:
@MrAlias @bogdandrutu can we move forward with this? We are waiting for the finalization of metrics ProtoBufs to be able to use them in the Collector.

MrAlias:
> @MrAlias @bogdandrutu can we move forward with this? We are waiting for the finalization of metrics ProtoBufs to be able to use them in the Collector.

Sorry, was blocked working on https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/229 for a few days, but will get back to this today.

Talking with @bogdandrutu last week it sounds like two things (related to this PR) should be done:

- The protobuf should describe data ***post*** aggregation (instead of pre-instrument). This means the blocking conversation could be resolved by keeping things as is, but ...
- The concept of `Monotonic` doesn't apply to all datapoints (`SummaryDataPoints`), therefore, we should move the concept directly to the `Type` itself.

Going to update later to remove the `Monotonic` field and change `Type` to:

```
// Type is the type of values a metric has.
enum Type {
  INVALID_TYPE = 0;

  INT64 = 1;
  MONOTONIC_INT64 = 2;

  DOUBLE = 3;
  MONOTONIC_DOUBLE = 4;

  HISTOGRAM = 5;
  SUMMARY = 6;
}
```

There was an idea to include a `MONOTONIC_HISTOGRAM` as well, but I'm not sure how useful it will be. I'm open to opinions on this.

MrAlias:
Looks like @jmacd has also been thinking about this: https://github.com/open-telemetry/opentelemetry-proto/issues/125#issuecomment-635196903. Going to wait until after the Metric SIG meeting today to make any changes to make sure we are all in sync.

jmacd:
I think we found resolution today in the SIG call.

bogdandrutu:
@jmacd can you document the proposed resolution?

MrAlias:
> @jmacd can you document the proposed resolution?

@bogdandrutu : The plan is to go with:

```
// Type is the type of values a metric has.
enum Type {
  INVALID_TYPE = 0;

  INT64 = 1;
  MONOTONIC_INT64 = 2;

  DOUBLE = 3;
  MONOTONIC_DOUBLE = 4;

  HISTOGRAM = 5;
  SUMMARY = 6;
}
```
thoughts?

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: Resolving with the decision having been to describe the measurement value as it is kept in this proto (instead of how it was provided to the measuring instrument), and that we move in a new direction to have the monotonic refinement directly linked to the type.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -123,28 +123,39 @@ message MetricDescriptor {
...
     // A Metric of this Type MUST store its values as Int64DataPoint.
     INT64 = 1;
 
-    // DOUBLE values are represents as double-precision floating-point
-    // numbers.
+    // MONOTONIC_INT64 values are monotonically increasing signed 64-bit
+    // integers.
+    //
+    // A Metric of this Type MUST store its values as Int64DataPoint.
+    MONOTONIC_INT64 = 2;
```
bogdandrutu, line 135: Should we document the interaction with temporality?

MrAlias: I'm not following, is that not already documented in and on the `Temporality` message itself already?

jmacd: I agree with this change about Monotonic representation. I've come to disagree with use of "Temporality" and do not believe we should use any of the terms Instantaneous, Cumulative, or Delta directly in the protocol.

I have outlined my latest opinion here: https://github.com/open-telemetry/opentelemetry-proto/issues/125#issuecomment-635196903

There I used the terms "Collective vs Additive" to distinguish two kinds of aggregation, but later debate has nixed the word "Collective". See https://github.com/open-telemetry/opentelemetry-specification/issues/625#issuecomment-636119812

I think we should replace the entire section on temporality with a two-state enum: Grouping or Additive.

Separately, also stated in the comment on proto issue 125, I think we should use the start and end time to convey whether reported data is delta or cumulative.

MrAlias: > I agree with this change about Monotonic representation. I've come to disagree with use of "Temporality" and do not believe we should use any of the terms Instantaneous, Cumulative, or Delta directly in the protocol.
> 
> I have outlined my latest opinion here: [#125 (comment)](https://github.com/open-telemetry/opentelemetry-proto/issues/125#issuecomment-635196903)
> 
> There I used the terms "Collective vs Additive" to distinguish two kinds of aggregation, but later debate has nixed the word "Collective". See [open-telemetry/opentelemetry-specification#625 (comment)](https://github.com/open-telemetry/opentelemetry-specification/issues/625#issuecomment-636119812)
> 
> I think we should replace the entire section on temporality with a two-state enum: Grouping or Additive.
> 
> Separately, also stated in the comment on proto issue 125, I think we should use the start and end time to convey whether reported data is delta or cumulative.

The more I've thought about the Temporality the more I think I agree with this. It seems like even with directly specifying it you should still be checking (start, end) against last seen to determine if the interval was reset (cumulative) or data was lost (delta).

I'm interested to talk more with @jkwatson about this though.

bogdandrutu: Is MONOTONIC_SUM_INT64 better name?

bogdandrutu: Probably not... Let's keep as it is.

bogdandrutu:
Can you please update the description?

PR #146, opened 2020-04-28
Recover changes from PR#144

bogdandrutu:
PR #144 was merged before #140, during rebase a mistake happen and changes from #144 were voided.


bogdandrutu:
@MrAlias no worries, we are humans :)

PR #151, opened 2020-05-14
Add messages to support Log data type

tigrannajaryan:
This definition is based on the accepted log data model:
https://github.com/open-telemetry/oteps/blob/master/text/logs/0097-log-data-model.md

LogsService is defined similarly to TraceService and MetricsService.


tigrannajaryan:

opentelemetry/proto/common/v1/common.proto
```diff
@@ -60,3 +60,50 @@ message InstrumentationLibrary {
...
+// AnyValue is a value that is used to store values that can be one of the supported
+// value types.
+message AnyValue {
+  // type of the value.
+  ValueType type = 1;
+
+  // Only one of the following fields is supposed to contain data (determined by `type` field).
+  // This is deliberately not using Protobuf `oneof` for performance reasons (verified by benchmarks).
+
+  bool bool_value = 2;
```
SergeyKanzhelev, line 84: nit: I wonder why the order of fields doesn't match the order in enum =).

opentelemetry/proto/common/v1/common.proto
```diff
@@ -60,3 +60,50 @@ message InstrumentationLibrary {
...
+  // case when the value is a primitive type.
+  ValueListOrMap list_or_map = 6;
+}
+
+// ValueListOrMap is used for storing a list or a map of values.
+message ValueListOrMap {
+  // AnyValue.type field defines which of the 2 fields in this message are set.
+
+  // A list of values.
+  repeated AnyValue list = 1;
```
SergeyKanzhelev, line 103: naming-wise, should it be `list_value`?

SergeyKanzhelev: or better `list_values`

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -0,0 +1,119 @@
...
+    ERROR4 = 20;
+    FATAL  = 21;
+    FATAL2 = 22;
+    FATAL3 = 23;
+    FATAL4 = 24;
+}
+
+// Masks for LogRecord.flags field.
+enum LogRecordFlags {
+    UNDEFINED_LOG_RECORD_FLAG = 0;
```
SergeyKanzhelev, line 68: why values are upper case? It's inconsistent with other enums

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -0,0 +1,119 @@
...
+    TRACE_FLAGS_MASK = 0x000000FF;
+}
+
+// A log record according to OpenTelemetry Log Data Model: https://github.com/open-telemetry/oteps/pull/97
```
SergeyKanzhelev, line 72: perhaps link the otep, not the PR?

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -0,0 +1,119 @@
...
+    // invalid. Can be set for logs that are part of a particular processing span.
+    // If span_id is present trace_id SHOULD be also present. Optional.
+    bytes span_id = 3;
+
+    // Optional flags, a bit field. 8 least significant bits are the trace flags as
+    // defined in W3C Trace Context specification. 24 most significant bits are reserved
+    // and must be set to 0. Readers that must not assume that 24 most significant bits
+    // will be zero and must correctly mask the bits when reading 8-bit trace flag (use
+    // flags & TRACE_FLAGS_MASK). Optional.
+    fixed32 flags = 4;
```
SergeyKanzhelev, line 96: if we carry flags, should we have a field for `tracestate`?

opentelemetry/proto/common/v1/common.proto
tigrannajaryan: It reduces memory usage because of better packing of fields due to struct field alignment rules.

tigrannajaryan:
@SergeyKanzhelev thanks for reviewing. I am experimenting with a slightly different approach. Let me mark this PR as WIP and I will update it once I finish the experiments.

opentelemetry/proto/common/v1/common.proto
SergeyKanzhelev: good to know. Worth mentioning in comments for future reference.

tigrannajaryan:
I will return to this once https://github.com/open-telemetry/opentelemetry-proto/pull/157 is merged.

tigrannajaryan:
@open-telemetry/specs-approvers please have a look. I updated the PR.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: We use UPPERCASE as a convention for enums in this repo, e.g. https://github.com/open-telemetry/opentelemetry-proto/blob/e43e1abc40428a6ee98e3bfd79bec1dfa2ed18cd/opentelemetry/proto/trace/v1/trace.proto#L103

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: Done.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: This was discussed in Log Data Model OTEP and at the time it was decided that it is not needed, see https://github.com/open-telemetry/oteps/blob/master/text/0097-log-data-model.md#traceflags-vs-traceparent-and-tracestate
I am open to reconsidering and we can add the tracestate as a field if needed (it will be a backward compatible change, so this PR is still OK to move forward with).

kumoroku: minor: the OTEP talks about `TraceFlags` but here we are not using `trace_flags` as I would have naively expected, but just `flags`. is this a previous convention?

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -0,0 +1,130 @@
...
+
+    // Numerical value of the severity, normalized to values described Log Data Model.
+    // Optional.
+    SeverityNumber severity_number = 5;
+
+    // The severity text (also known as log level). The original string representation as
+    // it is known at the source. Optional.
+    string severity_text = 6;
```
kumoroku, line 115: one more very nit-picky comment: order of fields and naming follows the OTEP very closely. here however, `severity_number` and `severity_text` are swapped in order vs the OTEP.

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -0,0 +1,130 @@
...
+option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/logs/v1";
+
+// A collection of InstrumentationLibraryLogs from a Resource.
+message ResourceLogs {
+    // The resource for the logs in this message.
+    // If this field is not set then no resource info is known.
+    opentelemetry.proto.resource.v1.Resource resource = 1;
+
+    // A list of InstrumentationLibraryLogs that originate from a resource.
+    repeated InstrumentationLibraryLogs instrumentation_library_spans = 2;
```
odsod, line 34: How come this isn't called `instrumentation_library_logs`?

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -0,0 +1,130 @@
...
+// Masks for LogRecord.flags field.
+enum LogRecordFlags {
+    UNDEFINED_LOG_RECORD_FLAG = 0;
+    TRACE_FLAGS_MASK = 0x000000FF;
+}
+
+// A log record according to OpenTelemetry Log Data Model:
+// https://github.com/open-telemetry/oteps/blob/master/text/logs/0097-log-data-model.md
+message LogRecord {
+    // time_unix_nano is the time when the event occurred.
```
odsod, line 85: There's a mismatch between the documented name `time_unix_nano` and the actual name `time_unixnano` here.

opentelemetry/proto/logs/v1/logs.proto
bogdandrutu: Copy paste :)

tigrannajaryan: Fixed, thanks.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: Fixed, thank you. `time_unix_nano` is the right name, uniform with span field naming.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: TraceFlags is a single byte. Protobufs do not have a single byte data type. To avoid wasting an entire fixed32 just for TraceFlags I introduced a general `flags` field where 8 least significant bits are the TraceFlag (as denoted by TRACE_FLAGS_MASK). The 24 most significant bits of `flags` are reserved for future use.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: This is done intentionally since it results in better packing and more compact in-memory representation dictated by field alignment rules in languages where the order of the in-memory fields is dictated by the order of declaration here (e.g. in Go).

Better packing in this case is achieved because both `flags` and `severity_number` have 32 bit alignment, while `severity_text` has 64 bit alignment (on 64 bit systems). This ordering of the fields results in no wasted space. If I place `severity_number` after `severity_text` we will waste 4 bytes per log record in memory.

The ordering of the fields has no other implications (other than placing logically related fields closer to each other for better readability, which we did not hurt in this case), so I believe very slight deviation from the order listed in the data model is warranted in this case.

tigrannajaryan:
@SergeyKanzhelev can you please review.

opentelemetry/proto/logs/v1/logs.proto
kumoroku: got it, thanks for the explanation. I promise to read the comment next time.

opentelemetry/proto/logs/v1/logs.proto
kumoroku: got it and thanks for the explanation.

opentelemetry/proto/collector/logs/v1/logs_service.proto
```diff
@@ -0,0 +1,48 @@
+// Copyright 2019, OpenTelemetry Authors
```
yurishkuro, line 1: 2020?

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -0,0 +1,130 @@
...
+
+    // A unique identifier for a span within a trace, assigned when the span
+    // is created. The ID is an 8-byte array. An ID with all zeroes is considered
+    // invalid. Can be set for logs that are part of a particular processing span.
+    // If span_id is present trace_id SHOULD be also present. Optional.
+    bytes span_id = 3;
+
+    // Optional flags, a bit field. 8 least significant bits are the trace flags as
+    // defined in W3C Trace Context specification. 24 most significant bits are reserved
+    // and must be set to 0. Readers that must not assume that 24 most significant bits
```
yurishkuro, line 104:
```suggestion
    // and must be set to 0. Readers must not assume that 24 most significant bits
```

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -0,0 +1,130 @@
...
+message LogRecord {
+    // time_unix_nano is the time when the event occurred.
+    // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
+    // Value of 0 indicates unknown or missing timestamp.
+    fixed64 time_unix_nano = 1;
+
+    // A unique identifier for a trace. All logs from the same trace share
+    // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes
+    // is considered invalid. Can be set for logs that are part of request processing
+    // and have an assigned trace id. Optional.
```
yurishkuro, line 93: I would recommend the optionality of the fields to be emphasized for better readability, e.g. as `[Optional]` prefix to the description (or suffix, as long as it's in brackets to draw attention).

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -0,0 +1,130 @@
...
+    bytes trace_id = 2;
```
yurishkuro, line 94: Is there a reason for having trace-related fields in the most prominent place of the struct? I would prefer the actually log-related fields to be at the top.

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -0,0 +1,130 @@
...
+    // Optional.
+    SeverityNumber severity_number = 5;
+
+    // The severity text (also known as log level). The original string representation as
+    // it is known at the source. Optional.
+    string severity_text = 6;
+
+    // Short event identifier that does not contain varying parts. Name describes
+    // what happened (e.g. "ProcessStarted"). Recommended to be no longer than 50
+    // characters. Not guaranteed to be unique in any way. Optional.
```
yurishkuro, line 119: Q: is the plan to have the proto serve as the de-facto spec for logging data, or will there be a separate document in the specification that explains the difference between `name` and `body`? For example, if I were to write an adapter for some standard logging API like Go's or SLF4J, how would I map `logger.info("hello")`? Is "hello" a name or a body?

opentelemetry/proto/collector/logs/v1/logs_service.proto
tigrannajaryan: Fixed.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: Fixed.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: Done.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: I followed the ordering with have in log data model OTEP: https://github.com/open-telemetry/oteps/blob/master/text/logs/0097-log-data-model.md

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: Yes, we have a separate doc that has more details: https://github.com/open-telemetry/oteps/blob/master/text/logs/0097-log-data-model.md
The doc includes sample mappings to help clarify how the fields are intended to be used.

tigrannajaryan:
@SergeyKanzhelev can you please have another look. You requested changes, I believe everything is addressed now.

opentelemetry/proto/logs/v1/logs.proto
yurishkuro: Will it be translated into spec? I view OTEPs more as WIP, regular user shouldn't be expected to read through them.

tigrannajaryan: Sure, I can do that.

opentelemetry/proto/logs/v1/logs.proto
yurishkuro: I would suggest moving trace context fields to the bottom of the struct. It doesn't even apply to large classes of logs.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: Done: https://github.com/open-telemetry/opentelemetry-specification/pull/694

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: I am a bit reluctant because of 2 reasons:
- It will be a deviation from the order we have in the data model.
- It will result in worse packing of the struct and increased memory usage due to alignment rules in certain languages (e.g. in Go, which probably matters most from memory usage perspective since it is used by Collector).

yurishkuro: Since these are byte arrays of unknown (to proto) length, why would they impact packing of struct?

Which data model are you referring to? Given that this is all new dev, I don't really buy the "legacy" argument. If I were a newcomer to OTEL logging, I would be confused by seeing trace fields given higher ranking than the actual log fields.

tigrannajaryan: > Since these are byte arrays of unknown (to proto) length, why would they impact packing of struct?
> 

There is no problem with moving `trace_id` and `span_id`. I can move them without impacting the packing (byte arrays are slices and have 64 bit alignment in Go 64-bit arch). `flags` is the problem because it is 32 bit. Currently it is tightly packed because it is followed by another 32 bit field, which means we have no wasted space. If I move the `trace_id`/`span_id`/`flags` to the end, we will have less tight packing.

However, I thought a bit more and I can do it slightly differently and still have good packing.

I can either keep `flags` where it is and only move `trace_id`/`span_id` to the end or simply change the order of these 3 fields to `flags`/`trace_id`/`span_id` and move all 3 to the end. This way 32 bit `flags` will pack tightly with preceding `dropped_attributes_count` which is also 32 bit.

If you are OK with either of these approaches I can do it.

> Which data model are you referring to? Given that this is all new dev, I don't really buy the "legacy" argument. If I were a newcomer to OTEL logging, I would be confused by seeing trace fields given higher ranking than the actual log fields.

I was referring our OpenTelemetry [log data model](https://github.com/open-telemetry/oteps/blob/master/text/logs/0097-log-data-model.md). This is a minor argument. I am fine to reorder if you think it looks bad.

yurishkuro: I'm ok with flags to be above trace_id.

tigrannajaryan: Done.

tigrannajaryan:

Issue #91, opened 2020-01-07
Publish compiled files for Python


hectorhdzg:
Similar to https://pypi.org/project/opencensus-proto/

PR #152, opened 2020-05-21
Adding make target for python code generation

codeboten:
Left the generated code out as it will live in the opentelemetry-python repository as a separate installable package.

Partially fixes https://github.com/open-telemetry/opentelemetry-proto/issues/91

### DIFF
--- a/makefile
+++ b/makefile
@@ -38,6 +38,13 @@ gen-java:
 	mkdir -p ./$(GENDIR)/java
 	$(foreach file,$(PROTO_FILES),$(call exec-command, protoc --java_out=./$(GENDIR)/java $(file)))
 
+# Generate ProtoBuf implementation for Python.
+.PHONY: gen-python
+gen-python:
+	rm -rf ./$(GENDIR)/python
+	mkdir -p ./$(GENDIR)/python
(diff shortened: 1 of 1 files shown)


tigrannajaryan:
> Left the generated code out as it will live in the opentelemetry-python repository as a separate installable package.

@codeboten It is not clear what is the purpose of the change if the generate code does not live in this repo. Are the files going to be generated in some other repository? If so why is the makefile here and not in that other repo?

Or is this purely for convenience as a development tool for people who actively work on protos and want to see what it looks like in Python?

I may be missing some context, please clarify.

codeboten:
@tigrannajaryan that's correct, this was submitted here as a convenience for devs wanting to generate the protos.

Ultimately, I think it might make the most sense for the opentelemetry-proto package for python to be created from this repository since this is where the code lives, but that would require the process for packaging and publishing to be available here. Is the current plan that the SIG for each language maintain that process?

tigrannajaryan:
> ... Is the current plan that the SIG for each language maintain that process?

I am not sure what the ProtoBuf generation process looks like for Python an other languages so I don't have a good answer for this.

PR #153, opened 2020-05-26
Rename Span Event name to short_name

tigrannajaryan:
This is part of the alignment effort defined here https://github.com/open-telemetry/opentelemetry-specification/issues/622

The comment added clarifies the semantics, but I believe this is compatible
with our previous understanding and doe not change that understanding.

This does not break the protocol since renaming the fields does not change
the wire format.

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -170,9 +170,11 @@ message Span {
     // time_unix_nano is the time the event occurred.
     fixed64 time_unix_nano = 1;
 
-    // name of the event.
-    // This field is semantically required to be set to non-empty string.
-    string name = 2;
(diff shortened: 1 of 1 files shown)


Oberon00:
IMHO we should rather strive to keep span name and event name aligned than log name and event name. Related https://github.com/open-telemetry/opentelemetry-specification/issues/557

tigrannajaryan:
I think I agree with you. I am not sure my original proposal is the right one. Perhaps `Name` is a better choice for the field name and instead we should rename Log data model to also use `Name` instead of `ShortName`. Let's what others think in https://github.com/open-telemetry/opentelemetry-specification/issues/622 and I can submit an amendment to the Log Data Model OTEP.

tigrannajaryan:
Closing this for now.

Issue #106, opened 2020-02-17
Support array valued attributes


SergeyKanzhelev:
See https://github.com/open-telemetry/opentelemetry-specification/pull/368

Oberon00:
Should this get a milestone? Seems quite overdue.

arminru:
> Should this get a milestone? Seems quite overdue.

Quite overdue (and apparently overlooked), yes. Currently there are no milestones defined/used in this repo, however.

tigrannajaryan:
I'll take this.

PR #157, opened 2020-06-08
Add support for arrays and maps for attribute values

tigrannajaryan:
## Summary

This adds support for arrays and maps to attribute values, including support for
nested values.

This is a breaking protocol change.

Resolves: https://github.com/open-telemetry/opentelemetry-specification/issues/376
Resolves: https://github.com/open-telemetry/opentelemetry-proto/issues/106

## Motivation

There are several reasons for this change:

- The API defines that attributes values [may contain arrays of values](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-attributes).
  However the protocol has no way of representing array values. We need to add
  such capability.

- We intend to support Log data type in the protocol, which also requires array
  values (it is a Log Data Model requirement). In addition, Log data type
  requires support of key-value lists (maps) as attribute values, including
  nested values.

- There are long-standing requests to support nested values, arrays and maps for
  attributes:
  https://github.com/open-telemetry/opentelemetry-specification/issues/376
  https://github.com/open-telemetry/opentelemetry-specification/pull/596

This change introduces AnyValue. AnyValue can represent arbitrary numeric,
boolean, string, arrays or maps of values and allows nesting.

AttributeKeyValue now uses AnyValue to store the "value" part.

Note: below "Current" refers to the state of the "master" branch before this
PR/commit is merged. "Proposed" refers to the schema suggested in this
PR/commit.

## Performance

This change has a negative impact on the performance when using canonical Go
ProtoBuf compiler (compared to current OTLP state):

```
BenchmarkEncode/Current/Trace/Attribs-8              	     813	   1479588 ns/op
BenchmarkEncode/Proposed/Trace/Attribs-8             	     417	   2873476 ns/op
BenchmarkEncode/OpenCensus/Trace/Attribs-8           	     162	   7354799 ns/op

BenchmarkDecode/Current/Trace/Attribs-8              	     460	   2646059 ns/op	 1867627 B/op	   36201 allocs/op
BenchmarkDecode/Proposed/Trace/Attribs-8             	     246	   4827671 ns/op	 2171734 B/op	   56209 allocs/op
BenchmarkDecode/OpenCensus/Trace/Attribs-8           	     154	   7560952 ns/op	 2775949 B/op	   76166 allocs/op
```

However, I do not think this is important for most applications. Serialization
CPU and Memory usage is going to be a tiny portion of consumed resources for
most applications, except certain specialized ones.

For the perspective I am also showing OpenCensus in the benchmark to make it
clear that we are still significantly faster than it despite becoming slower
compared to the current state.

More importantly, performance critical applications can use Gogo ProtoBuf
compiler (Collector does use it), which _gains_ performance due to this change:

```
BenchmarkEncode/Current(Gogo)/Trace/Attribs-8        	    1645	    705385 ns/op
BenchmarkEncode/Proposed(Gogo)/Trace/Attribs-8       	    1555	    698771 ns/op

BenchmarkDecode/Current(Gogo)/Trace/Attribs-8        	     537	   2241570 ns/op	 2139634 B/op	   36201 allocs/op
BenchmarkDecode/Proposed(Gogo)/Trace/Attribs-8       	     600	   2053120 ns/op	 1323287 B/op	   46205 allocs/op
```

With Gogo compiler proposed approach uses 40% less memory than the current
schema.

After considering all tradeoffs and alternates (see below) I believe this
proposal is the best overall approach for OTLP. It is idiomatic ProtoBuf, easy
to read and understand, is future-proof to adding new attribute types, has
enough flexibility to represent simple and complex attribute values for all
telemetry types and can be made fast by custom code generation for applications
where it matters using Gogo ProtoBuf compiler.

Note: all performance measurements are done for Go implementation only (although
it is expected that other languages should exhibit somewhat similar behavior).

## Alternates Considered

I also designed and benchmarked several alternate schemas, see below.

### Adding array value to AttributeKeyValue

This is the simplest approach. It doubles down on the current OTLP protocol
approach and simply adds "array_values" field to AttributeKeyValue, e.g.:

```proto
message AttributeKeyValue {
  // all existing fields here.

  // A list of values. "key" field of each element in the list is ignored.
  repeated AttributeKeyValue array_values = 7;
}
```

This eliminates the need to have a separate AnyValue message and has lower CPU
usage because it requires less indirections and less memory allocations per
value. However, this is semantically incorrect since the elements of the array
must actually be values not key-value pairs, which this schema violates. It also
uses more memory than the proposed approach:

```proto
BenchmarkEncode/Proposed/Trace/Attribs-8             	     400	   2869055 ns/op
BenchmarkEncode/MoreFieldsinAKV/Trace/Attribs-8      	     754	   1540978 ns/op

BenchmarkDecode/Proposed/Trace/Attribs-8             	     250	   4790010 ns/op	 2171741 B/op	   56209 allocs/op
BenchmarkDecode/MoreFieldsinAKV/Trace/Attribs-8      	     420	   2806918 ns/op	 2347827 B/op	   36201 allocs/op
```

It will become even worse memory-wise if in the future we need to add more data
types to attributes. This approach is not scalable for future needs and is
semantically wrong.

### Fat AnyValue instead of oneof.

In this approach AnyValue contains all possible field values (similarly to how
AttributeKeyValue is currently):

```proto
message AnyValue {
    ValueType type = 1;
    bool bool_value = 2;
    string string_value = 3;
    int64 int_value = 4;
    double double_value = 5;
    repeated AnyValue list_values = 6;
    repeated AttributeKeyValue kvlist_values = 7;
}
message AttributeKeyValue {
    string key = 1;
    AnyValue value = 2;
}
```

This results in significantly bigger AnyValue in-memory. In vast majority of
cases attribute values of produced telemetry are strings (see e.g. semantic
conventions for proof). Integer and boolean values are also used, although
significantly less frequently than strings. Floating point number, arrays and
maps are likely going to be diminishingly rare in the attributes. If we keep all
these value types in AnyValue we will pay the cost for all these fields although
almost always only string value would be present.

Here are benchmarks comparing proposed schema and schema with fat AnyValue and
using string and integer attributes in spans:

```
BenchmarkEncode/Proposed/Trace/Attribs-8             	     415	   2894513 ns/op	  456866 B/op	   10005 allocs/op
BenchmarkEncode/FatAnyValue/Trace/Attribs-8          	     646	   1885003 ns/op	  385024 B/op	       1 allocs/op

BenchmarkDecode/Proposed/Trace/Attribs-8             	     247	   4872270 ns/op	 2171746 B/op	   56209 allocs/op
BenchmarkDecode/FatAnyValue/Trace/Attribs-8          	     343	   3423494 ns/op	 2988081 B/op	   46205 allocs/op
```

Memory usage with this approach is much higher and it also will become worse as
we add more types.

### AnyValue plus ExoticValue

This is based on fat AnyValue approach but rarely used value types are moved to
a separate ExoticValue message that may be referenced from AnyValue if needed:

```proto
message AnyValue {
    ValueType type = 1;
    bool bool_value = 2;
    string string_value = 3;
    int64 int_value = 4;
    ExoticValue exotic_value = 5;
}
message ExoticValue {
    double double_value = 1;
    repeated AnyValue array_values = 2;
    repeated AttributeKeyValue kvlist_values = 3;
}
message AttributeKeyValue {
    string key = 1;
    AnyValue value = 2;
}
```

While this improves the performance (particularly lowers memory usage for most
frequently used types of attributes) it is awkward and sacrifices too much
readability and usability for small performance gains. Also for the rare cases
it is slow and uses even more memory so its edge case behavior is not desirable.

### Using different schema for log data type

I also considered using a different message definition for LogRecord attributes
and Spans. This would allow to eliminate some of the requirements that we do not
yet formally have for Span attributes (particularly the need to have maps of
nested values).

However, this does not help much in terms of performance, makes Span and
LogRecord attributes non-interchangeable and significantly increases the bloat
of code in applications that need to work with both Spans and Log records.


tigrannajaryan:
@open-telemetry/specs-approvers please review. I am working on signing the CLA.

opentelemetry/proto/common/v1/common.proto
```diff
@@ -21,30 +21,37 @@ option java_package = "io.opentelemetry.proto.common.v1";
 option java_outer_classname = "CommonProto";
 option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/common/v1";
 
+// AnyValue is used to represent any type of attribute value. AnyValue may contain a
+// primitive value such as a string or integer or it may contain an arbitrary nested
+// object containing arrays, key-value lists and primitives.
+message AnyValue {
+  // The value is one of the listed fields. It is valid for all values to be unspecified
+  // in which case this AnyValue is considered to be "null".
+  oneof value {
```
bogdandrutu, line 30: Can we avoid using oneof? It is bad for a lot of languages as well as the JSON representation.

opentelemetry/proto/common/v1/common.proto
```diff
@@ -21,30 +21,37 @@ option java_package = "io.opentelemetry.proto.common.v1";
...
+    AttributeKeyValueList kvlist_values = 6;
+  }
+}
+
+message ArrayValue {
+  // Array of values. The array may be empty (contain 0 elements).
+  repeated AnyValue array_values = 1;
+}
+
+message AttributeKeyValueList {
```
bogdandrutu, line 45: why is this message needed? can you have `repeated AttributeKeyValue kvlist_values = 6;` in the AnyValue?

opentelemetry/proto/common/v1/common.proto
```diff
@@ -21,30 +21,37 @@ option java_package = "io.opentelemetry.proto.common.v1";
...
+    int64 int_value = 3;
+    double double_value = 4;
+    ArrayValue array_values = 5;
+    AttributeKeyValueList kvlist_values = 6;
+  }
+}
+
+message ArrayValue {
+  // Array of values. The array may be empty (contain 0 elements).
+  repeated AnyValue array_values = 1;
```
bogdandrutu, line 42: why is this message needed? can you have `repeated ArrayValue array_values = 5;` in the AnyValue?

opentelemetry/proto/common/v1/common.proto
tigrannajaryan: I have tried a lot of versions without oneof but they are all either ugly or use too much memory. Can you point to the exact problems caused by oneof? OpenCensus proto used it, so I assume it is not that bad :-)

opentelemetry/proto/common/v1/common.proto
tigrannajaryan: This is required by oneof. You cannot have a repeated field inside oneof.

opentelemetry/proto/common/v1/common.proto
tigrannajaryan: Same, required by oneof.

opentelemetry/proto/common/v1/common.proto
```diff
@@ -21,30 +21,37 @@ option java_package = "io.opentelemetry.proto.common.v1";
...
+message AnyValue {
+  // The value is one of the listed fields. It is valid for all values to be unspecified
+  // in which case this AnyValue is considered to be "null".
+  oneof value {
+    string string_value = 1;
+    bool bool_value = 2;
+    int64 int_value = 3;
+    double double_value = 4;
+    ArrayValue array_values = 5;
+    AttributeKeyValueList kvlist_values = 6;
```
jmacd, line 36: There's a little inconsistency here. Can we use "array_value" instead of "array_values"? The array is the value and the type is "ArrayValue".

Can we use a type named "MapValue" or "DictionaryValue" instead of "AttributeKeyValueList"? Can we name the corresponding field "map_value" or "dictionary_value"? Use of "map" or "dictionary" implies that the keys will be unique, whereas "list" does not.

opentelemetry/proto/common/v1/common.proto
jmacd: Can we name this just "values"?  It's already inside a thing named "ArrayValue", so "values" seems like enough information.

opentelemetry/proto/common/v1/common.proto
```diff
@@ -21,30 +21,37 @@ option java_package = "io.opentelemetry.proto.common.v1";
...
+}
+
+message ArrayValue {
+  // Array of values. The array may be empty (contain 0 elements).
+  repeated AnyValue array_values = 1;
+}
+
+message AttributeKeyValueList {
+  // List of key-value pairs. The list may be empty (contain 0 elements).
+  repeated AttributeKeyValue kvlist_values = 1;
```
jmacd, line 47: Same comment as above. Can we name this "values"?

jmacd: (Oh, then I remember that protoc adds an "s" to the end of repeated fields. So maybe this should be just "value". Same above.)

opentelemetry/proto/common/v1/common.proto
tigrannajaryan: Renamed to `values`.

opentelemetry/proto/common/v1/common.proto
tigrannajaryan: Renamed to `array_value` and `kvlist_value`.

I am not sure about using the term `map` here. From protocols perspective this is a list of elements. See also trace.proto `Span.attributes` where we describe it as "a collection of key/value pairs" and avoid calling it a map.

The proto here avoids taking a stance and specifying whether it is a map or multimap (both are possible for a key-value pair list). I have a slight preference to keep it this way to be aligned with how `Span.attributes` are defined.

Thoughts?

opentelemetry/proto/common/v1/common.proto
tigrannajaryan: Changed to `values`. I think plural works better here. We use plural for all repeated fields elsewhere.

opentelemetry/proto/common/v1/common.proto
jmacd: Sounds good. thanks.

opentelemetry/proto/common/v1/common.proto
bogdandrutu: That was mostly a rumour I heard :)). But I couldn't find how oneof is represented as JSON. It may be good to look at generated code.

opentelemetry/proto/common/v1/common.proto
```diff
@@ -21,30 +21,37 @@ option java_package = "io.opentelemetry.proto.common.v1";
...
+    int64 int_value = 3;
+    double double_value = 4;
+    ArrayValue array_value = 5;
+    AttributeKeyValueList kvlist_value = 6;
+  }
+}
+
+message ArrayValue {
+  // Array of values. The array may be empty (contain 0 elements).
+  repeated AnyValue values = 1;
```
arminru, line 42: This would allow heterogeneous arrays, whereas the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/1729bc4337bdc1df83ff7eab53eb7405d6e75665/specification/trace/api.md#set-attributes) prescribes:
> The array MUST be homogeneous, i.e. it MUST NOT contain values of different types.

Should this be reflected in the protocol as well or are we fine with having this constraint ensured only by the SDK implementations?

jmacd: I'll advocate for changing the specification. I have never understood why it matters that arrays are homogenous when JSON values are what people use in the real world.

bogdandrutu: The protocol can be a bit more generic than the API. With this we can ensure that we have a path forward and we don't have to break the protocol if that changes.

opentelemetry/proto/common/v1/common.proto
tigrannajaryan: I went ahead and applied this PR's proposal on a branch in Collector. `oneof` seems to be doing fine. I tried gRPC-gateway which uses JSON and it works fine. This is what the payload looks like (see "attributes.value"):

```json
{
    "resource_spans": [
        {
            "resource": {
                "attributes": [
                    {
                        "key": "host.hostname",
                        "value": {
                            "stringValue": "testHost"
                        }
                    }
                ]
            },
            "instrumentation_library_spans": [
                {
                    "spans": [
                        {
                            "trace_id": "W47/95gDgQPSabYzgT/GDA==",
                            "span_id": "7uGbfsPBsXM=",
                            "name": "testSpan",
                            "start_time_unix_nano": 1544712660000000000,
                            "end_time_unix_nano": 1544712661000000000,
                            "attributes": [
                                {
                                    "key": "attr1",
                                    "value": {
                                        "intValue": 55
                                    }
                                }
                            ]
                        }
                    ]
                }
            ]
        }
    ]
}
```

opentelemetry/proto/common/v1/common.proto
tigrannajaryan: @arminru The protocol does not have to be an exact match of the spec. It needs to be able to efficiently and unambiguously represent what the spec defines. This schema can unambiguously represent homogeneous arrays. It happens to also be able to represent heterogeneous arrays, no harm from that. In fact, we gain ability to represent JSON-like data properly, which is a big benefit IMO.

@jmacd some motivation for homogeneous arrays in spec is here: https://github.com/open-telemetry/opentelemetry-specification/issues/341#issuecomment-559708227
I don't have a strong opinion on this.

tigrannajaryan:

opentelemetry/proto/common/v1/common.proto
tigrannajaryan: @open-telemetry/javascript-approvers any concerns about using `oneof`?

obecny: I think it looks straightforward :)

bogdandrutu:
@open-telemetry/specs-approvers would like to merge this soon. Please approve/comment

opentelemetry/proto/common/v1/common.proto
```diff
@@ -21,30 +21,37 @@ option java_package = "io.opentelemetry.proto.common.v1";
...
+}
+
+message AttributeKeyValueList {
```
yurishkuro, line 45: I'm somewhat bothered by the Attribute prefix here - do we intend to reuse the term "attribute" for every instance of key-value data? E.g. a field in the log entry, will it also be called "attribute"?

If not, then this type should be simply called KeyValueList and the next one KeyValue, to match AnyValue.

jmacd: +1 

Great point! Let's name it `KeyValueList`.

yurishkuro: I am guessing renaming AttrributeKeyValue->KeyValue will be a bigger deal since it's an existing name.

tigrannajaryan: This was annoying me too a bit. I can rename to KeyValue, no problem with that. We are breaking the proto anyway.

The only reason I was reluctant to use KeyValue was because we also have StringKeyValue. So we will have KeyValue which is a String:AnyValue pair and StringKeyValue which is a String:String pair.

Does this look OK to you?

I would not want to name it KeyAnyValue, it does not looks nice to me.

yurishkuro: KeyValue and StringKeyValue sgtm 👍

tigrannajaryan: OK, changing to KeyValue and KeyValueList.

tigrannajaryan: Done.

tigrannajaryan:
@yurishkuro PTAL.

tigrannajaryan:
@open-telemetry/specs-approvers  I need one more review/approval please.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -175,7 +175,7 @@ message Span {
     string name = 2;
 
     // attributes is a collection of attribute key/value pairs on the event.
-    repeated opentelemetry.proto.common.v1.AttributeKeyValue attributes = 3;
+    repeated opentelemetry.proto.common.v1.KeyValue attributes = 3;
```
bogdandrutu, line 178: It seems a bit confusing to use the repeated field here when there is a List message defined. I understand the reasoning but would be good to have a comment.

opentelemetry/proto/common/v1/common.proto
```diff
@@ -21,34 +21,41 @@ option java_package = "io.opentelemetry.proto.common.v1";
...
+message ArrayValue {
+  // Array of values. The array may be empty (contain 0 elements).
+  repeated AnyValue values = 1;
+}
 
-  // Only one of the following fields is supposed to contain data (determined by `type` field).
-  // This is deliberately not using Protobuf `oneof` for performance reasons (verified by benchmarks).
+message KeyValueList {
+  // List of key-value pairs. The list may be empty (contain 0 elements).
+  repeated KeyValue values = 1;
```
bogdandrutu, line 47: Is this real list or a map represented as a list? Please comment.

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: Added a comment near `KeyValueList` message definition.

opentelemetry/proto/common/v1/common.proto
tigrannajaryan: From protocol's perspective this is a list of elements. See also trace.proto Span.attributes where we describe it as "a collection of key/value pairs" and avoid calling it a map.

The proto here avoids taking a stance and specifying whether it is a map or multimap (both are possible for a key-value pair list). I have a slight preference to keep it this way to be aligned with how Span.attributes are defined.

I updated the comment here to match what we have for `Span.attributes`.

PR #158, opened 2020-06-15
Remove Temporality and add Structure

MrAlias:
[Consensus was reached](https://docs.google.com/document/d/1LfDVyBJlIewwm3a0JtDtEjkusZjzQE3IAix8b0Fxy3Y/edit?ts=5ee27efe#) that the description of the time interval over which data was measured can be determined as `DELTA` or `CUMULATIVE` by comparison of the timestamps sent with the data. The classification of `INSTANTANEOUS` has also been found irrelevant in meaningfully description of the data. For these reasons, `Temporality` is removed.

Conversely, an understanding of what operations can be performed on data is essential information to contain in this protocol so it can support systems of statistics or roll-up. To provide this support, the `Structure` enum is added and an appropriate field in the Metric descriptor included to capture
if the data is adding or grouping.

Relates to #125 
Related to https://github.com/open-telemetry/opentelemetry-specification/issues/625


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -161,82 +161,49 @@ message MetricDescriptor {
...
-    // it receives and reports the sum of these requests every second as a
-    // DELTA metric:
+  // Structure describes the relation between data of a Metric. This
+  // relation is expressed by the operation(s) supported by the data. The
+  // Structure of data is important to know if further analysis or roll-up
+  // of the data is to be performed. Without knowledge of data Structure
+  // meaningless results could be produced from these tasks.
+  //
+  // Note:
+  // Knowing the Structure of data does not guarantee meaningfully results
```
yurishkuro, line 171:
```suggestion
  // Knowing the Structure of data does not guarantee meaningful results
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -161,82 +161,49 @@ message MetricDescriptor {
...
+  // for arbitrary operations on the data, rather it means that the results
+  // of performing the operation(s) related to the data Structure COULD
+  // produce meaningful results.
+  //
+  // Take for example adding together two different the measurements of
```
yurishkuro, line 176:
```suggestion
  // Take for example adding together two different measurements of
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -161,82 +161,49 @@ message MetricDescriptor {
...
+    // does not support any operation(s) or it is unknown.
     //
-    // For example, consider a system measuring the number of requests that
-    // it receives and reports the sum of these requests every second as a
-    // CUMULATIVE metric:
+    // Examples of GROUPING data:
+    // - CPU core number that was the hottest.
+    // - If a system was stopped, crashed, or running (represented by the
+    //   integers '0', '1', or '2' respectively).
+    // - The average power consumption of a server.
```
yurishkuro, line 192: This could be transformed into average power per server across data center, is it not a "combination operation"?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -161,82 +161,49 @@ message MetricDescriptor {
...
+  // Knowing the Structure of data does not guarantee meaningfully results
+  // for arbitrary operations on the data, rather it means that the results
+  // of performing the operation(s) related to the data Structure COULD
+  // produce meaningful results.
+  //
+  // Take for example adding together two different the measurements of
+  // request counts from a server. This data has ADDING structure, but if
+  // the two measurements were for different services, or even different
+  // servers, this result is likely meaningless.
+  enum Structure {
```
yurishkuro, line 180: Is there a corresponding change to the API?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -161,82 +161,49 @@ message MetricDescriptor {
...
+  // Note:
+  // Knowing the Structure of data does not guarantee meaningfully results
+  // for arbitrary operations on the data, rather it means that the results
+  // of performing the operation(s) related to the data Structure COULD
+  // produce meaningful results.
+  //
+  // Take for example adding together two different the measurements of
+  // request counts from a server. This data has ADDING structure, but if
+  // the two measurements were for different services, or even different
+  // servers, this result is likely meaningless.
```
jmacd, line 179: I'm not sure I agree. It the metric counts requests, it's meaningful to combine them in as much as you get a combined count, which is well defined. If the metric is the same, it should be a meaningful aggregation: it's a count of requests from multiple servers, I think.

The best example, for me, of a structural property is with measurements from a GROUPING instrument. If you have one individual measurement, there is nothing you can combine with it to make it into a different individual measurement.  Combining measurements with a GROUPING instruments gives you more measurements.  (Not sure that helps, but it might be a better example?)

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: (This brings the protocol closer to the existing API specification.  ADDING structure is generated by Counter, UpDownCounter, SumObserver, and UpDownSumObserver instruments. GROUPING structure is generated by ValueObserver and ValueRecorder instruments.)

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: I agree. I think you can add power.
How about "request latency" instead?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -161,82 +161,49 @@ message MetricDescriptor {
...
-    //      t_0+2 with a value of 5.
-    //   9. The system experiences a fault and loses state.
-    //   10. The system recovers and resumes receiving at time=t_1.
-    //   11. A request is received, the system measures 1 request.
-    //   12. The 1 second collection cycle ends. A metric is exported for the
-    //      number of requests received over the interval of time t_1 to
-    //      t_0+1 with a value of 1.
-    CUMULATIVE = 3;
+    // Examples of ADDING data:
+    // - CPU core temperature.
```
jmacd, line 199: Temperature belongs in GROUPING, I think. We would average it, not sum it, I believe.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: Echoing what @jmacd said, this doesn't change the API. The added dimension allows the data to not lose meaning during transport.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: Removed.

tigrannajaryan:
@MrAlias can you please show how this would map to OpenCensus [Proto metric types](https://github.com/census-instrumentation/opencensus-proto/blob/be218fb6bd674af7519b1850cdf8410d8cbd48e8/src/opencensus/proto/metrics/v1/metrics.proto#L73)? We will need to do this translation in the Collector but I am not sure how it will need to be mapped.

MrAlias:
> @MrAlias can you please show how this would map to OpenCensus [Proto metric types](https://github.com/census-instrumentation/opencensus-proto/blob/be218fb6bd674af7519b1850cdf8410d8cbd48e8/src/opencensus/proto/metrics/v1/metrics.proto#L73)? We will need to do this translation in the Collector but I am not sure how it will need to be mapped.

@tigrannajaryan that is a good question. My understanding is that the `Structure` property describes a dimension to the data that was not previously included in the OpenCensus protocol.

@jmacd does this sound correct to you?

jmacd:
I am starting to have doubts about this change. Initially my goal was to provide some information about the original instrument--I'd like to know whether the inputs were ADDING or GROUPING, but I'm not sure it matters for the consumer of an aggregation whether the inputs were ADDING or GROUPING. I still think we should be able to know the kind of the source instrument, but it's not clear how the information will be used.

I have a different kind of doubt about the translation to and from the [OpenCensus format](https://github.com/census-instrumentation/opencensus-proto/blob/be218fb6bd674af7519b1850cdf8410d8cbd48e8/src/opencensus/proto/metrics/v1/metrics.proto#L73). I do not believe we have a correct conversion from OpenCensus into the protocol before this PR, much less after this PR. 

The OpenCensus GAUGE_INT64, GAUGE_DOUBLE, GAUGE_DISTRIBUTION would become Temporality=INSTANTANEOUS before this PR, Structure=GROUPING after this PR, I think. The OpenCensus CUMULATIVE_INT64 CUMULATIVE_DOUBLE CUMULATIVE_DISTRIBUTION could become either DELTA or CUMULATIVE. OpenCensus is not telling whether the caller should expect the next interval to reset the start time or not. In that sense, we're better off AFTER this PR because there is no more temporality field: OpenCensus is not telling us temporality.

One of my colleagues made an argument for keeping Temporality (at least the notion of Cumulative vs Delta) in OTLP. The argument goes: Sure, you can infer whether a point is DELTA or CUMULATIVE after you've seen two data points, but you're likely to want to validate the timestamps, and the validation logic will be different. Having explicit information will make processing CUMULATIVE and DELTA data easier, the argument goes, because an individual point will carry that information. (To repeat, OpenCensus doesn't tell us this.)

I am struggling to find a conclusion to this. I don't know that we need either Structure or Temporality. I still find myself wanting to know the kind of the original instrument, but I'm not convinced that's needed either.

MrAlias:
After the Metric SIG meeting today it sounds like we are circling on the correct direction here, but not strongly enough to move this forward without exploring other issues and approaches.

Going to close with the anticipation it will be retried in a week or so after we have further explored these topics.

PR #159, opened 2020-06-22
Add Exemplar support to Metrics proto

cnnradams:
Adds support for [OTEP#113](https://github.com/open-telemetry/oteps/pull/113). This should also handle [duplicate labels in exemplars](https://github.com/open-telemetry/oteps/pull/113#discussion_r438950970) by bringing labels up a level and making exemplars only hold `additional_labels`.

Proto question (since I'm new to this): I defined a `measurement_type` enum for the RawValue type, but couldn't create a new enum with just INT64 and DOUBLE (because I can't share names with other enums). So right now I'm using `Type`, which means things other than INT64 and DOUBLE can be picked for `measurement_type`. What is the right solution for this?


cnnradams:
Closing in favor of #162 (which implements exemplars)

jmacd:
I reopened this and put it in draft state. This has been referred to heavily in #168, and after #168 merges this PR should be restored to roughly its original state.

cnnradams:
updated to not make structual changes and to add raw_value_data_points. Still need to specify the type of RawValue in some way, which will be unblocked after #168

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -104,6 +133,7 @@ message Metric {
   repeated DoubleDataPoint double_data_points = 3;
   repeated HistogramDataPoint histogram_data_points = 4;
   repeated SummaryDataPoint summary_data_points = 5;
+  repeated RawValue raw_data_points = 6;
```
bogdandrutu, line 136: I think this should be removed for the moment, this is not exemplar, is supporting "RawMeasurements" which is out of scope for this PR.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -44,6 +44,35 @@ message InstrumentationLibraryMetrics {
   repeated Metric metrics = 2;
 }
 
+// A representation of raw measurements, which can have statistical meaning based
+// on how the measurement was sampled, as well as a span ID and trace ID of the
+// active span when the measurement was recorded
+message RawValue {
```
bogdandrutu, line 50: Would be good to be consistent for the ordering with data points, so maybe labels first, etc.

bogdandrutu:
The generated code was removed, please remove that as well. Rebase the PR please.

opentelemetry/proto/metrics/v1/metrics.proto
cnnradams: removed

opentelemetry/proto/metrics/v1/metrics.proto
cnnradams: fixed

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -44,6 +44,33 @@ message InstrumentationLibraryMetrics {
...
+  // The set of labels that were dropped by the aggregator, but recorded
+  // alongside the original measurement. Only labels that were dropped by the aggregator should be included
+  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1;
+
+  // Exact time that the measurement was recorded
+  fixed64 time_unix_nano = 2;
```
bogdandrutu, line 56: // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -44,6 +44,33 @@ message InstrumentationLibraryMetrics {
   repeated Metric metrics = 2;
 }
 
+// A representation of raw measurements, which can have statistical meaning based
+// on how the measurement was sampled, as well as a span ID and trace ID of the
+// active span when the measurement was recorded
+message RawValue {
+  // The set of labels that were dropped by the aggregator, but recorded
+  // alongside the original measurement. Only labels that were dropped by the aggregator should be included
+  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1;
```
bogdandrutu, line 53: What is the relationship between these labels and the labels in the DataPoint:
* Do we duplicate them?
* Do we extract these labels and the actual set of labels is the combination of these + datapoint.lables?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -44,6 +44,33 @@ message InstrumentationLibraryMetrics {
...
+  // Exact time that the measurement was recorded
+  fixed64 time_unix_nano = 2;
+
+  // Numerical value of the measurement that was recorded. Only one of these 
+  // two fields is used for the data, based on MetricDescriptor.measurement_value_type
+  double double_value = 3;
+  int64 int64_value = 4;
+
+  // (Optional) Span ID of the current trace
+  bytes span_id = 5;
```
bogdandrutu, line 64: Would be good to explain that this may be missing if the measurement is not recorded inside a "request"/"trace"

lubingfeng:
@jmacd  When can we get  #159 and #171 closed, which are blockers for getting stats proposal finalized, based on discussion two weeks ago.

opentelemetry/proto/metrics/v1/metrics.proto
cnnradams: For exemplars these labels are only the labels _not_ included in the DataPoint's labels. I would change this field to be `dropped_labels` but if RawValue is used as a data point itself the labels would include all labels

opentelemetry/proto/metrics/v1/metrics.proto
cnnradams: done

opentelemetry/proto/metrics/v1/metrics.proto
cnnradams: added comments

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: @jmacd I know you tried to share the messages, can you help define the behavior here?

bogdandrutu:
@lubingfeng 
> @jmacd When can we get #159 and #171 closed, which are blockers for getting stats proposal finalized, based on discussion two weeks ago.

Not sure I understand:
* 159 is this PR
* 171 is merged

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -452,6 +491,10 @@ message HistogramDataPoint {
   // If we decide to also support (a, b] intervals we should add support for these by defining
   // a boolean value which decides what type of intervals to use.
   repeated double explicit_bounds = 7;
+
+  // (Optional) List of exemplars collected from 
+  // measurements that were used to form the data point
+  repeated RawValue exemplars = 8;
```
aabmass, line 497: Can you have more than one exemplar per bucket, similar to how you made the exemplars repeated for int/double? If so the indices wouldn't line up anymore with `bucket_counts`.

Idk the performance implications, but you could also have a `message Bucket` with the counts + bounds + exemplar for that bucket?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -508,4 +551,8 @@ message SummaryDataPoint {
   // A list of values at different quantiles of the distribution calculated
   // from the current snapshot. The quantiles must be strictly increasing.
   repeated ValueAtQuantile quantile_values = 6;
+
+  // (Optional) List of exemplars collected from 
+  // measurements that were used to form the data point
+  repeated RawValue exemplars = 7;
```
aabmass, line 557: Would this just be a list of random samples from the whole window? [Open question](https://github.com/open-telemetry/oteps/blob/master/text/metrics/0113-exemplars.md#open-questions) in the OTEP:

> We don’t have a strong grasp on how the sketch aggregator works in terms of implementation - so we don’t have enough information to design how exemplars should work properly.

opentelemetry/proto/metrics/v1/metrics.proto
cnnradams: Exemplars are not sent as per-bucket - that can be determined afterwards if necessary. You can have multiple exemplars per bucket and they will all be combined here - indices not relevant.

opentelemetry/proto/metrics/v1/metrics.proto
cnnradams: The proto does not define how the exemplars were sampled, not sure your question?

aabmass: Nvm, I see now

jmacd:
I left a lengthy remark on this topic here: 
https://github.com/open-telemetry/opentelemetry-specification/issues/617#issuecomment-670818918

I am worried that my request for @cnnradams to explore and implement statistical sampling for exemplars has led to some confusion, and (with my apologies) I am willing to omit it, but as noted in the comment, there are many related questions and even if we take the statistical question out of it, we're left with tough questions.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: @bogdandrutu does this sound right to you, for now?

bogdandrutu: >We also agreed to move the Exemplars into the DataPoints so that they can refer to dropped labels, not include full label sets in each point.

I think we agreed that you will evaluate what is better for performance/semantics:
1. Having a `repeated RawValue exemplars` in the `Metric` that applies to all data points (user may need to do another remapping to every data point) vs `repeated RawValue exemplars` in every point (if we go with every point then "dropped_labels" is better name for that).

My point is that I don't have a strong opinion between the both, and was trying to make you investigate and decide which way. Here are my thoughts:
* Having `repeated RawValue exemplars` in the Metrics:
  * Pros:
    * Saves some memory in the internal representation (have extra 24 bytes per data point).
    * Same message may be able to be re-used with raw-measurements because labels don't represent dropped.
  * Cons:
     * Duplicate labels on the wire.
     * User needs to re-map every exemplar to the data point by doing the labels matching.

I feel `cons` are more "significant" than `pros`, so personally I would go with exemplars in every DataPoint as you suggested.

bogdandrutu: If we go with exemplars in every DataPoint I would say to rename the message to "Exemplar" :)

jmacd:
@cnnradams I know your internship is over, but let us know if you're willing to make these changes.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: This discussion makes me want a way to intern label sets to avoid the "re-map every exemplar" problem.

I don't feel inclined to invest time in this now, so we should probably choose "repeated RawValue exemplars in every point".

cnnradams:
sure. from reading the discussion, it seems like the only things I need to change are `RawValue` -> `Exemplar`, `labels` -> `dropped_labels`, and remove `sample_count`? Since there is already an exemplar set per datapoint?

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: > This discussion makes me want a way to intern label sets to avoid the "re-map every exemplar" problem.

Even with an "intern label" you still need to map every exemplar to a point (that mapping may be faster if we have "intern label" but still needs some work)

cnnradams:
> sure. from reading the discussion, it seems like the only things I need to change are `RawValue` -> `Exemplar`, `labels` -> `dropped_labels`, and remove `sample_count`? Since there is already an exemplar set per datapoint?

done all 3.

PR #163, opened 2020-07-03
Allow negative bucket bounds in explicit_bounds

### LINKED ISSUE #156: Allow histograms to specify negative bucket bounds
According to the proto documentation, histogram bucket bounds [can only be positive](https://github.com/open-telemetry/opentelemetry-proto/blob/b54688569186e0b862bf7462a983ccf2c50c0547/opentelemetry/proto/metrics/v1/metrics.proto#L362-L385), which makes it impossible for OpenTelemetry to be fully compatible with Prometheus, which [allows negative bounds as well](https://prometheus.io/docs/practices/histograms/#count-and-sum-of-observations).

Prometheus receiver doesn't respect this and sends negative bucket bounds down OpenTelemetry collector pipelines anyway. This can lead to bugs, as processors/exporters can make incorrect assumptions about the data (in fact, at least Stackdriver exporter already does).

This is a request to extend OpenTelemetry data model to treat first histogram bucket as `(-inf, buckets[0])` instead of `[0, buckets[0])`.

/cc @bogdandrutu @rghetia @nilebox @mayurkale22

x13n:
Fixes #156 

@nilebox @bogdandrutu @jmacd


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -373,14 +373,14 @@ message HistogramDataPoint {
   // This defines size(bounds) + 1 (= N) buckets. The boundaries for bucket
   // at index i are:
   //
-  // [0, bounds[i]) for i == 0
+  // (-infinity, bounds[i]) for i == 0
   // [bounds[i-1], bounds[i]) for 0 < i < N-1
   // [bounds[i], +infinity) for i == N-1
   // The values in bounds array must be strictly increasing and > 0.
```
nilebox, line 379: The requirement `and > 0` should probably be removed, since with `-infinity` lower bound there could be negative values?

x13n: Good catch, thanks! Removed.

bogdandrutu:
@jmacd You are the ultimate decider for this :)

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -373,14 +373,14 @@ message HistogramDataPoint {
...
   // [bounds[i-1], bounds[i]) for 0 < i < N-1
   // [bounds[i], +infinity) for i == N-1
-  // The values in bounds array must be strictly increasing and > 0.
+  // The values in bounds array must be strictly increasing.
   //
-  // Note: only [a, b) intervals are currently supported for each bucket. If we decides
-  // to also support (a, b] intervals we should add support for these by defining a boolean
-  // value which decides what type of intervals to use.
+  // Note: only [a, b) intervals are currently supported for each bucket except the first one.
+  // If we decide to also support (a, b] intervals we should add support for these by defining
```
jmacd, line 382: You could strike lines 383-383 if you want. 😀

PR #164, opened 2020-07-08
Update codeowners to be consistent with specs and oteps

bogdandrutu:


CODEOWNERS
```diff
@@ -12,4 +12,14 @@
 #  https://help.github.com/en/articles/about-code-owners
 #
 
-*  @bogdandrutu @carlosalberto @SergeyKanzhelev @yurishkuro @tedsuo @iredelmeier @arminru @c24t @reyang @tigrannajaryan @jmacd
+# Global owners, will be the owners for everything in the repo. Membership is tracked via https://github.com/open-telemetry/community/blob/master/community-members.md 
+*   @open-telemetry/specs-approvers
+
+# Trace owners (global + trace)
+opentelemetry/proto/trace/    @open-telemetry/specs-approvers @open-telemetry/specs-trace-approvers
```
tigrannajaryan, line 19: Can you also add `opentelemetry/proto/collector/trace/v1/` and similarly for other signals?

bogdandrutu: I feel there are very few changes there, they are mostly constant that files.

tigrannajaryan: OK, we can add later if needed.

Oberon00:
I am not sure if the collector/trace/ directories, etc. should also be added, but with the current content, it seems OK not to have the area-specific code owners applied there (since it would more likely belong to a "collector" group than a "trace" group for example).

PR #168, opened 2020-07-15
Replace Temporality with Kind; Type with ValueType

### LINKED ISSUE #78: MetricDescriptor.Type enum items do not match specification

### LINKED ISSUE #124: Label keys/values in diagram in metrics.proto do not match the proto specification

jmacd:
This is a partial re-do of #162 with only the Kind and ValueType changes and more extensive comments.

Type renamed ValueType
Temporality values (CUMULATIVE, DELTA, INSTANTANEOUS) move into Kind
Structure values (ADDING_MONOTONIC, ADDING, GROUPING) included in Kind
Continuity values (SNAPSHOT, CONTINUOUS) included in Kind
Monotonic becomes a Structure

Part of #125
Resolves #124
Resolves #78


opentelemetry/proto/metrics/v1/gen.go
```diff
@@ -0,0 +1,119 @@
...
+
+var (
+	cumulativeProps = []string{"Instantaneous", "Delta", "Cumulative"}
+	addingProps     = []string{"Adding", "Grouping"}
+	monoProps       = []string{"Monotonic", ""}
+	syncProps       = []string{"Synchronous", "Asynchronous"}
+)
+
+func main() {
+	fmt.Println(`package main
```
MrAlias, line 39: Building the output program in totality and then passing through [`format.Source`](https://golang.org/pkg/go/format/#Source) would account for the style differences this output has.

opentelemetry/proto/metrics/v1/gen.go
```diff
@@ -0,0 +1,119 @@
...
+func main() {
+`)
+
+	maxNameSize := 0
+
+	for _, fn := range fullnames {
+		if maxNameSize < len(fn) {
+			maxNameSize = len(fn)
+		}
+	}
```
MrAlias, line 113: Is this needed?

jmacd:
FYI in [162](https://github.com/open-telemetry/opentelemetry-proto/pull/162#issuecomment-658379573) I measured the [cost of a `oneof` -ike arrangement and we continue to rule out this approach](https://github.com/open-telemetry/opentelemetry-proto/pull/162#issuecomment-658379573).

opentelemetry/proto/metrics/v1/gen.go
jmacd: Fixed. (Thanks I was looking for that!)

opentelemetry/proto/metrics/v1/gen.go
jmacd: Fixed.

opentelemetry/proto/metrics/v1/gen.go
```diff
@@ -0,0 +1,123 @@
...
+					fullvalue := fmt.Sprint("", strings.ToUpper(a), "|", strings.ToUpper(c), mfrag, sfrag)
+					fullvalues = append(fullvalues, fullvalue)
+					buf.WriteString(fmt.Sprint(fullname, " Kind = ", fullvalue, "\n"))
+				}
+			}
+		}
+	}
+	buf.WriteString(`)
+
+func main() {
```
bogdandrutu, line 108: a bit confused of this:
1. if this is the code generator, I would suggest to have it in "cmd/..."
2. I don't see the new generated code, is that intentional?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -86,94 +90,142 @@ message InstrumentationLibraryMetrics {
...
+  // may be interpreted differently depending on Structure.  For
+  // example, a Histogram DataPoint may be computed for a Counter
+  // (ADDING) instrument or a ValueRecorder (GROUPING) instrument, and
+  // the Sum and Count of the resulting aggregation may be interpreted
+  // differently depending on structure.
+  //
+  // MONOTONIC is a boolean option that may be applied to
+  // ADDING-structure metrics.
+  //
+  // ASYNCHRONOUS is a boolean option that may be applied to all
```
bogdandrutu, line 191: Sorry that I am keep repeating this, but I fail to understand the benefit of this bit.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -86,94 +90,142 @@ message InstrumentationLibraryMetrics {
...
+  // metrics.  When set, an ASYNCHRONOUS kind indicates that the
+  // corresponding measurement was made through a callback called by
+  // the SDK.  TimeUnixNano is deliberate, not the result of an
+  // application-level event.
+  //
+  // When measurements originate from an OpenTelemetry API, the
+  // ASYNCHRONOUS kind allows the interpreter to recognize values that
+  // were emitted by the same callback invokation, as identified by
+  // equal an Resource and TimeUnixNano.  OpenTelemetry callbacks are
```
bogdandrutu, line 200: s/equal an/equal on?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -86,94 +90,142 @@ message InstrumentationLibraryMetrics {
...
+  // presence is always meaningful.
+  //
+  // ASYNCHRONOUS means that:
+  // - The count of events is a measure of the rate of SDK collection
+  //   times the aggregated cardinality, therefore cannot be used
+  //   to extrapolate application-specific rate information.
+  // - Measurements with the same Resource and TimeUnixNano
+  //   form a meaningful ratio, and the set of measurements at this
+  //   instant are coherent, resulting from a single callback.
+  // - Trace context MUST not be set in exemplars.
```
bogdandrutu, line 213: This probably does not have exemplars at all, correct?

bogdandrutu:
@jmacd you had a go mod/sum in the original PR consider to send a quick PR to add them to the repo (independent of any proto changes)

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: Not exemplars, but associated RawValues makes sense.

bogdandrutu: What is the difference between Exemplar and RawValue?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -232,61 +284,246 @@ message MetricDescriptor {
...
+  // equally the RAW value types):
+  //
+  // 1. SCALAR value type interpretation depends on
+  //    both temporality and structure
+  // 2. For GROUPING kinds:
+  //    - For INSTANTANEOUS temporality: An individual measurement
+  //    - For DELTA or CUMULATIVE temporality: The last measured individual value
+  //      in the collection interval.
+  // 3. For ADDING kinds:
+  //    - For INSTANTANEOUS temporality: An instantaneously measured change in a sum
```
tsloughter, line 383: You lost me. I get the need for including if a sum is a delta or cumulative but `INSTANTANEOUS`?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -232,61 +284,246 @@ message MetricDescriptor {
...
+    // is expressed as a scalar value that would be added into a Sum.
+    ADDING = 0x8;
+
+    // GROUPING structure means the value has been computed by
+    // combining individual values through a meaningful aggregation.
+    // GROUPING structure implies the sum of measurements is not
+    // necessarily meaningful.  These may be expressed as DELTA
+    // kind, indicating an aggregation of values since the last
+    // collection, or as CUMULATIVE kind, indicating an aggregation
+    // of values since the last StartTimeUnixNano reset.  These may
```
tsloughter, line 312: Why have a concept of grouping that is not based on the last collection? Why this separate reset option? And why is this the case only for `GROUPING` but not for `DELTA`?

opentelemetry/proto/metrics/v1/metrics.proto
tsloughter: "system knows not to display a rate of the collection events" -- Do you mean taking the rate of a metric collected on an interval like Otel observers "shouldn't" be done? I'm not sure that makes sense so maybe I'm misreading this.

jmacd: The "rate of a metric" is OK.

It's the "rate of events for a metric" (i.e., rate of the Count field for Histogram and Summary points), which is something that _is_ meaningful for an ordinary GROUPING metrics. This is where an ASYNCHRONOUS or not distinction can be used.

tsloughter: Aaaah, got it.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: This comment tells me you're not lost. :grinning:

There is something here, and I considered making an argument that ADDING/INSTANTANEOUS is either equivalent to ADDING/DELTA or to ADDING/CUMULATIVE, so why include this possibility?

I reasoned that ADDING/INSTANTANEOUS means "adding without a start time", meaning the ADDING/DELTA interpretation is the only good interpretation. 

I would consider adding more example to help here. The interpretation is:
- ADDING/INSTANTANEOUS an individual DELTA event as might emit from a stateless counter. There is no StartTimeUnixNano set in this case (and no synchronization required)
- ADDING/DELTA is a sum of 1 or more instantaneous events with StartTimeUnixNano set.
- ADDING/CUMULATIVE is a sum of 1 or more instantaneous events with StartTimeUnixNano set.

opentelemetry/proto/metrics/v1/metrics.proto
james-bebbington: Apologies if these are dumb questions, but I need more clarification:

1. Is it possible to have a Histogram or Summary aggregation for an asynchronous instrument where you might want to calculate the rate of events? I'm trying to think of a case where this would be useful but can't think of any off hand.

2. In what case is it not true that we can compute meaningful ratios for a simple aggregation type? It seems like the same holds true for a simple example like the count of requests (e.g. `Method={GET} / Method={*}`)

Overall, the distinction between sync/async still feels a little tenuous to me as well. Maybe the term itself makes me a feel a little uncomfortable, as it doesn't really seem like consumers should care about _how_ a metric was generated.

jmacd: For question (1), it's possible to compute a histogram or summary aggregation of an asynchronous instrument. The raw events in this case are observations, and the API specifies that multiple observations for a single label set by the same callback are considered replacement values, so the count of raw events per label set is either 0 or 1 at the moment when an asynchronous instrument is evaluated. The number of raw events equals the cardinality of the set for a given evaluation of the callback. You could aggregate away the labels of these data points, combining them so that Count becomes greater than 1, still equal to the number of distinct label sets that were recorded. 

Now you could read all the data points for this instrument across a range of time, and the Count aggregation now computes something, but what is the meaning of this aggregation? This Count is a function of the collection period, because collecting more frequently yields more data points.  That is the property that ASYNCHRONOUS describes. The ASYNCHRONOUS bit does not change any interpretation or alter the mathematics of processing metric data, but it does indicate when a particular derived computation may be meaningless.

jmacd: For question (2), it's possible with a synchronous instrument to compute meaningful ratios (e.g. Method={GET} / Method={*}), but it requires a time interval to evaluate this query. The result depends on the time interval again, a small window will give you one result, and a large window gives another. The time variable has to be smoothed out of this equation, whereas with an ASYNCHRONOUS instrument the semantics of the instrument give us a better answer to this sort of query. I've sometimes referred to these "point-in-time" ratios.

jmacd: Also you're leading me to believe that ASYNCHRONOUS is not the best term here. I am thinking of alternatives, and am open to suggestion. The term originates from the synchronous usage: you call a synchronous instrument at a meaningful time, whereas the SDK calls asynchronous instruments at a meaningless time.

jmacd: Another term, possibly: 
NONCHRONOMETRIC

james-bebbington: Hmm okay, the point in time calculation def makes sense to me 👍

Re naming, wouldn't it just be CHRONOMETRIC? That name does sound more appealing than ASYNCHRONOUS imo.

Re #1, I'd still be interested to hear an actual example of a CHRONOMETRIC (asynchronously recorded) histogram.

jmacd: I'm not sure because CHRONOMETER refers to a stopwatch which is used for finely measuring time, and it feels like the synchronous instruments are more chronometrical than the asynchronous instruments. 

Another term I would support is NONCHRONOLOGICAL, since there is no meaningful order between events observed in the same callback. This seems to suggest more than the meaning given to ASYNCHRONOUS (NONCHRONOLOGICAL also suggests that the events of different callbacks are not order-able?).

james-bebbington: Hmm okay, do asynchronous callbacks always occur at a regular time interval or is that not necessarily the case? (that's what led me to believe they are "chronometric" although perhaps that doesn't mean what I thought it does)

jmacd: There is no specification about collection intervals being regular, and there is definitely interest in making it configurable. See https://github.com/open-telemetry/oteps/pull/121.

To summarize, I believe we've agreed there is something different about these metrics created by Observer instruments, but it's come to light that "Asynchronous" has been badly applied. The term CHRONOLOGICAL is appealing to me, now, and is one I'm willing to back-port into the API specification. That is to suggest we replace the existing SYNCHRONOUS / ASYNCHRONOUS distinction with a SYNCHRONOUS / CHRONOLOGICAL distinction for instruments.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd:
```
  // - Measurements are not traced, by definition, and SHOULD NOT have
  //   associated trace context (e.g., no exemplars with trace_id,
  //   span_id).
```

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: I replaced the use of ASYNCHRONOUS with CHRONOLOGICAL.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: I added
```
  // About ADDING|INSTANTANEOUS vs ADDING|DELTA:
  //
  // 1. An INSTANTANEOUS|ADDING event is stateless, has no StartTimeUnixNano.
  // 2. A DELTA|ADDING event remembers its start time, sets StartTimeUnixNano.
  //
```

jmacd:
@open-telemetry/specs-metrics-approvers please take another look. Thanks.

jmacd:
Note: the term "ASYNCHRONOUS" was found to be quite confusing and I have replaced it with "CHRONOLOGICAL" to describe the kind of measurement that comes from an Observer instrument. I would port this terminology back to the API specification as well.

jmacd:
@lubingfeng I would like a better timeline, however I can say that there are four PRs that I want merged before we declare a new release of the protocol, after which I do not foresee more changes. Sorry for the delays.

#168 
#170 
#159 
#171

opentelemetry/proto/metrics/v1/metrics.proto
james-bebbington: Another possible suggestion: SNAPSHOT vs CONTINUOUS

jmacd:
> SNAPSHOT vs CONTINUOUS

I like these terms. Either way, this discussion is about a name that we could easily change w/o breaking the protocol in the future. I'm happy to switch CHRONOLOGICAL to SNAPSHOT, it's appropriate because the intention of this bit is to say that the observations were simultaneous, which is what snapshot means.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -232,61 +289,278 @@ message MetricDescriptor {
...
+    // kinds.  These may also be expressed as INSTANTANEOUS kind when
+    // reporting a scalar value type (e.g., for LastValue
+    // aggregation).  TODO(#159) or raw value (i.e., no aggregation).
+    GROUPING = 0x10;
+
+    // MONOTONIC may be set in conjunction with ADDING kinds.  When
+    // set, MONOTONIC indicates that the calculated sum is
+    // non-decreasing, therefore can be monitored as a non-negative
+    // rate of change.
+    MONOTONIC = 0x20;
```
bogdandrutu, line 325: I read this, and I think we can better group all these 3 bits into an enum with 3 valid values:

```proto
enum Structure {
ADDING
MONOTONIC_ADDING
GROUPING
}
```

The main advantage is that we represent only valid combinations (no GROUPING_MONOTONIC, or at least not yet) and we use 1 less bit.

This also forces users when parsing the kind to switch on the 3 possible values for the "structure" property.

bogdandrutu: Also no possible way to have ADDING_GROUPING (both bits set).

In terms of representing, even if we go with a bit for every possible combination (the current ADDING and GROUPING), we can replace the MONOTONIC bit to ADDING_MONOTONIC and achieve the same results as the enum, the main point is to unify the structure + monotonic concept into 3 possible values.

bogdandrutu:
@jmacd as you see the summary changes are going fast, would you mind proposing the scalar (int and double unification) changes in a separate PR. I promised that we don't do release until this gets in so no worry about inconsistencies, and also scalar change is independent of most of other changes in this PR.

bogdandrutu:
> The decision to unify the two SCALAR value types into a single data point message type is justified as follows: the marginal memory cost of carrying both a int64 and double value in the same structure is minor, because that type already contains three other fields, one of them repeated. E.g., we have two 64 bit timestamps, one slice (3 x 64 bits), plus with the Google protoc-gen-go v1.23.0 we have an additional pointer, uint32, and slice known as MessageState, SizeCache, and UnknownFields. Taken together, the Go ScalarDataPoint is less than 10% larger than the former Int64DataPoint.

This is a justification that it does not impact that much memory but not a justification why it benefits

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: This sounds good to me. I will apply this change.

lubingfeng:
> @lubingfeng I would like a better timeline, however I can say that there are four PRs that I want merged before we declare a new release of the protocol, after which I do not foresee more changes. Sorry for the delays.
> 
> #168
> #170
> #159
> #171

Thanks Josh. I am wondering if we have a timeline (one week, two weeks, etc) to get those four PRs will be merged. We would like to use this protocol to develop OT receiver by end of August.

opentelemetry/proto/metrics/v1/metrics.proto

jmacd:
@open-telemetry/specs-metrics-approvers This has been udpated:

Use `SNAPSHOT` kind element for Observer instrument semantics
Make `ADDING_MONOTONIC` a Structure kind element
Restore Int64DataPoint/DoubleDataPoint.

jmacd:
> @lubingfeng 

I sure hope we have resolved the issues and can get these changes merged in 2 weeks. @open-telemetry/specs-metrics-approvers Let's Really Do This.

lubingfeng:
> @lubingfeng
> 
> I sure hope we have resolved the issues and can get these changes merged in 2 weeks. @open-telemetry/specs-metrics-approvers Let's Really Do This.


This is great. I plan to attend this Thursday's SIG meeting for metrics and would be good to hear latest progress / status on those for pull requests: 
#168
#170
#159
#171

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -104,76 +117,116 @@ message Metric {
...
-    // Summary value. Some frameworks implemented Histograms as a summary of observations
-    // (usually things like request durations and response sizes). While it
-    // also provides a total count of observations and a sum of all observed
-    // values, it calculates configurable percentiles over a sliding time
-    // window.
-    // Corresponding values are stored in SummaryDataPoint.
-    SUMMARY = 6;
+    // Summary implies that the value is found in
+    // Metric.summary_data_points[].summary.
+    SUMMARY = 4;
```
cnnradams, line 161: Thinking ahead to the exemplars changes; in #162 we had `SUMMARY_INT64` and `SUMMARY_DOUBLE` which lets us determine the data type of exemplars (int or double). But if we only specify `double` types for summary, does it make sense to have a `SUMMARY_INT64` and `SUMMARY_DOUBLE` type for only exemplars or should there be a different enum for raw measurement type? Same thing for histogram

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -232,11 +285,248 @@ message MetricDescriptor {
...
+  //    - For INSTANTANEOUS temporality: An instantaneously measured change in a sum
+  //    - For DELTA temporality: An aggregated change in a sum
+  //    - For CUMULATIVE temporality: An aggregated sum.
+  // 4. INSTANTANEOUS kinds are always scalar values (TODO(#159) or raw).
+  //
+  // About CUMULATIVE kinds:
+  //
+  // 1. CUMULATIVE GROUPING kinds typically use Histogram data points
+  // 2. CUMULATIVE is rarely used with Summary data points
+  enum Kind {
```
bogdandrutu, line 396: Spent time during the weekend trying this new proposal, and found a lot of things that don't make a lot of sense.

See more details here and a proposed different approach of thinking about the data model:
https://docs.google.com/document/d/1UBpke_e_HWl7UV3piNIMAZYw8eNjfGjucdwwsEOqUI4/edit#

bogdandrutu: Please don't hesitate to comment in the doc or here. Will try to answer all the comments :)

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -104,76 +117,116 @@ message Metric {
...
+  // This property can be safetly disregarded and not set when
+  // importing data from other sources, but when set its presence is
+  // always meaningful.
+  //
+  // When measurements originate from an OpenTelemetry API, the
+  // SNAPSHOT kind allows the interpreter to recognize values
+  // that were emitted by the same callback invocation, as identified
+  // by equal Resource and TimeUnixNano values.  SNAPSHOT metrics
+  // MUST NOT have more than one data point per TimeUnixNano,
+  // Resource, and LabelSet.  If the interpreter of this data finds
```
bogdandrutu, line 202: > SNAPSHOT metrics MUST NOT have more than one data point per TimeUnixNano, Resource, and LabelSet.

This is true for every metric correct?

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: I was on the fence about creating _INT64 and _DOUBLE variations of Histogram and Summary data points. I wouldn't object, but I fear this is more detail than most people want. I draw a distinction between single-value data points (SCALAR, RAW) and multi-value data points (HISTOGRAM, SUMMARY). I think it's more OK to support only floats for multi-value data points, basically, and really not OK to do that for scalars and raw values. 

I asked around whether this was a real problem at Google, and in cases it was. If you count your network traffic in bits per second, your router is likely to overflow a 2^50 counter but not a 2^64 counter.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -232,61 +289,278 @@ message MetricDescriptor {
...
+    // with asynchronous semantics specified by the OpenTelemetry API,
+    // indicating the value was collected through a callback.
+    // Generally this is the temporal difference between values of an
+    // OpenTelemetry SumObserver instrument.  Typically expressed as a
+    // scalar or histogram value.
+    ADDING_MONOTONIC_DELTA_CHRONOLOGICAL         = 0x6a; // ADDING|DELTA|MONOTONIC|CHRONOLOGICAL
+
+    // ADDING_MONOTONIC_CUMULATIVE describes the current value of a
+    // sum accumulated from the reset time (StartTimeUnixNano) through
+    // the measured time (TimeUnixNano).  The value must be
```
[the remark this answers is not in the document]
jmacd: I integrated #178.

jmacd:
See #182.

lubingfeng:
> > I sure hope we have resolved the issues and can get these changes merged in 2 weeks. @open-telemetry/specs-metrics-approvers Let's Really Do This.
> 
> This is great. I plan to attend this Thursday's SIG meeting for metrics and would be good to hear latest progress / status on those four pull requests:
> #168
> #170
> #159
> #171

I see the closure of #168, #170.  #159 is still in draft stage while #171 is merged. Are we close to get #159 and #171 closed this week or next week?

PR #170, opened 2020-07-18
Specify MinMaxLastSumCount aggregation be exposed as fields of SummaryDataPoint

jmacd:
This corresponds with OTEP https://github.com/open-telemetry/oteps/pull/117.


bogdandrutu:
I feel that it is a small conflict between this and #171. Do you want to end up with quantiles (0,1) plus min and max?

jmacd:
> Do you want to end up with quantiles (0,1) plus min and max?

Yes.

bogdandrutu:
@jmacd I talked to someone from Splunk (good at math and statistics) to help us clarify if quantiles 0 and 1 can be treated as min and max. Will wait Monday to have an official answer, but he pointed me to some examples where this is already happening:
1. https://numpy.org/doc/stable/reference/generated/numpy.quantile.html
2. https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/quantile

Will wait to chat more with him Monday and ask him to post an answer here.

Hope you don't feel bad that I asked for some external help.

jmacd:
@bogdandrutu That's great. I find the definition of quantile to be quite confusing, but it's the same as the definition for percentile. A question on this topic was raised in the SDK specification draft much earlier, and I did some research:
https://github.com/open-telemetry/opentelemetry-specification/pull/347/files#diff-20cdaa336161f6e1bd3a18da7174deceR262

jmacd:
More justification related to min/max value encoding: https://github.com/open-telemetry/opentelemetry-proto/pull/171#issuecomment-660853620

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -403,25 +403,44 @@ message SummaryDataPoint {
...
+  double min = 8;
+
+  // The maximum of recorded values since start_time.  If the count
+  // was omitted or zero because its true value was 1, this field
+  // shall be omitted by the writer and the value of `max` shall be
+  // set by the reader to value of `last`.  Otherwise, this field MUST
+  // be set to the maximum of the measurements reflected in count.
+  double max = 9;
+
+  // Represents the value at a given percentile of a distribution.  to
```
MrAlias, line 438:
```suggestion
  // Represents the value at a given percentile of a distribution.  To
```

jmacd: Done.

bogdandrutu:
Couple of suggestions:
* Last value needs to wait until that OTEP is merged.
* Explicit Min/Max vs quantiles 0/1. I think this is what you try to address in this PR. If that is the case we should remove the last value change :)

joe-sfx:
While it's true the extreme quantiles are the only ones that can be computed exactly by aggregating those quantiles calculated separately on the elements of a partition, in general aggregated quantiles do possess a meaning. More precisely, if an ordered set S is expressed as a disjoint union of (ordered) subsets S_1, ..., S_n,  and a particular quantile (e.g., P90) is estimated on each, producing values q(S_1), ..., q(S_n), then the range [min(q(S_1), ..., q(S_n)), max(q(S_1), ..., q(S_n))] is guaranteed to contain the true quantile (i.e., that of S itself). It just so happens that for q=0 and q=1 we can do a little better and ignore most of the range.

A sample quantile is based on one (in rank-based methods) or two (if one wishes to interpolate) order statistics. Hyndman and Fan ("Sample Quantiles in Statistical Packages") contains a theoretical discussion of various approaches. There is no ambiguity about which order statistic should participate for q=0 or q=1.

PR #171, opened 2020-07-18
Replace percentile with quantile

jmacd:
The use of percentiles to convey summary statistics is likely to lead to errors, since readers and writers must multiply or divide by 100.0 to compute a population ratio. Percentiles are good for display. Quantiles are good for math. This proposal just replaces percentile with quantile and replaces [0, 100] with [0, 1].


bogdandrutu:
@open-telemetry/specs-metrics-approvers please review. This is not very controversial

bogdandrutu:
@jmacd please rebase and make CI happy and I will merge

jmacd:
> However, I am surprised we have these values in the protocol in the first place. 

These values are needed in the protocol where the server scrapes a Prometheus Summary metric, for example, but this is another reason that Min and Max should be their own field: the extreme quantiles can be re-aggregated correctly.

bogdandrutu:
They can only be merged and not subtracted (from cumulative to delta) but it is still good. Also probably we should always send deltas and let the backend deal with this problem.

jmacd:
I've been having trouble getting the build to pass. I believe we can stop checking-in generated code and can remove the call top `verify-go.sh`. https://github.com/open-telemetry/opentelemetry-proto/pull/173

lubingfeng:
When can we get this one (#171) and #159 closed, which are blockers for getting stats proposal finalized.

PR #173, opened 2020-07-20
Stop checking-in generated Golang code

### LINKED ISSUE #79: Remove generated code from repository

jmacd:
I believe this will fix the build in all the open PRs.
I believe both the OTel-Go SDK and Collector have stopped using these generated files.

Resolves #79


jmacd:
The OTel-Go fix: https://github.com/open-telemetry/opentelemetry-go/pull/942

MrAlias:
Can we gate this on a release of the Go SIG library to ensure downstream has a versoned source for this code? I can get this release done this afternoon.

jmacd:
@MrAlias Is there going to be a problem? I would have assumed the Go dependency was pinned to an older release that had the generated code.

yurishkuro:
@MrAlias sorry, didn't notice your comment. Hope it's not a problem, can revert it is.

MrAlias:
> @MrAlias Is there going to be a problem? I would have assumed the Go dependency was pinned to an older release that had the generated code.

I think the worry was [this](https://github.com/open-telemetry/opentelemetry-proto/issues/79#issuecomment-659643521), but I don't think it is critical. I'll just make sure to get the release out today.

MrAlias:
> @MrAlias sorry, didn't notice your comment. Hope it's not a problem, can revert it is.

Should be okay, not worth a revert.

PR #174, opened 2020-07-20
Add error to span, deprecate status

codeboten:
This changes the mechanism to report an error on a Span from using status.code to using error. As per the discussion in this spec issue: https://github.com/open-telemetry/opentelemetry-specification/issues/706#issuecomment-659567198

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -223,7 +223,14 @@ message Span {
   // An optional final status for this span. Semantically when Status
   // wasn't set it is means span ended without errors and assume
   // Status.Ok (code = 0).
-  Status status = 15;
+  // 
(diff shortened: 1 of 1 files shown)


yurishkuro:
+1, but I think we also need a corresponding Spec change (which technically should precede the proto change).

codeboten:
Completely agree on waiting for the spec change, hoping this PR can help driving that conversation.

tigrannajaryan:
I see this was already mentioned but out of caution I want to reiterate one more time: we declared the Trace portion of the protocol stable. We will need to make sure we don't break this promise. Any implementation of v0.4.0 of the protocol will need to stay compatible (i.e. be able to interoperate) with the implementations of the future versions of the protocol.

codeboten:
Closing this PR as Status will stay.

PR #179, opened 2020-07-21
Change Metric type to be a more detailed structure with details about the aggregation and temporality.

bogdandrutu:
This is an alternative to PR #168. It takes a more descriptive approach where the Aggregation (the last applied aggregation if any) is clearly defined and Temporality is available only where it makes sense. This can help clearly identify what are the possible values and properties for different types of Aggregations.

There are some things that can be considered (TODOs left in the code):
* Histogram/Summary sum - monotonicity?
* Previous aggregation measurements: raw measurements or "derived" measurements (results of another aggregation).

IMPORTANT: This PR and #168 are not equivalent, this is inspired by that PR but makes some changes that are not compatible with that PR.


bogdandrutu:
Closing this PR to have people focused on #182. Will leave it around for reference.

PR #182, opened 2020-07-29
Add Gauge. Change Metric type to be a more detailed structure with details about the aggregation and temporality.

bogdandrutu:
This PR takes a more descriptive approach where the Aggregation (the last applied aggregation if any) is clearly defined and Temporality is available only where it makes sense. This can help clearly identify what are the possible values and properties for different types of Aggregations.

There are some things that can be considered (TODOs left in the code):

* https://github.com/open-telemetry/opentelemetry-specification/issues/725
* Histogram/Summary sum - monotonicity?
* Previous aggregation measurements: raw measurements or "derived" measurements (results of another aggregation).
* Support for RawMeasurements (recorded via the Sync Instruments)
* Decide if support for INSTANTANEOUS temporality is still needed.

IMPORTANT:
* This PR is not equivalent with #168 or #181, this is inspired by these PRs but makes some changes that are not compatible with these PRs.
* This PR is an incremental update, without any significant semantic changes (except adding the notion of GAUGE), from the current state,
  more changes will be added in the future to resolve the TODOs.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -52,9 +52,10 @@ message InstrumentationLibraryMetrics {
 // to refer to any one of the lists of points contained in the Metric.
 //
 // - Metric is composed of a MetricDescriptor and a list of data points.
-// - MetricDescriptor contains a name, description, unit, type, and temporarility.
+// - MetricDescriptor contains a name, description, unit, and type.
 // - Points is a list of DataPoints (shown vertically).
-// - DataPoint contains timestamps, labels, and one of the possible value type fields.
+// - DataPoint contains timestamps, labels, and one of the possible value type
+//   fields.
 //
 //     Metric
 //  +----------+         +------------------------+
@@ -63,8 +64,8 @@ message InstrumentationLibraryMetrics {
 //  |          |         | description            |
 //  |          |         | unit                   |
 //  |    points|--+      | type                   |
-//  +----------+  |      | temporarility          |
-//                |      +------------------------+
+//  +----------+  |      +------------------------+
+//                |
 //                |
 //                |      +---------------------------+
 //                |      |DataPoint 1                |
@@ -98,7 +99,7 @@ message InstrumentationLibraryMetrics {
 //   set for individual points.
 // - TimeUnixNano MUST be set to:
 //   - the end of the interval (CUMULATIVE or DELTA)
-//   - the instantaneous time of the event (INSTANTANEOUS).
+//   - the instantaneous time of the event.
 message Metric {
   // metric_descriptor describes the Metric.
   MetricDescriptor metric_descriptor = 1;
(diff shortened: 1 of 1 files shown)


bogdandrutu:
@james-bebbington 
> - Temp = Instantaneous only seems to be relevant for  "Last" (i.e. Summary?) or Raw values, whereas Temp = Delta/Cumulative seems to be only relevant for Gauge/Histogram. Should these concepts be separated?
> - I'm not across the exemplar stuff, but I would prefer to keep individual measurements (raw measurements?) as a separate category similar to your not-current PR. I think sharing the same data structure for exemplars and Instantaneous aggregations is confusing and unnecessary.

@bogdandrutu 
- Temp - there is a todo to be discussed in a followup PR, trying to limit changes in this
- Exemplar - Not yet finalized, but if we support RawMeasurements and Exemplars will be done as in the previous PR that I close to keep people focus, based on your feedback

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -124,48 +125,133 @@ message MetricDescriptor {
   // described by http://unitsofmeasure.org/ucum.html.
   string unit = 3;
 
-  // Type is the type of values a metric has.
-  enum Type {
-    // INVALID_TYPE is the default Type, it MUST not be used.
-    INVALID_TYPE = 0;
+  // MeasurementValueType determines the value type for a measurement.
```
jmacd, line 128: This may determine whether Int64DataPoint or DoubleDataPoint is used. 

TODO: There is an open question about whether this should control int64 vs. double for Histogram and Summary. There are good arguments on both sides of this.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -124,48 +125,133 @@ message MetricDescriptor {
...
+    // The scalar value type is an int64.
+    INT64 = 1;
+    // The scalar value type is a floating point number.
+    DOUBLE = 2;
+  }
 
-    // INT64 values are signed 64-bit integers.
+  // TODO: Decide if support for RawMeasurements (measurements recorded using
+  // the synchronous instruments) is necessary. It can be used to delegate the
+  // aggregation from the application to the agent/collector.
```
jmacd, line 141: FWIW https://github.com/open-telemetry/opentelemetry-specification/issues/617

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -124,48 +125,133 @@ message MetricDescriptor {
...
-    // Corresponding values are stored in SummaryDataPoint.
-    SUMMARY = 6;
+    // Determines the value type of the exemplars.
+    MeasurementValueType measurement_value_type = 1;
+
+    // Temporality is the temporal quality values of a metric have.
+    Temporality temporality = 2;
+
+    // TODO: Decide if knowing that the aggregated measurements are raw
+    // measurements is important or not. Equivalent of "Snapshot" in #162.
```
jmacd, line 224: I realize now the source of confusion for me w.r.t. Temporality is that it sometimes applies to the individual Metric (i.e., is this a change in a sum or a total sum?) and it sometimes applies to the collection strategy (i.e., am I returning all "active" Metrics since the beginning of the process or since the last report).

I think we should find a way to distinguish these two ideas with terminology. A practical way to probe this issue is that I might like to encode the following things:

1. A Histogram of deltas from the start of the process (dimensionally reduced or not)
2. A Histogram of cumulatives from the start of the process (probably dimensionally reduced otherwise just one value)
3. A Histogram of deltas from a recent interval (dimensionally reduced or not)
4. A Histogram of cumulatives at the end of a recent interval (probably dimensionally reduced otherwise just one value)

There's a question as to whether (2) and (4) are any different, probably they are not, but you can imagine different interpretations because in (2) you can't forget any timeseries, and in (4) you can omit timeseries that hvaen't been "active" in the latest window. 

So, two ways to interpret Cumulative.

OTOH, I think we're having trouble distinguishing (1) and (2).

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -124,48 +125,133 @@ message MetricDescriptor {
...
+
+    // TODO: Decide if knowing that the Sum is monotonic is important or not,
```
jmacd, line 226: :+1: this is a minor point, I agree.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -124,48 +125,133 @@ message MetricDescriptor {
...
+  //   ----------------------------------------------
+  //   Counter            Sum(temporality=delta;is_monotonic=true)
+  //   UpDownCounter      Sum(temporality=delta;is_monotonic=false)
+  //   ValueRecorder      Summary(temporality=delta)
+  //   SumObserver        Sum(temporality=cumulative;is_monotonic=true)
+  //   UpDownSumObserver  Sum(temporality=cumulative;is_monotonic=false)
+  //   ValueObserver      Gauge()
+  oneof type {
+    // TODO: Determine if encoding all possible values in a uint64 bitset
+    // improves performance significantly and propose that if that is the case.
```
jmacd, line 249: Good. We have pretty strong evidence that this is problematic, but it will vary by workload.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -175,6 +261,7 @@ message MetricDescriptor {
     // used.
     INVALID_TEMPORALITY = 0;
 
+    // TODO: Re-evaluate if this is still needed.
```
jmacd, line 264: +1 it looks like this can be dropped.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Added the TODO. We should discuss, also once this PR is merged will make sure we have an issue for every TODO in the code to keep track. Not creating issues now since this is not approved.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Done, thanks.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Created https://github.com/open-telemetry/opentelemetry-proto/issues/184 to track this. I want us to make progress with the overall model.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Yes, but wanted to point it so clarifies that we can add if we want.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Indeed, left a TODO to evaluate after we finish all other issues and see how much we can benefit.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: I want to do this in a separate PR because I want to explain better why, and only that explanation is more than this entire PR most likely.

bogdandrutu: Also I don't think it affects the overall structure, so easy to change in a separate PR.

bogdandrutu:
@open-telemetry/specs-metrics-approvers please review this 👍

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -124,48 +125,133 @@ message MetricDescriptor {
...
-    DOUBLE = 3;
+    // Determines the value type of the exemplars.
+    MeasurementValueType measurement_value_type = 1;
+
+    // Temporality is the temporal quality values of a metric have.
+    Temporality temporality = 2;
 
-    // MONOTONIC_DOUBLE values are monotonically increasing double-precision
-    // floating-point numbers.
+    // TODO: Decide if knowing that the aggregated measurements are raw
```
james-bebbington, line 206: > Decide if knowing that the aggregated measurements are raw measurements

Should this be "decide if knowing that the metric values are aggregated from raw measurements is important" (similar for below)

james-bebbington:
(please remember to tag with "Metrics" label)

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Done.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -124,48 +125,137 @@ message MetricDescriptor {
...
 
-  // type is the type of values this metric has.
-  Type type = 4;
+  // Type is the type of the metric and determines what are the type of the
+  // reported value, as well as the relatationship to the time interval over
+  // which they are reported.
+  //
+  // TODO: Update table after the decision on:
+  // https://github.com/open-telemetry/opentelemetry-specification/issues/731.
+  // Be default metrics recording using the OpenTelemetry API are exported as
```
aabmass, line 240: Typo?

```suggestion
  // By default, metrics recording using the OpenTelemetry API are exported as
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -124,48 +125,137 @@ message MetricDescriptor {
...
-    // A Metric of this Type MUST store its values as Int64DataPoint.
-    INT64 = 1;
+    // Determines if the points are Int64DataPoint or DoubleDataPoint, as well
+    // as the value type of the exemplars.
+    MeasurementValueType measurement_value_type = 1;
+  }
 
-    // MONOTONIC_INT64 values are monotonically increasing signed 64-bit
-    // integers.
+  // Represents the type of a numeric scalar metric that is calculated as a sum
```
MrAlias, line 166: Similar here?

```suggestion
  // Sum represents the type of a numeric scalar metric that is calculated as a sum
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -124,48 +125,137 @@ message MetricDescriptor {
...
+    // TODO: Decide if knowing that the metric values are aggregated from raw
+    // measurements is important.
+
+    // TODO: Decide if knowing that the Sum is monotonic is important or not,
+    // may be derived from the buckets definition.
   }
 
-  // type is the type of values this metric has.
-  Type type = 4;
+  // Type is the type of the metric and determines what are the type of the
```
MrAlias, line 234: Can we make the three different uses of the word "type" here more explicit about what "type" they describe?

The last use is the one I'm confused on its meaning currently.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -124,48 +125,137 @@ message MetricDescriptor {
...
+  // (the table does not include MeasurementValueType to avoid extra rows):
+  //
+  //   Instrument         Type
+  //   ----------------------------------------------
+  //   Counter            Sum(temporality=delta;is_monotonic=true)
+  //   UpDownCounter      Sum(temporality=delta;is_monotonic=false)
+  //   ValueRecorder      Summary(temporality=delta)
+  //   SumObserver        Sum(temporality=cumulative;is_monotonic=true)
+  //   UpDownSumObserver  Sum(temporality=cumulative;is_monotonic=false)
+  //   ValueObserver      Gauge()
```
MrAlias, line 250: This is great :+1:

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -124,48 +125,137 @@ message MetricDescriptor {
...
+    DOUBLE = 2;
+  }
 
-    // INT64 values are signed 64-bit integers.
+  // TODO: Decide if support for RawMeasurements (measurements recorded using
+  // the synchronous instruments) is necessary. It can be used to delegate the
+  // aggregation from the application to the agent/collector. See
+  // https://github.com/open-telemetry/opentelemetry-specification/issues/617
+
+  // Represents the type of a scalar metric that always exports the "current
```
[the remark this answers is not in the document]
bogdandrutu: Very minor to me. If we go this path (go like) we should just make sure we enforce it. I used Google's C++ style for comments (probably the 8y spent there changed me).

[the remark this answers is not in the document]
bogdandrutu: It looks like the only example that I found on the official website uses the syntax that you recommended:
https://developers.google.com/protocol-buffers/docs/proto3#adding_comments

I am fine to go with this, was just trying to find some style guide but could not find anything about comments.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Tried a new comment, please check.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Yes :)

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
MrAlias: Sounds good to me. I agree very minor.

opentelemetry/proto/metrics/v1/metrics.proto
MrAlias: Looks great :+1:, thanks.

bogdandrutu:
This PR has 6 approves, and 3 owners approves, + myself so I think it is good to be merged.

PR #185, opened 2020-07-29
Declare common and resource directories stable

tigrannajaryan:
We declared traces part of the protocol stable. Traces use common
and resource messages so we have to declare these as stable as well
otherwise the declaration of trace part as stable does not achieve
the goal (the stable interoperability).

Obviously any new messages can be freely added to common and resource,
"stable" label does not prevent that.


bogdandrutu:
@open-telemetry/technical-committee @open-telemetry/specs-approvers please review :)

PR #186, opened 2020-07-30
Use the new otel/build-protobuf docker

bogdandrutu:


tigrannajaryan:
Just to clarify: do we still need generated files in this repo? The Collector no longer uses these generated files and if I remember correctly Go SDK was also intending to generate its own.

bogdandrutu:
@tigrannajaryan we don't generate, but what we run the generator to ensure code compiles in the CI. Also projects can use the same make rules to generate code when using git submodules.

tigrannajaryan:
> @tigrannajaryan we don't generate, but what we run the generator to ensure code compiles in the CI. Also projects can use the same make rules to generate code when using git submodules.

In that case should we delete the generated files from the repo or you want to do it in a separate PR?

bogdandrutu:
@tigrannajaryan no code generated in this repo (submitted).

tigrannajaryan:
> @tigrannajaryan no code generated in this repo (submitted).

My bad, I did not realize it is already deleted.

bogdandrutu:
@open-telemetry/technical-committee @open-telemetry/specs-approvers please review :)

PR #111, opened 2020-02-27
Fix Go package name to 'trace'

nilebox:
Fixes #109, assuming it's a typo.

NOTE: this is a breaking change in the Go library, so we may consider if this change is safe to do. It seems worth doing for consistency at the early stage.

PR #190, opened 2020-07-31
Change trace -> traces for non collector

bogdandrutu:
This is a backwards compatible change from the wire format perspective. Changing the collector gRPC service name and package + the grpc gateway path will be breaking changes for the protocol.

Would like to make a decision on this.

### DIFF
--- a/opentelemetry/proto/collector/trace/v1/trace_service.proto
+++ b/opentelemetry/proto/collector/trace/v1/trace_service.proto
@@ -16,7 +16,7 @@ syntax = "proto3";
 
 package opentelemetry.proto.collector.trace.v1;
 
-import "opentelemetry/proto/trace/v1/trace.proto";
+import "opentelemetry/proto/traces/v1/traces.proto";
 
 option java_multiple_files = true;
 option java_package = "io.opentelemetry.proto.collector.trace.v1";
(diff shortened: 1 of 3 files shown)


opentelemetry/proto/collector/trace/v1/trace_service.proto
```diff
@@ -16,7 +16,7 @@ syntax = "proto3";
 
 package opentelemetry.proto.collector.trace.v1;
 
-import "opentelemetry/proto/trace/v1/trace.proto";
+import "opentelemetry/proto/traces/v1/traces.proto";
 
 option java_multiple_files = true;
 option java_package = "io.opentelemetry.proto.collector.trace.v1";
```
Oberon00, line 22: Should we not also change this? See #111. If this would become inconsistent, I'm against this PR.

bogdandrutu: I can change that namespace in Java. I cannot change Lin 17 because that will be backwards incompatible

Oberon00: Oh right. With that the folder structure does not match the package name anymore. Then I really think we should not rename the folder.

bogdandrutu:
Please don't merge this until we decide what to do with the grpc package

tigrannajaryan:
@bogdandrutu based on Spec discussion close this?

tigrannajaryan:
In spec discussion we decided that since there is no good way to make the renaming everywhere we will not do the renaming at all.

PR #191, opened 2020-08-04
Cleanup comments for time members in the points

bogdandrutu:
No real data model changes in this PR, just updates to the start_time_unix_nanos and time_unix_nano fields, to reflect that they are related to the aggregation value.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -340,19 +340,24 @@ message Int64DataPoint {
   // The set of labels that uniquely identify this timeseries.
   repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1;
 
-  // start_time_unix_nano is the time when the cumulative value was reset to zero.
-  // This is used for Counter type only. For Gauge the value is not specified and
-  // defaults to 0.
+  // start_time_unix_nano is the last time when the aggregation value was reset
+  // to "zero". For some metric types this is ignored, see MetricsDescriptor
+  // types for more details.
   //
-  // The cumulative value is over the time interval (start_time_unix_nano, time_unix_nano].
-  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
+  // The aggregation value is over the time interval (start_time_unix_nano,
+  // time_unix_nano].
+  // 
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
+  // 1970.
   //
-  // Value of 0 indicates that the timestamp is unspecified. In that case the timestamp
-  // may be decided by the backend.
+  // Value of 0 indicates that the timestamp is unspecified. In that case the
+  // timestamp may be decided by the backend.
   fixed64 start_time_unix_nano = 2;
 
-  // time_unix_nano is the moment when this value was recorded.
-  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
+  // time_unix_nano is the moment when this aggregation value was reported.
+  // 
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
+  // 1970.
   fixed64 time_unix_nano = 3;
 
   // value itself.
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -340,19 +340,24 @@ message Int64DataPoint {
...
   //
-  // Value of 0 indicates that the timestamp is unspecified. In that case the timestamp
-  // may be decided by the backend.
+  // Value of 0 indicates that the timestamp is unspecified. In that case the
+  // timestamp may be decided by the backend.
   fixed64 start_time_unix_nano = 2;
 
-  // time_unix_nano is the moment when this value was recorded.
-  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
+  // time_unix_nano is the moment when this aggregation value was reported.
```
james-bebbington, line 357: Why change this to "reported"? To me that sounds like it's the time you send the request. Isn't "recorded" more correct?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -340,19 +340,24 @@ message Int64DataPoint {
   // The set of labels that uniquely identify this timeseries.
   repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1;
 
-  // start_time_unix_nano is the time when the cumulative value was reset to zero.
-  // This is used for Counter type only. For Gauge the value is not specified and
-  // defaults to 0.
+  // start_time_unix_nano is the last time when the value was reset to "zero".
```
aabmass, line 343:
```suggestion
  // start_time_unix_nano is the last time when the aggregation value was reset to "zero".
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -365,19 +370,24 @@ message DoubleDataPoint {
...
```
aabmass, line 373:
```suggestion
  // start_time_unix_nano is the last time when the aggregation value was reset to "zero".
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -391,18 +401,24 @@ message HistogramDataPoint {
   // The set of labels that uniquely identify this timeseries.
   repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1;
 
-  // start_time_unix_nano is the time when the cumulative value was reset to zero.
+  // start_time_unix_nano is the last time when the value was reset to "zero".
```
aabmass, line 404:
```suggestion
  // start_time_unix_nano is the last time when the aggregation value was reset to "zero".
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -485,17 +501,24 @@ message SummaryDataPoint {
...
```
aabmass, line 504:
```suggestion
  // start_time_unix_nano is the last time when the aggregation value was reset to "zero".
```

opentelemetry/proto/metrics/v1/metrics.proto
aabmass: I think it's to be consistent with `ReportStrategy` in #192. I agree though, I thought the same thing

opentelemetry/proto/metrics/v1/metrics.proto
aabmass: I think it's a little weird to say `value was reset to "zero"` for a Histogram/Summary but I know that's why you quoted it.

I would also be in favor of changing all of the comments to `start_time_unix_nano is the last time when the aggregation was reset`.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: My problem with "recorded" was that it hints (maybe hidden hint) that this is the time when the measurement was recorded using the SDK instruments, which is not the case

opentelemetry/proto/metrics/v1/metrics.proto

opentelemetry/proto/metrics/v1/metrics.proto

opentelemetry/proto/metrics/v1/metrics.proto

PR #193, opened 2020-08-04
Remove Exemplars from Histogram, will be added later for all points

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-proto/issues/81


bogdandrutu:
Not planning to declare proto stable now. But definitely we understood that this is not well defined. So prefer to remove it now to remove one blocker from stable proto

Issue #143, opened 2020-04-27
Add comment about CUMULATIVE edge case


MrAlias:
One edge case that probably we should document is if one implementation decides to use `CUMULATIVE` and resets the `start_time` every export interval making this type essentially a `DELTA`. What should we recommend/do when in this case?

_Originally posted by @bogdandrutu in https://github.com/open-telemetry/opentelemetry-proto/pull/140_

jmacd:
It's not clear why we should be concerned about this. This looks like a valid transformation, although it is sure to cause performance pains in some systems, for two reasons:

1. The client will have to remember the last value to correctly transform the cumulative into the delta representation.
2. A server that expects cumulative inputs will see each delta as a new timeseries, unless it has code to convert deltas back to cumulatives by comparing start_times and appending when possible.

Nevertheless, it's possible to reconstruct the correct state of the world from the data in this case. WAI?

bogdandrutu:
My idea was just to add a statement that clarifies that this is valid but not recommended.

PR #195, opened 2020-08-06
Add a note that it is not recommended to use CUMULATIVE when DELTA can be used

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-proto/issues/143

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -331,7 +331,7 @@ message MetricDescriptor {
     //      number of requests received over the interval of time t_1 to
     //      t_0+1 with a value of 1.
     //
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -330,6 +330,11 @@ message MetricDescriptor {
     //   12. The 1 second collection cycle ends. A metric is exported for the
     //      number of requests received over the interval of time t_1 to
     //      t_0+1 with a value of 1.
+    //
+    // Note: Even though when reporting changes since last report time, using
```
jkwatson, line 334:
```suggestion
    // Note: Even though, when reporting changes since last report time, using 
```

Issue #73, opened 2019-11-25
Support optional sample-weight factors for all metrics Value count fields


jmacd:
As discussed in #59, there is a question of how to report metrics that have been sampled. Although the metrics API currently does not specify a way to **input** sampled metrics, certain exporters may wish to perform sampling themselves. Here we propose a solution for representing sampled metrics in the protocol without touching on whether the API should support sampled inputs.

Each of the Int64Value, DoubleValue, SummaryValue, HistogramValue, and HistogramValue.Bucket types would support a new optional `double` value named `weight`, which, if omitted, has a default equal to 1.0, meaning that no sampling took place.

For Int64Value and DoubleValue types, there is an implicit count of 1 already, as these represent one measured value.  The other types have explicit `count` fields. In each case, the value of `weight` is logically multiplied by the count, yielding an adjusted count due to sampling.

Consumers SHOULD use the value of `count` times `weight` as the total number of metric events being reported.  Consumers MUST interpret `count` as the actual number of events collected, whereas `weight` MUST be considered a count multiplier due to sampling in the metrics export pipeline.

PR #202, opened 2020-08-18
Merge descriptor type and data points into Data.

bogdandrutu:
This PR does not change any semantics, it just re-organize the protos:
* Removes MetricDescriptor and adds fields to the Metric directly
* Combines MetricDescriptor.Type with specific DataPoints.

This has the following advantages:
* Require only one allocation more than the initial version (v0.4.0). When using gogo proto.
* In-memory representation is smaller than the initial version (v0.4.0).
* Extensible: allows to add new data types in the Metric.
* Groups together metadata and points about different aggregation types.


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -151,87 +115,14 @@ message MetricDescriptor {
...
-    // aggregation.
-    //
-    // Determines the value type of the exemplars.
-    MeasurementValueType measurement_value_type = 1;
-
-    // aggregation_temporality describes if the aggregator reports delta changes
-    // since last report time, or cumulative changes since a fixed start time.
-    AggregationTemporality aggregation_temporality = 2;
-  }
+  // TODO: Should MeasurementValueType be here instead of in every Aggregation?
```
bogdandrutu, line 123: Before merging this PR, I need help to make a decision on this.

bogdandrutu:
@tigrannajaryan in the Attributes you decided to use oneof instead of the list of all possibilities, I did the same here because most likely we will have a list of more than 5 items as possibilities for the data type, and I remember that you said is ok.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -469,3 +440,30 @@ message HistogramDataPoint {
...
+  //
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
+  fixed64 time_unix_nano = 2;
+
+  // Numerical value of the measurement that was recorded. Only one of these
+  // two fields is used for the data, based on MeasurementValueType.
+  double double_value = 3;
+  int64 int64_value = 4;
+
+  // (Optional) Span ID of the current trace.
```
tigrannajaryan, line 462:
```suggestion
  // (Optional) Span ID of the exemplar trace.
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -469,3 +440,30 @@ message HistogramDataPoint {
...
+  // span_id may be missing if the measurement is not recorded inside a trace or if the trace is not sampled.
+  bytes span_id = 5;
+
+  // (Optional) Trace ID of the current trace.
```
tigrannajaryan, line 466:
```suggestion
  // (Optional) Trace ID of the exemplar trace.
```

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Done.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Done.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: I agree with Tigdan's point below on this question:
https://github.com/open-telemetry/opentelemetry-proto/pull/202/files#r473095541

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -248,84 +139,164 @@ message MetricDescriptor {
...
+  // Determines the value type of the exemplars.
+  MeasurementValueType measurement_value_type = 2;
+
+  // aggregation_temporality describes if the aggregator reports delta changes
+  // since last report time, or cumulative changes since a fixed start time.
+  AggregationTemporality aggregation_temporality = 3;
+}
+
+// MeasurementValueType determines the value type for a measurement.
+// TODO: There is an open question about whether this should control int64 vs
```
jmacd, line 219: To be clear this refers to the Sum fields. If inputs are doubles, the sum is probably a double. If inputs are integers and the numbers are large, they'll run into a loss of precision.

There's a separate question about whether Count fields can be doubles, which they would be if generating a histogram from sampled metrics (as in statsd's `@sample-rate` parameter, or if a sampled trace was converted into a metric event). 

I wish to not hold up the specification over either of these questions, so perhaps we could go with a single Double-value Histogram for now, and if ever we decide to add an Integer histogram add it and rename the existing one DoubleHistogram.

If ever we decide to embrace sampled metric events (which statsd will push us to do, I think), then having an integer count field is still OK, we can extend the struct with a double to indicate the degree of sampling that took place. Discussed here:
https://github.com/open-telemetry/opentelemetry-proto/issues/73

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -469,3 +440,34 @@ message HistogramDataPoint {
...
+  repeated opentelemetry.proto.common.v1.StringKeyValue dropped_labels = 1;
+
+  // time_unix_nano is the exact time when this exemplar was recorded
+  //
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
+  fixed64 time_unix_nano = 2;
+
+  // Numerical value of the measurement that was recorded. Only one of these
+  // two fields is used for the data, based on MeasurementValueType.
+  double double_value = 3;
```
jmacd, line 461: Similar to Tigran's question above, should we distinguish raw integer and raw double points?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -248,84 +139,164 @@ message MetricDescriptor {
   //   SumObserver        Sum(aggregation_temporality=cumulative;is_monotonic=true)
   //   UpDownSumObserver  Sum(aggregation_temporality=cumulative;is_monotonic=false)
   //   ValueObserver      Gauge()
-  oneof type {
-    // TODO: Determine if encoding all possible values in a uint64 bitset
-    // improves performance significantly and propose that if that is the case.
+  oneof data {
```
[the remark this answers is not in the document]
bogdandrutu: Some thoughts here:
1. MeasurementValueType cannot be completely removed because still needed in places like Histogram, unless we double that type as well. There may be other aggregations like that that will not have to expose different types based on the measurement value type (e.g. Summary that will most likely be re-added).
2. I do agree that having a GaugeInt/GaugeDouble is simple enough, for Sum I am a bit worried to duplicate all the fields.

In general case we will have >5 data points per metrics (I remember this from a previous backend that I used), so indeed we may argue that splitting into int/double data points may save some memory for the cost of maintenance of duplicate code.

Another option is to merge https://github.com/open-telemetry/opentelemetry-proto/pull/172 and simply don't care about the in-memory extra 8 bytes per point (no effect on the marshal/unmarshal or network size).

Any ideas?

My preference would be to simplify the model and merge #172 and keep the current MeasurementValueType as part of the Sum/Gauge/Histogram (not in the metric because it may not apply to some aggregation)

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: This was just moved from the top of the file. But I tend to agree with #172 proposal you had, and just ignore 8bytes in memory overhead in order to simplify and not duplicate all the fields.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: #205 created to remove this TODO.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Moved discussion there.

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
tigrannajaryan: >MeasurementValueType cannot be completely removed

Correct. What I wanted to say is to remove the `measurement_value_type` field from Gauge/GaugeInt, not remove the MeasurementValueType enum definition.

>I do agree that having a GaugeInt/GaugeDouble is simple enough, for Sum I am a bit worried to duplicate all the fields.

Sum is currently 5 fields and I suggest to remove 2. Only 3 fields will remain, 2 of which will be the same (duplicate) for SumInt64 and SumDouble. Not big deal unless we think more common fields will be added in the future.

>Another option is to merge #172 and simply don't care about the in-memory extra 8 bytes per point (no effect on the marshal/unmarshal or network size).

That works too. Did you do any benchmarking to see if it can hint to a more performant approach? Other than performance I don't have a strong preference for one of the discussed options, we can use either, both have their own benefits and drawbacks.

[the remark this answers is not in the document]
bogdandrutu: I promise a follow-up immediately when this is merged to resolve this with benchmark numbers. For the moment the main change proposed here is to unify metadata about aggregation with the data so I want to move forward with that.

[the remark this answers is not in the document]
bogdandrutu: Filed #206 to track this.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -44,103 +44,67 @@ message InstrumentationLibraryMetrics {
...
-//                       |+-----+                    |
-//                       +---------------------------+
+//  +------------+
+//  |name        |
+//  |description |
+//  |unit        |     +---------------------------+
+//  |data        |---> |Gauge, Sum, Histogram, ... |
+//  +------------+     +---------------------------+ 
+//
+//    Data
```
jkwatson, line 68:
```suggestion
//    Data. [One of Gauge, Sum, Histogram, ...]
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -245,84 +137,161 @@ message MetricDescriptor {
...
+// Sum represents the type of a numeric scalar metric that is calculated as a
+// sum of all reported measurements over a time interval.
+//
+// A Metric of this Type MUST store its values as Int64DataPoint or
+// DoubleDataPoint.
+message Sum {
+  repeated Int64DataPoint int64_data_points = 1;
+  repeated DoubleDataPoint double_data_points = 2;
+
+  // It describes the value type of the measurement used to build this
```
anuraaga, line 180: I don't think it's for this PR, but I notice there are three patterns for comments

- Go-style, repeating name in javadoc e.g. Sum represents the type
- Starting with `It`, otherwise similar to go-style
- Java-style, documenting behavior directly e.g. Represents the type of a metric

Should this be made consistent? I have a personal preference to Java-style for protos since I've seen that more in the past, but it looks like we use Go-style more here so consolidating on that would make sense too.

bogdandrutu: Please file an issue so we don't forget about this

PR #208, opened 2020-08-21
Merge Int64DataPoint and DoubleDataPoint into ScalarDataPoint and support different sum types in Histogram

### LINKED ISSUE #206: Decide if for Gauge/Sum use measurement type or different types IntGauge/DoubleGauge
See https://github.com/open-telemetry/opentelemetry-proto/pull/202#discussion_r473095541

bogdandrutu:
This is the first option to fix #206  

Cannot see any significant difference between merged messages vs split messages, code is here https://github.com/bogdandrutu/metrics-proto

```
goos: darwin
goarch: amd64
pkg: github.com/bogdandrutu/metrics-proto/benchmark
BenchmarkEncodeDecode
BenchmarkEncodeDecode/merged
BenchmarkEncodeDecode/merged-16         	   34406	     33744 ns/op	   30288 B/op	     523 allocs/op
BenchmarkEncodeDecode/unmerged
BenchmarkEncodeDecode/unmerged-16       	   36130	     33190 ns/op	   29168 B/op	     523 allocs/op
```

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -152,36 +152,34 @@ message Metric {
 // AggregationTemporality is not included. Consequently, this also means
 // "StartTimeUnixNano" is ignored for all data points.
 message Gauge {
-  repeated Int64DataPoint int64_data_points = 1;
-  repeated DoubleDataPoint double_data_points = 2;
+  repeated ScalarDataPoint data_points = 1;
 
   // It describes the value type of the measurement used to build this
   // aggregation.
   //
   // Determines if the points are Int64DataPoint or DoubleDataPoint, as well
   // as the value type of the exemplars.
-  MeasurementValueType measurement_value_type = 3;
+  MeasurementValueType measurement_value_type = 2;
 }
 
 // Sum represents the type of a numeric scalar metric that is calculated as a
 // sum of all reported measurements over a time interval.
 message Sum {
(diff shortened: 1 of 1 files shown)


james-bebbington (approved):
I would value the readability & maintainability of this proto going forward (which might allow faster iteration & increase adoption) more highly than a <1% performance difference in this case, so prefer this over https://github.com/open-telemetry/opentelemetry-proto/pull/209.

jmacd:
@jkwatson put in a vote for this PR in Gitter. I'm on the fence, but I should mention that I proposed a similar PR to this one in the past: https://github.com/open-telemetry/opentelemetry-proto/pull/172

I would prefer that this is settled than to debate these options, though. Can we move forward with this version?

tigrannajaryan:
I agree performance difference is negligible and we need to choose based on usability or some other criteria. 
@bogdandrutu can you also assign the value to `measurement_value_type` where appropriate to make the code complete?

tigrannajaryan:
Does `Log10kDPS` measure CPU/RAM usage only of otelcol process? We do not see the resource usage by Fluentbit in the test, right?

bogdandrutu:
Thought about this more and I prefer #209 for the reasons I put in gitter: extensibility (can add new types that work only on some measurement types like strings if we decide to support that); no corner cases like measurement type is long and exemplars are using double values.

It seems there is no clear winner so I would want to move forward with that one

bogdandrutu:
I am going to close this and ask everyone to review #209.

PR #213, opened 2020-08-25
metrics proto: rename dropped labels to filtered labels

tsloughter:
In other protos "dropped" is used to refer to counts of
how many of an item were discarded because did not meet
the spec (such as an attribute key being too long) or
because there is a limit on the total number.

This field is not labels dropped for not conforming to
a spec but labels filtered out because they were not
included in the aggregations definition of labels to
include. So this commit renames to `filtered` instead
of `dropped` to avoid confusion.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -429,10 +429,10 @@ message HistogramDataPoint {
 // was recorded, for example the span and trace ID of the active span when the
 // exemplar was recorded.
 message Exemplar {
-  // The set of labels that were dropped by the aggregator, but recorded
-  // alongside the original measurement. Only labels that were dropped by the
-  // aggregator should be included
-  repeated opentelemetry.proto.common.v1.StringKeyValue dropped_labels = 1;
+  // The set of labels that were filtered out by the aggregator, but recorded
+  // alongside the original measurement. Only labels that were filtered out
+  // by the aggregator should be included
+  repeated opentelemetry.proto.common.v1.StringKeyValue filtered_labels = 1;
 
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -429,10 +429,10 @@ message HistogramDataPoint {
 // was recorded, for example the span and trace ID of the active span when the
 // exemplar was recorded.
 message Exemplar {
-  // The set of labels that were dropped by the aggregator, but recorded
-  // alongside the original measurement. Only labels that were dropped by the
-  // aggregator should be included
-  repeated opentelemetry.proto.common.v1.StringKeyValue dropped_labels = 1;
+  // The set of labels that were filtered out by the aggregator, but recorded
```
tigrannajaryan, line 432: Does this even have to be labels that are filtered by the aggregator? This assumes that there is an original set of key/value lables and the aggregator keeps some as metric labels, while the rest are recorded in the exemplar.

Is this the only use case for this field? Should we not generally allow any span _attributes_ that describe the exemplar span? Does it matter how exactly the labels end up in this field? I think what's more important is that they describe the span.

I could be wrong, but would like to understand.

bogdandrutu: They don't necessarily describe a span. User may record more labels in the metrics API call, then configure an aggregator to use only a subset to build the metric (reduce cardinality). In this case they are not at all related to span

tigrannajaryan: Thanks. I mistakenly assumed `Exemplar` means an exemplar span.

tsloughter:
I'm not authorized to merge, so if this is good someone can pls merge.

Issue #211, opened 2020-08-21
Check usage of int64 vs sfixed64 in metrics proto and uint64 vs fixed64


bogdandrutu:

tigrannajaryan:
Do you want to make sure we use them consistently or you want to check to see if there are any benefits and drawbacks?

They are all the same in-memory size, `int` versions may be more compact on the wire (for small values) and may be a tiny bit slower than `fixed` versions when marshaling/unmarshaling (likely negligible).

bogdandrutu:
/private/var/folders/5c/5p_3jmvd6qx0rsvmb9j7c_s00000gn/T/___BenchmarkEncodeDecode_in_github_com_bogdandrutu_metrics_proto_benchmark__3_ -test.v -test.bench ^\QBenchmarkEncodeDecode\E$ -test.run ^$
goos: darwin
goarch: amd64
pkg: github.com/bogdandrutu/metrics-proto/benchmark
BenchmarkEncodeDecode
BenchmarkEncodeDecode/merged
BenchmarkEncodeDecode/merged-16         	   35745	     33669 ns/op	   30288 B/op	     523 allocs/op
BenchmarkEncodeDecode/mergedfixed
BenchmarkEncodeDecode/mergedfixed-16    	   36036	     33310 ns/op	   30416 B/op	     523 allocs/op
BenchmarkEncodeDecode/unmerged
BenchmarkEncodeDecode/unmerged-16       	   36104	     32878 ns/op	   29168 B/op	     523 allocs/op
BenchmarkEncodeDecode/unmergedfixed
BenchmarkEncodeDecode/unmergedfixed-16  	   36522	     32474 ns/op	   29296 B/op	     523 allocs/op
PASS

Process finished with exit code 0

jmacd:
This was discussed and we did not find a lot of support.

Issue #11, opened 2019-04-17
Review Data Model (Protos)


tedsuo:
Everyone, please review the current protobuf and metadata definitions and leave comments here.

tmc:
It's a bit unfortunate that there isn't a great way to attach structured nested metadata as attributes. Perhaps consider supporting Any values as a means of extension?

bogdandrutu:
@tmc we discuss  to add some support of AttributeValue to be another Map<String, AttributeValue> is that enough?

We do not want to add Any because that means that a backend cannot analyze the data, maybe not even display them.

tmc:
I imagine that for many use cases allowing maps would suffice.

tete17:
Sorry to step into this conversation but is the idea of this project that all other language implementations to use this definition as a basis of is it only for the Java one?

I do not see it in use as a submodule for the node-js one

SergeyKanzhelev:
Seems to be complete. See https://github.com/open-telemetry/oteps/pull/59 where we extensively reviewed it.

Closing.

PR #214, opened 2020-08-26
Use fixed64 and sfixed64 for the sums and counts

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-proto/issues/211

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -315,7 +315,7 @@ message IntDataPoint {
   fixed64 time_unix_nano = 3;
 
   // value itself.
-  int64 value = 4;
+  sfixed64 value = 4;
 
   // (Optional) List of exemplars collected from
   // measurements that were used to form the data point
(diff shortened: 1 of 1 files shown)


jmacd:
I feel like #11 didn't provide justification for this. It's smaller on the wire to and slower to decode, and #211 says that therefore we should make the trade?

bogdandrutu:
@jmacd I think you wanted to reference #211 not #11. Happy to drop this, but in my opinion it is a small win that we get on the CPU by trading a smaller network traffic.

Not sure what to do, but feels like more natural to encode the numbers this way, this will speedup also functions like "size()".

anuraaga:
If I'm not mistaken, pretty large values like 34 billion can be stored in 5 bytes with varint, vs 8 bytes with fixed which is about 30% savings.

https://github.com/protocolbuffers/protobuf/blob/367f3d831ee3186035a14f871cbcbc1b028b6b9a/python/google/protobuf/internal/encoder.py#L103

It seems relatively rare to cross such a huge number, and very common to be even smaller. Won't the I/O time vastly outweigh the varint processing in practical cases?

> this will speedup also functions like "size()".

Correct me if I'm wrong, but I don't think there are any proto implementations that read directly from the serialized state. Since they copy into an arraylike, `size()` shouldn't be affected.

bogdandrutu:
@anuraaga not sure, for 100 data points (timeseries) the extra bytes is at most 700, which is less than the 1/2 MTU, so it is 50% chance no more messages, and 50% chance one extra message. Not sure how to prove this, and happy to drop it, because you seem to have a more strong opinion.

anuraaga (approved):
Ah, thanks for clarifying size is about the serialized size, yeah that does seem like it can get significant. And found that capn proto uses fixed size too so this seems to make sense :) We're in a different world in terms of network / CPU size than when proto was first developed I guess.

https://stackoverflow.com/a/24642169

PR #222, opened 2020-09-17
Modify status code enum according to OTEP 0136

tigrannajaryan:
See OTEP 0136 here:
https://github.com/open-telemetry/oteps/blob/master/text/trace/0136-error_flagging.md

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -221,8 +221,8 @@ message Span {
   uint32 dropped_links_count = 14;
 
   // An optional final status for this span. Semantically when Status
-  // wasn't set it is means span ended without errors and assume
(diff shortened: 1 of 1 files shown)


bogdandrutu (approved):
We should wait for the specs before merging, since the otep says some of the concrete things may be finalize in the specs.

Oberon00 (commented):
I'm against this breaking change.
The protocol is stable, so unless we want to make a special exception here, we should introduce a new field and/or introduce a mapping from the new to old status code.

tigrannajaryan:
> I'm against this breaking change.
> The protocol is stable, so unless we want to make a special exception here, we should introduce a new field and/or introduce a mapping from the new to old status code.

This is a good point, I rushed this. Closing for now, needs more thought.

PR #224, opened 2020-09-25
Modify Span status according to specification

tigrannajaryan:
1. Renamed Status.code field to Status.deprecated_code.
2. Renamed StatusCode enum to Status.DeprecatedStatusCode.
3. Introduce new Status.code with corresponding StatusCode enum.
4. Added guidelines about how to use the fields to ensure backward compatibility.

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -220,7 +220,7 @@ message Span {
   // enforced. If this value is 0, then no links were dropped.
   uint32 dropped_links_count = 14;
 
-  // An optional final status for this span. Semantically when Status isn't set it means
+  // An optional final status for this span. Semantically when Status isn't set, it means
   // span's status code is unset, i.e. assume STATUS_CANONICAL_CODE_UNSET (code = 0).
   Status status = 15;
 }
@@ -251,7 +251,7 @@ message Status {
 
   // The deprecated status code. This is an optional field.
   //
-  // This field is deprecated and is replaced by `code` field below. See backward
+  // This field is deprecated and is replaced by the `code` field below. See backward
   // compatibility notes below. According to our stability guarantees this field
   // will be removed in 12 months, on Sep 26, 2021. All usage of old senders and
   // receivers that do not understand the `code` field MUST be phased out by then.
@@ -277,27 +277,27 @@ message Status {
 
   // IMPORTANT: Backward compatibility notes:
   //
-  // To ensure any two pair of senders and receivers continues to correctly signal and
-  // interpret erroneous situation the senders and receivers SHOULD follow these rules:
+  // To ensure any pair of senders and receivers continues to correctly signal and
+  // interpret erroneous situations, the senders and receivers SHOULD follow these rules:
   //
-  // 1. Old senders and receives that are not aware of `code` field will continue using
-  // `deprecated_code` field to signal and interpret erroneous situation.
+  // 1. Old senders and receivers that are not aware of `code` field will continue using
+  // the `deprecated_code` field to signal and interpret erroneous situation.
   //
-  // 2. New senders, which are aware of `code` field should set both `deprecated_code`
-  // and `code` fields according to the following rules:
+  // 2. New senders, which are aware of the `code` field SHOULD set both the
+  // `deprecated_code` and `code` fields according to the following rules:
   //
-  //   if code==STATUS_CANONICAL_CODE_UNSET then `deprecated_code` must be
+  //   if code==STATUS_CANONICAL_CODE_UNSET then `deprecated_code` MUST be
   //   set to DEPRECATED_STATUS_CODE_OK.
   //
-  //   if code==STATUS_CANONICAL_CODE_ERROR then `deprecated_code` must be
+  //   if code==STATUS_CANONICAL_CODE_OK then `deprecated_code` MUST be
+  //   set to DEPRECATED_STATUS_CODE_OK.
+  //
+  //   if code==STATUS_CANONICAL_CODE_ERROR then `deprecated_code` MUST be
   //   set to DEPRECATED_STATUS_CODE_UNKNOWN_ERROR.
-  //
-  //   if code==STATUS_CANONICAL_CODE_OK then `deprecated_code` must be
-  //   set to DEPRECATED_STATUS_CODE_OK.
   //
   // These rules allow old receivers to correctly interpret data received from new senders.
   //
-  // 3. New receivers should look both at `code` and `deprecated_code` fields in order
+  // 3. New receivers should look at both the `code` and `deprecated_code` fields in order
   // to interpret the overall status:
   //
   //   If code==STATUS_CANONICAL_CODE_UNSET then the value of `deprecated_code` is the
@@ -310,7 +310,7 @@ message Status {
   //     the overall status to be STATUS_CANONICAL_CODE_ERROR.
   //
   //   If code!=STATUS_CANONICAL_CODE_UNSET then the value of `deprecated_code` should be
-  //   ignored, the `code` field is the sole carrier of status.
+  //   ignored, the `code` field is the sole carrier of the status.
   //
   // These rules allow new receivers to correctly interpret data received from old senders.
 }
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -221,7 +221,7 @@ message Span {
   uint32 dropped_links_count = 14;
 
   // An optional final status for this span. Semantically when Status isn't set, it means
-  // span's status code is unset, i.e. assume STATUS_CANONICAL_CODE_UNSET (code = 0).
+  // span's status code is unset, i.e. assume STATUS_CANONICAL_CODE_UNSET (canonical_code = 0).
   Status status = 15;
 }
 
@@ -229,33 +229,34 @@ message Span {
 // programming environments, including REST APIs and RPC APIs.
 message Status {
 
-  enum DeprecatedStatusCode {
-    DEPRECATED_STATUS_CODE_OK                  = 0;
-    DEPRECATED_STATUS_CODE_CANCELLED           = 1;
-    DEPRECATED_STATUS_CODE_UNKNOWN_ERROR       = 2;
-    DEPRECATED_STATUS_CODE_INVALID_ARGUMENT    = 3;
-    DEPRECATED_STATUS_CODE_DEADLINE_EXCEEDED   = 4;
-    DEPRECATED_STATUS_CODE_NOT_FOUND           = 5;
-    DEPRECATED_STATUS_CODE_ALREADY_EXISTS      = 6;
-    DEPRECATED_STATUS_CODE_PERMISSION_DENIED   = 7;
-    DEPRECATED_STATUS_CODE_RESOURCE_EXHAUSTED  = 8;
-    DEPRECATED_STATUS_CODE_FAILED_PRECONDITION = 9;
-    DEPRECATED_STATUS_CODE_ABORTED             = 10;
-    DEPRECATED_STATUS_CODE_OUT_OF_RANGE        = 11;
-    DEPRECATED_STATUS_CODE_UNIMPLEMENTED       = 12;
-    DEPRECATED_STATUS_CODE_INTERNAL_ERROR      = 13;
-    DEPRECATED_STATUS_CODE_UNAVAILABLE         = 14;
-    DEPRECATED_STATUS_CODE_DATA_LOSS           = 15;
-    DEPRECATED_STATUS_CODE_UNAUTHENTICATED     = 16;
+  enum StatusCode {
+    STATUS_CODE_OK                  = 0;
+    STATUS_CODE_CANCELLED           = 1;
+    STATUS_CODE_UNKNOWN_ERROR       = 2;
+    STATUS_CODE_INVALID_ARGUMENT    = 3;
+    STATUS_CODE_DEADLINE_EXCEEDED   = 4;
+    STATUS_CODE_NOT_FOUND           = 5;
+    STATUS_CODE_ALREADY_EXISTS      = 6;
+    STATUS_CODE_PERMISSION_DENIED   = 7;
+    STATUS_CODE_RESOURCE_EXHAUSTED  = 8;
+    STATUS_CODE_FAILED_PRECONDITION = 9;
+    STATUS_CODE_ABORTED             = 10;
+    STATUS_CODE_OUT_OF_RANGE        = 11;
+    STATUS_CODE_UNIMPLEMENTED       = 12;
+    STATUS_CODE_INTERNAL_ERROR      = 13;
+    STATUS_CODE_UNAVAILABLE         = 14;
+    STATUS_CODE_DATA_LOSS           = 15;
+    STATUS_CODE_UNAUTHENTICATED     = 16;
   };
 
   // The deprecated status code. This is an optional field.
   //
-  // This field is deprecated and is replaced by the `code` field below. See backward
-  // compatibility notes below. According to our stability guarantees this field
+  // This field is deprecated and is replaced by the `canonical_code` field below. See
+  // backward compatibility notes below. According to our stability guarantees this field
   // will be removed in 12 months, on Sep 26, 2021. All usage of old senders and
-  // receivers that do not understand the `code` field MUST be phased out by then.
-  DeprecatedStatusCode deprecated_code = 1 [deprecated=true];
+  // receivers that do not understand the `canonical_code` field MUST be phased out
+  // by then.
+  StatusCode code = 1 [deprecated=true];
 
   // A developer-facing human readable error message.
   string message = 2;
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,93 @@ message Span {
...
+  // receivers that do not understand the `code` field MUST be phased out by then.
+  DeprecatedStatusCode deprecated_code = 1 [deprecated=true];
 
   // A developer-facing human readable error message.
   string message = 2;
+
+  // For the semantics of status codes see
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#statuscanonicalcode
+  enum StatusCanonicalCode {
+    STATUS_CANONICAL_CODE_UNSET               = 0;
```
bogdandrutu, line 266: Comments for the values?

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,93 @@ message Span {
...
+    STATUS_CANONICAL_CODE_ERROR               = 1;
+    STATUS_CANONICAL_CODE_OK                  = 2;
```
bogdandrutu, line 268: Can this be make 1? I am confident that if we add more values they will be errors not ok so they will be grouped

bogdandrutu (approved):

jkwatson (approved):
Looks great!

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: Yeah, I was considering that too. Changing.

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: Added.

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: Done.

iNikem (approved):

bryannaegele (commented):
How does this avoid being a breaking change of a stable API when the name of the field and its values are being changed?

Oberon00:
Also, since JSON is allowed, renaming is not even wire-compatible.

bogdandrutu:
JSON is not stable

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,97 @@ message Span {
...
   };
 
-  // The status code. This is optional field. It is safe to assume 0 (OK)
-  // when not set.
-  StatusCode code = 1;
+  // The deprecated status code. This is an optional field.
+  //
+  // This field is deprecated and is replaced by `code` field below. See backward
+  // compatibility notes below. According to our stability guarantees this field
+  // will be removed in 12 months, on Sep 26, 2021. All usage of old senders and
```
anuraaga, line 256: Not related to this PR but just noticed since it's defined very clearly here with a date - 12 months seems like a pretty short time for API stability post-GA.

anuraaga (approved):

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,97 @@ message Span {
...
+    // The Span contains an error.
+    STATUS_CANONICAL_CODE_ERROR               = 2;
+  };
+
+  // The status code.
+  StatusCanonicalCode code = 3;
+
+  // IMPORTANT: Backward compatibility notes:
+  //
+  // To ensure any two pair of senders and receivers continues to correctly signal and
```
arminru, line 280:
```suggestion
  // To ensure any pair of senders and receivers continues to correctly signal and
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,97 @@ message Span {
...
+  // interpret erroneous situation the senders and receivers SHOULD follow these rules:
```
arminru, line 281:
```suggestion
  // interpret erroneous situations, the senders and receivers SHOULD follow these rules:
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,97 @@ message Span {
...
+  //
+  // 1. Old senders and receives that are not aware of `code` field will continue using
```
arminru, line 283:
```suggestion
  // 1. Old senders and receivers that are not aware of the `code` field will continue using
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,97 @@ message Span {
...
+  // `deprecated_code` field to signal and interpret erroneous situation.
```
arminru, line 284:
```suggestion
  // the `deprecated_code` field to signal and interpret erroneous situation.
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,97 @@ message Span {
...
+  //
+  // 2. New senders, which are aware of `code` field should set both `deprecated_code`
```
arminru, line 286:
```suggestion
  // 2. New senders, which are aware of the `code` field SHOULD set both the `deprecated_code`
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,97 @@ message Span {
...
+  // and `code` fields according to the following rules:
+  //
+  //   if code==STATUS_CANONICAL_CODE_UNSET then `deprecated_code` must be
```
arminru, line 289:
```suggestion
  //   if code==STATUS_CANONICAL_CODE_UNSET then `deprecated_code` MUST be
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,97 @@ message Span {
...
+  //   set to DEPRECATED_STATUS_CODE_OK.
+  //
+  //   if code==STATUS_CANONICAL_CODE_ERROR then `deprecated_code` must be
```
arminru, line 292:
```suggestion
  //   if code==STATUS_CANONICAL_CODE_ERROR then `deprecated_code` MUST be
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,97 @@ message Span {
...
+  //   set to DEPRECATED_STATUS_CODE_UNKNOWN_ERROR.
+  //
+  //   if code==STATUS_CANONICAL_CODE_OK then `deprecated_code` must be
```
arminru, line 295:
```suggestion
  //   if code==STATUS_CANONICAL_CODE_OK then `deprecated_code` MUST be
```

You could move that before the error rule to match the sorting order of the definition above.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,97 @@ message Span {
...
+  //   set to DEPRECATED_STATUS_CODE_OK.
+  //
+  // These rules allow old receivers to correctly interpret data received from new senders.
+  //
+  // 3. New receivers should look both at `code` and `deprecated_code` fields in order
```
arminru, line 300:
```suggestion
  // 3. New receivers should look at both the `code` and `deprecated_code` fields in order
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,97 @@ message Span {
...
+  //   carrier of the overall status according to these rules:
+  //
+  //     if deprecated_code==DEPRECATED_STATUS_CODE_OK then the receiver should interpret
+  //     the overall status to be STATUS_CANONICAL_CODE_UNSET.
+  //
+  //     if deprecated_code!=DEPRECATED_STATUS_CODE_OK then the receiver should interpret
+  //     the overall status to be STATUS_CANONICAL_CODE_ERROR.
+  //
+  //   If code!=STATUS_CANONICAL_CODE_UNSET then the value of `deprecated_code` should be
+  //   ignored, the `code` field is the sole carrier of status.
```
arminru, line 313:
```suggestion
  //   ignored, the `code` field is the sole carrier of the status.
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,97 @@ message Span {
   // enforced. If this value is 0, then no links were dropped.
   uint32 dropped_links_count = 14;
 
-  // An optional final status for this span. Semantically when Status
-  // wasn't set it is means span ended without errors and assume
-  // Status.Ok (code = 0).
+  // An optional final status for this span. Semantically when Status isn't set it means
```
arminru, line 223:
```suggestion
  // An optional final status for this span. Semantically when Status isn't set, it means
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,97 @@ message Span {
...
+    DEPRECATED_STATUS_CODE_DATA_LOSS           = 15;
+    DEPRECATED_STATUS_CODE_UNAUTHENTICATED     = 16;
   };
 
-  // The status code. This is optional field. It is safe to assume 0 (OK)
-  // when not set.
-  StatusCode code = 1;
+  // The deprecated status code. This is an optional field.
+  //
+  // This field is deprecated and is replaced by `code` field below. See backward
```
arminru, line 254:
```suggestion
  // This field is deprecated and is replaced by the `code` field below. See backward
```

arminru (dismissed):

tigrannajaryan:
> How does this avoid being a breaking change of a stable API when the name of the field and its values are being changed?

Names of the fields are not exposed OpenTelemetry API. Compatibility from protocol perspective means wire compatibility. names of the fields do not exist on the wire level.

We could keep the name of the field and of the enum values if there is a big desire to keep compatibility on the build level. I am not sure it is worth it and it was not part of the compatibility promise in the past (Collector codebase hides these details on purpose so that usage of OTLP protobuf messages in the codebase is not affected).

I am open to re-considering this and it does not affect the logic.

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: 12 months is our formal definition of "stable": https://github.com/open-telemetry/community/blob/47813530864b9fe5a5146f466a58bd2bb94edc72/maturity-matrix.yaml#L57

Oberon00:
> Names of the fields are not exposed in OpenTelemetry API. Compatibility from protocol perspective means wire compatibility. Names of the fields do not exist on the wire level.

What about OTLP/JSON?

tigrannajaryan:
> > Names of the fields are not exposed in OpenTelemetry API. Compatibility from protocol perspective means wire compatibility. Names of the fields do not exist on the wire level.
> 
> What about OTLP/JSON?

Yes, in OTLP/JSON the Protobuf field names are directly exposed on the wire. However, OTLP/JSON is currently in "Alpha", we are free to change it:
image

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: Done, thanks.

bryannaegele:
@tigrannajaryan tooling does rely on these names, so it is a breaking change for every library that relies on the proto names to have to update. When these files (Common and Trace) were officially marked stable in August, I assumed that included stable for tooling, as well.

What I'm not clear on is why this enum is now being deprecated at all, nor why it's being marked as deprecated by prefixing the word to everything. The comment directly above this code states `The Status type defines a logical error model that is suitable for different programming environments, including REST APIs and RPC APIs.` but now those codes are marked deprecated and not replaced by anything equivalent.

It's unclear to me in https://github.com/open-telemetry/oteps/blob/master/text/trace/0136-error_flagging.md#status-mapping-schema if these codes are actually intended to be removed at this time. 

> As part of the specification, OpenTelemetry provides a mapping of semantic conventions to status codes. This removes any ambiguity as to what OpenTelemetry ships with out of the box.

What I did get from the OTEP was the addition of a code field `Status.Code` which has been named `StatusCanonicalCode in here and the addition of a `Status.Source` field which seems to not be in the PR.

I could be missing something as it's difficult to keep up. If so, I'd love to have more context on this PR. I greatly welcome the additions but I'm unclear on the motivation of the deprecation.

opentelemetry/proto/trace/v1/trace.proto
bogdandrutu: I think we need a new stability called "GA", that has stronger requirements.

tigrannajaryan:
>It's unclear to me in https://github.com/open-telemetry/oteps/blob/master/text/trace/0136-error_flagging.md#status-mapping-schema if these codes are actually intended to be removed at this time.

@bryannaegele the OTEP's language may not be clear but the specification is clear: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#status

There is only one code - the StatusCanonicalCode, there is no other code in the spec that we need to also support in the protocol.

There is certainly no intent to have 2 different codes in the API or in the protocol. The OTEP's motivation was that the old enum did not make sense. This has been discussed in multiple meetings and in the OTEP.

>tooling does rely on these names, so it is a breaking change for every library that relies on the proto names to have to update. When these files (Common and Trace) were officially marked stable in August, I assumed that included stable for tooling, as well.

I am not strongly opposed to keeping the old field name and old enum name and value names. The reason I used "deprecated_" as a prefix was to discourage its usage. 

If there is no strong oposition I am happy to modify this PR to restore the names of the deprecated field and use new names for the new field and enum.

bryannaegele:
@tigrannajaryan That doc changed 7 days ago where the API spec clearly stated that these status codes were to be used and thus available to other uses, such as messaging https://github.com/open-telemetry/opentelemetry-specification/blame/3e66e6fcc3bb8070db7ccbdd30005ae9f35ac29c/specification/trace/semantic_conventions/http.md#L40-L63. So the ability to set the status with these semantic conventions were in the API spec and does introduce a breaking change.

These status codes being set are also still in the referenced document for HTTP clients https://github.com/open-telemetry/opentelemetry-specification/blame/master/specification/trace/semantic_conventions/http.md#L109-L116.

bryannaegele:
> If there is no strong oposition I am happy to modify this PR to restore the names of the deprecated field and use new names for the new field and enum.

If this change must be done for some reason, I'm simply asking what is the replacement for marking statuses on a span beyond `error`? Are users meant to devise their own schemes going forward?

bryannaegele:
I just caught up and found https://github.com/open-telemetry/opentelemetry-specification/pull/966/files from 4 days ago, so this conversation is apparently moot and we'll just have to deal with the breaking changes.

tigrannajaryan:
> I'm simply asking what is the replacement for marking statuses on a span beyond error?

The updated spec does not define anything for this. These has been discussed extensively in the spec SIG and error SIG. This PR is a consequence, it is not something new. This simply makes the corresponding change in the protocol.

>we'll just have to deal with the breaking changes.
Again, I am happy to keep the old field names in the protocol, but language SDK's have to deal with the spec change anyway. OTLP exporters in SDK will also have to comply with Compatibility requirements for senders that I added in this PR before GA. I don't think we want to release an SDK that has outdated understanding of Span Status.

bryannaegele:
We can't all be in every SIG meeting :)

We'll deal with it. As you noted, we're going to have to remove all this from the SDK anyhow, so it's better to just get the pain over with.

I'm just looking for guidance at this point on how to guide users on how to substitute the functionality. For users that are already leveraging this functionality, what is the upgrade path for them? Are we just saying "Hey, error status codes aren't a thing anymore so you'll have to figure something out" or is there guidance/patterns we can provide? Asking as a user _and_ a maintainer 😆

Oberon00:
Maybe we should add a compatibility semantic convention to ease transition at least?

tigrannajaryan:
> I'm just looking for guidance at this point on how to guide users on how to substitute the functionality. For users that are already leveraging this functionality, what is the upgrade path for them? Are we just saying "Hey, error status codes aren't a thing anymore so you'll have to figure something out" or is there guidance/patterns we can provide? Asking as a user _and_ a maintainer 😆

@bryannaegele best to ask this in spec repo or gitter where it will get more eyeballs.

tigrannajaryan:
> Maybe we should add a compatibility semantic convention to ease transition at least?

Can you please clarify what do you mean by this?

(I was about to merge this PR, just noticed you comment, so will hold for now).

tigrannajaryan:
> Maybe we should add a compatibility semantic convention to ease transition at least?

@Oberon00  Can you please clarify what do you mean by this?

(I was about to merge this PR, just noticed you comment, so will hold for now).

tigrannajaryan:
I have put a bit more thought in to this and I believe the transition in codebases will be easier to do if we keep the names of the deprecated field and enum unchanged. I am going to revert that part of the change. @bryannaegele FYI.

I am going to also wait for https://github.com/open-telemetry/opentelemetry-specification/issues/1069 resolution and do it in one PR. There is no point in introducing the field with wrong names and risk someone start using it and then immediately break it.

tigrannajaryan:
I update this PR and removed "canonical" adjective to be inline with https://github.com/open-telemetry/opentelemetry-specification/pull/1081

I am using "code value" since "code" is unfortunately already used by a deprecated field which have to keep around for now.

tigrannajaryan:
We have the final version of the filed in the spec: `StatusCode`. Unfortunately this is exactly the name of the existing enum which we cannot use. I am forced to keep what I invented - CodeValue. If there are better suggestions for the field and enum names please speak.
I am removing the WIP label, this is ready for final review.

tigrannajaryan:
I would like to know what others think about the field names. The new name looks awkward to me. I would rather prefer to rename the old fields and use `StatusCode` for the new field/enum.

I am going to open an unrelated PR which legitimizes field renaming as soon as it does not break wire compatibility (which is fine up until we say OTLP/JSON is stable).

tigrannajaryan:
FYI, regarding field name guarantees: https://github.com/open-telemetry/opentelemetry-proto/pull/225

tigrannajaryan:
Given that https://github.com/open-telemetry/opentelemetry-proto/pull/225 got several approvals and is about to be merged I am inclined more and more to rename the fields since we no longer are limited by the compatibility guarantees. I am going to that unless I hear objections.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,97 @@ message Span {
...
+    // completed successfully.
+    STATUS_CODE_OK                  = 1;
+    // The Span contains an error.
+    STATUS_CODE_ERROR               = 2;
+  };
+
+  // The status code.
+  StatusCode code = 3;
+
+  // IMPORTANT: Backward compatibility notes:
```
Oberon00, line 278: Having this floating around at the end looks odd. Maybe move it to the docs of StatusCode? Or at least mention "see compatibility notes below" for both code and deprecated_code.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -220,42 +220,97 @@ message Span {
...
+  //
+  // To ensure any pair of senders and receivers continues to correctly signal and
+  // interpret erroneous situations, the senders and receivers SHOULD follow these rules:
```
Oberon00, line 281: >  SHOULD follow these rules

Either this should be a MUST or we need to describe what other alternatives are valid.

Oberon00: Also you use MUST within this should, which is odd (the other way round would be more understandable to me)

tigrannajaryan:
@open-telemetry/specs-approvers @open-telemetry/specs-trace-approvers please review. I made a few changes back and forth in this PR. The PR description now reflects the latest state.

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: Moved to top of Status.

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: Changed to MUST.

Oberon00 (approved):

bogdandrutu (approved):

tigrannajaryan:
@open-telemetry/specs-approvers @open-telemetry/specs-trace-approvers please review. Implementation in the Collector cannot proceed until this is merged. The Collector is currently non-compliant with the spec because of this.

Oberon00:
I think there are already enough approvals, aren't there?

tigrannajaryan:
> I think there are already enough approvals, aren't there?

@Oberon00 I re-requested reviews. Approvers previously saw and approved a different version of this PR and I think differences are pretty significant, so it is important for them to confirm they are still OK.

Oberon00:
@tigrannajaryan Then you should dismiss the reviews in question to make this clearer.

tigrannajaryan:
@arminru @anuraaga @jkwatson PTAL, I need to move forward, this is blocking Collector work. If I don't see any objections I will merge this today.

jkwatson (approved):

PR #225, opened 2020-10-16
Clarify maturity guarantees

tigrannajaryan:
This clarifies what is and is not part of the promises.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,13 @@ All messages                         | Alpha    |
 (See [maturity-matrix.yaml](https://github.com/open-telemetry/community/blob/47813530864b9fe5a5146f466a58bd2bb94edc72/maturity-matrix.yaml#L57)
 for definition of maturity levels).
 
+Note that maturity guarantees apply only to wire-level compatibility. Message, field, enum
+names of Protobuf messages are not visible on the wire and are not considered part of the
+guarantees. We are free to make change to the names.
+
+In the future when OTLP/JSON is declared stable field names will also become part of
+the maturity guarantees since field names are visible on the wire for JSON encoding.
+
 ## Experiments
 
 In some cases we are trying to experiment with different features. In this case
--- a/README.md
+++ b/README.md
@@ -24,12 +24,13 @@ All messages                         | Alpha    |
 (See [maturity-matrix.yaml](https://github.com/open-telemetry/community/blob/47813530864b9fe5a5146f466a58bd2bb94edc72/maturity-matrix.yaml#L57)
 for definition of maturity levels).
 
-Note that maturity guarantees apply only to wire-level compatibility. Message, field, enum
-names of Protobuf messages are not visible on the wire and are not considered part of the
-guarantees. We are free to make change to the names.
+Note that maturity guarantees apply only to wire-level compatibility for the binary
+Protobuf serialization. Neither message, field, nor enum names of Protobuf messages
+are visible on the wire and are not considered part of the guarantees. We are free
+to make change to the names.
 
-In the future when OTLP/JSON is declared stable field names will also become part of
-the maturity guarantees since field names are visible on the wire for JSON encoding.
+In the future when OTLP/JSON is declared stable, field names will also become part of
+the maturity guarantees, since field names are visible on the wire for JSON encoding.
 
 ## Experiments
 


README.md
```diff
@@ -24,6 +24,13 @@ All messages                         | Alpha    |
 (See [maturity-matrix.yaml](https://github.com/open-telemetry/community/blob/47813530864b9fe5a5146f466a58bd2bb94edc72/maturity-matrix.yaml#L57)
 for definition of maturity levels).
 
+Note that maturity guarantees apply only to wire-level compatibility. Message, field, enum
+names of Protobuf messages are not visible on the wire and are not considered part of the
```
seh, line 28:
```suggestion
Note that maturity guarantees apply only to wire-level compatibility. Neither message, field,
nor enum names of Protobuf messages are visible on the wire and are not considered part of the
```

README.md
```diff
@@ -24,6 +24,13 @@ All messages                         | Alpha    |
...
+guarantees. We are free to make change to the names.
+
+In the future when OTLP/JSON is declared stable field names will also become part of
+the maturity guarantees since field names are visible on the wire for JSON encoding.
```
seh, line 32:
```suggestion
In the future when OTLP/JSON is declared stable, field names will also become part of
the maturity guarantees, since field names are visible on the wire for JSON encoding.
```

MrAlias (approved):

README.md
anuraaga: Would we raise the major version number of the otlp artifacts when changing field names? Since we publish code, if a user uses it for example in integration tests, they will have compile breakage that does not follow semver if we don't consider it a major change. While I'm ok with renames before GA, I feel as if marking stable is independent of the JSON protocol since we publish code - since the world isn't one big monorepo we can't really rely on protobuf to help with that.

arminru (approved):

README.md
tigrannajaryan: Fixed.

README.md
tigrannajaryan: @seh fixed, thanks.

tigrannajaryan: @anuraaga we probably should bump the version number if we make compile-time breaking changes to the proto file _once_ we are at 1.0. Since we are still on 0.x then I think it is not yet necessary. The intent of declaring the protocol parts stable was because we recognized people are using it in the wild and breaking wire compatibility is a big deal. Breaking compile time compatibility is annoying but manageable (especially since it is really us, Otel contributors who need to fix the resulting problems).

bogdandrutu (approved):

tigrannajaryan:

README.md
tigrannajaryan: >I feel as if marking stable is independent of the JSON protocol since we publish code

I agree, we may want to make that decision before declaring JSON encoding stable. However, once we decide that JSON encoding needs to be declared stable we will be forced to at least declare that field names are stable since they are visible on the wire. Note that there are part of Protobufs that are not visible on the wire in JSON either - e.g. enum names, so we do need to differentiate on-the-wire and build-time compatibility.

anuraaga: It sounds like the plan is JSON being stable before 1.x wasn't sure of that. If that's the case, no problem at all :)

anuraaga (approved):

README.md
```diff
@@ -24,6 +24,13 @@ All messages                         | Alpha    |
 (See [maturity-matrix.yaml](https://github.com/open-telemetry/community/blob/47813530864b9fe5a5146f466a58bd2bb94edc72/maturity-matrix.yaml#L57)
 for definition of maturity levels).
 
+Note that maturity guarantees apply only to wire-level compatibility. Neither message,
```
Oberon00, line 27:
```suggestion
Note that maturity guarantees apply only to wire-level compatibility for the binary protobuf serialization. Neither message,
```

README.md
```diff
@@ -24,6 +24,13 @@ All messages                         | Alpha    |
...
+field, nor enum names of Protobuf messages are visible on the wire and are not
+considered part of the guarantees. We are free to make change to the names.
+
+In the future when OTLP/JSON is declared stable, field names will also become part of
```
Oberon00, line 31: I think OTLP/JSON should just always be left unstable. I think the cost/benefit ratio to declaring this stable is just not right. Are there any strong reasons for supporting JSON at all?

Oberon00 (approved):
I'm only a specs-trace-approver, so not sure if my approval formally counts, but LGTM!

README.md
tigrannajaryan: Short answer: yes, we want it, but I am not sure we want to open this discussion here in this PR. OTLP/JSON is already part of the specification: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/protocol/otlp.md#request

README.md
tigrannajaryan: Thanks, fixed.

README.md
bogdandrutu: I think @Oberon00 has good arguments (probably more in his mind). Would be good to have this discussion, also may be good to discuss if JSON should be based on these protos or we defined it separately.

PR #227, opened 2020-10-28
Add SummaryDataPoint support to Metrics proto (opentelemetry/opentele…

gcacace:
…metry-specification#1146)

* Add IntSummary and DoubleSummary to data
* Add IntSummaryDataPoint and DoubleSummaryDataPoint
* Comments

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -61,11 +61,11 @@ message InstrumentationLibraryMetrics {
 //  +------------+
 //  |name        |
 //  |description |
-//  |unit        |     +---------------------------+
-//  |data        |---> |Gauge, Sum, Histogram, ... |
-//  +------------+     +---------------------------+ 
+//  |unit        |     +------------------------------------+
+//  |data        |---> |Gauge, Sum, Histogram, Summary, ... |
+//  +------------+     +------------------------------------+
 //
-//    Data [One of Gauge, Sum, Histogram, ...]
+//    Data [One of Gauge, Sum, Histogram, Summary, ...]
 //  +-----------+
 //  |...        |  // Metadata about the Data.
 //  |points     |--+
@@ -142,6 +142,7 @@ message Metric {
     DoubleSum double_sum = 7;
     IntHistogram int_histogram = 8;
     DoubleHistogram double_histogram = 9;
+    DoubleSummary double_summary = 11;
   }
 }
 
@@ -217,6 +218,14 @@ message DoubleHistogram {
   AggregationTemporality aggregation_temporality = 2;
 }
 
+// DoubleSummary metric data are used to convey quantile summaries,
+// a Prometheus and OpenMetrics data type. These data points cannot
+// always be merged in a meaningful way. While they can be useful in some
+// applications, histogram data points are recommended for new applications.
+message DoubleSummary {
+  repeated DoubleSummaryDataPoint data_points = 1;
+}
+
 // AggregationTemporality defines how a metric aggregator reports aggregated
 // values. It describes how those values relate to the time interval over
 // which they are aggregated.
@@ -250,7 +259,7 @@ enum AggregationTemporality {
   //      t_0+2 with a value of 2.
   AGGREGATION_TEMPORALITY_DELTA = 1;
 
-  // CUMULATIVE is an AggregationTemporality for a metic aggregator which
+  // CUMULATIVE is an AggregationTemporality for a metric aggregator which
   // reports changes since a fixed start time. This means that current values
   // of a CUMULATIVE metric depend on all previous measurements since the
   // start time. Because of this, the sender is required to retain this state
@@ -506,6 +515,61 @@ message DoubleHistogramDataPoint {
   // (Optional) List of exemplars collected from
   // measurements that were used to form the data point
   repeated DoubleExemplar exemplars = 8;
+}
+
+// SummaryDataPoint is a single data point in a timeseries that describes the
+// time-varying values of a Summary metric.
+message DoubleSummaryDataPoint {
+  // The set of labels that uniquely identify this timeseries.
+  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1;
+
+  // start_time_unix_nano is the last time when the aggregation value was reset
+  // to "zero". For some metric types this is ignored, see data types for more
+  // details.
+  //
+  // The aggregation value is over the time interval (start_time_unix_nano,
+  // time_unix_nano].
+  //
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
+  // 1970.
+  //
+  // Value of 0 indicates that the timestamp is unspecified. In that case the
+  // timestamp may be decided by the backend.
+  fixed64 start_time_unix_nano = 2;
+
+  // time_unix_nano is the moment when this aggregation value was reported.
+  //
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
+  // 1970.
+  fixed64 time_unix_nano = 3;
+
+  // count is the number of values in the population. Must be non-negative.
+  fixed64 count = 4;
+
+  // sum of the values in the population. If count is zero then this field
+  // must be zero.
+  double sum = 5;
+
+  // Represents the value at a given quantile of a distribution.
+  //
+  // To record Min and Max values following conventions are used:
+  // - The 1.0 quantile is equivalent to the maximum value observed.
+  // - The 0.0 quantile is equivalent to the minimum value observed.
+  //
+  // See the following issue for more context:
+  // https://github.com/open-telemetry/opentelemetry-proto/issues/125
+  message ValueAtQuantile {
+    // The quantile of a distribution. Must be in the interval
+    // [0.0, 1.0].
+    double quantile = 1;
+
+    // The value at the given quantile of a distribution.
+    double value = 2;
+  }
+
+  // (Optional) list of values at different quantiles of the distribution calculated
+  // from the current snapshot. The quantiles must be strictly increasing.
+  repeated ValueAtQuantile quantile_values = 6;
 }
 
 // A representation of an exemplar, which is a sample input int measurement.
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -219,9 +219,11 @@ message DoubleHistogram {
 }
 
 // DoubleSummary metric data are used to convey quantile summaries,
-// a Prometheus and OpenMetrics data type. These data points cannot
-// always be merged in a meaningful way. While they can be useful in some
-// applications, histogram data points are recommended for new applications.
+// a Prometheus (see: https://prometheus.io/docs/concepts/metric_types/#summary)
+// and OpenMetrics (see: https://github.com/OpenObservability/OpenMetrics/blob/4dbf6075567ab43296eed941037c12951faafb92/protos/prometheus.proto#L45)
+// data type. These data points cannot always be merged in a meaningful way.
+// While they can be useful in some applications, histogram data points are
+// recommended for new applications.
 message DoubleSummary {
   repeated DoubleSummaryDataPoint data_points = 1;
 }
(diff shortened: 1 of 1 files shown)


gcacace:
Please consider the main use case for `IntSummary` and `DoubleSummary` is for transport purposes (not aggregation). In this specific case I've removed `AggregationTemporality`, addressing concerns expressed in open-telemetry/opentelemetry-proto#199.

I've added `min` and `max`.

Not sure if we need to support both `IntSummary` and `DoubleSummary`, or the Double version is enough (I've included both for model completeness).

Not sure if we need to support `exemplars`.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -217,6 +219,14 @@ message DoubleHistogram {
   AggregationTemporality aggregation_temporality = 2;
 }
 
+message IntSummary {
```
JasonXZLiu, line 222: Following a previous discussion of the summary metric (in the metrics SIG) and [here](https://github.com/open-telemetry/opentelemetry-proto/pull/199), I believe @bogdandrutu suggested to mark this data format as Legacy. This would help avoid confusion and prevent future instruments from being built on the summary (as we're mainly using it to transport Prometheus summary metrics).

@jmacd @bogdandrutu Would you have any opinions here?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -508,6 +518,136 @@ message DoubleHistogramDataPoint {
...
+  // sum of the values in the population. If count is zero then this field
+  // must be zero.
+  sfixed64 sum = 5;
+
+  // min of the values in the population.
+  sfixed64 min = 6;
+
+  // max of the values in the population.
+  sfixed64 max = 7;
+
```
bogdandrutu, line 559: I would not add this for the moment, we can followup. The purpose of this is to be compatible with OpenMetrics/Prometheus and they don't define that.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: I'll accept "Legacy", yes.
Because of this legacy, should we only support Double summaries?
The precise wording can be:

// DoubleSummary metric data are used to convey quantile summaries, 
// a Prometheus and OpenMetrics data type. These data points cannot 
// always be merged in a meaningful way. While they can be useful in some
// applications, histogram data points are recommended for new applications.

opentelemetry/proto/metrics/v1/metrics.proto
gcacace: I've removed them, thanks for the comment

opentelemetry/proto/metrics/v1/metrics.proto
gcacace: I've removed `IntSummary` and added the suggested comment the `DoubleSummary`.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -508,6 +517,65 @@ message DoubleHistogramDataPoint {
...
+    double value = 2;
+  }
+
+  // (Optional) list of values at different quantiles of the distribution calculated
+  // from the current snapshot. The quantiles must be strictly increasing.
+  repeated ValueAtQuantile quantile_values = 6;
+
+  // (Optional) List of exemplars collected from
+  // measurements that were used to form the data point
+  repeated DoubleExemplar exemplars = 7;
```
gcacace, line 576: I'm wondering if we need also exemplars or we can get rid of them

gcacace: I'm gonna get rid of that since it's not required for Prometheus use case.

jmacd (approved):

bogdandrutu (approved):

JasonXZLiu (approved):

gcacace:
Can someone with merge permissions push and close this change please? When it is going to be released in Maven?

bogdandrutu:
We don't release this repo to maven (that is a Java specific repo, if I am not wrong). The other projects consume this repo as git submodule and generates their code.

bogdandrutu:
@open-telemetry/specs-metrics-approvers please take a look.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -217,6 +218,14 @@ message DoubleHistogram {
   AggregationTemporality aggregation_temporality = 2;
 }
 
+// DoubleSummary metric data are used to convey quantile summaries,
+// a Prometheus and OpenMetrics data type. These data points cannot
+// always be merged in a meaningful way. While they can be useful in some
+// applications, histogram data points are recommended for new applications.
+message DoubleSummary {
```
rakyll, line 225: Prometheus doesn't have support for int summaries but I'm seeing OTLP represents int histograms. Are we planning to do the same here? If yes, it would be good to add a TODO at least.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -508,6 +517,61 @@ message DoubleHistogramDataPoint {
...
+    // [0.0, 1.0].
+    double quantile = 1;
+
+    // The value at the given quantile of a distribution.
+    double value = 2;
+  }
+
+  // (Optional) list of values at different quantiles of the distribution calculated
+  // from the current snapshot. The quantiles must be strictly increasing.
+  repeated ValueAtQuantile quantile_values = 6;
```
rakyll, line 572: Does Prometheus summary type support exemplars yet?

rakyll (commented):
I'm not familiar with the OTLP yet to give a more comprehensive review but I am leaving two high level questions.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Not that I know of

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: We are not planning to support the int version since this is mostly for backwards compatibility with Prometheus

opentelemetry/proto/metrics/v1/metrics.proto
JasonXZLiu: Looking at the [Prometheus client proto](https://github.com/prometheus/client_model/blob/master/go/metrics.pb.go), there is no support for exemplars with the Prometheus Summary type. (It's only available for Counters and Histograms.)

bogdandrutu: If they add that we will also add it.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: If they add that we will add it

gcacace:
> We don't release this repo to maven (that is a Java specific repo, if I am not wrong). The other projects consume this repo as git submodule and generates their code.

Who should I ask in order to have a released version on this code in Maven then? https://mvnrepository.com/artifact/io.opentelemetry/opentelemetry-proto

bogdandrutu:
@gcacace that is generated by the opentelemetry-java repo. So once this is merged, and the otel java repo resyncs will export this to maven.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -508,6 +517,61 @@ message DoubleHistogramDataPoint {
   repeated DoubleExemplar exemplars = 8;
 }
 
+// SummaryDataPoint is a single data point in a timeseries that describes the
```
MrAlias, line 520:
```suggestion
// DoubleSummaryDataPoint is a single data point in a timeseries that describes the
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -217,6 +218,14 @@ message DoubleHistogram {
   AggregationTemporality aggregation_temporality = 2;
 }
 
+// DoubleSummary metric data are used to convey quantile summaries,
+// a Prometheus and OpenMetrics data type. These data points cannot
```
MrAlias, line 222: Should we include that these are used to also record `MinMaxSumCount` values?

MrAlias (approved):

opentelemetry/proto/metrics/v1/metrics.proto
gcacace: I'm following the same convention on comments of the other data structures

opentelemetry/proto/metrics/v1/metrics.proto
JasonXZLiu: I'm not too sure, but I believe there wasn't a final decision on whether the DoubleSummary would be used to record MinMaxSumCount values. I believe it's only used for backwards compatibility with the Prometheus Summary metric for now.

opentelemetry/proto/metrics/v1/metrics.proto
tigrannajaryan: I see 
```
// IntDataPoint is a single data point in a timeseries that describes the
// time-varying values of a int64 metric.
message IntDataPoint {
```
and
```
// DoubleDataPoint is a single data point in a timeseries that describes the
// time-varying value of a double metric.
message DoubleDataPoint {
```
The convention should be to use the full name of the message in the comment. If elsewhere it is not so it needs to be fixed too. DoubleHistogramDataPoint appears to be violating the convention, can be fixed in a future PR.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -508,6 +517,61 @@ message DoubleHistogramDataPoint {
...
+  fixed64 start_time_unix_nano = 2;
+
+  // time_unix_nano is the moment when this aggregation value was reported.
+  //
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
+  // 1970.
+  fixed64 time_unix_nano = 3;
+
+  // count is the number of values in the population. Must be non-negative.
+  fixed64 count = 4;
```
tigrannajaryan, line 547: uint64 is possibly a better choice since counts tend to be relatively small numbers which will pack better on the wire. I see histogram does this too, so it is probably best to change it in a separate PR if we want to do it.

opentelemetry/proto/metrics/v1/metrics.proto
tigrannajaryan, line 222: Can we leave out the references to Prometheus and OpenMetrics? Do they add anything of value? If it is important for understanding the semantics then it is best to include a link to Prometheus and OpenMetrics that explain it.

gcacace:
How can I get more traction on closing out this PR?

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Yes, @MrAlias we decided to postpone that decision for the moment.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: We had this discussion before. Let's keep fixed64 and rethink this everywhere instead of having an inconsistency

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Making this a link is better. We want to keep the reference because as the PR explains the summary is mostly to be able to carry openmetrics/Prometheus summaries not for us to produce them

bogdandrutu:
@gcacace i think the author needs to respond and fix some of the comments.

opentelemetry/proto/metrics/v1/metrics.proto
gcacace: Just to get some clarity on the ask, do you want me to add links to https://prometheus.io/ and https://openmetrics.io/ ?

tigrannajaryan: Yes, the links would be very useful. Also add a relevant OpenMetrics link.

gcacace:
All comments addressed.

tigrannajaryan (approved):
LGTM

JasonXZLiu:
Would it be possible to make a release with these changes right after merging this PR? We have some blocked [PR](https://github.com/open-telemetry/opentelemetry-collector/pull/2048)'s waiting for this datapoint to be added to the proto.

bogdandrutu:
Don't need a release just update the submodule to point to the merged commit

PR #233, opened 2020-11-09
Annotate proto code documentation

aheadmode:
Annotate pseudo-code in proto code documentation

https://github.com/open-telemetry/opentelemetry-rust/pull/352


arminru:
@aheadmode Could you please explain why this is necessary? From the PR you linked in the description it looks as if "tonic" would somehow try to parse the comments and expects a particular syntax there but I don't understand what it would do with these comments.

aheadmode:
https://github.com/open-telemetry/opentelemetry-proto/blob/master/opentelemetry/proto/trace/v1/trace.proto#L242-L249

aheadmode:
https://github.com/danburkert/prost/pull/360

PR #240, opened 2020-11-18
Move codeowners in .github directory

bogdandrutu:


yurishkuro:
Unless this is functionally necessary (which I don't think it is), I prefer these metadata files in the root where they are easier to find.

Oberon00 (approved):
I prefer these files in the .github repository where they don't clutter the root. Also I think they are easy enough to find if they are all in one place.
(Note: I'm only a traces approver, so disregard the green color of the checkmark)

arminru (approved):
I agree with @Oberon00 but don't have a strong opinion - either is fine with me.

bogdandrutu:
@yurishkuro my argument is mostly consistency, we have this everywhere and users would know the file is there.

PR #245, opened 2020-11-30
Clarify what empty instrumentation name means, and missing instrumentation library

bogdandrutu:

### DIFF
--- a/opentelemetry/proto/common/v1/common.proto
+++ b/opentelemetry/proto/common/v1/common.proto
@@ -72,6 +72,8 @@ message StringKeyValue {
 // InstrumentationLibrary is a message representing the instrumentation library information
 // such as the fully qualified name and version. 
 message InstrumentationLibrary {
+  // An empty instrumentation library name is equivalent with an "unknown" 
+  // instrumentation library.
   string name = 1;
   string version = 2;
 }
--- a/opentelemetry/proto/common/v1/common.proto
+++ b/opentelemetry/proto/common/v1/common.proto
@@ -73,7 +73,6 @@ message StringKeyValue {
 // such as the fully qualified name and version. 
 message InstrumentationLibrary {
   // An empty instrumentation library name means the name is unknown. 
-  // instrumentation library.
   string name = 1;
   string version = 2;
 }
--- a/opentelemetry/proto/common/v1/common.proto
+++ b/opentelemetry/proto/common/v1/common.proto
@@ -72,7 +72,7 @@ message StringKeyValue {
 // InstrumentationLibrary is a message representing the instrumentation library information
 // such as the fully qualified name and version. 
 message InstrumentationLibrary {
-  // An empty instrumentation library name is equivalent with an "unknown" 
+  // An empty instrumentation library name means the name is unknown. 
   // instrumentation library.
   string name = 1;
   string version = 2;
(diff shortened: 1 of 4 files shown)


opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -26,106 +26,107 @@ option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/logs/v
 
 // A collection of InstrumentationLibraryLogs from a Resource.
 message ResourceLogs {
-    // The resource for the logs in this message.
-    // If this field is not set then no resource info is known.
-    opentelemetry.proto.resource.v1.Resource resource = 1;
+  // The resource for the logs in this message.
+  // If this field is not set then no resource info is known.
```
yurishkuro, line 30:
```suggestion
  // If this field is not set then no resource info is unknown.
```

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -26,106 +26,107 @@ option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/logs/v
...
-    SEVERITY_NUMBER_WARN4  = 16;
-    SEVERITY_NUMBER_ERROR  = 17;
-    SEVERITY_NUMBER_ERROR2 = 18;
-    SEVERITY_NUMBER_ERROR3 = 19;
-    SEVERITY_NUMBER_ERROR4 = 20;
-    SEVERITY_NUMBER_FATAL  = 21;
-    SEVERITY_NUMBER_FATAL2 = 22;
-    SEVERITY_NUMBER_FATAL3 = 23;
-    SEVERITY_NUMBER_FATAL4 = 24;
+  // UNSPECIFIED is the default SeverityNumber, it MUST not be used.
```
yurishkuro, line 50:
```suggestion
  // UNSPECIFIED is the default SeverityNumber, it MUST NOT be used.
```

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -27,7 +27,7 @@ option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/logs/v
 // A collection of InstrumentationLibraryLogs from a Resource.
 message ResourceLogs {
   // The resource for the logs in this message.
-  // If this field is not set then no resource info is known.
+  // If this field is not set then no resource info is unknown.
```
Oberon00, line 30: Too many negations here: "If this field is **not** set then **no** resource info is **un**known.

yurishkuro: good catch
```suggestion
  // If this field is not set then resource info is unknown.
```

yurishkuro (approved):

opentelemetry/proto/common/v1/common.proto
```diff
@@ -72,6 +72,8 @@ message StringKeyValue {
 // InstrumentationLibrary is a message representing the instrumentation library information
 // such as the fully qualified name and version. 
 message InstrumentationLibrary {
+  // An empty instrumentation library name is equivalent with an "unknown"
```
tigrannajaryan, line 75: I think this is confusing.It is not literally equivalent to to a library with the name equal to string "unknown", right? Can we remove the double quote characters and just say:
```suggestion
  // An empty instrumentation library name means the name is unknown. 
```

opentelemetry/proto/common/v1/common.proto
```diff
@@ -72,6 +72,8 @@ message StringKeyValue {
 // InstrumentationLibrary is a message representing the instrumentation library information
 // such as the fully qualified name and version. 
 message InstrumentationLibrary {
+  // An empty instrumentation library name means the name is unknown. 
+  // instrumentation library.
```
tigrannajaryan, line 76: Line not needed anymore.

tigrannajaryan (approved):

opentelemetry/proto/common/v1/common.proto
bogdandrutu: Done.

PR #255, opened 2021-01-29
Deprecate old explicit bounds, add message for explicit bounds

bogdandrutu:
This is required in order for the protocol to develope in the future and support adding
different other types of bounds. With this change it is backwards compatible to add multiple
bound types and embed them into an oneof (wire compatible).

On any service (receiver) if the deprecated field is received the only operation required
is to reassign the deprecated field to the new embeded repeated field in the message.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -368,6 +368,23 @@ message DoubleDataPoint {
   repeated DoubleExemplar exemplars = 5;
 }
 
+// ExplicitBounds specifies buckets with explicitly defined bounds for values.
+// The bucket boundaries are described by "bounds" field.
+message ExplicitBounds {
+  // This defines size(bounds) + 1 (= N) buckets. The boundaries for bucket
+  // at index i are:
+  //
+  // (-infinity, bounds[i]) for i == 0
+  // [bounds[i-1], bounds[i]) for 0 < i < N-1
+  // [bounds[i], +infinity) for i == N-1
+  // The values in bounds array must be strictly increasing.
+  //
+  // Note: only [a, b) intervals are currently supported for each bucket except the first one.
+  // If we decide to also support (a, b] intervals we should add support for these by defining
+  // a boolean value which decides what type of intervals to use.
+  repeated double explicit_bounds = 7;
+}
+
 // IntHistogramDataPoint is a single data point in a timeseries that describes
 // the time-varying values of a Histogram of int values. A Histogram contains
 // summary statistics for a population of values, it may optionally contain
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -496,24 +500,11 @@ message DoubleHistogramDataPoint {
   // Otherwise all option fields and "buckets" field must be omitted in which case the
   // distribution of values in the histogram is unknown and only the total count and sum are known.
 
+  repeated double deprecated_explicit_bounds = 7 [deprecated=true];
```
jsuereth, line 503: For my own understanding, we're *not* supporting JSON compatibility yet, so this kind of deprecation (with an immediate rename) is ok?

tigrannajaryan: This will also break Collector's build because the field name is used in the codebase, but that's acceptable for now.
What's probably more important is that once we update Collector translation code to use the new explicit_bounds field we are going to break semantic compatibility with existing senders / receivers. (Alternatively, we can look at the deprecated field as well, but not sure if we want to do that).

bogdandrutu: I am deprecating the field and plan to accept it in the receiver and translate it into the new format

tigrannajaryan (approved):

opentelemetry/proto/metrics/v1/metrics.proto
jsuereth: I'm not sure this answered my question.  Can we rename here?  Does this break a JSON implementation?

I understand from a PROTO standpoint, this isn't breaking, but this would break text-based version (IIUC).   Do we not have text-based usage yet so only proto matters?

tigrannajaryan: It does break JSON. This is acceptable, since JSON is currently declared Alpha: https://github.com/open-telemetry/opentelemetry-proto#maturity-level and we allow breaking changes while in Alpha.

bogdandrutu:
Needs a decision on https://github.com/open-telemetry/opentelemetry-proto/issues/259 and https://github.com/open-telemetry/opentelemetry-proto/issues/258

jmacd:
I intend to send an alternate to this PR, in which I propose the approach described here: https://github.com/open-telemetry/opentelemetry-proto/issues/259#issuecomment-784850753

Specifically, this would:
1. make the `oneof` cover "buckets" which include both boundaries and counts in a specific format
2. make counts be double, not uint64, to support sampling use-cases (e.g., Statsd).

jmacd:
The group has discussed support for #272.

PR #270, opened 2021-03-03
Deprecate IntHistogram, rename DoubleHistogram to Histogram

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-proto/issues/257

Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>

Issue #264, opened 2021-02-17
Decouple Number Type from Point Kind in OTLP protocol 


victlu:
The OTLP message Metric has the following definition for instrument types...

[spec](https://github.com/open-telemetry/opentelemetry-proto/blob/92d15b79b5f24c711cf2c00fee366904069bcb6b/opentelemetry/proto/metrics/v1/metrics.proto#L139)
```
  oneof data {
    IntGauge int_gauge = 4;
    DoubleGauge double_gauge = 5;
    IntSum int_sum = 6;
    DoubleSum double_sum = 7;
    IntHistogram int_histogram = 8;
    DoubleHistogram double_histogram = 9;
    DoubleSummary double_summary = 11;
  }
```

I propose we only define Instrument Types (i.e. Gauge, Sum, etc...) at the message Metric level.

The data type (i.e. int, double, etc...) should be lower level, maybe at the message DataPoint level.

i.e.
  ```
  message DataPoint {
    // ...

    // value itself.
    // sfixed64 value = 4;     <= Replace

    oneof value {
       sfixed64 Int64Value = 4;
       double doubleValue = 5;
    }
  }
  ```

jmacd:
I've proposed similar, #172, but at this point I'm happy with what we have.

bogdandrutu:
@victlu the problem with this approach, is that if there are systems that consider the data point kind as part of the identity for a metric will be impossible to have that. For example stackdriver in GCP does that, they consider an int sum different than and double sum, and they need to know if the metric will have one or the other.

victlu:
@bogdandrutu, In your case, I would think it serves them better to actually name their metrics accordingly (as they want two time series) rather than by the data type.

I think this question is related to #1366.  How do we consider the "name" to be unique in identifying a time series.

victlu:
> This would require at least one data point to arrive in order for exporters to create metric descriptors in the backend. What are the use cases where the current data model is not working for you?

@rakyll Why can't metric descriptors be created (without data type)?  Or are you saying a descriptor MUST include the data type?  In which case, that's the argument I'm making here...

IMHO, metric backend systems ultimately shows/graphs/alerts/store/etc... numeric based Time-Series.  And for performance and efficiency, they want to use the most efficient representation of the "number".  Thus, promotion of data types are common especially for storage considerations.  The important thing is that it's just a number regardless if its transmitted as a byte, short, ushort, int, uint, long, ulong, float, double, etc...

[...]

PR #272, opened 2021-03-04
Create a oneof for Histogram buckets; add ExplicitBuckets; deprecate DoubleHistogram

### LINKED ISSUE #259: Histogram support multiple boundaries types vs different types of histogram aggregation
During the meeting on Thu, Feb 11, there was a discussion about our data model supporting new aggregation types, and the answer was clearly we do support that, but that comment triggered me a new idea.

Instead of deprecating the current explicit boundaries from our current definition of Histogram, what if we just add a new type ExponentialHistogram, and that combined with #256 and #257 will keep the number of types at a reasonable number.

Advantages of this solution is that #226 can be a simple addition to the current data model, and also can improve the way how counts are stored for large number of buckets (like using int64 instead of fixed64, or even using other encoding that does not require to have a value for 0 buckets, etc.).

Another advantage is that we don't have to deal with situation where points in the same histogram metric have different boundaries types.

I would like to propose this for discussion and make a decision about this. I am not 100% convinced on one or the other, but I try to propose this possibility.

/cc @jmacd @open-telemetry/specs-metrics-approvers

jmacd:
This deprecates the existing DoubleHistogram and creates a new one with room for more histogram types to be added in the future.

The oneof proposed here was first described in the issue https://github.com/open-telemetry/opentelemetry-proto/issues/259#issuecomment-784850753 and has support from contributors working on #226, e.g., https://github.com/open-telemetry/opentelemetry-proto/pull/226#issuecomment-776526864.

Fixes #259.
Alternate to #255.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -421,7 +421,6 @@ message IntHistogramDataPoint {
   // distribution of values in the histogram is unknown and only the total count and sum are known.
 
   // explicit_bounds is the only supported bucket option currently.
-  // TODO: Add more bucket options.
 
   // explicit_bounds specifies buckets with explicitly defined bounds for values.
   //
@@ -495,9 +494,6 @@ message DoubleHistogramDataPoint {
   // In that case one of the option fields below and "buckets" field both must be defined.
   // Otherwise all option fields and "buckets" field must be omitted in which case the
   // distribution of values in the histogram is unknown and only the total count and sum are known.
-
-  // explicit_bounds is the only supported bucket option currently.
-  // TODO: Add more bucket options.
 
   // explicit_bounds specifies buckets with explicitly defined bounds for values.
   //
@@ -633,4 +629,89 @@ message DoubleExemplar {
   // trace_id may be missing if the measurement is not recorded inside a trace
   // or if the trace is not sampled.
   bytes trace_id = 5;
+}
+
+// Deprecated DoubleHistogram => DeprecatedHistogram
+
+// Represents the type of a metric that is calculated by aggregating as a
+// Histogram of all reported double measurements over a time interval.
+message DeprecatedHistogram {
+  repeated DeprecatedHistogramDataPoint data_points = 1;
+
+  // aggregation_temporality describes if the aggregator reports delta changes
+  // since last report time, or cumulative changes since a fixed start time.
+  AggregationTemporality aggregation_temporality = 2;
+}
+
+// HistogramDataPoint is a single data point in a timeseries that describes the
+// time-varying values of a Histogram of double values. A Histogram contains
+// summary statistics for a population of values, it may optionally contain the
+// distribution of those values across a set of buckets.
+message DeprecatedHistogramDataPoint {
+  // The set of labels that uniquely identify this timeseries.
+  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1;
+
+  // start_time_unix_nano is the last time when the aggregation value was reset
+  // to "zero". For some metric types this is ignored, see data types for more
+  // details.
+  //
+  // The aggregation value is over the time interval (start_time_unix_nano,
+  // time_unix_nano].
+  // 
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
+  // 1970.
+  //
+  // Value of 0 indicates that the timestamp is unspecified. In that case the
+  // timestamp may be decided by the backend.
+  fixed64 start_time_unix_nano = 2;
+
+  // time_unix_nano is the moment when this aggregation value was reported.
+  // 
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
+  // 1970.
+  fixed64 time_unix_nano = 3;
+
+  // count is the number of values in the population. Must be non-negative. This
+  // value must be equal to the sum of the "count" fields in buckets if a
+  // histogram is provided.
+  fixed64 count = 4;
+
+  // sum of the values in the population. If count is zero then this field
+  // must be zero. This value must be equal to the sum of the "sum" fields in
+  // buckets if a histogram is provided.
+  double sum = 5;
+
+  // bucket_counts is an optional field contains the count values of histogram
+  // for each bucket.
+  //
+  // The sum of the bucket_counts must equal the value in the count field.
+  //
+  // The number of elements in bucket_counts array must be by one greater than
+  // the number of elements in explicit_bounds array.
+  repeated fixed64 bucket_counts = 6;
+
+  // A histogram may optionally contain the distribution of the values in the population.
+  // In that case one of the option fields below and "buckets" field both must be defined.
+  // Otherwise all option fields and "buckets" field must be omitted in which case the
+  // distribution of values in the histogram is unknown and only the total count and sum are known.
+
+  // explicit_bounds specifies buckets with explicitly defined bounds for values.
+  //
+  // This defines size(explicit_bounds) + 1 (= N) buckets. The boundaries for
+  // bucket at index i are:
+  //
+  // (-infinity, explicit_bounds[i]] for i == 0
+  // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < N-1
+  // (explicit_bounds[i], +infinity) for i == N-1
+  // 
+  // The values in the explicit_bounds array must be strictly increasing.
+  //
+  // Histogram buckets are inclusive of their upper boundary, except the last
+  // bucket where the boundary is at infinity. This format is intentionally
+  // compatible with the OpenMetrics histogram definition.
+  repeated double explicit_bounds = 7;
+
+  // (Optional) List of exemplars collected from
+  // measurements that were used to form the data point
+  repeated DoubleExemplar exemplars = 8;
 }
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -490,20 +490,9 @@ message HistogramDataPoint {
   //
   // If this field is omitted, there is an implied single bucket
   // spanning (-Inf,+Inf) containing the whole population.
-  //
-  // Each of the supported buckets in the oneof below are expected
-  // to state under which circumstances buckets of that type are
-  // considered Mergeable in a non-lossy way with buckets of the same
-  // type.
-  //
-  // OTLP does not allow two `Metric` instances having `Histogram`s
-  // with unmergable bucket types to be merged.  These will remain
-  // disjoint `Metric` instances through any processors within
-  // OpenTelemetry.
   oneof buckets {
     // ExplicitBuckets support encoding arbitrary boundaries and
-    // counts.  Histograms with these buckets are mergeable when
-    // they have identical boundaries, otherwise are not mergeable.
+    // counts.
     ExplicitBuckets explicit = 9;
   }
 
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -141,8 +141,14 @@ message Metric {
     DoubleGauge double_gauge = 5;
     IntSum int_sum = 6;
     DoubleSum double_sum = 7;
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -142,7 +142,8 @@ message Metric {
     IntSum int_sum = 6;
     DoubleSum double_sum = 7;
     IntHistogram int_histogram = 8;
-    DoubleHistogram double_histogram = 9;
+    DeprecatedHistogram deprecated_histogram = 9;
```
jsuereth, line 145: Nit: Can you add details for how consumers of this should behave (similar to what Bogdan has in his PR)

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -482,42 +480,59 @@ message DoubleHistogramDataPoint {
...
-  // the number of elements in explicit_bounds array.
-  repeated fixed64 bucket_counts = 6;
+  // Each of the supported buckets in the oneof below are expected
+  // to state under which circumstances buckets of that type are
+  // considered Mergeable in a non-lossy way with buckets of the same
+  // type.
+  //
+  // When a single Metric uses Histograms with buckets of that are not
+  // mergeable, they will pass through OpenTelemetry collectors as
+  // distinct Metric series, as they cannot be merged.
```
jsuereth, line 495: Nit: Grammar (drop `of`)

```
When a single Metric uses Histograms with buckets that are not mergeable, they will pass through OpenTelemetry collectors as distinct Metric series, as they cannot be merged.
```

Second nit:

> Pass through OpenTelemetry collectors as distinct metric series

I assume this means in the context of OTLP.   A backend exporter should be free to do whatever they think best for their users and backend, but OTLP will guarantee separate series being pushed.

I think a better way to phrase this:

```
OTLP does not allow two `Metric` instances having `Histogram`s with unmergable bucket types to be merged.  These will remain disjoint `Metric` instances through any processors within OpenTelemetry.

File: opentelemetry/proto/metrics/v1/metrics.proto (hunk 1 above)
Comment (jmacd) [reply opentelemetry/proto/metrics/v1/metrics.proto:145]: Sure. I'll let that PR merge first, then update the comments. The comment written in that PR more-or-less applies.
I think I should revert "DeprecatedHistogram" back to "DoubleHistogram", leave it the same way IntHistogram keeps its name.

File: opentelemetry/proto/metrics/v1/metrics.proto (hunk 2 above)

File: opentelemetry/proto/metrics/v1/metrics.proto [hunk 3]
```diff
@@ -482,42 +486,60 @@ message DoubleHistogramDataPoint {
   // buckets if a histogram is provided.
   double sum = 5;
 
-  // bucket_counts is an optional field contains the count values of histogram
+  // buckets are one of several defined bucketing strategies
+  //
+  // If this field is omitted, there is an implied single bucket
+  // spanning (-Inf,+Inf) containing the whole population.
+  //
+  // Each of the supported buckets in the oneof below are expected
```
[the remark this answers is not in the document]
jmacd: Yes, I agree. I've removed the text about what is and is not mergeable.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -482,42 +486,60 @@ message DoubleHistogramDataPoint {
...
+  // type.
+  //
+  // OTLP does not allow two `Metric` instances having `Histogram`s
+  // with unmergable bucket types to be merged.  These will remain
+  // disjoint `Metric` instances through any processors within
+  // OpenTelemetry.
+  oneof buckets {
+    // ExplicitBuckets support encoding arbitrary boundaries and
+    // counts.  Histograms with these buckets are mergeable when
+    // they have identical boundaries, otherwise are not mergeable.
```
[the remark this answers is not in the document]
jmacd: :+1:

bogdandrutu:
@jmacd and some folks had an offline chat, and I want us to make a decision between multiple bucket types vs multiple histogram types:
* Pros for this: Simplify the number of types we will have in the metrics.
* Cons for this: Extra allocations (one extra allocation in every point); Have to deal with a corner case where points have different bucket types.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -482,42 +486,49 @@ message DoubleHistogramDataPoint {
...
-  // Otherwise all option fields and "buckets" field must be omitted in which case the
-  // distribution of values in the histogram is unknown and only the total count and sum are known.
+  // (Optional) List of exemplars collected from
+  // measurements that were used to form the data point
+  repeated DoubleExemplar exemplars = 8;
+}
 
-  // explicit_bounds is the only supported bucket option currently.
-  // TODO: Add more bucket options.
+// ExplicitBuckets is a standard way of represeting arbitrary
```
aabmass, line 504: typo
```suggestion
// ExplicitBuckets is a standard way of representing arbitrary
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -482,42 +486,49 @@ message DoubleHistogramDataPoint {
...
+// histogram data using explicit boundaries.  It combines an
+// array of N-1 boundaries and an array of N counts.
+message ExplicitBuckets {
+  // counts is an optional field contains the count values of histogram
+  // for each bucket.
+  //
+  // The sum of the counts must equal the value in the count field.
```
aabmass, line 511: nit
```suggestion
  // The sum of the counts must equal the value in the HistogramDataPoint's
 count field.
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -482,42 +486,49 @@ message DoubleHistogramDataPoint {
...
+  //
+  // The number of elements in counts array must be by one greater than
```
aabmass, line 513: nit feel free to disregard
```suggestion
  // The number of elements in counts array must be one greater than
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -482,42 +486,49 @@ message DoubleHistogramDataPoint {
   // buckets if a histogram is provided.
   double sum = 5;
 
-  // bucket_counts is an optional field contains the count values of histogram
-  // for each bucket.
-  //
-  // The sum of the bucket_counts must equal the value in the count field.
+  // buckets are one of several defined bucketing strategies
```
aabmass, line 489: nit, i didn't realize `buckets` was referring to the field at first
```suggestion
  // buckets is one of several defined bucketing strategies
```

jmacd:
> Extra allocations (one extra allocation in every point)

These allocations will be real costs in the collector, but they're not a big deal in SDKs where the export frequency is relatively low.

A fancy protocol buffer implementation would help, for the collector case. 

> Have to deal with a corner case where points have different bucket types

Note that in https://github.com/open-telemetry/opentelemetry-proto/issues/264#issuecomment-789241623 is a proposal to place parallel repeated fields into the Gauge, Sum, and Histogram points to reflect number-type variation. In #259, you propose to add per-bucket-style Histogram value types to the `Metric` oneof, where such variation currently resides. If we follow @victlu's proposal for histograms, to be consistent, we should place the parallel repeated fields inside a single `Histogram`. Note @victlu as you may be prototyping, we no longer desire to support number-type variation for histograms, but we do intend to support bucket-style variation. @bogdandrutu do you think it would reduce complexity if the (single) `Histogram` point had parallel fields, as follows?

```
message Histogram {

  repeated ExplicitHistogramDataPoint explicit_data_points = 1;
  repeated ExpontentialHistogramDataPoint exponential_data_points = 3;
}
```

In both cases, with number-type variation and with bucket-style variation, we do not see a semantic conflict but there is loss of information if we convert between them. By supporting parallel fields we reduce the number of overall oneofs, and we prevent users from having to observe parallel histograms or sums/gauges when there is only bucket-style or number-type variation. We avoid an allocation (however the points are correspondingly larger by 24 bytes in the collector case.).

hdost:
Is the `ExpontentialHistogramDataPoint` the concept of the "HDR" histogram. As i believe the explicit bounds could technically hold a set of Exponential buckets as well. 
Perhaps instead `DynamicHistogramDataPoint` or something to that effect.

yzhuge:
@hdost Exponential buckets will handle high dynamic range. It is different from http://hdrhistogram.org/ in that http://hdrhistogram.org/ use "log linear" (linear subbuckets in log buckets), while Exponential will use log buckets only, for wider compatibility. We might add support for log linear in the future.

0674021379 (approved):

yzhuge:
See https://github.com/open-telemetry/oteps/pull/149 (Add exponential bucketing to histogram protobuf)
The new format is cleaner and simpler than the one proposed in https://github.com/open-telemetry/opentelemetry-proto/pull/226, because the count array is no longer shared across bucket formats

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -482,42 +486,50 @@ message DoubleHistogramDataPoint {
...
-  // TODO: Add more bucket options.
+// ExplicitBuckets is a standard way of representing arbitrary
+// histogram data using explicit boundaries.  It combines an
+// array of N-1 boundaries and an array of N counts.
+message ExplicitBuckets {
+  // counts is an optional field contains the count values of histogram
+  // for each bucket.
+  //
+  // The sum of the counts must equal the value in the HistogramDataPoint's
+ count field.
```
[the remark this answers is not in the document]
bogdandrutu: I mean the fact that you missed " // " on the line 512.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -482,42 +486,50 @@ message DoubleHistogramDataPoint {
...
   //
   // Histogram buckets are inclusive of their upper boundary, except the last
   // bucket where the boundary is at infinity. This format is intentionally
   // compatible with the OpenMetrics histogram definition.
-  repeated double explicit_bounds = 7;
-
-  // (Optional) List of exemplars collected from
-  // measurements that were used to form the data point
-  repeated DoubleExemplar exemplars = 8;
+  repeated double boundaries = 7;
```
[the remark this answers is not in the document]
jmacd: For reference, that's https://github.com/OpenObservability/OpenMetrics/blob/142172882e57a52a26ec32ddfb43b15c42d7f520/proto/openmetrics_data_model.proto#L141

I have a slight preference for two parallel arrays, because the boundaries are constant and they ought to compress better (or could be pre-serialized). I'm definitely open to what the community wants, here.

jmacd:
@open-telemetry/specs-metrics-approvers note this PR fully incorporates #270.

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
jsuereth: Given examplar's already are different from OM (with trace_id embedded too), then I think we'll have special handling here either way. 

I agree w/ @jmacd that repeated primitives may (or can) compress better.

jsuereth (approved):

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -497,42 +497,50 @@ message HistogramDataPoint {
...
+  // measurements that were used to form the data point
+  repeated DoubleExemplar exemplars = 8;
+}
 
-  // explicit_bounds is the only supported bucket option currently.
-  // TODO: Add more bucket options.
+// ExplicitBuckets is a standard way of representing arbitrary
+// histogram data using explicit boundaries.  It combines an
+// array of N-1 boundaries and an array of N counts.
+message ExplicitBuckets {
```
gavindoudou, line 518: Nice to see we revised `histogram` type, wonder do we plan to have any default boundaries or we plan to let each receiver to design themselves?

bogdandrutu: This seems to be a different topic, and I would encourage you to file a different issue for this.

gavindoudou: Added here: https://github.com/open-telemetry/opentelemetry-proto/issues/280

jmacd:
We have decided to let each distinct form of histogram have its own data point kind.

PR #279, opened 2021-03-15
 Add min/max fields to HistogramDataPoint

### LINKED ISSUE #266: Add min and max values to the histogram data points
Today, OpenTelemetry doesn't report min and max values collected for a histogram data point and is breaking the capability to query absolute min and max values from a time series. We propose to add min and max values to the histogram data types. When users set boundaries explicitly, not reporting precise min and max doesn't allow them to see whether their bucket boundaries are well-chosen and precisely set.

Here are some use cases where being able to query absolute min and max values from a series is critical:

* In latency analysis, the bucket boundaries often starts at 0 even though the smallest latency will never be 0. Being able to see the absolute minimum allows users to evaluate the minimum value. This sometimes helps users to find bugs! For example, if your min never went below 40ms for a case where theoretical latency could be 20 ms, it may help you to identify something went wrong.
* Being able to see the max allow users to see how far it is from the 95th/99th percentile. Seldom spikes can add noise to the highest buckets, reporting maximum allows users to adjust bucket boundaries if needed. 
* Outliers are easier to detect with max. For example 1 TB of nearly full disk usage on a few machines will be invisible in the aggregate buckets but are important to identity.
* When doing trend analysis over time, min and max (when collected) are easy to query and allows you to precisely see the boundaries of the population.

jmacd:
Adds a definition for min and max in the Histogram data point. 
See the [definition of these fields in the data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#histogram).
Fixes #266.


jmacd:
Note the idea would be to apply this to the Histogram point _after_ PR #272.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -518,6 +518,28 @@ message DoubleHistogramDataPoint {
...
+  // SHOULD be up to 5-10 minutes or for as long as data is available.
+  // https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#summary
+  
+  // min is the minimum value over a recent window of time.
+  //
+  // The interpretation is based on Metric's temporality.
+  // - delta temporality: the min value MUST be computed over over [start_time, end_time]
+  // - cumulative temporality: the min value SHOULD be computed over [now - WIDTH, end_time]
+  //   for WIDTH in the 5-10 minute range.
+  double min = 13;
```
jsuereth, line 535: I kind of view this as an "exemplar with default 'min' or 'max' labels", given we can represent exemplars outside of buckets.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -518,6 +518,28 @@ message DoubleHistogramDataPoint {
...
+
+  // max is the maximum value over a recent window of time.
+  //
+  // The interpretation is based on Metric's temporality.
+  // - delta temporality: the max value MUST be computed over over [start_time, end_time]
```
jsuereth, line 540: Nit: just once over, not twice over.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -518,6 +518,28 @@ message DoubleHistogramDataPoint {
...
   repeated DoubleExemplar exemplars = 8;
+
+  // Note: the definition used to for min and max fields below is
+  // based on temporality in a way meant to match the Prometheus
+  // Summary data type used with quantiles 0 and 1, respectively, when
+  // paired with cumulative temporality.  This uses a fixed width that
+  // SHOULD be up to 5-10 minutes or for as long as data is available.
+  // https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#summary
+  
+  // min is the minimum value over a recent window of time.
```
jsuereth, line 529: Nit: Declare min/max as (Optional) in the docs?

jsuereth (approved):

MrAlias (approved):

jmacd:
This is stale and we may revisit this topic after other histogram work is finished.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -518,6 +518,28 @@ message DoubleHistogramDataPoint {
   // (Optional) List of exemplars collected from
   // measurements that were used to form the data point
   repeated DoubleExemplar exemplars = 8;
+
+  // Note: the definition used to for min and max fields below is
+  // based on temporality in a way meant to match the Prometheus
+  // Summary data type used with quantiles 0 and 1, respectively, when
+  // paired with cumulative temporality.  This uses a fixed width that
+  // SHOULD be up to 5-10 minutes or for as long as data is available.
+  // https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#summary
```
jack-berg, line 527: The OpenMetrics doc [link](https://github.com/OpenObservability/OpenMetrics/blame/main/specification/OpenMetrics.md#L293) says:
> Quantile values SHOULD represent the recent values. Commonly this would be over the last 5-10 minutes.

Perhaps we should adjust the wording to say something similar, for instance: "Prometheus quantiles are computed on observations more recent than a fixed max age, which is configured in clients but not known to prometheus. This interval width SHOULD be set to represent recent values. It is commonly set to 10 minutes."

I know the OpenMetrics doc says 5-10 minutes, but it defaults to 10 minutes in the prometheus client implementations:
- [java client](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Summary.java#L72-L73)
- [go client](https://github.com/prometheus/client_golang/blob/master/prometheus/summary.go#L75)
- [python client](https://github.com/prometheus/client_python/blame/9118c02a9dfee530af9e83924a88be1ac4994bc0/README.md#L150) doesn't have quantile support
- [ruby client](https://github.com/prometheus/client_ruby/blame/master/README.md#L171) doesn't have quantile support

opentelemetry/proto/metrics/v1/metrics.proto
jack-berg: As discussed in the metrics SIG, Min and max can be made optional by adding flags to [DataPointFlags](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L300-L316).

I propose:
```
  // This HistogramDataPoint contains no recorded min value.  
  // This value SHOULD be used to reflect the intentional omission 
  // of the optional histogram min field, or that no values were 
  // recorded in the window of time.
  FLAG_HISTOGRAM_NO_MIN_VALUE = 2;

  // This HistogramDataPoint contains no recorded max value.  
  // This value SHOULD be used to reflect the intentional omission 
  // of the optional histogram max field, or that no values were 
  // recorded in the window of time.
  FLAG_HISTOGRAM_NO_MIN_VALUE = 2;
```

jmacd: cc: @jsuereth mentioned the idea of using field presence for this, which I think I would prefer (if we can agree).

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: I changed this to say 10 minutes or as long as data is available (i.e., "up to 10 minutes").

jmacd:
I modified this PR to use proto3 "optional" which means to use explicit presence, see this document: 
https://chromium.googlesource.com/external/github.com/protocolbuffers/protobuf/+/refs/heads/master/docs/field_presence.md

opentelemetry/proto/metrics/v1/metrics.proto
jsuereth: Like it!

jsuereth (approved):

mx-psi (commented):
Now that #322 has been merged, should this PR add min/max fields for `ExponentialHistogram` too?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -446,6 +446,29 @@ message HistogramDataPoint {
...
+  // SHOULD be up to 10 minutes or for as long as data is available.
+  // https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#summary
+  
+  // min is the minimum value over a recent window of time.
+  //
+  // The interpretation is based on Metric's temporality.
+  // - delta temporality: the min value MUST be computed over [start_time, end_time]
+  // - cumulative temporality: the min value SHOULD be computed over [now - WIDTH, end_time]
+  //   for WIDTH up to 10 minutes.
+  optional double min = 11;
```
bogdandrutu, line 463: FYI the use of "optional" will make the collector impossible to upgrade to this version right away (will take couple of months probably). Not suggesting to not do the correct thing, but setting the right expectations.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -446,6 +446,29 @@ message HistogramDataPoint {
...
+  // paired with cumulative temporality.  This uses a fixed width that
+  // SHOULD be up to 10 minutes or for as long as data is available.
+  // https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#summary
+  
+  // min is the minimum value over a recent window of time.
+  //
+  // The interpretation is based on Metric's temporality.
+  // - delta temporality: the min value MUST be computed over [start_time, end_time]
+  // - cumulative temporality: the min value SHOULD be computed over [now - WIDTH, end_time]
+  //   for WIDTH up to 10 minutes.
```
bogdandrutu, line 462: Did we not say that we will not do this "width" thing?

bogdandrutu (dismissed):

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: Since the spec PR has added these fields to the data model, now I feel very little has to be said in the protocol. PTAL.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: This was discussed and accepted. We think it is worth handling "optional" fields.

opentelemetry/proto/metrics/v1/metrics.proto

jmacd:
The build says:
```
opentelemetry/proto/metrics/v1/metrics.proto: This file contains proto3 optional fields, but --experimental_allow_proto3_optional was not set.
```

cijothomas (approved):

codeboten:
I believe with https://github.com/open-telemetry/opentelemetry-collector/pull/4929 this is now fully unblocked. @jmacd please rebase the branch

codeboten (approved):

jmacd:
Thank you @codeboten!

codeboten:
@bogdandrutu please review

bogdandrutu:
@jmacd @codeboten as in any ideal world, we should start with fixing a bug first https://github.com/open-telemetry/opentelemetry-proto/pull/336 than adding new functionality. Should we merge the `optional sum`?

codeboten:
Either one is fine with me @bogdandrutu, I don't think I have permissions to push changes to #336 to unblock the PR 😬

bogdandrutu:
Let’s fix that first. Maybe clone that PR and open a new one with necessary changes.

codeboten:
@bogdandrutu see https://github.com/open-telemetry/opentelemetry-proto/pull/366

bogdandrutu:
@codeboten I want to do a release https://github.com/open-telemetry/opentelemetry-proto/pull/367 then start using this in collector to prove everything is fine, then will merge this.

Falmarri:
> I modified this PR to use proto3 "optional" which means to use explicit presence, see this document: https://chromium.googlesource.com/external/github.com/protocolbuffers/protobuf/+/refs/heads/master/docs/field_presence.md

This is really annoying. It means opentelemetry isn't build-able or usable on ubuntu bionic with upstream provided packages.

PR #282, opened 2021-03-22
Add date when the fields will be removed

bogdandrutu:

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -130,8 +130,10 @@ message Metric {
     // and `sum` fields.
     // 3. New receivers, which are aware of this change MUST convert these into
     // `gauge` and `sum` by using the provided as_int field in the oneof values.
+    // This field will be removed in ~3 months, on July 1, 2021.
     IntGauge int_gauge = 4 [deprecated = true];
     Gauge gauge = 5;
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -130,8 +130,10 @@ message Metric {
     // and `sum` fields.
     // 3. New receivers, which are aware of this change MUST convert these into
     // `gauge` and `sum` by using the provided as_int field in the oneof values.
+    // This field will be removed in ~3 months, on July 1, 2021.
```
jsuereth, line 133: Is this enough time for SDKs THEN collector?

bogdandrutu: Do you think 6 months will be better?

jmacd: I imagine the whole project is better off if we move quickly. I support 3 months.

jmacd (approved):

jsuereth (approved):

opentelemetry/proto/metrics/v1/metrics.proto
jsuereth: We discussed this in the SiG.  I'll open an OTEP around attempting to use real-usage to determine migration timelines vs. hard dates and we can discuss my concerns on that.   This is a non-blocking concern on this PR.

Issue #275, opened 2021-03-06
metrics: Meta issue about Data Model Stability Guarantees


hdost:
There are a couple of issues which are looking at breaking changes.

According to the [README.md](https://github.com/open-telemetry/opentelemetry-proto#maturity-level) right now. Metrics is considered _Beta_. However, as part of the [0.5.0 release](https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v0.5.0) there is a mention:
> This is a Release Candidate to declare Metrics part of the protocol **Stable**.

I believe from what I understand that it should still be considered _Beta_ given it's state. And the fact that only as of 0.7.0 was it even marked as _Beta_

Related Change Propositions:
#257 
#264 
#272 

As mentioned in https://github.com/open-telemetry/opentelemetry-proto/issues/264#issuecomment-791629884 we should try to make the transition smooth.

Taking a cursory look at the release history it looks like the last breaking change release was made 2020-08-31 in the form of release 0.5.0. So according to the [maturity matrix](https://github.com/open-telemetry/community/blob/47813530864b9fe5a5146f466a58bd2bb94edc72/maturity-matrix.yaml#L70) we are free to make a release which is not backwards compatible for metrics when we want.

My proposal would be: 

1. Add target types and deprecations to the the protocol
2. Release the "compatible" pre-release version 0.8.0. 
3. Incorporate the incompatible changes to the protocol.
4. Release the "incompatible" version whether that be 0.9.0 or which ever.  And this could be within a few days of each other possibly.

This change would also model for metrics to actually be _Stable_ according to the maturity matrix.

If something like this doesn't occur there is a strong risk of slippage on the whole metric roadmap. Which wouldn't necessarily be the end of the world, but estimates might need to be re-considered.

hdost:
I also discovered (#269) PR which changes the name of Summary. This is something that would techncially break compatibility. Likely if we want to follow the above proposal we should consider reverting it before Step 2. Or more than likely re-add it and have the plain summary. The `DoubleSummary` should be marked deprecated, and then for the case of Step 3. it could then be removed.

hdost:
cc: @jmacd , @bogdandrutu

jmacd:
[...]

PR #283, opened 2021-03-22
Metrics: Replace <string, string> labels with attributes

bogdandrutu:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ Full list of differences found in [this compare.](https://github.com/open-teleme
 * :stop_sign: [DEPRECATION] Deprecate IntSum, IntGauge, and IntDataPoint (#278)
 * :stop_sign: [DEPRECATION] Deprecate IntExemplar (#281)
 * :stop_sign: [DEPRECATION] Deprecate IntHistogram (#270)
+* :stop_sign: [BREAKING] Deprecate `labels` field from NumberDataPoint, HistogramDataPoint, SummaryDataPoint and add equivalent `attributes` field (#283)
+* :stop_sign: [BREAKING] Deprecate `filtered_labels` field from Exemplars and add equivalent `filtered_attributes` field (#283)
 * :stop_sign: [BREAKING] Rename DoubleGauge to Gauge (#278)
 * :stop_sign: [BREAKING] Rename DoubleSum to Sum (#278)
 * :stop_sign: [BREAKING] Rename DoubleDataPoint to NumberDataPoint (#278)
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -352,8 +352,19 @@ message IntDataPoint {
 // NumberDataPoint is a single data point in a timeseries that describes the
 // time-varying value of a double metric.
 message NumberDataPoint {
-  // The set of labels that uniquely identify this timeseries.
-  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1;
+  // The set of key/value pairs that uniquely identify the timeseries from 
+  // where this point belongs. The list may be empty (may contain 0 elements).
+  repeated opentelemetry.proto.common.v1.KeyValue attributes = 7;
+
+  // Labels is deprecated and will be removed soon.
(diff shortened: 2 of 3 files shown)


tigrannajaryan:
It would be great to benchmark this to see the performance impact.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -352,8 +352,19 @@ message IntDataPoint {
 // NumberDataPoint is a single data point in a timeseries that describes the
 // time-varying value of a double metric.
 message NumberDataPoint {
-  // The set of labels that uniquely identify this timeseries.
-  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1;
+  // The set of key/value pairs that uniquely identify the timeseries from
```
jsuereth, line 355: Nit:  Should we call out error scenarios and what to do with competing types in same-label-key across different Metrics, or just but a TODO to call that out in the DataModel Spec?

Oberon00:
Since according to our maturity definition for beta (which metrics are) only allows one breaking change every 3 months, do we want to downgrade to alpha? I don't think we want to wait till end of June with releasing this?

jsuereth (approved):

tigrannajaryan:
@bogdandrutu can you also link for the record to a specification issue or PR where the decision was made to move from strings to full attributes?

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: I am comfortable with a TODO to call this out in the data model spec. As it stands, this file could use a complete rewrite of its comments, but I'd rather discuss and debate what those comments should really say in the data model document, and then copy sections of the data model back into the .proto at the end.

jmacd (approved):

jsuereth:
@tigrannajaryan Here's the recorded decision for this: https://github.com/open-telemetry/opentelemetry-specification/issues/1113#issuecomment-800525450

tigrannajaryan:
> @tigrannajaryan Here's the recorded decision for this: [open-telemetry/opentelemetry-specification#1113 (comment)](https://github.com/open-telemetry/opentelemetry-specification/issues/1113#issuecomment-800525450)

@jsuereth thanks for recording the decision, this is very helpful.

jsuereth:
Please add to description: Fixes #1113

carlosalberto (approved):

hdost:
> Since according to our maturity definition for beta (which metrics are) only allows one breaking change every 3 months, do we want to downgrade to alpha? I don't think we want to wait till end of June with releasing this?

This is why I had brought up #275 I'm sure what the rush being getting 0.8 was if guarantees aren't being guaranteed then they don't mean anything. But also agree i wouldn't have even considered it Beta yet based on the discussions.

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ Full list of differences found in [this compare.](https://github.com/open-teleme
 * :stop_sign: [DEPRECATION] Deprecate IntSum, IntGauge, and IntDataPoint (#278)
 * :stop_sign: [DEPRECATION] Deprecate IntExemplar (#281)
 * :stop_sign: [DEPRECATION] Deprecate IntHistogram (#270)
+* :stop_sign: [BREAKING] Deprecate `labels` field from NumberDataPoint, HistogramDataPoint, SummaryDataPoint and add equivalent `attributes` field (#283)
```
jsuereth, line 18: Discussed in Data Model SiG:  This isn't a breaking change because the deprecated fields are still included in the protocol.   Once removed, that would be breaking.

reyang (approved):
LGTM.

jmacd:
I would like to know from @bogdandrutu and @tigrannajaryan how they think we should decide to merge or not merge this PR.

tigrannajaryan:
> I would like to know from @bogdandrutu and @tigrannajaryan how they think we should decide to merge or not merge this PR.

My understanding is that Metrics SIG decided that metrics labels should become attributes. I trust you have considered all aspects and made this decision, so I have no objections.

My only ask would be that after merging this PR and any other final changes planned, before making a release to benchmark the proto and do an optimization pass to see if we can improve the proto performance wise, while maintaining the same semantics.

PR #288, opened 2021-03-29
Renames buckets to bucket_counts in comments for HistogramDataPoint

maxedmands:
As far as I understand, this is what the comments intended to say. Am I correct in this assumption?


bogdandrutu (approved):

bogdandrutu:
This was fixed in https://github.com/open-telemetry/opentelemetry-proto/pull/291

PR #291, opened 2021-04-01
Fix histogram data point comment about supported boundaries

bogdandrutu:
Fixes: https://github.com/open-telemetry/opentelemetry-proto/issues/259

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -396,6 +396,12 @@ message NumberDataPoint {
 // the time-varying values of a Histogram of int values. A Histogram contains
 // summary statistics for a population of values, it may optionally contain
 // the distribution of those values across a set of buckets.
+//
+// If the histogram contains the distribution of values, then both
+// "explicit_bounds" and "bucket counts" fields must be defined.   
+// If the histogram does not contain the distribution of values, then both
+// "explicit_bounds" and "bucket_counts" must be omitted and only "count" and
+// "sum" are known.
 message IntHistogramDataPoint {
   option deprecated = true;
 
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -432,6 +432,12 @@ message IntHistogramDataPoint {
   // buckets if a histogram is provided.
   sfixed64 sum = 5;
 
+  // A histogram may optionally contain the distribution of the values in the
+  // population. In that case one the "explicit_bounds" and "bucket_counts"
+  // fields must be defined. Otherwise both fields must be omitted in which case
+  // the distribution of values in the histogram is unknown and only the total
+  // count and sum are known.
```
jsuereth, line 439: Nit: This is a bit hard to parse.

Proposal:
```
If the histogram contains a distribution of values then one of the "explicit_bounds" or 
"bucket counts" fields must be defined.   
If the histogram is not a distribution of values, then both "explicit_bounds" and "bucket_counts"
must be omitted and only "count" and "sum" are known.
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -508,6 +506,12 @@ message HistogramDataPoint {
   // buckets if a histogram is provided.
   double sum = 5;
 
+  // A histogram may optionally contain the distribution of the values in the
```
jsuereth, line 509: Same comment as above.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -432,6 +432,12 @@ message IntHistogramDataPoint {
   // buckets if a histogram is provided.
   sfixed64 sum = 5;
 
+  // A histogram may optionally contain the distribution of the values in the
```
jsuereth, line 435: Does this belong as a message-level comment?

jsuereth (changes requested):

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Realized that the message was also wrong, we need both.

opentelemetry/proto/metrics/v1/metrics.proto

opentelemetry/proto/metrics/v1/metrics.proto

jsuereth (approved):

jmacd (approved):

PR #298, opened 2021-04-30
Add schema_url fields as described in OTEP 0152

tigrannajaryan:
This adds schema_url fields introduced in OTEP 0152 (Telemetry Schemas):
https://github.com/open-telemetry/oteps/blob/main/text/0152-telemetry-schemas.md#otlp-changes

### DIFF
--- a/opentelemetry/proto/logs/v1/logs.proto
+++ b/opentelemetry/proto/logs/v1/logs.proto
@@ -33,9 +33,9 @@ message ResourceLogs {
   // A list of InstrumentationLibraryLogs that originate from a resource.
   repeated InstrumentationLibraryLogs instrumentation_library_logs = 2;
 
-  // This schema_url applies to the "resource" field and to all logs in the
-  // "instrumentation_library_logs" except the logs which have a schema_url
-  // specified in the nested InstrumentationLibraryLogs message.
+  // This schema_url applies to the data in the "resource" field. It does not apply
+  // to the data in the "instrumentation_library_logs" field which have their own
+  // schema_url field.
   string schema_url = 3;
 }
 
@@ -49,8 +49,7 @@ message InstrumentationLibraryLogs {
   // A list of log records.
   repeated LogRecord logs = 2;
 
-  // This schema_url applies to all logs in the "logs" field regardless of the
-  // value of the schema_url field in the outer ResourceLogs message.
+  // This schema_url applies to all logs in the "logs" field.
   string schema_url = 3;
 }
 
--- a/opentelemetry/proto/logs/v1/logs.proto
+++ b/opentelemetry/proto/logs/v1/logs.proto
@@ -32,6 +32,11 @@ message ResourceLogs {
 
   // A list of InstrumentationLibraryLogs that originate from a resource.
   repeated InstrumentationLibraryLogs instrumentation_library_logs = 2;
+
+  // This schema_url applies to the "resource" field and to all logs in the
+  // "instrumentation_library_logs" except the logs which have a schema_url
+  // specified in the nested InstrumentationLibraryLogs message.
+  string schema_url = 3;
 }
 
 // A collection of Logs produced by an InstrumentationLibrary.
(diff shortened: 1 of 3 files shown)


jmacd (approved):

jmacd (approved):
Looks good!

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -43,6 +48,10 @@ message InstrumentationLibraryLogs {
 
   // A list of log records.
   repeated LogRecord logs = 2;
+
+  // This schema_url applies to all logs in the "logs" field regardless of the
+  // value of the schema_url field in the outer ResourceLogs message.
+  string schema_url = 3;
```
bogdandrutu, line 54: How does this impact the API? Why is it not part of the `InstrumentationLibrary`?

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -32,6 +32,11 @@ message ResourceLogs {
 
   // A list of InstrumentationLibraryLogs that originate from a resource.
   repeated InstrumentationLibraryLogs instrumentation_library_logs = 2;
+
+  // This schema_url applies to the "resource" field and to all logs in the
+  // "instrumentation_library_logs" except the logs which have a schema_url
+  // specified in the nested InstrumentationLibraryLogs message.
+  string schema_url = 3;
```
bogdandrutu, line 39: Can we start without this optimization? How is this set in the API/SDK?

tigrannajaryan: See application-wide schema URL in SDK: https://github.com/open-telemetry/opentelemetry-specification/pull/1666/files

tigrannajaryan: This is not an optimization. It is _required_ to describe the attributes of the Resource.

gramidt (approved):
Looks good!

jsuereth (approved):

tigrannajaryan:
I have just chatted with @bogdandrutu and he had some ideas that may result in putting schema_url inside InstrumentationLibrary message instead of InstrumentationLibraryXXX message (no difference in semantics).
Please don't merge this yet, while I am thinking about it.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: The API impact is shown here https://github.com/open-telemetry/opentelemetry-specification/pull/1666

I would prefer to keep `schema_url` inside InstrumentationLibrarySpans since it is not a property of the instrumentation library, it is a “meta” field which describes other fields in InstrumentationLibrarySpans and I think putting it inside InstrumentationLibrary is obscure. It would also look a bit strange to use a field in message A (InstrumentationLibrary) to describe fields in message B (repeated Span), when A and B are siblings and do not have parent-child relationship.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: After discussing with @bogdandrutu and putting some more though into this I am going to make a change here. The `schema_url` here will describe the Resource, but will not have any effect on the spans (the overriding logic that it has in the current form). This is unnecessary complication, it saves maybe a tiny bit of data in a request, but creates confusion.
I am going to delete this "overriding" logic and simplify it. Will post a commit in a bit.

tigrannajaryan: OK, see the new commit https://github.com/open-telemetry/opentelemetry-proto/commit/dd92c019a527e55d1f7b926ed4618de7ef4a9369 that does this. Only the comments are changed (which changes the semantic of this field).
I believe this is simpler and easier to understand.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: Discussed with @bogdandrutu one more time and I believe what we have is the right approach. Resolving.

tigrannajaryan:
Reviewers, please have another look. I will keep this open for 2 days just in case there are objections to the last change.

Aneurysm9 (approved):

jsuereth (approved):

jmacd:
@tigrannajaryan Do you want this to merge and release in v0.9? The metrics group would like to release v0.9 very soon. Thanks!

tigrannajaryan:
We have the necessary number of approvals and all comments are resolved, so merging.

PR #307, opened 2021-05-17
Setup github action for auto-assign

bogdandrutu:

### DIFF
--- a/.github/workflows/auto-assign-owners.yml
+++ b/.github/workflows/auto-assign-owners.yml
@@ -1,0 +1,14 @@
+name: 'Auto Assign'
+on:
+  pull_request_target:
+    types: [opened, ready_for_review]
+
+jobs:
+  add-owner:
+    runs-on: ubuntu-latest
+    steps:
+      - name: run
+        uses: kentaro-m/auto-assign-action@v1.1.2
+        with:
+          configuration-path: ".github/auto_assign.yml"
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev (approved):

carlosalberto (approved):

.github/workflows/auto-assign-owners.yml
```diff
@@ -0,0 +1,14 @@
+name: 'Auto Assign'
+on:
+  pull_request_target:
```
arminru, line 3: For reference: The `pull_request` trigger stated in the action's [docs](https://github.com/kentaro-m/auto-assign-action#arrow_forward-usage) will not work for PRs from forks since a GITHUB_TOKEN with write access to the target repository is required. See https://github.com/kentaro-m/auto-assign-action/issues/29.

arminru (approved):

jsuereth:
Can you delete `.github/auto-assign.yml` here then too?

jsuereth (approved):

.github/workflows/auto-assign-owners.yml
bogdandrutu: I know, I think I was the one found the solution :)

arminru: Yep indeed, that was not directed to you but other curious readers 😄

Issue #309, opened 2021-06-01
Should we use bit fields to flag data point properties?


victlu:
We need to flag a few properties in OTLP protocol. Currently, we have a bool for monotonic. 

Should we use a 32-bit word (as a bit flag) now to allow future growth?

Concerns:
- Performance of handling a bit-field vs individual bools.
- Backward and forward compatibility.

victlu:
Initial results on DotNet for Summary kind.
- Generating Summary with Field cost more cycles and is also 9% larger in byte size.


### EncodeSumFlag uses...

```
message SumFlag {
  repeated NumberDataPoint data_points = 1;

  // aggregation_temporality describes if the aggregator reports delta changes
  // since last report time, or cumulative changes since a fixed start time.
  AggregationTemporality aggregation_temporality = 2;

  uint32 flag = 3;
}
```

### EncodeSumField uses...

```
message SumField {
  repeated NumberDataPoint data_points = 1;

  // aggregation_temporality describes if the aggregator reports delta changes
  // since last report time, or cumulative changes since a fixed start time.
  AggregationTemporality aggregation_temporality = 2;

  bool isFlag0 = 10;
  bool isFlag1 = 11;
  bool isFlag2 = 12;
  bool isFlag3 = 13;
  bool isFlag4 = 14;
  bool isFlag5 = 15;
  bool isFlag6 = 16;
  bool isFlag7 = 17;
  bool isFlag8 = 18;
  bool isFlag9 = 19;
  bool isFlag10 = 20;
  bool isFlag11 = 21;
  bool isFlag12 = 22;
  bool isFlag13 = 23;
  bool isFlag14 = 24;
  bool isFlag15 = 25;
  bool isFlag16 = 26;
  bool isFlag17 = 27;
  bool isFlag18 = 28;
  bool isFlag19 = 29;
  bool isFlag20 = 30;
  bool isFlag21 = 31;
  bool isFlag22 = 32;
  bool isFlag23 = 33;
  bool isFlag24 = 34;
  bool isFlag25 = 35;
  bool isFlag26 = 36;
  bool isFlag27 = 37;
  bool isFlag28 = 38;
  bool isFlag29 = 39;
  bool isFlag30 = 40;
  bool isFlag31 = 41;
}
```

flags are set as boolean flags...
```
                       sum.Flag = 
                        (1<<0) | 
                        (1<<1) | 
                        (1<<2) | 
                        (1<<3) | 
[...]

PR #310, opened 2021-06-02
Introduce DataPointFlags bit-field to HistogramDataPoint

victlu:
#309 

Introducing a 32-bit bit-field to allow setting boolean flags (i.e. FLAG_NO_RECORDED_VALUE)  for a Histogram DataPoint. The remaining bits are reserved for future use without needing further modification to protobuf messages.


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -526,6 +566,11 @@ message HistogramDataPoint {
   // 1970.
   fixed64 time_unix_nano = 3;
 
+  // Field "flags" is used as a bit-field as defined in enum DataPointFlags.
+  // If FLAG_NO_RECORDED_VALUE is set, the "count", "sum", "bucket_counts",
+  // "explicit_bounds" fields are undefined.
+  uint32 flags = 10;
```
tigrannajaryan, line 572: I think uint32 packed between two fixed64 will result in wasted 4 bytes in memory due to alignment rules in most (?) compiled languages. Probably better to move this to the end. It won't help with space usage right now (since the last field is also 64bit-aligned and the entire struct is likely too large and most heap allocators will pad it at least to 64bit boundary anyway), but in the future if we add any other field with 32bit or smaller alignment size they will get packed better together.

tigrannajaryan:
Please re-generate (`make gen-all` should do it I believe).

opentelemetry/proto/metrics/v1/metrics.proto
victlu: I think protobuf packs based on ids, and thus flags is at the end at the moment (id = 10).  Although in the middle (for logical organization) in source code.

victlu: Humm, I guess not.

```
When a message is serialized, there is no guaranteed order for how its known or unknown fields should be written. Serialization order is an implementation detail and the details of any particular implementation may change in the future. Therefore, protocol buffer parsers must be able to parse fields in any order.
```

I also don't know if the order in the source code dictates the order of packing. Given serialization is implementation details, I don't know we should worry about this as I don't know how to guarantee this between different protobuf implementations.

tigrannajaryan: Sorry, I was not clear. I am talking about in-memory representation. At least in Go the in-memory representation is a struct with the field listed in the same order they are listed in the proto (regardless of the id numbers). Due to struct field alignment rules this will result in an unused hole of 4 bytes between `flags` and `count` fields.

victlu: I see.

jmacd:
In our group discussion last week, we keeping the NO_RECORDED_VALUE bit from this PR.
The Sum-has-negative-values aspect of this change, as we discussed, belongs at a higher level in the protocol (in the Histogram message, not the Histogram data point).
I will close this PR, until then.

jmacd:
Moves to #316, solving #315.

PR #312, opened 2021-06-08
moved deprecated messages to the bottom of metrics.proto to help with readability

lakamsani:


tigrannajaryan:
Benefits if the change are arguable. Do majority of approvers find this useful?

MrAlias:
> Do majority of approvers find this useful?

Personally, I do not.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -716,3 +546,157 @@ message Exemplar {
   // or if the trace is not sampled.
   bytes trace_id = 5;
 }
+
+//
+// Add deperecated messages below this line
```
Oberon00, line 551:
```suggestion
// Move deprecated messages below this line
```

lakamsani: Hi @Oberon00  sounds good. Will make that change once the majority of approvers decide this is useful. For those of us have  of APM/metrics background but very new to OTel metrics

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -573,7 +433,7 @@ message HistogramDataPoint {
 }
 
 // SummaryDataPoint is a single data point in a timeseries that describes the
-// time-varying values of a Summary metric.
+// time-varying values of a Summary metric. It is considered legacy. See comments for Summary
```
jmacd, line 436: Please undo this minor addition. This is an opinion that is not universally held within the Prometheus community, and until the OTel histogram options are extended beyond explicit boundaries, the Histogram points are not an alternative to Summary points.

jmacd (approved):
I think this is a helpful way to make the `.proto` file more approachable for new readers. I've wanted to make this change myself!

opentelemetry/proto/metrics/v1/metrics.proto
lakamsani: @jmacd  done. Just for my understanding then, as `Summary` is the only user of `SummaryDataPoint`, are we saying that there is disagreement about both or are there situations where `SummaryDataPoint` could be used independent of `Summary` in the future?

jsuereth (approved):

cijothomas (approved):

PR #313, opened 2021-06-17
Fix comment on explicit_bounds

victlu:
Fix bug in comment for explicit_bounds.

If I gave explicit_bounds of...

explicit_bounds  = [ 10, 20, 30 ]

`N` = 4 (sizeof(explicit_bounds) + 1)
valid `i` will include 0, 1, 2.

Given previous statement of...
(explicit_bounds[i], +infinity) for `i` == `N`-1
would cause a out-of-bound error. `N`-1= 3.  explicit_bounds[3] is out-of-bound.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -417,7 +417,7 @@ message HistogramDataPoint {
   //
   // (-infinity, explicit_bounds[i]] for i == 0
   // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < size(explicit_bounds)
-  // (explicit_bounds[i], +infinity) for i == size(explicit_bounds)
+  // (explicit_bounds[i-1], +infinity) for i == size(explicit_bounds)
   //
   // The values in the explicit_bounds array must be strictly increasing.
   //
@@ -644,7 +644,7 @@ message IntHistogramDataPoint {
   //
   // (-infinity, explicit_bounds[i]] for i == 0
   // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < size(explicit_bounds)
-  // (explicit_bounds[i], +infinity) for i == size(explicit_bounds)
+  // (explicit_bounds[i-1], +infinity) for i == size(explicit_bounds)
   //
   // The values in the explicit_bounds array must be strictly increasing.
   //
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -413,12 +413,11 @@ message HistogramDataPoint {
 
   // explicit_bounds specifies buckets with explicitly defined bounds for values.
   //
-  // This defines size(explicit_bounds) + 1 (= N) buckets. The boundaries for
-  // bucket at index i are:
+  // The boundaries for bucket at index i are:
   //
   // (-infinity, explicit_bounds[i]] for i == 0
-  // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < N-1
-  // (explicit_bounds[i-1], +infinity) for i == N-1
+  // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < size(explicit_bounds)
+  // (explicit_bounds[i], +infinity) for i == size(explicit_bounds)
   //
   // The values in the explicit_bounds array must be strictly increasing.
   //
@@ -641,12 +640,11 @@ message IntHistogramDataPoint {
 
   // explicit_bounds specifies buckets with explicitly defined bounds for values.
   //
-  // This defines size(explicit_bounds) + 1 (= N) buckets. The boundaries for
-  // bucket at index i are:
+  // The boundaries for bucket at index i are:
   //
   // (-infinity, explicit_bounds[i]] for i == 0
-  // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < N-1
-  // (explicit_bounds[i-1], +infinity) for i == N-1
+  // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < size(explicit_bounds)
+  // (explicit_bounds[i], +infinity) for i == size(explicit_bounds)
   //
   // The values in the explicit_bounds array must be strictly increasing.
   //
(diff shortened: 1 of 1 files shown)


jmacd (approved):

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -646,7 +646,7 @@ message IntHistogramDataPoint {
   //
   // (-infinity, explicit_bounds[i]] for i == 0
   // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < N-1
-  // (explicit_bounds[i], +infinity) for i == N-1
+  // (explicit_bounds[i-1], +infinity) for i == N-1
```
bogdandrutu, line 649: I think this is wrong. Lets assume only one bound [1.0]:

(-infinity, explicit_bounds[i]] for i == 0 -> (-infinity, 5.0]
(explicit_bounds[i-1], +infinity) for i == N-1 (segfault, infinity) (segfault because N=1 then N-1=0 then i=-1

bogdandrutu: May want to use the values for i where i is fixed:

```
  // (-infinity, explicit_bounds[0]]
  // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < N-1
  // (explicit_bounds[N-1], +infinity)
```

pirgeo:
I think @victlu's solution works. Consider the following code:
```go
bounds := []float64{1.0}
N := len(bounds) + 1

fmt.Printf("N=%d;\tN-1=%d\n", N, N-1)
for i := 0; i < N; i++ {
	if i == 0 {
		// first bound
		fmt.Printf("(-infinity, %f]", bounds[i])
	} else if i < N-1 {
		// all bounds with 0 < i < N-1
		fmt.Printf("(%f, %f]", bounds[i-1], bounds[i])
	} else if i == N-1 {
		// last bound (i == N-1)
		fmt.Printf("(%f, +infinity)", bounds[i-1])
	}
	// Print the indices
	fmt.Printf("\t\ti:% 2d\ti-1: % 2d", i, i-1)

	fmt.Println()
}
```
results in:
```
N=2;    N-1=1
(-infinity, 1.000000]           i: 0    i-1: -1
(1.000000, +infinity)           i: 1    i-1:  0
```
which is what we want, I think. For longer `bounds` arrays, say `bounds := []float64{1.0, 2.0, 3.0}` this prints:
```
N=4;    N-1=3
(-infinity, 1.000000]           i: 0    i-1: -1
(1.000000, 2.000000]            i: 1    i-1:  0
(2.000000, 3.000000]            i: 2    i-1:  1
(3.000000, +infinity)           i: 3    i-1:  2
```
### Alternative

If we define `N` to be `len(bounds)`, we would have to do less `-1` manipulation:
```go
// Not +1
N = len(bounds)
fmt.Printf("N=%d;\tN-1=%d\n", N, N-1)
// less than or equal to instead of less than
for i := 0; i <= N; i++ {
	if i == 0 {
		// first bound
		fmt.Printf("(-infinity, %f]", bounds[i])
	} else if i < N {
		// all bounds with 0 < i < N
		fmt.Printf("(%f, %f]", bounds[i-1], bounds[i])
	} else if i == N {
		// last bound (i == N)
		fmt.Printf("(%f, +infinity)", bounds[i-1])
	}
	fmt.Println()
}
```

which prints (for `bounds := []float64{1.0}`)
```
N=1;    N-1=0
(-infinity, 1.000000]           i: 0    i-1: -1
(1.000000, +infinity)           i: 1    i-1:  0
```
and for `bounds := []float64{1.0, 2.0, 3.0}`: 
```
N=3;    N-1=2
(-infinity, 1.000000]           i: 0    i-1: -1
(1.000000, 2.000000]            i: 1    i-1:  0
(2.000000, 3.000000]            i: 2    i-1:  1
(3.000000, +infinity)           i: 3    i-1:  2
```
and the comment for the `.proto` could therefore be:
```
  // (-infinity, explicit_bounds[i]] for i == 0
  // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < N
  // (explicit_bounds[i], +infinity) for i == N
```

victlu:
I like @pirgeo suggestion for defining N to be size(array). Is there other reasons why N needs to be size(array)+1?

bogdandrutu:
@victlu @pirgeo but I don't see what was wrong before, this is a clarification not a fix correct?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -413,12 +413,11 @@ message HistogramDataPoint {
...
   //
-  // This defines size(explicit_bounds) + 1 (= N) buckets. The boundaries for
-  // bucket at index i are:
+  // The boundaries for bucket at index i are:
   //
   // (-infinity, explicit_bounds[i]] for i == 0
-  // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < N-1
-  // (explicit_bounds[i], +infinity) for i == N-1
+  // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < size(explicit_bounds)
+  // (explicit_bounds[i], +infinity) for i == size(explicit_bounds)
```
bogdandrutu, line 420: is this not obvious wrong now? you access explicit_bounds[size(explicit_bounds)]?

pirgeo: Yes, you are right. The code also accesses `bounds[i-1]` for `i == N`, I missed that when writing up the proposed comment. I think it should be
```suggestion
  // (explicit_bounds[i-1], +infinity) for i == size(explicit_bounds)
```

victlu:
### Using this definition:

(-infinity, explicit_bounds[i]] for i == 0
(explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < size(explicit_bounds)
(explicit_bounds[i-1], +infinity) for i == size(explicit_bounds)

### For 1 item...

explicit_bounds = [ 10 ]
size(explicit_bound) = 1

for i = 0
(-infinity, explicit_bounds[i]] ==> (-infinity, 10]

for i = 1 (i == size(explicit_bounds))
(explicit_bounds[i-1], +infinity) ==> (0, +infinity)

### For 2 items...

explicit_bounds = [ 10, 20 ]
size(explicit_bound) = 2

for i = 0
(-infinity, explicit_bounds[i]] ==> (-infinity, 10]

for i = 1 (0 < i < size(explicit_bounds))
(explicit_bounds[i-1], explicit_bounds[i]) ==> (10, 20]

for i = 2 (i == size(explicit_bounds))
(explicit_bounds[i-1], +infinity) ==> (20, +infinity)

bogdandrutu (approved):
Thanks a lot :)

PR #316, opened 2021-06-23
Add optional data point flags to all metrics data points, support staleness markers

### LINKED ISSUE #309: Should we use bit fields to flag data point properties?

### LINKED ISSUE #315: Support for Prometheus staleness markers
There are several issues in the Prometheus WG that cannot be addressed without specifying how to treat staleness markers. 

Mentioned in https://github.com/open-telemetry/wg-prometheus/issues/59
Mentioned in https://github.com/open-telemetry/wg-prometheus/issues/38

None of these issues actually state that we need a way to represent the NaN value, particularly for Histogram data points which do not always have a semantically meaningful floating point value where a NaN value could go (the `sum` is not always meaningful, see issues #303, #314).

Prometheus used a *specific* IEEE 754 NaN value. In #309 we've discussed using a bit for this.

jmacd:
Adds a bit to represent stale points they way Prometheus does when scraping for metrics yields no results. This avoids the use of IEEE 754 NaN values in data points that do not necessarily have a floating point field that can be used.

Resolves #315.

Related: https://github.com/open-telemetry/opentelemetry-specification/issues/1078

Partial copy from #310.
Resolves #309.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -298,7 +298,12 @@ enum AggregationTemporality {
 }
 
 // DataPointFlags is defined as a protobuf 'uint32' type and is to be used as a
-// bit-field representing 32 distinct boolean flags.
+// bit-field representing 32 distinct boolean flags.  Each flag defined in this
+// enum is a bit-mask.  To test the presence of a single flag in the flags of
+// a data point, for example, use an expression like:
+//
+//   (point.flags & FLAG_NO_RECORDED_VALUE) == FLAG_NO_RECORDED_VALUE
+//
 enum DataPointFlags {
   FLAG_NONE = 0;
 
@@ -307,7 +312,7 @@ enum DataPointFlags {
   // for an equivalent to the Prometheus "staleness marker".
   FLAG_NO_RECORDED_VALUE = 1;
 
-  // Bits 2-32 are reserved for future use.
+  // Bits 2-31 are reserved for future use.
 }
 
 // NumberDataPoint is a single data point in a timeseries that describes the
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -351,8 +351,9 @@ message NumberDataPoint {
   // measurements that were used to form the data point
   repeated Exemplar exemplars = 5;
 
-  // Flags that apply to this specific data point.
-  DataPointFlags flags = 8;
+  // Flags that apply to this specific data point.  See DataPointFlags
+  // for the available flags and their meaning.
+  uint32 flags = 8;
 }
 
 // HistogramDataPoint is a single data point in a timeseries that describes the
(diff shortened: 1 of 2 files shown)


weyert (approved):

jsuereth (approved):

jmacd:
@bogdandrutu @open-telemetry/specs-metrics-approvers Please take a look.

reyang (approved):

cijothomas (approved):

bogdandrutu (commented):
LGTM, except one question:
In protobuf is the enum equivalent with an "int32"? If not how would we intend to add new values? An enum has the property that only one value can be sent, if we add a new bit "foo" then are we going to change them to 4 new values NO_FOO_NONE, NO_FOO_NO_RECORD, FOO_NONE, etc.?

Do we need the value in the points to be uint32 and the enum be the bitset mask? Probably I am missing something...

bogdandrutu:
@jmacd also this needs a rebase

jmacd:
> In protobuf is the enum equivalent with an "int32"?

See https://developers.google.com/protocol-buffers/docs/proto3#enum

jmacd:
I switched this to `uint32` based on @bogdandrutu's feedback.

MrAlias (approved):

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -297,6 +297,19 @@ enum AggregationTemporality {
...
 
+// DataPointFlags is defined as a protobuf 'uint32' type and is to be used as a
+// bit-field representing 32 distinct boolean flags.
+enum DataPointFlags {
+  FLAG_NONE = 0;
+
+  // This DataPoint is valid but has no recorded value.  This value
+  // SHOULD be used to reflect explicitly missing data in a series, as
+  // for an equivalent to the Prometheus "staleness marker".
+  FLAG_NO_RECORDED_VALUE = 1;
```
bogdandrutu, line 308: Is this the mask or the bit position? How do we want people to use these values?

`flags & FLAG_NO_RECORDED_VALUE == 1` // The new value will be FLAG_FOO = 2, FLAG_BAR = 4
OR
`flags & (1 << FLAG_NO_RECORDED_VALUE) == 1` // The current value needs to change to 0 and we have values from 0 to 31.

bogdandrutu (approved):
Last comment :)

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: I was assuming these are masks. I've added a note.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -297,6 +297,24 @@ enum AggregationTemporality {
   AGGREGATION_TEMPORALITY_CUMULATIVE = 2;
 }
 
+// DataPointFlags is defined as a protobuf 'uint32' type and is to be used as a
+// bit-field representing 32 distinct boolean flags.  Each flag defined in this
```
tigrannajaryan, line 301: Nit: it should not necessarily be boolean flags only in the future. If needed some flags may use more than one bit. It is probably obvious so I think what is written is fine.

tigrannajaryan (approved):

jmacd:
:tada:

PR #320, opened 2021-07-20
Allow non-monontonic and empty sums on Histogram data points.

### LINKED ISSUE #303: Allow `sum` to present with negative measurements in Histogram
Follow on to #187 - We currently limit the sum field in Histogram to be 100% compatible with OpenMetrics/Prometheus histograms so we know we can do a conversion when it exists.   However, we think there's value to having sum available when there are negative measurements as well.

Tentative proposal:

```
message Histogram {.
  .. 
  // When this is true, a histogram allows negative measurements.  This means
  // the sum is non monotonic, and may be 0 even if there are recorded events.
  bool has_negative_measurements = ?;
}
```

jsuereth:
Fixes #303

- Add a new enumeration to Histogram denoting the type of sum for that histogram.
- Update documentation to denote this new enumeration is used for interpretataion of the sum field.
- Default the value of this enumeration to "monotonic sum" for backwards compatiblity.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -219,6 +219,18 @@ message Histogram {
   // aggregation_temporality describes if the aggregator reports delta changes
   // since last report time, or cumulative changes since a fixed start time.
   AggregationTemporality aggregation_temporality = 2;
+
+  enum SumType {
+    // The histogram sum is a monotonic value and will not decrease.
+    // Note: This is the default for histograms if unspecified.
+    MONOTONIC = 0;
+    // The histogram sum may increasee or decrease.
+    NON_MONOTONIC = 1;
+    // The histogram sum is not meaningful and should be ignored.
+    UNKNOWN_SUM = 2;
+  }
+  // How to interpret the sum field on histogram data points.
+  SumType sum_type = 3;
 }
 
 // Summary metric data are used to convey quantile summaries,
@@ -393,13 +405,10 @@ message HistogramDataPoint {
 
   // sum of the values in the population. If count is zero then this field
   // must be zero. This value must be equal to the sum of the "sum" fields in
-  // buckets if a histogram is provided.
+  // buckets of a histogram if provided.
   //
-  // Note: Sum should only be filled out when measuring non-negative discrete
-  // events, and is assumed to be monotonic over the values of these events.
-  // Negative events *can* be recorded, but sum should not be filled out when
-  // doing so.  This is specifically to enforce compatibility w/ OpenMetrics,
-  // see: https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#histogram
+  // Additionally, this field should abide by the interpretation specified
+  // in Histogram.sum_type.  
   double sum = 5;
 
   // bucket_counts is an optional field contains the count values of histogram
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Full list of differences found in [this compare.](https://github.com/open-teleme
 
 ### Added
 
-* Remove if no changes for this section before release.
+* Added an enum to Histogram to denote the meaning of the sum field.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Full list of differences found in [this compare.](https://github.com/open-teleme
 
 ### Added
 
-* Added an enum to Histogram to denote the meaning of the sum field.
+* Added an enum to Histogram to denote the meaning of the sum field. (#320)
 
 ### Removed
 


jsuereth:
FYI - I think this is what we discussed in the last metrics SiG: @jmacd @victlu @bogdandrutu

jmacd (approved):

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -219,6 +219,18 @@ message Histogram {
   // aggregation_temporality describes if the aggregator reports delta changes
   // since last report time, or cumulative changes since a fixed start time.
   AggregationTemporality aggregation_temporality = 2;
+
+  enum SumType {
+    // The histogram sum is a monotonic value and will not decrease.
+    // Note: This is the default for histograms if unspecified.
+    MONOTONIC = 0;
```
victlu, line 226: Should we avoid 0 as we may not know if it's set or missing?

Maybe...
NONE = 0
MONOTONIC = 1
NON_MONOTONIC = 2
UNKNOWN_SUM = 3

victlu (approved):

opentelemetry/proto/metrics/v1/metrics.proto
jsuereth: Backwards-commpatiblity means previous users would NOT be specifying this value, which defaults to 0.  So we should treat that as MONOTONIC.

cijothomas (approved):

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: Is it that before we had sum as monotonic?

jsuereth: Yes, sum was monotonic before

bogdandrutu: From here, that is not clear:

```
```

bogdandrutu (commented):
Instead of making "sum" special in terms of having an extra metadata, have we consider the following:
1. Deprecating the current "sum" - keep it for backwards compatibility for 1.5y as we did for StatusCode in trace.
2. Add a new sum with the type "google.protobuf.DoubleValue"?

The advantage is that this is a point scope property not a metric scope property? Is this important?

opentelemetry/proto/metrics/v1/metrics.proto
jsuereth: See the linked bug.  Basically we can't *not* fill out a sum, so due to the phrasing, we're always providing a sum and it's always monotonic.

jsuereth:
@bogdandrutu 
> Instead of making "sum" special in terms of having an extra metadata, have we consider the following:
> 1. Deprecating the current "sum" - keep it for backwards compatibility for 1.5y as we did for StatusCode in trace.

Didn't entertain removing sum completely.  AFAICT something like it is expected as a value here.   Care to comment on the viablity concerns to that approach?

> 2. Add a new sum with the type "google.protobuf.DoubleValue"?

This doesn't solve the issue of understanding if the sum is monotonic or non-monotonic (which lead to the existing wonky wording we have).   We'd like to keep sum, we just also need to know how to export it to prometheus so we need some kind of monotonicity marker.

> The advantage is that this is a point scope property not a metric scope property? Is this important?

@jmacd pushed for the semantic of sum to be a metric-scope property vs. point scope (which was an earlier proposal).   I could go either way myself, and see enough flexibility in metric-scope (given InstrumentationLibrary) that I think we're ok here.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -214,6 +214,18 @@ message Histogram {
...
+    // The histogram sum may increasee or decrease.
```
reyang, line 222:
```suggestion
    // The histogram sum may increase or decrease.
```

bogdandrutu:
@jsuereth 

> @jmacd pushed for the semantic of sum to be a metric-scope property vs. point scope (which was an earlier proposal). I could go either way myself, and see enough flexibility in metric-scope (given InstrumentationLibrary) that I think we're ok here.

For me this seems to be inconsistent with the flags. Why is a difference there?

jmacd:
In #316, the one new `DataPointFlags` value is about a condition that is specific to the point.

If there is a distinction to be made about the monotonicity of the Sum, then it appears to be a metric-level property, since it relates to the difference between points. I can't imagine what it means to have a stream of points where some declare their sum to be monotonic and some do not--those appear to be different streams, so I think the distinction belongs in the metric not the point.

I have in the past attempted to optimize the number of fields in these protocols by taking a cross-product of existing enums. We already have an `AggregationTemporality` field with two values, and now a `SumType` field with three values, (and I can imagine one more), so we could combine them as follows:

```
enum HistogramType {
  Unspecified = 0
  DeltaMonotonicSum = 1 // Backwards => AggregationTemporality_Delta w/ monotonic
  CumulativeMonotonicSum = 2 // Backwards => AggregationTemporality_Cumulative w/ monotonic
  DeltaUpDownSum = 3
  DeltaGaugeSum = 4
  CumulativeUpDownSum = 5
  CumulativeGaugeSum = 6
}
```

@bogdandrutu would that alleviate your concern about creating a new field?

jsuereth:
Going to propose we migrate to optional in proto, with justifcation.  PR pending.

PR #322, opened 2021-07-23
Base-2 exponential histogram protocol support

jmacd:
Adds a base-2 exponential histogram to OTLP metrics.

This protocol is copied from the OpenMetrics/Prometheus draft created by @beorn7, with comments written by me to summarize discussion in https://github.com/open-telemetry/opentelemetry-specification/issues/1776 (particularly around the meaning of the zero bucket), following the design effort in [OTEP 149](https://github.com/open-telemetry/oteps/pull/149) led by @yzhuge. 

Note: Past debate and discussion in the OTel Metrics SIG has led us to use different types for different kinds of histogram, as opposed to burying a new `oneof` in the existing types. This proposal follows that decision, and I would prefer not to revisit that topic. The new ExponentialHistogram and ExponentialHistogramDataPoint types mimic the existing explicit histogram types for most of their fields, including attributes, timestamps, sum, count, and exemplars fields. As such, the real review here is on lines 483 onward, starting from the `scale` field.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -447,39 +447,11 @@ message HistogramDataPoint {
 // distribution of those values across a set of buckets.
 //
 message ExponentialHistogramDataPoint {
-  // The set of key/value pairs that uniquely identify the timeseries from
-  // where this point belongs. The list may be empty (may contain 0 elements).
-  repeated opentelemetry.proto.common.v1.KeyValue attributes = 1;
-
-  // StartTimeUnixNano is optional but strongly encouraged, see the
-  // the detiled comments above Metric.
-  //
-  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
-  // 1970.
-  fixed64 start_time_unix_nano = 2;
-
-  // TimeUnixNano is required, see the detailed comments above Metric.
-  //
-  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
-  // 1970.
-  fixed64 time_unix_nano = 3;
-
-  // count is the number of values in the population. Must be non-negative. This
-  // value must be equal to the sum of the "count" fields in buckets if a
-  // histogram is provided.
-  fixed64 count = 4;
-
-  // sum of the values in the population. If count is zero then this field
-  // must be zero. This value must be equal to the sum of the "sum" fields in
-  // buckets if a histogram is provided.
-  //
-  // Note: Sum should only be filled out when measuring non-negative discrete
-  // events, and is assumed to be monotonic over the values of these events.
-  // Negative events *can* be recorded, but sum should not be filled out when
-  // doing so.  This is specifically to enforce compatibility w/ OpenMetrics,
-  // see: https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#histogram
-  double sum = 5;
-
+  // T.B.D. The Attributes, Timestamps, Flags, Count, and Sum fields will
+  // be copied from HistohgramDataPoint before merging.  These fields arte
+  // identical between these two types, and both have pending changes in the
+  // opentelemetry-proto repository.
+  
   // scale describes the resolution of the histogram.  Boundaries are
   // located at powers of the base, where:
   //
@@ -509,66 +481,35 @@ message ExponentialHistogramDataPoint {
   // mass equal to (zero_count / count).
   fixed64 zero_count = 7;
 
-  // zero_tolerance may be optionally set to convey the width of the
-  // zero region.  When this is not set, implementations may assume
-  // the zero tolerance equals the smallest representable positive
-  // IEEE 754 double-precision floating point number, which is (2^−1074).
-  //
-  // Producers can be configured to round measurements to zero that fall
-  // below the zero tolerance.  Users should avoid producing measurements
-  // that have meaningless precision near zero, as exponential buckets
-  // become increasingly dense in this region.
-  //
-  // Implementations MAY consider the zero bucket to have probability
-  // density equal to (zero_count / 2*zero_tolerance), if set.
-  double zero_tolerance = 8;
-
   // positive carries the positive range of exponential bucket counts.
-  SparseBuckets positive = 9;
+  Span positive = 8;
 
   // negative carries the negative range of exponential bucket counts.
-  SparseBuckets negative = 10;
+  Span negative = 9;
 
-  // SparseBuckets are a sparse set of bucket counts, compactly encoded.
-  message SparseBuckets {
-    // Span encodes a run of contiguous histogram buckets.
-    message Span {
-      sint32 offset = 1; // Gap to previous span, or starting point for 1st span (which can be negative).
-      uint32 length = 2; // Length of consecutive buckets.
-    }
+  // Buckets are a set of bucket counts, encoded in a contiguous array
+  // of counts.
+  message Span {
+    // Offset is the index of the first count expressed in this span.
+    // 
+    // Note: This uses a varint encoding as a simple form of compression.
+    sint64 offset = 1;
 
-    // span is a repeated list of Spans that encodes a range of
-    // exponential buckets, positive or negative.  The offset of the
-    // first Span indicates the smallest index in the Histogram with
-    // non-zero count.  Subsequent contiguous buckets will be included
-    // in the same Span, and subsequent Spans encode additional
-    // buckets separated by gaps where the histogram has zero count.
+    // Count is an array of counts, where count[i] carries the count
+    // of the bucket at index (offset+i), i.e., count[i] is the count
+    // of values less than or equal to base^(offset+i) and greater
+    // than base^(offset+i-1).
     //
-    // The starting index of a Span equals its own offset plus the
-    // starting index of the previous Span plus the previous Span's
-    // length.  As a recursive formula:
-    //
-    //    starting_index(span[x+1]) =
-    //       starting_index(span[x]) +
-    //       span[x+1].offset +
-    //       span[x].length
-    //
-    // For example, the two Spans [ {offset: 10, length: 3}, {offset:
-    // 5, length: 8} ] indicate a total of 11 buckets with indices at
-    // [10, 11, 12] and [18, 19, 20, 21, 22, 23, 24, 25].
-    repeated Span span = 1;
-
-    // delta is an array of counts corresponding to the buckets
-    // described by `span`, encoded as deltas.  The indicated count 
-    // for bucket[x] is defined as a recursive formula:
-    //
-    //    bucket_count(x+1) = bucket_count(x) + delta[x+1]
-    repeated sint64 delta = 2;
+    // Note: By contrast, the explicit HistogramDataPoint uses
+    // fixedt64.  This field is expected to have many buckets,
+    // especially zeros, so has been selected to ensure varint
+    // encoding.
+    repeated uint64 bucket_counts = 6;
   } 
 
   // (Optional) List of exemplars collected from
   // measurements that were used to form the data point
-  repeated Exemplar exemplars = 11;
+  repeated Exemplar exemplars = 10;
 }
 
 // SummaryDataPoint is a single data point in a timeseries that describes the
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -176,7 +176,7 @@ message Metric {
     // This field will be removed in ~3 months, on July 1, 2021.
     IntHistogram int_histogram = 8 [deprecated = true];
     Histogram histogram = 9;
-    ExponentialHistogram exponential_histogram = 12;
+    ExponentialHistogram exponential_histogram = 10;
     Summary summary = 11;
   }
 }
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -534,7 +534,7 @@ message ExponentialHistogramDataPoint {
   // Buckets are a set of bucket counts, encoded in a contiguous array
   // of counts.
   message Buckets {
-    // Offset is the index of the first count expressed in this span.
+    // Offset is the bucket index of the first entry in the bucket_counts array.
     // 
     // Note: This uses a varint encoding as a simple form of compression.
     sint32 offset = 1;
@@ -545,9 +545,9 @@ message ExponentialHistogramDataPoint {
     // base^(offset+i+1).
     //
     // Note: By contrast, the explicit HistogramDataPoint uses
-    // fixedt64.  This field is expected to have many buckets,
-    // especially zeros, so has been selected to ensure varint
-    // encoding.
+    // fixed64.  This field is expected to have many buckets,
+    // especially zeros, so uint64 has been selected to ensure
+    // varint encoding.
     repeated uint64 bucket_counts = 2;
   } 
 
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -416,8 +416,7 @@ message HistogramDataPoint {
   fixed64 count = 4;
 
   // sum of the values in the population. If count is zero then this field
-  // must be zero. This value must be equal to the sum of the "sum" fields in
-  // buckets if a histogram is provided.
+  // must be zero.
   //
   // Note: Sum should only be filled out when measuring non-negative discrete
   // events, and is assumed to be monotonic over the values of these events.
@@ -482,14 +481,13 @@ message ExponentialHistogramDataPoint {
   // 1970.
   fixed64 time_unix_nano = 3;
 
-  // count is the number of values in the population. Must be non-negative. This
-  // value must be equal to the sum of the "count" fields in buckets if a
-  // histogram is provided.
+  // count is the number of values in the population. Must be
+  // non-negative. This value must be equal to the sum of the "bucket_counts"
+  // values in the positive and negative Buckets plus the "zero_count" field.
   fixed64 count = 4;
 
   // sum of the values in the population. If count is zero then this field
-  // must be zero. This value must be equal to the sum of the "sum" fields in
-  // buckets if a histogram is provided.
+  // must be zero.
   //
   // Note: Sum should only be filled out when measuring non-negative discrete
   // events, and is assumed to be monotonic over the values of these events.
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -527,6 +527,10 @@ message ExponentialHistogramDataPoint {
   // (Optional) List of exemplars collected from
   // measurements that were used to form the data point
   repeated Exemplar exemplars = 10;
+
+  // Flags that apply to this specific data point.  See DataPointFlags
+  // for the available flags and their meaning.
+  uint32 flags = 11;
 }
 
 // SummaryDataPoint is a single data point in a timeseries that describes the
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -566,7 +566,7 @@ message ExponentialHistogramDataPoint {
 
   // (Optional) List of exemplars collected from
   // measurements that were used to form the data point
-  repeated Exemplar exemplars = 8;
+  repeated Exemplar exemplars = 11;
 }
 
 // SummaryDataPoint is a single data point in a timeseries that describes the
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -490,9 +490,11 @@ message ExponentialHistogramDataPoint {
   // greater than (base^(index-1)).
   //
   // The positive and negative ranges of the histogram are expressed
-  // separately but both use the same scale to locate their boundaries.
+  // separately.  Negative values are mapped by their absolute value
+  // into the negative range using the same scale as the positive range.
   //
-  // scale is valid in the range -4 through +8 inclusive.
+  // scale is not restricted by the protocol, as the permissible
+  // values depend on the range of the data.
   sint32 scale = 6;
 
   // zero_count is the count of values that are either exactly zero or
@@ -504,7 +506,7 @@ message ExponentialHistogramDataPoint {
   // the width of the zero region.
   //
   // Implementations MAY consider the zero bucket to have probability
-  // mass equal to (zero_count / sum).
+  // mass equal to (zero_count / count).
   fixed64 zero_count = 7;
 
   // zero_tolerance may be optionally set to convey the width of the
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -464,10 +464,10 @@ message HistogramDataPoint {
 // distribution of those values across a set of buckets.
 //
 message ExponentialHistogramDataPoint {
-  // T.B.D. The Attributes, Timestamps, Flags, Count, and Sum fields will
-  // be copied from HistohgramDataPoint before merging.  These fields arte
-  // identical between these two types, and both have pending changes in the
-  // opentelemetry-proto repository.
+  // T.B.D. The Attributes, Timestamps, Count, and Sum fields will be
+  // copied with their comments from HistohgramDataPoint before
+  // merging.  These fields arte identical between these two types,
+  // and have a pending change in the opentelemetry-proto repository.
   
   // scale describes the resolution of the histogram.  Boundaries are
   // located at powers of the base, where:
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -344,7 +344,7 @@ message NumberDataPoint {
   repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1 [deprecated = true];
 
   // StartTimeUnixNano is optional but strongly encouraged, see the
-  // the detiled comments above Metric.
+  // the detailed comments above Metric.
   //
   // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
   // 1970.
@@ -398,7 +398,7 @@ message HistogramDataPoint {
   repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1 [deprecated = true];
 
   // StartTimeUnixNano is optional but strongly encouraged, see the
-  // the detiled comments above Metric.
+  // the detailed comments above Metric.
   //
   // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
   // 1970.
@@ -469,7 +469,7 @@ message ExponentialHistogramDataPoint {
   repeated opentelemetry.proto.common.v1.KeyValue attributes = 1;
 
   // StartTimeUnixNano is optional but strongly encouraged, see the
-  // the detiled comments above Metric.
+  // the detailed comments above Metric.
   //
   // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
   // 1970.
@@ -576,7 +576,7 @@ message SummaryDataPoint {
   repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1 [deprecated = true];
 
   // StartTimeUnixNano is optional but strongly encouraged, see the
-  // the detiled comments above Metric.
+  // the detailed comments above Metric.
   //
   // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
   // 1970.
@@ -687,7 +687,7 @@ message IntDataPoint {
   repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1;
 
   // StartTimeUnixNano is optional but strongly encouraged, see the
-  // the detiled comments above Metric.
+  // the detailed comments above Metric.
   //
   // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
   // 1970.
@@ -736,7 +736,7 @@ message IntHistogramDataPoint {
   repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1;
 
   // StartTimeUnixNano is optional but strongly encouraged, see the
-  // the detiled comments above Metric.
+  // the detailed comments above Metric.
   //
   // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
   // 1970.
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -413,12 +413,11 @@ message HistogramDataPoint {
 
   // explicit_bounds specifies buckets with explicitly defined bounds for values.
   //
-  // This defines size(explicit_bounds) + 1 (= N) buckets. The boundaries for
-  // bucket at index i are:
+  // The boundaries for bucket at index i are:
   //
   // (-infinity, explicit_bounds[i]] for i == 0
-  // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < N-1
-  // (explicit_bounds[i], +infinity) for i == N-1
+  // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < size(explicit_bounds)
+  // (explicit_bounds[i-1], +infinity) for i == size(explicit_bounds)
   //
   // The values in the explicit_bounds array must be strictly increasing.
   //
@@ -641,12 +640,11 @@ message IntHistogramDataPoint {
 
   // explicit_bounds specifies buckets with explicitly defined bounds for values.
   //
-  // This defines size(explicit_bounds) + 1 (= N) buckets. The boundaries for
-  // bucket at index i are:
+  // The boundaries for bucket at index i are:
   //
   // (-infinity, explicit_bounds[i]] for i == 0
-  // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < N-1
-  // (explicit_bounds[i], +infinity) for i == N-1
+  // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < size(explicit_bounds)
+  // (explicit_bounds[i-1], +infinity) for i == size(explicit_bounds)
   //
   // The values in the explicit_bounds array must be strictly increasing.
   //
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -180,22 +180,7 @@ message Metric {
   }
 }
 
-// IntGauge is deprecated.  Use Gauge with an integer value in NumberDataPoint.
-//
-// IntGauge represents the type of a int scalar metric that always exports the
-// "current value" for every data point. It should be used for an "unknown"
-// aggregation.
-//
-// A Gauge does not support different aggregation temporalities. Given the
-// aggregation is unknown, points cannot be combined using the same
-// aggregation, regardless of aggregation temporalities. Therefore,
-// AggregationTemporality is not included. Consequently, this also means
-// "StartTimeUnixNano" is ignored for all data points.
-message IntGauge {
-  option deprecated = true;
 
-  repeated IntDataPoint data_points = 1;
-}
 
 // Gauge represents the type of a double scalar metric that always exports the
 // "current value" for every data point. It should be used for an "unknown"
@@ -210,22 +195,6 @@ message Gauge {
   repeated NumberDataPoint data_points = 1;
 }
 
-// IntSum is deprecated.  Use Sum with an integer value in NumberDataPoint.
-//
-// IntSum represents the type of a numeric int scalar metric that is calculated as
-// a sum of all reported measurements over a time interval.
-message IntSum {
-  option deprecated = true;
-
-  repeated IntDataPoint data_points = 1;
-
-  // aggregation_temporality describes if the aggregator reports delta changes
-  // since last report time, or cumulative changes since a fixed start time.
-  AggregationTemporality aggregation_temporality = 2;
-
-  // If "true" means that the sum is monotonic.
-  bool is_monotonic = 3;
-}
 
 // Sum represents the type of a numeric double scalar metric that is calculated
 // as a sum of all reported measurements over a time interval.
@@ -240,20 +209,7 @@ message Sum {
   bool is_monotonic = 3;
 }
 
-// IntHistogram is deprecated, replaced by Histogram points using double-
-// valued exemplars.
-//
-// This represents the type of a metric that is calculated by aggregating as a
-// Histogram of all reported int measurements over a time interval.
-message IntHistogram {
-  option deprecated = true;
 
-  repeated IntHistogramDataPoint data_points = 1;
-
-  // aggregation_temporality describes if the aggregator reports delta changes
-  // since last report time, or cumulative changes since a fixed start time.
-  AggregationTemporality aggregation_temporality = 2;
-}
 
 // Histogram represents the type of a metric that is calculated by aggregating
 // as a Histogram of all reported double measurements over a time interval.
@@ -346,34 +302,7 @@ enum AggregationTemporality {
   AGGREGATION_TEMPORALITY_CUMULATIVE = 2;
 }
 
-// IntDataPoint is a single data point in a timeseries that describes the
-// time-varying values of a int64 metric.
-message IntDataPoint {
-  option deprecated = true;
 
-  // The set of labels that uniquely identify this timeseries.
-  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1;
-
-  // StartTimeUnixNano is optional but strongly encouraged, see the
-  // the detiled comments above Metric.
-  //
-  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
-  // 1970.
-  fixed64 start_time_unix_nano = 2;
-
-  // TimeUnixNano is required, see the detailed comments above Metric.
-  //
-  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
-  // 1970.
-  fixed64 time_unix_nano = 3;
-
-  // value itself.
-  sfixed64 value = 4;
-
-  // (Optional) List of exemplars collected from
-  // measurements that were used to form the data point
-  repeated IntExemplar exemplars = 5;
-}
 
 // NumberDataPoint is a single data point in a timeseries that describes the
 // time-varying value of a double metric.
@@ -417,76 +346,7 @@ message NumberDataPoint {
   repeated Exemplar exemplars = 5;
 }
 
-// IntHistogramDataPoint is deprecated; use HistogramDataPoint.
-//
-// This is a single data point in a timeseries that describes
-// the time-varying values of a Histogram of int values. A Histogram contains
-// summary statistics for a population of values, it may optionally contain
-// the distribution of those values across a set of buckets.
-//
-// If the histogram contains the distribution of values, then both
-// "explicit_bounds" and "bucket counts" fields must be defined.
-// If the histogram does not contain the distribution of values, then both
-// "explicit_bounds" and "bucket_counts" must be omitted and only "count" and
-// "sum" are known.
-message IntHistogramDataPoint {
-  option deprecated = true;
 
-  // The set of labels that uniquely identify this timeseries.
-  repeated opentelemetry.proto.common.v1.StringKeyValue labels = 1;
-
-  // StartTimeUnixNano is optional but strongly encouraged, see the
-  // the detiled comments above Metric.
-  //
-  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
-  // 1970.
-  fixed64 start_time_unix_nano = 2;
-
-  // TimeUnixNano is required, see the detailed comments above Metric.
-  //
-  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
-  // 1970.
-  fixed64 time_unix_nano = 3;
-
-  // count is the number of values in the population. Must be non-negative. This
-  // value must be equal to the sum of the "count" fields in buckets if a
-  // histogram is provided.
-  fixed64 count = 4;
-
-  // sum of the values in the population. If count is zero then this field
-  // must be zero. This value must be equal to the sum of the "sum" fields in
-  // buckets if a histogram is provided.
-  sfixed64 sum = 5;
-
-  // bucket_counts is an optional field contains the count values of histogram
-  // for each bucket.
-  //
-  // The sum of the bucket_counts must equal the value in the count field.
-  //
-  // The number of elements in bucket_counts array must be by one greater than
-  // the number of elements in explicit_bounds array.
-  repeated fixed64 bucket_counts = 6;
-
-  // explicit_bounds specifies buckets with explicitly defined bounds for values.
-  //
-  // This defines size(explicit_bounds) + 1 (= N) buckets. The boundaries for
-  // bucket at index i are:
-  //
-  // (-infinity, explicit_bounds[i]] for i == 0
-  // (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < N-1
-  // (explicit_bounds[i], +infinity) for i == N-1
-  //
-  // The values in the explicit_bounds array must be strictly increasing.
-  //
-  // Histogram buckets are inclusive of their upper boundary, except the last
-  // bucket where the boundary is at infinity. This format is intentionally
-  // compatible with the OpenMetrics histogram definition.
-  repeated double explicit_bounds = 7;
-
-  // (Optional) List of exemplars collected from
-  // measurements that were used to form the data point
-  repeated IntExemplar exemplars = 8;
-}
 
 // HistogramDataPoint is a single data point in a timeseries that describes the
 // time-varying values of a Histogram of double values. A Histogram contains
@@ -639,37 +499,7 @@ message SummaryDataPoint {
   repeated ValueAtQuantile quantile_values = 6;
 }
 
-// A representation of an exemplar, which is a sample input int measurement.
-// Exemplars also hold information about the environment when the measurement
-// was recorded, for example the span and trace ID of the active span when the
-// exemplar was recorded.
-message IntExemplar {
-  option deprecated = true;
 
-  // The set of labels that were filtered out by the aggregator, but recorded
-  // alongside the original measurement. Only labels that were filtered out
-  // by the aggregator should be included
-  repeated opentelemetry.proto.common.v1.StringKeyValue filtered_labels = 1;
-
-  // time_unix_nano is the exact time when this exemplar was recorded
-  //
-  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
-  // 1970.
-  fixed64 time_unix_nano = 2;
-
-  // Numerical int value of the measurement that was recorded.
-  sfixed64 value = 3;
-
-  // (Optional) Span ID of the exemplar trace.
-  // span_id may be missing if the measurement is not recorded inside a trace
-  // or if the trace is not sampled.
-  bytes span_id = 4;
-
-  // (Optional) Trace ID of the exemplar trace.
-  // trace_id may be missing if the measurement is not recorded inside a trace
-  // or if the trace is not sampled.
-  bytes trace_id = 5;
-}
 
 // A representation of an exemplar, which is a sample input measurement.
 // Exemplars also hold information about the environment when the measurement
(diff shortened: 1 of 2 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -431,6 +441,134 @@ message HistogramDataPoint {
...
+    // described by `span`, encoded as deltas.  The indicated count 
+    // for bucket[x] is defined as a recursive formula:
+    //
+    //    bucket_count(x+1) = bucket_count(x) + delta[x+1]
+    repeated sint64 delta = 2;
+  } 
+
+  // (Optional) List of exemplars collected from
+  // measurements that were used to form the data point
+  repeated Exemplar exemplars = 8;
```
jsuereth, line 569: This should be something other than 8 (probably 11 if you're going in order)

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -431,6 +441,134 @@ message HistogramDataPoint {
...
+  // count is the number of values in the population. Must be non-negative. This
+  // value must be equal to the sum of the "count" fields in buckets if a
+  // histogram is provided.
+  fixed64 count = 4;
+
+  // sum of the values in the population. If count is zero then this field
+  // must be zero. This value must be equal to the sum of the "sum" fields in
+  // buckets if a histogram is provided.
+  //
+  // Note: Sum should only be filled out when measuring non-negative discrete
```
jsuereth, line 476: We know this line is false (if the field is not filled out it we default to 0).  I'd suggest adding a bit about a value of 0 represents an unknown sum and cannot be interpreted as the actual sum.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -431,6 +441,134 @@ message HistogramDataPoint {
...
+  // events, and is assumed to be monotonic over the values of these events.
+  // Negative events *can* be recorded, but sum should not be filled out when
+  // doing so.  This is specifically to enforce compatibility w/ OpenMetrics,
+  // see: https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#histogram
```
beorn7, line 480: Note that the OM spec cited here is for the conventional histograms (and the requirement of monotonic increase comes from Prometheus's habit of treating this series as an independent counter).

With the new sparse high-resolution histograms, Prometheus has to look at a histogram sample as a whole (rather than as a set of independent time series), so I don't expect that this annoying restrictions will make it into the new histograms.

jmacd:
@jsuereth and @beorn7 thank you. About the `sum` field, I copied the existing Histogram data types and modified the bucketing strategy, so this `sum` is the same as `HistogramDataPoint`. If #320 merges first, this PR would be updated to again be a copy of `HistogramDataPoint` with the same comment about `sum`.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: https://github.com/open-telemetry/opentelemetry-proto/pull/322#issuecomment-887103548

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: See https://github.com/open-telemetry/opentelemetry-proto/pull/322#issuecomment-887103548

yzhuge:
Besides my in-line comments. I have a major one: 

I am not convinced of the value of the sparse array encoding. It adds a lot of complexity. The counter array is separate from the span list. So it is easy to misalign in encoding or decoding. Adding delta encoding on counters just adds another layer of complexity. Implementations will be error prone and more expensive than simple array (basically one span, no delta encoding). Having no other encoding options requires all producers and consumers to implement this encoding/decoding scheme. I feel this is undue burden. 

For array compaction, there is always the trade off between cpu, complexity and space. I remember in previous discussions, some people are even concerned on the cpu cost of using varint (Protobuf default encoding of integer arrays) and favor "repeated fixed64 bucket_counts", which is the current spec for explicit bucket histograms. 

For final on-disk storage, heavy compression at the cost of cpu may be worth it. The Prometheus/OM sparse array appears like such a scheme. For intermediate result and transportation, we need to balance cpu and space. In fact, for streaming, where the processor typically holds one value at a time, cpu (ie. throughput) maybe favored over space. In streaming, an object may need to pass though multiple stages of processing, going through deserialization and re-serialization at each stage. In such cases, reducing cpu cost is even more critical. 

My preference for OTel bucket counters is to use simple array of "repeated sint64 bucket_counts" (one span, no delta, varint). We get "varint" for free from Protobuf. In NrSketch (https://github.com/newrelic-experimental/newrelic-sketch-java), the default 320 buckets works very well in New Relic production as a "one size fits all" size. Our serialization uses variable bytes per bucket, though still requires same number of bytes on all buckets. It is not even as aggressive as varint on individual bucket. In practice, typical serialized size is a few hundred bytes. At this level of size, fancy"sparse array" compression is not cost (cpu and complexity) effective. 

In the spirit of keeping it simple, I'd like to see simple arrays in the OTel standard first. There are many sparse array encoding methods. In the future, if we do go that way, we have to be very careful choosing one.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -431,6 +441,134 @@ message HistogramDataPoint {
...
+  double sum = 5;
+
+  // scale describes the resolution of the histogram.  Boundaries are
+  // located at powers of the base, where:
+  //
+  //   base = (2^(2^-scale))
+  //
+  // The histogram bucket identified by `index`, a signed integer,
+  // contains values that are less than or equal to (base^index) and
```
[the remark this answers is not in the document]
beorn7: I don't follow the argument of "more intuitive for humans". In fact, I see it the other way around.

The computational argument is minor and shouldn't tip the balance. Positive scales will be the exception, and even then, the actual difference in computational or algorithmic complexity doesn't really make a dent.

More generally (and along the lines I have explained before), this is at a level of detail that I don't feel qualified to discuss while we are still experimenting to get the basics right. Little knobs like this can easily be tweaked later in the process, based in the insights gained.

Having said that, if I had to take a guess, I would say it's likely Prometheus will stick to the "less or equal" upper bounds because it is consistent with Prometheus's existing histograms so that creating compatibility mappings is easier. There needed to be stronger arguments than "I save a single `if frac == 0.5 { sparseKey-- }` line in the code in the rare case of very low resolution histograms" to tip the balance.

Finally, to avoid misunderstandings: Everything you see in a "sparsehistogram" branch in a Prometheus repo is part of our experiment. There is no direct link to OpenMetrics, i.e. this is not part of some kind of "OpenMetrics 2.0 draft". First these experiments have to be done and insights have to be gained. Only then can we think about the actual implementation in a future Prometheus release, which will in turn inform how OpenMetrics will embrace sparse high-resolution histograms.

jmacd:
> In the spirit of keeping it simple, I'd like to see simple arrays in the OTel standard first.

You wrote a very convincing explanation, and you have my support. I will update this PR, unless you would like to take over @yzhuge? The key point is that we have to convert losslessly into the OpenMetrics protocol, and it looks like this just means agreeing to an optional width for the zero bucket.

yzhuge:
There is another decision to make before we proceed: 
Whether base^index is the lower or upper bound of the bucket at “index”.

The discussion started at https://github.com/open-telemetry/opentelemetry-proto/pull/322#discussion_r677865687. I am continuing here so that it doesn’t get buried or lost in the in-line discussion. I favor the lower bound option. It has the pros:
- It is mathematically cleaner. The mathematical neutral point of index space (0) and log space (1) coincide perfectly. Bucket 0 and above map to values 1 and above; negative buckets map to values less than 1. Bucket 0 covers [1, base), rather than (1/base, 1]. 
- Both [NrSketch](https://github.com/newrelic-experimental/newrelic-sketch-java/blob/main/src/main/java/com/newrelic/nrsketch/indexer/LogIndexer.java#L22) and [DDSketch](https://github.com/DataDog/sketches-java/blob/master/src/main/java/com/datadoghq/sketch/ddsketch/mapping/LogLikeIndexMapping.java#L82) use lower bound
- It is often simpler to code. "Round down" via right shifting an integer, or casting a floating point number to integer is straightforward.  The reduction in code complexity (therefore less chance of a bug) is more important here than the runtime gain of one or a few instructions. 

Now the cons:
- Different from the "le" upper bound convention in Prometheus explicit histograms. Note that the Prometheus exponential histogram project is still going on. They do not have a final decision on upper vs lower bound. If OTel follows the older Prometheus convention, and newer Prometheus histogram goes the other way, it would be funny.

So I am suggesting that OTel go with lower bound. If we have to do "off by one" conversion from Prometheus index, no big deal. We need to look into the future more than the past.

yzhuge:
A minor point: bucket index needs to be 64 bit int. At higher scale, index can go beyond 32 bit. 

"double" has 64 bit, therefore about 2^63 distinct values (roughly, as there are special code points for Inf, Nan, etc.) each in positive and negative space. It is perfectly valid to divide the values into more than 2^32 buckets.  As a standard, we need to handle extreme cases. 

Such high scale use cases are more likely the case where contrast (dynamic range) of dataset is very small, rather than somebody really needing .00001% relative error. For example, if dataset ranges from 1000 to 1000.1, with 1.0001 contrast, we really need high scales to have enough buckets to split the small range. 

In such use cases, the exponential buckets actually resemble linear buckets more. In the example above, the bucket at end of the range is only 1.0001 wider than the one at the beginning. This shows that although exponential histogram is designed for high dynamic range dataset, it can handle low dynamic range dataset too. 

For something like [NrSketch](https://github.com/newrelic-experimental/newrelic-sketch-java) with a high enough default initial scale, it can handle low range dataset with default config. Right now default initial scale is 12, at base of 1.00016, which cannot quite resolve a range of 1.0001. This made me think about raising the default. In New Relic production, default init scale is 32, which has a base of 1 + 1.6E-10. So it can handle ranges as low as 1 + 1E-8 out of the box. The cost of higher initial scale is that it has to downscale to the more common scale of 4 or 5 in most cases (ie. cost on common case so that we can cover extreme cases). But the number of downscaling is finite and small compare to the millions and billions of numbers inserted to a high throughput histogram.

jmacd:
>  So I am suggesting that OTel go with lower bound. If we have to do "off by one" conversion from Prometheus index, no big deal. We need to look into the future more than the past.

@beorn7 I'm curious what you think about this?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -431,6 +441,134 @@ message HistogramDataPoint {
...
+  // IEEE 754 double-precision floating point number, which is (2^−1074).
+  //
+  // Producers can be configured to round measurements to zero that fall
+  // below the zero tolerance.  Users should avoid producing measurements
+  // that have meaningless precision near zero, as exponential buckets
+  // become increasingly dense in this region.
+  //
+  // Implementations MAY consider the zero bucket to have probability
+  // density equal to (zero_count / 2*zero_tolerance), if set.
+  double zero_tolerance = 8;
```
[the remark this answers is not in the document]
oertl: I am also not sure if the `zero_tolerance` field is really needed. However, if it is needed, I would rather define it as the largest nonnegative double-precision value that maps to the zero bucket. (Currently, it is defined as the smallest nonnegative value that maps to a regular bucket. ) As a consequence, the default value would be 0 and not the "magic number" `2^-1074`.

This definition would also be more consistent as we would have exclusive lower bounds for all buckets. Consider the first regular bucket to which values may be mapped to. With the proposed definition of `zero_tolerance`, this bucket would contain all values from the interval `(zero_tolerance, base^i]` where `i` is the bucket index. Hence, we would have again an exclusive lower bound as for all other buckets which contain values from `(base^(i-1), base^i]`.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -431,6 +441,134 @@ message HistogramDataPoint {
...
+  SparseBuckets positive = 9;
+
+  // negative carries the negative range of exponential bucket counts.
+  SparseBuckets negative = 10;
+
+  // SparseBuckets are a sparse set of bucket counts, compactly encoded.
+  message SparseBuckets {
+    // Span encodes a run of contiguous histogram buckets.
+    message Span {
+      sint32 offset = 1; // Gap to previous span, or starting point for 1st span (which can be negative).
```
aabmass, line 534: Same comment as Yuke made below about varint, it would be worth calling out this is a delta encoding for compression.

jmacd:
I've reviewed all the feedback, and it doesn't seem like more is coming.

> I am not convinced of the value of the sparse array encoding.

I think I agree.

> Whether base^index is the lower or upper bound of the bucket at “index”.

I'm inclined to follow your guidance, @yzhuge.

> Different from the "le" upper bound convention in Prometheus explicit histograms.

We've repeatedly made statements that the boundary condition doesn't matter in theory or in practice, which is how we justified simply changing that OTLP definition to match Prometheus for explicit boundaries, which made sense. I think we can use the same argument to go the other way in this case. 😀 

> I would rather define it as the largest nonnegative double-precision value that maps to the zero bucket.

This makes sense! Thanks @oertl.

This implies a number of changes in this protocol, which I will gladly follow up on next week.

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
jmacd: @yzhuge I was on the fence about this issue -- whether the upper or lower bound should be inclusive -- until the comment from @oertl here: https://github.com/open-telemetry/opentelemetry-proto/pull/322/files/e680b5938be129ac8410f7ca66f0a50bcc4c4b52#r681484304

You have both recommended against the zero-tolerance field described here, and the remark by @oertl shows how it can be added later in a way that is backwards compatible with exclusive lower-bound buckets. I recommend that we avoid unnecessary conflict with the Prometheus group on this and keep the "less-than-or-equal" upper bound.

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
jmacd: I like the suggestion by @oertl that zero_tolerance could be defined so that the zero value (default) is an inclusive upper bound.

I've removed this field from the protocol because it can be added later, by this logic, at which time the can of worms described by @yzhuge will have to be addressed. To add this field we will need to pin down more detail about how to merge histograms with different zero tolerances, for example.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: I've accepted @yzhuge's larger point that allowing multiple Spans is a lot of added complexity for a transport protocol. I reworked the protocol to be simpler, but called out two places where we explicitly choose a varint encoding in the protocol, because of this concern.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -431,6 +441,134 @@ message HistogramDataPoint {
...
+      uint32 length = 2; // Length of consecutive buckets.
+    }
+
+    // span is a repeated list of Spans that encodes a range of
+    // exponential buckets, positive or negative.  The offset of the
+    // first Span indicates the smallest index in the Histogram with
+    // non-zero count.  Subsequent contiguous buckets will be included
+    // in the same Span, and subsequent Spans encode additional
+    // buckets separated by gaps where the histogram has zero count.
```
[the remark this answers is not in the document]
jmacd: I've removed this complexity on your recommendation.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -431,6 +441,134 @@ message HistogramDataPoint {
...
+    //
+    // The starting index of a Span equals its own offset plus the
+    // starting index of the previous Span plus the previous Span's
+    // length.  As a recursive formula:
+    //
+    //    starting_index(span[x+1]) =
```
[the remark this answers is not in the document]
jmacd: I've removed this complexity on your recommendation.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -431,6 +441,134 @@ message HistogramDataPoint {
...
+    // 5, length: 8} ] indicate a total of 11 buckets with indices at
+    // [10, 11, 12] and [18, 19, 20, 21, 22, 23, 24, 25].
+    repeated Span span = 1;
+
+    // delta is an array of counts corresponding to the buckets
+    // described by `span`, encoded as deltas.  The indicated count 
+    // for bucket[x] is defined as a recursive formula:
+    //
+    //    bucket_count(x+1) = bucket_count(x) + delta[x+1]
+    repeated sint64 delta = 2;
```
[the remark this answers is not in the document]
jmacd: (Done.)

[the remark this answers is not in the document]
jmacd: I've removed this complexity on your recommendation.

jmacd:
@open-telemetry/specs-metrics-approvers please take a look.

weyert (approved):

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -216,6 +216,16 @@ message Histogram {
   AggregationTemporality aggregation_temporality = 2;
 }
 
+// ExponentialHistogram represents the type of a metric that is calculated by aggregating
+// as a ExponentialHistogram of all reported double measurements over a time interval.
+message ExponentialHistogram {
```
jsuereth, line 221: Should this get added to `Metric.data` oneof above?

lquerel: +1

lquerel (commented):
LGTM once the `oneof` will be fixed.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: Done.

jmacd:
FYI I renamed "Span" to "Buckets" -- that was taken from the Prometheus prototype, confusing in an OTel context. I've begun modifying OTC to serve as a prototype for use in a metrics receiver. See https://github.com/open-telemetry/opentelemetry-collector/pull/3841 for the `pdata` additions.

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
yzhuge: First: "inclusiveness/exclusiveness is irrelevant in practice". See previous discussion at https://github.com/open-telemetry/oteps/pull/149#pullrequestreview-614123499 (click on "outdated" to show the comment "CharlesMasson reviewed on Mar 17"). We should avoid inclusiveness/exclusiveness constraint in OTel standard.

I don't see why bucket index (base^index as upper vs lower bound) is relevant to the zero region width. The indexed buckets are "regular" buckets that meet the bound = base ^ i formula. The zero region bucket is a special bucket for zero and values near zero. https://github.com/open-telemetry/opentelemetry-proto/pull/322/files/e680b5938be129ac8410f7ca66f0a50bcc4c4b52#r681484304 seems to be introducing a third category of bucket [zero_tolerance, lowest_regular_bound]. This will add complexity. There was a proposal of linear buckets near zero, followed by exponential buckets further out. For now, I'd like to just stick to a single zero-region bucket and a sequence of regular exponential buckets.

Previous version says: "  
  // zero_tolerance may be optionally set to convey the width of the
  // zero region.  When this is not set, implementations may assume
  // the zero tolerance equals the smallest representable positive
  // IEEE 754 double-precision floating point number, which is (2^−1074)."
  
This only "suggests" that default is 2^-1074 (smallest positive number "double" can represent). It does not define the field as "the smallest nonnegative value that maps to a regular bucket". When it is not set, most producers mean "I don't know" or "I don't care". The consumer using 2^-1074 is assuming the narrowest zero region, which may not may not be what the producer meant. The key issue with default and backward compatibility is that Protobuf default value is 0, but zero_tolerance default is "unknown". If we remap 0 to unknown, we need another value for logical 0. There are plenty of choices, like -1. But all are confusing when physical 0 is not logical 0.  

A better solution is to use an addition boolean field for "zero_tolerance_is_set". Booleans default to false in protobuf. So we know it defaults to "not set" (ie. "unknown"). When it is set, we use zero_tolerance value as is (including the protobuf default of 0) for the real zero_tolerance. Logically, this achieves a default of "Null".

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -448,6 +459,109 @@ message HistogramDataPoint {
...
+    sint64 offset = 1;
+
+    // Count is an array of counts, where count[i] carries the count
+    // of the bucket at index (offset+i), i.e., count[i] is the count
+    // of values less than or equal to base^(offset+i) and greater
+    // than base^(offset+i-1).
+    //
+    // Note: By contrast, the explicit HistogramDataPoint uses
+    // fixedt64.  This field is expected to have many buckets,
+    // especially zeros, so has been selected to ensure varint
```
yzhuge, line 551: Did you mean "so "uint64" has been selected to ensure varint encoding?"

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -448,6 +459,109 @@ message HistogramDataPoint {
...
+  // negative carries the negative range of exponential bucket counts.
+  Buckets negative = 9;
+
+  // Buckets are a set of bucket counts, encoded in a contiguous array
+  // of counts.
+  message Buckets {
+    // Offset is the index of the first count expressed in this span.
+    // 
+    // Note: This uses a varint encoding as a simple form of compression.
+    sint64 offset = 1;
```
yzhuge, line 542: I don't have a strong opinion on this one. Is it necessary to use varint here even though it is a single integer? Elsewhere we seem to favor fixed64

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -448,6 +459,109 @@ message HistogramDataPoint {
...
+
+    // Count is an array of counts, where count[i] carries the count
+    // of the bucket at index (offset+i), i.e., count[i] is the count
+    // of values less than or equal to base^(offset+i) and greater
+    // than base^(offset+i-1).
+    //
+    // Note: By contrast, the explicit HistogramDataPoint uses
+    // fixedt64.  This field is expected to have many buckets,
```
bogdandrutu, line 550: s/fixedt64/fixed64

jmacd:
In today's Metrics SIG, @bogdandrutu suggested that we could eliminate the `zero_count` field and postpone any discussion about adding it and the corresponding tolerance issue. The Histogram already has a total count field, and if we subtract all the positive and negative buckets from the total we are left with a count for the zero "region". When more development has been done, perhaps stating the rules for merging histograms with unequal zero bucket tolerances are specified, we can introduce fields to this data type provided backward compatibility.

opentelemetry/proto/metrics/v1/metrics.proto
weyert: nit: looks `fixedt64` is a typo for `fixed64`?

weyert:
> In today's Metrics SIG, @bogdandrutu suggested that we could eliminate the `zero_count` field and postpone any discussion about adding it and the corresponding tolerance issue. The Histogram already has a total count field, and if we subtract all the positive and negative buckets from the total we are left with a count for the zero "region". When more development has been done, perhaps stating the rules for merging histograms with unequal zero bucket tolerances are specified, we can introduce fields to this data type provided backward compatibility.

Sounds like a reasonable middle ground to progress with this issue

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
jsuereth: If we need to carry the explicit flag (see flags below) for "zero_tolerance_enabled", it's somewhat akin to what we proposed for handling monotonicity of sums in histograms.  I'd suggest a similar approach if we need it.

From our (Google) perspective, we have an explicit zero bucket that ends where the first exponential bucket starts, so we don't really need an explicit tolerance here.

jsuereth (approved):
Outside of the minor language nit brought up by @yzhuge on bucket offset, this LGTM.

yzhuge:
I lean toward keeping the zero_count field. Removing it does not eliminate the concept of zero region bucket. We still need to state that "total count - sum of all bucket counts" is the zero region count, and consumer will need to sum up bucket counts to get zero count. The main benefit of removing it is to save space in the message, at the cost of computing the field on the receiving side. 

The bigger question is whether base^index is the upper or lower bound of the bucket at index. I still strongly favor lower bound. I previously reasoned on this. In practice, both NrSketch and DDSketch use lower bound. Is there any implementation that actually use upper bound (the Prometheus project is still work in progress and they only say they are "likely" to use upper bound)?

"From our (Google) perspective, we have an explicit zero bucket that ends where the first exponential bucket starts". In this case, the zero region is [0, lowest_exponential_bound]. In the lower bound world, lowest_exponential_bound is conveniently base^Buckets.offset. In the upper bound world, it will be base^(Buckets.offset -1). This is one more place where lower bound is more intuitive. If we did go with upper bound, implementations are more likely to make "off by 1" errors because programmers don't expect such unusual pattern.

To map a value to an exponential bucket, the mathematical formula is log(base, value). Because log result is a float point, we need to convert it to an integer for bucket index. In computing, it is customary to round down to convert a float point to integer. This is the foundation of the lower bound approach (and why many implementations went with lower bound). If we deviate from the customary approach, we risk getting into an anti-pattern and having to do plus-one or minus-one compensation in many places.

oertl:
@yzhuge The discussion about the inclusiveness only makes sense if the boundaries are known exactly up to machine precision. If you are using the logarithm for the evaluation as you mentioned, it is very hard to implement a platform-independent mapping that is consistent with the boundaries as the logarithm will introduce machine-dependent numerical errors.

The nice thing about the proposed specification is that it is platform independent if it is interpreted as I do (@jmacd maybe the spec could be more clear on that). The proposed spec says:
```
  // scale describes the resolution of the histogram.  Boundaries are
  // located at powers of the base, where:
  //
  //   base = (2^(2^-scale))
  //
  // The histogram bucket identified by `index`, a signed integer,
  // contains values that are less than or equal to (base^index) and
  // greater than (base^(index-1)).
```
If you are recording double-precision floating-point numbers, this means that the effective `i`-th bucket boundary is given by the largest double-precision floating-point number that is less than or equal to `2^(i*2^-scale)`. **This value needs to be interpreted as real number!** Therefore, this value cannot be easily computed using just the power function as this would introduce numerical platform-dependent errors.

Therefore, it is important that the boundaries are carefully precomputed with higher precision than possible with double-precision. Luckily, it is sufficient to know only the mantissa values of all boundaries, if the mapping is implemented as I proposed [here](https://github.com/dynatrace-oss/dynahist/blob/master/src/main/java/com/dynatrace/dynahist/layout/OpenTelemetryExponentialBucketsLayout.java). Hence, once these values are precomputed, they can be added to any implementation as hard-coded constants such that platform independence will be achieved. Maybe, it would make sense to list those values in the spec at least for the most important precisions.

Please note, that my implementation does not yet use the correct values for the boundaries. The `boundaries` array still needs to be filled with the more precise constants.

yzhuge:
It appears dynahist uses "biased" index that shifts all bucket indexes for positive numbers to positive range, and uses negative indexes for negative numbers. See https://github.com/dynatrace-oss/dynahist/blob/master/src/test/java/com/dynatrace/dynahist/layout/OpenTelemetryExponentialBucketsLayoutTest.java#L54. So the upper vs lower bound issue is moot for dynahist. It is easy to map dynahist index to OTel index though. We just need to apply a bias offset to the index.

Though with different bound inclusiveness, the bucket counts will be different. Inclusive/exclusiveness has no impact on relative error when it comes percentile/quantile calculation. For example, when there is a 10% population exactly at a bound, depending on inclusiveness, the population may be included in bucket n or bucket n+1. Assuming that population up to bucket n is 45% (excluding the 10% bound population), a histogram counting the 10% in bucket n would return 50% percentile at mid point of bucket n. A histogram counting the 10% in bucket n+1 would return 50% percentile at mid point of bucket n+1. Because the actual percentile is exactly at the bound (where population crosses the 50% threshold), whether reported percentile is in bucket n or n+1, the distance between the reported value and the actual value is the same 1/2 of bucket width. Thus the relative error is the same. 

If the bound population is included in bucket n or n+1 not because inclusiveness, but because of floating point calculation errors, the impact (or lack or impact, actually) on relative error still stands. So platform difference in float point calculation error is not as bad as it seems. 

It is still good to have total consistency across platforms though. With lookup table method used in [dynahist](https://github.com/dynatrace-oss/dynahist/blob/master/src/main/java/com/dynatrace/dynahist/layout/OpenTelemetryExponentialBucketsLayout.java) or [NrSketch](https://github.com/newrelic-experimental/newrelic-sketch-java/blob/main/src/main/java/com/newrelic/nrsketch/indexer/SubBucketLookupIndexer.java), we can achieve platform independence with published bound tables. This can be done in the future.

oertl:
@yzhuge 

> It appears dynahist uses "biased" index that shifts all bucket indexes for positive numbers to positive range, and uses negative indexes for negative numbers. 

Yes, you are right. However, the mapping is a little bit more complex than just shifting as Dynahist uses a non-consecutive numbering of the buckets. The zero bucket has index 0. Then the positive buckets are numbered in ascending order. However, buckets for which it is impossible that a double-precision floating point number is mapped into (which may happen in the subnormal value range) are skipped. For example, there is no double precision value between Double.MIN_VALUE and 2*Double.MIN_VALUE. However, there will be many buckets in-between which will always be empty.

oertl:
@jmacd, I have updated the Dynahist implementation. Now the mantissa values for all boundaries are hard-coded. Those constants allow platform independent implementations of the bucket mapping.

@yzhuge, after reconsidering my implementation with respect to inclusiveness of boundaries, I realized that inclusive lower bounds allow a slightly faster mapping of values to buckets. The reason is that the floating-point representations of the values in the range [1,2) have all the same exponent while this is not the case for the values in the range (1,2]. Therefore, the mapping of values to buckets will be more expensive as some additional CPU instructions will be needed to cover that. The current Dynahist implementation uses inclusive lower bounds.

oertl:
I agree that for large precisions precomputed arrays are not useful. The support of large precisions and platform independence are possibly difficult to reconcile. However, the purpose of precisions much greater than 10 is not clear to me as histograms are primarily used to compress the data.

@yzhuge, according to https://github.com/newrelic-experimental/newrelic-sketch-java#internal-classes,
> At the default initial scale of 12, AUTO_SELECT starts with SubBucketLogIndexer. At the default 320 buckets, most datasets will quickly downscale to scale 6 (when contrast exceeds 5.66) and start using the more efficient SubBucketLookupIndexer.

the NRSketch uses a precision of 12 by default, which is still managable with precalculated arrays (if NRSketch would use the same indexer instances across different sketches with the same precision like Dynahist does). And as you mentioned, you often end up with a precision of 6. Is there really a need of a precision equal to 20 as you described in your example?
> At very high scales (for data set with small contrast, like 1M to 1.001M), like scale 20, lookup table would have 1M entries. 

Does it really make sense to use hundreds of buckets to cover the range 1M to 1.001M? If the minimum and the maximum are explicitly tracked, you already have the information that all values are in the range between 1M and 1.001M. Using hundreds of buckets only makes sense if you are interested in the shape of the distribution of the values within that interval. Assuming those values represent response times, and you already know that all values are in the range 1M to 1.001M, are you really interested in whether the distribution of values within that interval is more like a uniform distribution or a normal distribution? A single outlier value falling outside this interval would reduce the resolution of the histogram, but is the distribution of values in the range 1M to 1.001M then less interesting? I would be interested in a monitoring use case for which this is important. 

By the way, with a precision of 20 or more, the number of possible buckets is already in the same order of magnitude as the number of possible single-precision floating point numbers. Storing the individual values as single-precision floating-point numbers in combination with some lossless delta compression could be similarly compact and would also be platform independent.

In summary, I think platform independence is much more important than supporting extremely high precisions, as this would help avoid inconsistencies that would require special handling when evaluating histograms on different platforms.

yzhuge:
@oertl  Admittedly, higher scale (10+) is rare, especially in telemetry. One "strawman" high scale use case I can think of is precision engineering, where a 1 meter part has the tolerance of .1mm or less. As a generally purpose library, NrSketch is made to be able to handle high scale histograms.  

For a particular histogram producer, it is totally OK to say "limit max scale at 10", or "we fold subnormals into the zero region". As an industrial standard, OTel transport protocol should support the full range of possibilities though.

oertl:
@yzhuge

> As an industrial standard, OTel transport protocol should support the full range of possibilities though.

The support of arbitrary large precisions is just one of many desired features. Platform independence and compact encoding are others which unfortunately cannot all be reconciled. For high precisions platform independence is probably not feasible as I am not aware of any efficient way to map values to bucket indices. Furthermore, high precisions require 64-bit integers for bucket indices which is disadvantageous for the in-memory representation of histograms, especially when encoding sparse histograms.

From a practical point of view, I think that we should introduce a limit to the precision. At least we should limit the precision such that the number of buckets is always smaller than the maximum value of a signed 32-bit integer. (This is the maximum size of a native array in Java.). Furthermore, all possible bucket indices should be addressable using 32-bit integers. (There are languages like Java that do not support unsigned native integer types.) From that, a precision limit can be derived. I have not computed it yet exactly and it might also depend on how buckets are numbered but it should be around 20.

Limiting the precision would have the following benefits:
* 32-bit integers can be used for bucket indices which helps reducing the in-memory representation of histograms, especially for sparse histograms
* platform independent implementations are possible for all allowed precisions. (A precision of 20 would require lookup arrays with 1M entries requiring multiple MB of memory which is not nice but also not impossible as a single lookup array can be used by all histogram instances. Furthermore, as already discussed above, such high precisions are rarely needed anyway.)

yzhuge:
@oertl In https://github.com/open-telemetry/opentelemetry-proto/pull/322#issuecomment-902145604 I discussed effect of float point calculation error on quantile/percentile relative error. The conclusion is that "platform difference in float point calculation error is not as bad as it seems.". Two platforms may give different answers on percentiles, but both answers are likely to be within (or close to) nominal relative error of the histogram. For example, when true percentile is at 100, and relative error is 3%, one platform may produce a histogram that would return a percentile at 97, another may return 103. They are different, but both within error margin.

We cannot and shall not require an implementation to use the lookup method and a published lookup table. For reasons like memory constraint, or simplicity, an implementation may call log() function for value to bucket mapping. We can recommend lookup method, but cannot prevent people from using log() call and submit histograms to OTel API. So limiting to lower scales will not eliminate platform difference.

For higher scales, 64 bit integer is required for cover bucket index space. This is why I stressed that OTel protocol should use 64 bit integer for counter array index offset. The actual counters are in a Protobuf "repeated" field. Protobuf says "this field can be repeated any number of times". Whatever that means. In practice, it is probably at least signed 32 bit int times, which is fine for practical matter. As long as the 64 bit offset can shift the window anywhere in the logical 64 bit space, we are good.

[NrSketch](https://github.com/newrelic-experimental/newrelic-sketch-java#windowedcounterarray) has a WindowedCounterArray that implements a window (up to signed int32 elements) in a 64 bit index space. So it can handle arbitrary scales. There are many other ways to implement 64 bit logical index arrays. And many platforms natively supports larger than 2G arrays. Java array limitation is just a vestige of the (now ancient) 32bit platforms.

Thus I think OTel should stick to 64 bit index space and no limit on scales.

oertl:
@yzhuge I know that platform independence has not a big impact on the estimation error of quantiles. However, it avoids inconsistencies with the minimum, maximum, or other exemplar values (see https://github.com/open-telemetry/opentelemetry-proto/pull/322#issuecomment-906165606). Consistency requires that:
* The minimum, maximum, or exemplar values map to nonempty buckets.
* The bucket that contains the minimum value is the leftmost nonempty bucket.
* The bucket that contains the maximum value is the rightmost nonempty bucket.

If the mapping depends on the platform, such inconsistencies may occur which make the histogram evaluation on a different platform more complicated. (E.g., if you want to guarantee that all estimated quantiles are greater than or equal to the minimum while having equality for the 0-quantile.) Dynahist cures such inconsistencies during deserialization.

Platform independence is not strictly necessary, but since it turns out that there is a fast, platform-independent mapping of values to bucket indexes for common precisions, it makes sense to demand platform independence to avoid inconsistencies. By the way, I thought you would also prefer platform-independent mappings. Otherwise, your strong opinion on inclusive lower bounds makes no sense.

Regarding bucket indexing: It seems that
* OpenHistogram (total number of buckets is a 32bit integer, see https://github.com/openhistogram/libcircllhist/blob/b86ff46c7bc25301f539e258fd07be99b67b5709/src/circllhist.h#L152)
* HdrHistogram (https://github.com/HdrHistogram/HdrHistogram/blob/9af106907eb6618c4c3fb5ac0da773ad0fee4f13/src/main/java/org/HdrHistogram/AbstractHistogram.java#L159)
* DDSketch (https://github.com/DataDog/sketches-java/blob/d1db3f49ab49e1f047cdb41ba38d7f16f49f686d/src/main/java/com/datadoghq/sketch/ddsketch/mapping/IndexMapping.java#L98)
* Dynahist (https://github.com/dynatrace-oss/dynahist/blob/844a121ad6de2452d3a1650ada6cb33b8804c4e4/src/main/java/com/dynatrace/dynahist/layout/Layout.java#L48)
 
all work with 32bit bucket indices, I really would be interested in a use case that requires more buckets than the number of different single-precision floating-point values and that justifies wasting space on 64bit bucket indices.

jmacd:
I've read through the feedback above. Thank you @oertl and @yzhuge. 

About the indexing scheme: having worked through the lookup table generation logic, and having seen the DynaTrace prototype's special handling of subnormal values, I think I prefer to avoid the fancy mapping function. 

> We cannot and shall not require an implementation to use the lookup method and a published lookup table.

I agree. Using log() is fine. I expect the implementation that uses log() will never be "off" the fully computed-with-arbitrary-precision bucket index by more than one.  The decision to accept conversion between <= and >= as opposed to supporting two modes also introduces an off-by-one situation, which I find acceptable. I think users should choose a scale that gives the precision they want so that an off-by-one condition is OK.

About the range of permitted scale values and the related precision question: I believe it is uncommon, these scenarios where a measurement is both very large in magnitude and also very high in precision. If there is a need to store many buckets between 1M and 1.001M units, then I would advise users to divide their number by 1e6 and change the units to mega-units in the range 1 to 1.001.  If this measurement is combined with measurements with a different orders of magnitude, the resulting histogram either has to lose precision or use 64-bit bucket indexes. I think I prefer the 32-bit indexes for this reason, it saves slightly on the in-memory cost of parsed metric data AND it encourages users to keep their data in an appropriate range.

Can we agree, therefore, to proceed with the simple(r) indexing strategy, no limit on scale, no requirement to use a lookup table?  

Is the only remaining questions whether the `offset` field of the exponential range is 32bits or 64bits? We specified the use of a varint encoding for that field, so the wire size would not matter. The in-memory footprint of the parsed message is somewhat larger, this is true. I am willing to change this to 32bits.

oertl:
@jmacd As I said before, platform independence is not a must but a way to avoid inconsistencies.  If we allow "approximate" mappings to buckets, we should raise awareness of potential inconsistencies by mentioning them in the protocol. If a message is received on a different platform, the message might look inconsistent. In particular, from the receiver's point of view, min/max/exemplar values may map to different and potentially empty buckets. These inconsistencies might need special treatment when evaluating histograms.

32-bit indices imply a precision limit, if the whole range of double-precision floating-point numbers needs to be covered. The smallest positive double-precision floating-point number is `2^(-1074)`, which maps to the index `i = -1074*2^p`. In order to represent this index with a (signed) 32-bit integer we need to have `i >= -2^31`. This implies `p <= 20`. If we also want to allow representing counts of all postive/negative buckets as arrays with maximum size `< 2^31` (Java array limit) then we should even have `p <= 19`. `p <= 19` would also ensure that all buckets (negative ones, postive ones, and zero bucket) can be addressed using a single 32-bit value. `p <= 18` further guarantees that the total number of buckets (including negative and positive ones) is `< 2^31` which would potentially allow storing the counts of all buckets in a single array in Java. Due to these considerations and since there are no known use cases anyway that require such high precisions, I am in favor of specifying an upper bound for the precision that is at most 18. Furthermore, I guess it is easier later to extend the protocol to higher precisions than to restrict it to smaller precisions.

yzhuge:
First, for floating point histograms, only the relative size of min vs. max matters, not the absolute size. I use the term "contrast" (max divided by min) for this purpose. So a dataset of 1M to 1.001M has the same contrast as one from 1 to 1.001. The table below shows dataset contrast at various scales. At higher scale, base is close to 1, so I use base - 1 to show the value. Similarly I show contrast  - 1. And I also show 1/(contrast - 1) as "reverse contrast". For example, a contrast of 1.01 has a a reverse contrast of 100, meaning this can be a dataset from 100 to 101. 

| scale | base - 1 | contrast\_at\_320\_buckets - 1 | 1/(contrast - 1) |
| ----- | -------- | ------------------------------ | ---------------- |
| 50    | 6.66E-16 | 1.97E-13                       | 5.08E+12         |
| 40    | 6.30E-13 | 2.02E-10                       | 4.96E+09         |
| 30    | 6.46E-10 | 2.07E-07                       | 4.84E+06         |
| 20    | 6.61E-07 | 2.12E-04                       | 4.73E+03         |
| 19    | 1.32E-06 | 4.23E-04                       | 2.36E+03         |
| 18    | 2.64E-06 | 8.46E-04                       | 1.18E+03         |
| 17    | 5.29E-06 | 1.69E-03                       | 5.90E+02         |
| 16    | 1.06E-05 | 3.39E-03                       | 2.95E+02         |
| 15    | 2.12E-05 | 6.79E-03                       | 1.47E+02         |
| 14    | 4.23E-05 | 1.36E-02                       | 7.34E+01         |
| 13    | 8.46E-05 | 2.74E-02                       | 3.64E+01         |
| 12    | 1.69E-04 | 5.56E-02                       | 1.80E+01         |

Though we usually handle high contrast dataset, like response time. There are low contrast dataset in telemetry. For example, tracking message size in network apps. Imaging that I am instrumenting a network app for message size, and I sell this app to my customers, who have diverse use of the app. Some may have high contrast message size from 10byte to 10MB (contrast of 1M). Some may have low contrast message size at 10MB +/- 100byte (reverse contrast of 10^5). Some may have both kind of data, depending on which of their app is active on a given day. In such case, a one size fits all metric (ie. an auto scaling histogram) will be much appreciated. You may ask what is the value of tracking 100MB +/- 100byte dataset with details. Such details can be valuable for debugging. For example, if you expect size to be exactly at 100MB, but they are not, info on the +/- 100 byte range will help you debug. If the histogram has lower scale, it may lump all data into one bucket and not give you any insight.

If we use 32 bit indexes, we will be limiting scale to around 20, which supports reverse contrast around 1000, not enough for the use case above. Thus I am having second thought on limiting indexes to 32 bits. One thing I have learned from my career is that users use your software in ways you cannot imagine in the lab. Now we have at least one legit use case on high scales. And in terms of OTel protocol here, we are not going out of our way to support 64 bit index spaces. In fact, limiting it to 32 bit is "extra work". So I'd like to keep the 64bit space support.

Here is another case of "users do things you cannot image". At New Relic, we thought few people would submit negative numbers to histogram. But we were wrong. A good number of legit use cases send negative numbers. One of them is cell phone signal strength, which is typically in negative range when measured in dBm. 

In New Relic production system, the initial scale of a histogram is 32, which requires an index space around 42 bits. Such high initial scale is intended for zero config. It deals with many kinds of data. We don't have interface to config it. Supporting higher scales is good for such zero config use cases.

As for computation errors, I think we should have a separate doc discussing ways to reduce errors, like the lookup method. The protocol code is not a good place for implementation suggestions. Naive implementations have noticeable errors even at scales below 5, as shown in https://github.com/newrelic-experimental/newrelic-sketch-java/issues/5, which I have addressed via https://github.com/newrelic-experimental/newrelic-sketch-java/pull/6. Subtle difference in calling Math.log() can have big impact on errors. A separated doc will give us the space to discuss these things.

oertl:
@yzhuge 

> Such details can be valuable for debugging. For example, if you expect size to be exactly at 100MB, but they are not, info on the +/- 100 byte range will help you debug. If the histogram has lower scale, it may lump all data into one bucket and not give you any insight.

It would be sufficient to track the minimum and maximum to know that not all values are equal even if they all fall into the same bucket.

In my opinion, the main reason for using an exponential histogram is to significantly reduce the original data size while still having at least some guarantees on the relative error. For high precisions, I wonder if it is not more space-efficient to just compress the raw values directly (maybe after first reducing them to single-precision). If you are interested in the shape of distributions with small contrasts, the KLL sketch (https://datasketches.apache.org/docs/KLL/KLLSketch.html) or something similar could be also a better choice. 

I have the feeling that we are not discussing a histogram format here, but rather the definition of an arbitrary-precision floating-point number format that fills the gap between the IEEE single and double precision formats.

yzhuge:
@oertl min/max won't tell you what is between the min and max. Is it spread out in the whole range, or having a few spikes? The distribution info will help debugging.

For low contrast dataset, exponential histogram is not the best tool. Simple equal width buckets will do (and arguably do better). For a dataset of known properties, or one that can be examined multiple times, you have the luxury to choose the best tool or even preprocess it (like rescale, or apply an offset) for best view. But for the telemetry use case, dataset is often of unknown property, or is too expensive to configure histogram type and parameters for individual data streams. Thus the advantage of a one size fits all tool that can have "wide angle view" on high contrast data, and "microscopic view" on low contrast data. In telemetry, you often don't have the luxury to examine the data for a 2nd time. A problem is sometimes unreproduceable, or too risky to reproduce (you don't want to risk company production system to debug every problem). So pre-wired general purpose instrumentation is important.

This reminds me of pet vs cattle. For pets, you can afford custom collar for each one, for cattle, you just want a standard collar on each cow. Here we actually have an auto sizing collar (auto scaling histogram), which is not as good as custom collar for each animal, but is way better than static sized ones (hardwired data range or scale).

jmacd:
@oertl and @yzhuge I updated this text to use the inclusive lower bound. 

TL;DR I will give detailed feedback in this comment and a summary in the following comment.

A number of the questions being discussed at the end of this thread can be treated not as protocol issues, but as issues present at runtime and/or issues with the compiled code. We are discussing the advisable ranges for `scale`, the width of the bucket index, and how to handle hardware imprecision, all of them interrelated topics.

For scale, I don't see anything mathematically wrong with large values (or negative values), just a limited range of useful outcomes. 

For the bucket index, we've already specified varint encoding for the relevant `offset`, so any decision about 32-bit vs 64-bit width is made by the protocol compiler, and we can generate new code to support wider values. We don't need to change the protocol to take advantage of large indexes, but we do need to change the generated code.  If we went ahead with `sint32` as the type for the offset field, then OTel collectors would not be able to manipulate or inspect large-scale histograms without code changes.

These are connected by the hardware precision issue, because at certain scales we have a limited ability to represent the entire range of IEEE floating point values and/or we can represent distinctions that the IEEE floating point values cannot distinguish.

@yzhuge has explained his motivation for not restricting scale to scale or bucket index.  @oertl points out that indexes below -2^63 and above 2^63-1 will be difficult to work with. Whether indexes are limited to 32-bit or 64-bit, the hypothetical auto-ranging histogram has to be aware that the maximum scale, subject to index-width constraints, is limited by the exponent of the data (i.e., your best scale depends on the data and the width of the index, but there was always going to be a limit). I believe a good auto-ranging histogram can be built for 32-bit indexes. If we have such a thing and users are unsatisfied with it because it does not have enough precision for their data ranges, then it will not require a protocol change to widen the bucket index. It will only require recompiling a bunch of code. I suspect it is likely that before we need 64-bit index support, we will want to support a sparse encoding the way Prometheus has proposed.

@yzhuge for example if New Relic wants to support 64-bit bucket indexes internally it can maintain a copy of the `.proto` file with `sint32 offset = 1;` replaced by `sint64 offset = 1;` and nothing will break. This would be a good way to convince the community to adopt 64-bit bucket widths in the OTel collector.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -448,6 +458,109 @@ message HistogramDataPoint {
...
+  // positive carries the positive range of exponential bucket counts.
+  Buckets positive = 8;
+
+  // negative carries the negative range of exponential bucket counts.
+  Buckets negative = 9;
+
+  // Buckets are a set of bucket counts, encoded in a contiguous array
+  // of counts.
+  message Buckets {
+    // Offset is the index of the first count expressed in this span.
```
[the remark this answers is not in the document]

opentelemetry/proto/metrics/v1/metrics.proto

opentelemetry/proto/metrics/v1/metrics.proto

jmacd:
@open-telemetry/specs-metrics-approvers to summarize the current state of this PR:

We have deliberated on every detail and I believe this is in its final form. There are questions about number widths in the `.proto` that can change in the future because we are using varint-encoding for the fields in question. 

A number of questions have come up that I think should be specified in the data model, not the protocol. 

An example: implementations are free to use `log()` to compute the bucket index, which may lead to inconsistent bucket calculations across platforms and could lead to inconsistencies between exemplar data and the buckets (e.g., the bucket index calculated on the server for the `max` is an empty bucket, because the client calculated the adjacent bucket). 

A second example: implementations may reject data with scale or bucket index that are too large or small for them to handle. We are trying to avoid being overly prescriptive about maximum scale, but when histogram buckets map onto values that cannot be expressed by an available IEEE floating point value, I think implementations should be free to decide how to proceed.

I think we should accept this PR in its current form and end the debate. We have demonstrated three prototypes. I'll follow up next week with proposed text for the [data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md). Please review.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -448,6 +459,107 @@ message HistogramDataPoint {
...
+  fixed64 start_time_unix_nano = 2;
+
+  // TimeUnixNano is required, see the detailed comments above Metric.
+  //
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
+  // 1970.
+  fixed64 time_unix_nano = 3;
+
+  // count is the number of values in the population. Must be non-negative. This
+  // value must be equal to the sum of the "count" fields in buckets if a
```
yzhuge, line 486: shall we explicitly say that "count" also includes "zero_count"?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -176,6 +176,7 @@ message Metric {
     // This field will be removed in ~3 months, on July 1, 2021.
     IntHistogram int_histogram = 8 [deprecated = true];
     Histogram histogram = 9;
+    ExponentialHistogram exponential_histogram = 12;
```
reyang, line 179: Curious, why is this one not following the number ordering?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -448,6 +459,107 @@ message HistogramDataPoint {
...
+// summary statistics for a population of values, it may optionally contain the
+// distribution of those values across a set of buckets.
+//
+message ExponentialHistogramDataPoint {
+  // The set of key/value pairs that uniquely identify the timeseries from
+  // where this point belongs. The list may be empty (may contain 0 elements).
+  repeated opentelemetry.proto.common.v1.KeyValue attributes = 1;
+
+  // StartTimeUnixNano is optional but strongly encouraged, see the
+  // the detiled comments above Metric.
```
reyang, line 473:
```suggestion
  // the detailed comments above Metric.
```

reyang (approved):

jsuereth:
Still approve, and +1 to hashing out further specification on acceptable precision + calculation consistency in the datamodel specification.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: Yes. I had copied the comments for `count` and `sum` from the explicit histogram point, and on re-review I see there are a couple of extraneous parts. The explicit histogram point mentions "if provided" for the bucket counts, and we do permit explicit histograms to be lacking in buckets, in which case the histogram becomes a sum/count data point.

I've removed the "if provided" from this point--we may as well use the explicit histogram point if a sum/count data point is needed. There was also a mention of a "sum" field in the buckets themselves, which is a reference to something very old and no longer applicable.

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: oh, I didn't realize--I've re-used number 10 so the fields are in order. (Was 10 ever used?)

opentelemetry/proto/metrics/v1/metrics.proto
jmacd: This had been copied a bunch of places, now all fixed!

jamesmoessis:
I'm trying to understand the description of `bucket_counts`. 

```
    // Count is an array of counts, where count[i] carries the count
    // of the bucket at index (offset+i).  count[i] is the count of
    // values greater than or equal to base^(offset+i) and less than
    // base^(offset+i+1).
```

For simplicity, assume my offset and scale are both 0. So my bucket lower bounds would be 1, 2, 4, 8, 16, etc. Does this mean that values below 1 and greater than 0 don't fall into any bucket? Is this what you are describing as "subnormal values"? Certainly possible I'm missing someone super obvious here so someone please point it out if I am!

yzhuge:
@jamesmoessis index and offset can be negative. base^index will be less than 1 when index is negative. For example: 2^-1 = 0.5.  It will never reach 0 though. Thus we have a separate count for 0.

jamesmoessis:
@yzhuge thankyou for the explanation, that makes sense.

jmacd:
This issue captures the need to specify more data model.
https://github.com/open-telemetry/opentelemetry-specification/issues/1932

I propose we merge this after there's a PR up for review on 1932.

jmacd:
I am working on adding this to the Metrics data model here:
https://github.com/open-telemetry/opentelemetry-specification/pull/1935

It is still a work in progress, but I feel we can merge this PR!

jsuereth:
Agreed this can be merged and we can hash out information in the DataModel portion.

jmacd:
Thank you to @yzhuge, @oertl, @beorn7, @giltene and the OpenTelemetry community who contributed to this design. :rocket:

PR #327, opened 2021-08-31
Remove outdated comment on attributes in trace.proto

arminru:
One can just look up the definition of `KeyValue` and `AnyValue` in `common.proto` which will stay up to date.

`AnyValue` is defined as:
https://github.com/open-telemetry/opentelemetry-proto/blob/e57e8494433f26e9050835c76e74282b0cd718a8/opentelemetry/proto/common/v1/common.proto#L30-L38

The Tracing API spec further restricts it to this subset:
https://github.com/open-telemetry/opentelemetry-specification/blob/eb44020eaf959464f2395023ab7c93a709de1725/specification/common/common.md#L21-L25
This could be stated in the comment instead if preferred.

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -158,8 +158,7 @@ message Span {
   // This field is semantically required and it is expected that end_time >= start_time.
   fixed64 end_time_unix_nano = 8;
 
-  // attributes is a collection of key/value pairs. The value can be a string,
-  // an integer, a double or the Boolean values `true` or `false`. Note, global attributes
+  // attributes is a collection of key/value pairs. Note, global attributes
   // like server name can be set using the resource API. Examples of attributes:
   //
   //     "/http/user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36"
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -158,8 +158,7 @@ message Span {
   // This field is semantically required and it is expected that end_time >= start_time.
   fixed64 end_time_unix_nano = 8;
 
-  // attributes is a collection of key/value pairs. The value can be a string,
-  // an integer, a double or the Boolean values `true` or `false`. Note, global attributes
+  // attributes is a collection of key/value pairs. Note, global attributes
```
Oberon00, line 161: The attribute in common.proto allows more than the OTel trace API spec. Not sure if this restriction applies to OTLP in general if used without OTel theoretically. But in doubt I would point to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes for the allowed subset.

tigrannajaryan (approved):

opentelemetry/proto/trace/v1/trace.proto
arminru: Good point, I added a reference.

SergeyKanzhelev (approved):

SergeyKanzhelev:
non controversial, merging

Oberon00:
> non controversial, merging

I think this is controversial and posted a comment saying so. While it was wrong previously, now it's just misleading since e.g. mixed arrays are not actually allowed.

PR #328, opened 2021-09-01
Remove outdated comment on empty span name

pmm-sumo:
Remove outdated comment on empty span name and replace it with recommendations from specification

Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/1872

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -99,10 +99,11 @@ message Span {
   // the same display name at the same call point in an application.
   // This makes it easier to correlate spans in different traces.
   //
-  // This field is semantically required to be set to non-empty string.
-  // When null or empty string received - receiver may use string "name"
-  // as a replacement. There might be smarted algorithms implemented by
-  // receiver to fix the empty span name.
+  // The span name SHOULD be the most general string that identifies a (statistically)
+  // interesting class of Spans, rather than individual Span instances while still
+  // being human-readable. That is, "get_user" is a reasonable name, while
+  // "get_user/314159", where "314159" is a user ID, is not a good name due to its high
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -99,10 +99,11 @@ message Span {
   // the same display name at the same call point in an application.
   // This makes it easier to correlate spans in different traces.
   //
-  // This field is semantically required to be set to non-empty string.
-  // When null or empty string received - receiver may use string "name"
-  // as a replacement. There might be smarted algorithms implemented by
-  // receiver to fix the empty span name.
```
bogdandrutu, line 105: Why is this outdated? There may be source that do not follow the semantics, in that case the receiver does either reject the span or treats that with a "default"

pmm-sumo: Such behavior (replacing empty span name with default or rejecting the span) is not mentioned anywhere in the specification. Also, it does not seem to be implemented anywhere

bogdandrutu: That is not true since in OpenCensus we had a TrcuncatableString which was a message so was able to be null

bogdandrutu: This was added to support languages like Java where strings can be null but in prptobufs this cannot be possible

pmm-sumo: Redone this, so I believe this is no longer relevant and for clarity I will resolve this conversation

bogdandrutu (approved):

bogdandrutu:
@open-telemetry/specs-approvers @open-telemetry/technical-committee please review

anuraaga (approved):

tigrannajaryan (approved):

PR #330, opened 2021-09-03
Add a note about restrictions of metric attribute values

pirgeo:
Adds a note about the restrictions for Attribute values, both in the Resource and for the data points. This should ensure that users and readers of the protobuf are aware of the restrictions (e.g. arbitrarily nested values are not allowed). This note has already been added to trace attributes.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -28,6 +28,9 @@ option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/metric
 message ResourceMetrics {
   // The resource for the metrics in this message.
   // If this field is not set then no resource info is known.
+  // The OpenTelemetry API specification further restricts the allowed value
+  // types for Attributes of the Resource:
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes
   opentelemetry.proto.resource.v1.Resource resource = 1;
 
   // A list of metrics that originate from a resource.
@@ -320,6 +323,8 @@ enum DataPointFlags {
 message NumberDataPoint {
   // The set of key/value pairs that uniquely identify the timeseries from
   // where this point belongs. The list may be empty (may contain 0 elements).
+  // The OpenTelemetry API specification further restricts the allowed value types:
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes
   repeated opentelemetry.proto.common.v1.KeyValue attributes = 7;
 
   // Labels is deprecated and will be removed soon.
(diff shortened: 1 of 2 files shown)


arminru (approved):

jmacd (approved):

cijothomas (approved):

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -28,6 +28,9 @@ option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/metric
 message ResourceMetrics {
   // The resource for the metrics in this message.
   // If this field is not set then no resource info is known.
+  // The OpenTelemetry API specification further restricts the allowed value
+  // types for Attributes of the Resource:
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes
```
Oberon00, line 33: Why here and not inside Resource?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -320,6 +323,8 @@ enum DataPointFlags {
 message NumberDataPoint {
   // The set of key/value pairs that uniquely identify the timeseries from
   // where this point belongs. The list may be empty (may contain 0 elements).
+  // The OpenTelemetry API specification further restricts the allowed value types:
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes
```
Oberon00, line 327: Since we have this so many times, I think it would be better to move this comment to the definition of KeyValue.

opentelemetry/proto/metrics/v1/metrics.proto
arminru: Good point, I moved the comment there in 39680c since that applies to traces, metrics and logs.
(jumping in for @pirgeo who's on vacation 🙂)

opentelemetry/proto/metrics/v1/metrics.proto
arminru: `KeyValue` is also used in `logs.proto` which does not have this restriction (logs were the reason why `KeyValue` was defined this permissively in first place). The comments only apply to `trace.proto` and `metrics.proto` but I agree the duplication is not really nice.

tigrannajaryan (changes requested):
I do not think these comments belong to the proto. They are already in the API spec and they belong there.

The proto allows to represent more than the API allows. This is an intentional design for 2 reasons:
1. Make proto future-proof when the API is extended.
2. To allow to represent telemetry data that comes from other protocols. For example Jaeger support BINARY data type that OTLP also supports but our API does not.

By adding this comments to the proto we create the false sense for the receivers that they won't see such data (they will, e.g. if using Collector with Jaeger receiver and OTLP exporter).

We also unnecessarily restrict exporter implementations which want to support the full range of possible values (OTLP exporter in Collector does).

We should not assume that Otel API is the only possible source of OTLP data. Collector is a source itself and it can and does produce data that does not fit the Otel API restrictions.

jsuereth:
I'm in agreement with @tigrannajaryan.  The spec limitation applies to our SDK but not our transport protocol.

I think we *do* need to specify how to deal with complex-valued attributes in metrics-that-expect-string-labels for some commonality in known implementations, but this doesn't belong in our transport protocol.

pirgeo:
I see. I do, however, agree that we should offer some guidance on how to deal with complex-valued attributes _somewhere_. Maybe an addition to the spec?

PR #331, opened 2021-09-08
Change "null" to "empty" when referencing AnyValue with no value set

Aneurysm9:
Signed-off-by: Anthony J Mirabella <a9@aneurysm9.com>

### DIFF
--- a/opentelemetry/proto/common/v1/common.proto
+++ b/opentelemetry/proto/common/v1/common.proto
@@ -26,7 +26,7 @@ option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/common
 // object containing arrays, key-value lists and primitives.
 message AnyValue {
   // The value is one of the listed fields. It is valid for all values to be unspecified
-  // in which case this AnyValue is considered to be "null".
+  // in which case this AnyValue is considered to be "empty".
   oneof value {
     string string_value = 1;
     bool bool_value = 2;


bogdandrutu (approved):

tigrannajaryan (approved):

anuraaga:
We mention attribute value of `null` in the spec

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes

Isn't it more consistent for this proto to use the same term here?

Though since `null` aren't allowed anyways, it's not that important just confirming.

tigrannajaryan:
It seems like we have `null` in more places than we had `empty`. Probably was more worth doing the opposite and renaming `empty` to `null` everywhere. `null` also seems a bit more natural to me.

Aneurysm9:
The spec very explicitly says that null values are not valid:

> Attribute values of `null` are not valid and attempting to set a `null` value is undefined behavior.
> 
> `null` values SHOULD NOT be allowed in arrays.

It also explicitly countenances empty values:

> Attribute values expressing a numerical value of zero, an empty string, or an empty array are considered meaningful and MUST be stored and passed on to processors / exporters.

The comment as it existed in the proto doesn't make any sense as it talks about valid, unspecified values but the spec says that `null` values are invalid.  It can only make sense for an unspecified `AnyValue` attribute to be considered empty.

anuraaga:
But what does empty mean specifically? `null` indeed may not be great but I don't think empty improves much since it's still a not precise term here.

Maybe it should be updated to say that one field must be set - if a receiver gets a message with no fields it should just ignore the message. This would be clearer than trying to use a semantic term for a state that doesn't actually have any semantic meaning.

Aneurysm9:
I don't think "empty" is either imprecise or without semantic meaning.  The `AnyValue` type is a container that can hold zero or one of a variety of types of values.  It it contains zero values it is empty, much the same way that an empty string is a string that contains zero characters and an empty array is an array that contains zero values.

Changing the specification and protocol to require that a value be set would be a much more significant change than clarifying a comment.  I don't think that is the correct course of action here.

tigrannajaryan:
>Changing the specification and protocol to require that a value be set would be a much more significant change than clarifying a comment.

It is also not what we want. Null (or empty) values are necessary to support JSON-like data structures which are a valid use case for logs.

anuraaga:
> It is also not what we want. Null (or empty) values are necessary to support JSON-like data structures which are a valid use case for logs.

Just to confirm would these be OTel attributes or something else? It does sound like it's needed to represent arbitrary JSON, it would be nice to use the google proto `Value` type instead if at all possible for that use case though since then the JSON version is actually mostly the original JSON :) I don't think `AnyValue` can support JSON objects without extension.

> The AnyValue type is a container that can hold zero or one of a variety of types of values. 

I wouldn't consider `AnyValue` a container - the name says it all, it's a value. proto doesn't have polymorphic or union typed values so using a message with a single oneof is sort of a workaround on how to represent such a value. But it's still a single value - empty would only apply if the type happened to be `string` or `array` but if it's one of the scalar types I don't see empty applying.

For reference, the official type which this one is based on actually has an explicit field for Null and explicitly says it's an error not to set any field

https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/struct.proto#L59

Changing our proto to be simpler I think would be compatible enough for our current usage of this proto, but it is effort though. So instead, representing `null` with our convention of just not setting any field seems like a way to do it too. But I don't think this is the concept of emptiness.

Aneurysm9:
> A producer of value is expected to set one of these
> variants. Absence of any variant indicates an error.

Absence indicating an error is different, I think, than absence being invalid.  It may indicate that an error happened at the producer, but it is not a protocol level error or malformed data.  The protocol definition used by `AnyValue` has different connotations for a message with no variant set in its single `oneof` field.

> I wouldn't consider AnyValue a container - the name says it all, it's a value. proto doesn't have polymorphic or union typed values so using a message with a single oneof is sort of a workaround on how to represent such a value. But it's still a single value

I don't see how it can't be viewed as a container.  It contains zero or one values in the current specification.  Is a `[]interface{}` not a container if it is meant to represent a string value, treating each element as either a `rune` or a `byte`?  Is a `int**` a container because it is meant to be read as a series of `int*` until a `\0` is encountered?  

Form and function cannot always be cleanly separated.  `AnyValue` as defined is a container that may be empty but I do not see any way for there to be a `null` `AnyValue`.  In a specific application there may be a pointer to memory that is intended to represent an `AnyValue`, and that pointer may have the value `\0`, but that does not make a null `AnyValue`.  It only represents the lack of existence of an `AnyValue` referenced by that pointer.

anuraaga:
> []interface{} not a container if it is meant to represent a string value, treating each element as either a rune or a byte? Is a int** a container because it is meant to be read as a series of int* until a \0 is encountered?

Not sure, but both of these seem to be array types where empty does indeed make sense. But AnyValue isn't an array type, it's scalar - it is comprised of a single field.

Here we're defining a meaning for a protobuf message - it can have any meaning we want it to have. We need to step back and look at how it's used, as the type for a field in a different message.

```
AnyValue cool_value = 2;
```

By setting one of the fields in the `AnyValue` assigned to `cool_value` we can represent int, string, bool, array types. Asides from these, to support "null", a concept we use in the spec to refer to a common language idiom of an unset value, we currently specify it as a `AnyValue` message with no fields set. Another approach could be to expliitly add `NullValue` like the referenced proto. Or another could be to use the lack of presence of `cool_value` itself. I think the reason that isn't used is to allow an array with null values.

```
repeated AnyValue cool_values = 3;
```

Here, the intent of the proto is for [{}, {string_value: "foo"}, {int_value: 3}] to correspond to an array (not every language can handle this) like [null, "foo", 3]. [{}, {}, {}] isn't an array of empties, it's an array of nulls.

tigrannajaryan:
> I don't think `AnyValue` can support JSON objects without extension, like the `Struct` type field on `Value`.

@anuraaga `AnyValue` can store a `KeyValueList` which allows representing JSON objects.

>it would be nice to use the google proto Value type instead

We can't. OTLP is declared stable for traces and google proto Value is not wire compatible with AnyValue. It would be a breaking change which is not allowed according to our stability guarantees.

>Here, the intent of the proto is for [{}, {string_value: "foo"}, {int_value: 3}] to correspond to an array (not every language can handle this) like [null, "foo", 3].

Correct. This is exactly the intended use.

jmacd (approved):

jsuereth:
Just to chime in on the `struct.proto` and `Null`-as-a-value discussion, I think the resolution is the following:

- We need the current rewording for logs, and we're trying our AnyValue like JSON, so empty is like any empty object (or null-object), vs. the actual value null in JSON.
- We *can* expand AnyValue with an explicit `null` if we really feel we need it.
- `struct.proto`, while nice, would be a breaking change which we cannot allow (especially in AnyValue).   

I gave this a week to see if there were any more comments before chiming in.  I think plan forward is to add explicit `null` type to AnyValue if we feel we need it.  Merging, and please file follow up if we need explicit Null or empty object.

PR #332, opened 2021-09-15
Add proto messages for signals data independent of OTLP protocol

bogdandrutu:
The main motivation for this is to ensure that other protocols can use it to
send/receive OTLP data, as well as for persistent storages such as kafka, disk, etc.

Unfortunately we cannot use the same message in the OTLP request messages since it will be a breaking change,
but we can ensure that all fields present in the data message will be added to the request message as well.

### DIFF
--- a/opentelemetry/proto/logs/v1/logs.proto
+++ b/opentelemetry/proto/logs/v1/logs.proto
@@ -24,6 +24,25 @@ option java_package = "io.opentelemetry.proto.logs.v1";
 option java_outer_classname = "LogsProto";
 option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/logs/v1";
 
+// LogsData represents the logs data that can be stored in a persistent storage,
+// OR can be embedded by other protocols that transfer OTLP logs data but do not
+// implement the OTLP protocol.
+//
+// The main difference between this message and collector protocol is that
+// in this message there will not be any "control" or "metadata" specific to
+// OTLP protocol.
+//
+// When new fields are added into this message, the OTLP request MUST be updated
+// as well.
+message LogsData {
+  // An array of ResourceLogs.
+  // For data coming from a single resource this array will typically contain
+  // one element. Intermediary nodes that receive data from multiple origins
+  // typically batch the data before forwarding further and in that case this
+  // array will contain multiple elements.
+  repeated ResourceLogs resource_logs = 1;
+}
+
 // A collection of InstrumentationLibraryLogs from a Resource.
 message ResourceLogs {
   // The resource for the logs in this message.
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -24,6 +24,25 @@ option java_package = "io.opentelemetry.proto.metrics.v1";
 option java_outer_classname = "MetricsProto";
 option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/metrics/v1";
 
+// MetricsData represents the metrics data that can be stored in a persistent
+// storage, OR can be embedded by other protocols that transfer OTLP metrics
+// data but do not implement the OTLP protocol.
+//
+// The main difference between this message and collector protocol is that
+// in this message there will not be any "control" or "metadata" specific to
+// OTLP protocol.
+//
+// When new fields are added into this message, the OTLP request MUST be updated
+// as well.
+message MetricsData {
+  // An array of ResourceMetrics.
+  // For data coming from a single resource this array will typically contain
+  // one element. Intermediary nodes that receive data from multiple origins
+  // typically batch the data before forwarding further and in that case this
+  // array will contain multiple elements.
+  repeated ResourceMetrics resource_metrics = 1;
+}
+
 // A collection of InstrumentationLibraryMetrics from a Resource.
 message ResourceMetrics {
   // The resource for the metrics in this message.
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -24,6 +24,25 @@ option java_package = "io.opentelemetry.proto.trace.v1";
 option java_outer_classname = "TraceProto";
 option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/trace/v1";
 
+// TracesData represents the traces data that can be stored in a persistent storage,
+// OR can be embedded by other protocols that transfer OTLP traces data but do
+// not implement the OTLP protocol.
+//
+// The main difference between this message and collector protocol is that
+// in this message there will not be any "control" or "metadata" specific to
+// OTLP protocol.
+//
+// When new fields are added into this message, the OTLP request MUST be updated
+// as well.
+message TracesData {
+  // An array of ResourceSpans.
+  // For data coming from a single resource this array will typically contain
+  // one element. Intermediary nodes that receive data from multiple origins
+  // typically batch the data before forwarding further and in that case this
+  // array will contain multiple elements.
+  repeated ResourceSpans resource_spans = 1;
+}
+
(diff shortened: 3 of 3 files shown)


jpkrohling (approved):

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -24,6 +24,25 @@ option java_package = "io.opentelemetry.proto.logs.v1";
 option java_outer_classname = "LogsProto";
 option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/logs/v1";
 
+// Logs represents the logs data that can be stored in a persistent storage,
+// OR can be embedded by other protocols that transfer OTLP logs data but do not
+// implement the OTLP protocol.
+//
+// The main difference between this message and collector protocol is that
```
tigrannajaryan, line 31:
```suggestion
// The main difference between this message and ExportLogsServiceRequest is that
```

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -24,6 +24,25 @@ option java_package = "io.opentelemetry.proto.logs.v1";
...
+// in this message there will not be any "control" or "metadata" specific to
+// OTLP protocol.
+//
+// When new fields are added into this message, the OTLP request MUST be updated
```
tigrannajaryan, line 35:
```suggestion
// When new fields are added into this message, ExportLogsServiceRequest MUST be updated
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -24,6 +24,25 @@ option java_package = "io.opentelemetry.proto.metrics.v1";
 option java_outer_classname = "MetricsProto";
 option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/metrics/v1";
 
+// Metrics represents the metrics data that can be stored in a persistent
+// storage, OR can be embedded by other protocols that transfer OTLP metrics
+// data but do not implement the OTLP protocol.
+//
+// The main difference between this message and collector protocol is that
```
tigrannajaryan, line 31:
```suggestion
// The main difference between this message and ExportMetricsServiceRequest is that
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -24,6 +24,25 @@ option java_package = "io.opentelemetry.proto.metrics.v1";
...
+// in this message there will not be any "control" or "metadata" specific to
+// OTLP protocol.
+//
+// When new fields are added into this message, the OTLP request MUST be updated
```
tigrannajaryan, line 35:
```suggestion
// When new fields are added into this message, ExportMetricsServiceRequest MUST be updated
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -24,6 +24,25 @@ option java_package = "io.opentelemetry.proto.trace.v1";
 option java_outer_classname = "TraceProto";
 option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/trace/v1";
 
+// Traces represents the traces data that can be stored in a persistent storage,
+// OR can be embedded by other protocols that transfer OTLP traces data but do
+// not implement the OTLP protocol.
+//
+// The main difference between this message and collector protocol is that
```
tigrannajaryan, line 31:
```suggestion
// The main difference between this message and ExportTraceServiceRequest is that
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -24,6 +24,25 @@ option java_package = "io.opentelemetry.proto.trace.v1";
...
+// in this message there will not be any "control" or "metadata" specific to
+// OTLP protocol.
+//
+// When new fields are added into this message, the OTLP request MUST be updated
```
tigrannajaryan, line 35:
```suggestion
// When new fields are added into this message, ExportTraceServiceRequest MUST be updated
```

bogdandrutu:
@tigrannajaryan PTAL

tigrannajaryan (approved):
LGTM

Aneurysm9 (approved):

jmacd (approved):

carlosalberto (approved):

jmacd:
Please add a CHANGELOG entry for the release notes.

jsuereth (approved):

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -24,6 +24,25 @@ option java_package = "io.opentelemetry.proto.trace.v1";
...
+// OR can be embedded by other protocols that transfer OTLP traces data but do
+// not implement the OTLP protocol.
+//
+// The main difference between this message and ExportTraceServiceRequest is that
+// in this message there will not be any "control" or "metadata" specific to
+// OTLP protocol.
+//
+// When new fields are added into this message, ExportTraceServiceRequest MUST be updated
+// as well.
+message Traces {
```
anuraaga, line 37: Is it ok that a structure `Traces` will generally not contain full traces, just the part that is local to a component? It seems like it could be confusing (`ExportTraceRequest` doesn't have this problem since it's explicitly about exporting).

bogdandrutu: Indeed it may be a bit confusing, but on the other side I really like the consistency.

Maybe rename everything with a suffix 'Data'?

anuraaga: `Data` came to mind to and might be OK. Though the `List` suffix seems dead simple and pretty clear too

tigrannajaryan: Yeah, `List` or `Collection` or similar should work.

jmacd: I like `List`.

bogdandrutu: ResourceSpansList pe TracesList? Not sure I understand the proposal exactly

pirgeo:
I am not sure what the motivation for this is. Could somebody expand on why this is important?

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: TracesList sounds like it is full traces like @anuraaga said. ResourceSpansList seems safer.

bogdandrutu: ResourceSpansList is bad because the whole idea is that we may possibly add extra things. Naming very specific like that is not my top preference

Oberon00:
> The main motivation for this is to ensure that other protocols can use it to
send/receive OTLP data, as well as for persistent storages such as kafka, disk, etc.

Why are the existing messages not suitable for that?

bogdandrutu:
@Oberon00 that is kind of missunderstanding of the motivation, or me missunderstanding your suggestion. I explained also in the meeting, currently we recommend users (including the collector) to serialize the Request object on the disk or in Kafka, but this seems wrong if the request will include protocol specific data. Also others are not willing to embed or Request object in their protocols see Jaeger example in their query API, these new messages will be guaranteed to contain only the data and nothing else more. Users can still include extra things as siblings to the data. Another reason is that if we add something else to data and users will include only the repeated list they will have a hard time upgrading (not going to happen automatically). 

So it is critical for us to separate Data from protocol specific metadata in our protos

opentelemetry/proto/trace/v1/trace.proto
bogdandrutu: Picked "TracesData" since again see my previous comment, one of the reason to have this as independent message is to ensure that people embedding this message will have all the "data" fields that a signal may have, excluding protocol specific metadata.

carlosalberto:
Change name looks fine!

anuraaga (approved):

PR #336, opened 2021-10-04
Allow field-presence to be used in OTLP

### LINKED ISSUE #314: Cannot distinguish Histogram's Sum field between 0 vs not assigned.
proto3 does not support a has_*() method. Thus, code cannot detect the difference between a field that is not assigned vs a field that has a default value (i.e. 0).

Current spec for Histogram's Sum field has a note to not assign the field in some condition. This condition not detectable by code (i.e. Sum is assigned to zero).

```
```

jsuereth:
DO NOT MERGE UNTIL GOGOPROTO SOLUTION IS FOUND

Update the build to allow field presence to be used in OTLP.

- Enable the optional field-presence flag.
- Fix #314 by marking sum as optional (non-breaking change, as 0/unset cannot be meaningfully interpreted in OTLP <0.9).

NOTE: GoGoProto does not support optional field presence.  Investigating how long adding that would take, but it would necessitate forking GoGo proto.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -442,7 +442,7 @@ message HistogramDataPoint {
   // Negative events *can* be recorded, but sum should not be filled out when
   // doing so.  This is specifically to enforce compatibility w/ OpenMetrics,
   // see: https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#histogram
-  double sum = 5;
+  optional double sum = 5;
 
   // bucket_counts is an optional field contains the count values of histogram
   // for each bucket.
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ gen-all: gen-cpp gen-csharp gen-go gen-java gen-objc gen-openapi gen-php gen-pyt
 OTEL_DOCKER_PROTOBUF ?= otel/build-protobuf:0.4.0
 PROTOC := docker run --rm -u ${shell id -u} -v${PWD}:${PWD} -w${PWD} ${OTEL_DOCKER_PROTOBUF} --proto_path=${PWD}
 PROTO_INCLUDES := -I/usr/include/github.com/gogo/protobuf
+PROTO_FLAGS := --experimental_allow_proto3_optional
 
 PROTO_GEN_CPP_DIR ?= $(GENDIR)/cpp
 PROTO_GEN_CSHARP_DIR ?= $(GENDIR)/csharp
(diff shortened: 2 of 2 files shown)


jsuereth:
cc @jack-berg This is also blocking min/max on histogram using optional.  Same fix for builds for non-go libraries

tigrannajaryan (changes requested):
This relies on an experimental capability of protoc. My understanding is that capability can be removed/changed anytime and thus does not belong to a stable protocol specification.

The official definition of the `optional` tag is the [following:](https://developers.google.com/protocol-buffers/docs/overview#optional)
>When a message is parsed, if it does not contain an optional element, the corresponding field in the parsed object is set to the default value for that field.

My reading of this is that the missing optional field cannot be distinguished from a present field that has the default value. So, if I understand correctly this proposal cannot solve https://github.com/open-telemetry/opentelemetry-proto/pull/336 regardless of what Gogo proto does.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -442,7 +442,7 @@ message HistogramDataPoint {
   // Negative events *can* be recorded, but sum should not be filled out when
   // doing so.  This is specifically to enforce compatibility w/ OpenMetrics,
   // see: https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#histogram
-  double sum = 5;
+  optional double sum = 5;
```
anuraaga, line 445: Out of curiosity (and not having read any prior context), is there a semantic meaning to optional fields in OTLP? i.e., why would we allow SDKs to not populate this, given it can provide value? Or if it doesn't provide value why do we not just deprecate it for removal?

If it's about some fields being possibly useful if the backend understands it, then it's a concept I haven't seen before in OTel, which generally leans toward sending as much data as possible. If it is about that sort of efficiency though, I think we first need some sort of negotiation protocol defined, where the required fields (and semantic attributes!) are negotiated beforehand before sending - that would then make this sort of optionality make sense.

tigrannajaryan:
A possible alternate solution is to add a bit to flags to indicate whether the sum is 0 or unassigned. (Does it need to be different from FLAG_NO_RECORDED_VALUE?)

opentelemetry/proto/metrics/v1/metrics.proto
jsuereth: Unfortunately, this one is about Prometheus compatibility.   Prometheus histogram sum only allows monotonic sums, and OpenTelemetry can compute histograms with non-monotonic sums.   This bit of OTLP only exists for prometheus compatibility, specifically knowing that sum-in-histogram can be a sum-in-prometheus.   There are alternative ways to solve this that were discarded.    If you read the description of the proto, this change merely brings this in line with what was decided for the v0.9 version of OTLP.

jsuereth:
@tigrannajaryan As of  this v3.15, field presence is part of protocol buffers (no longer experimental), see [these notes](https://github.com/protocolbuffers/protobuf/blob/master/docs/field_presence.md).

> A possible alternate solution is to add a bit to flags to indicate whether the sum is 0 or unassigned. 

We'd actually like to do better, but that PR was rejected for various reasons, see: #320


As a quick note, two important points:  

1. The optional bit is *not* the important part to the fix.  I just needed an optional to demonstrate this issue.
2. The important aspect is whether or not we CAN rely on field-presence to evolve our protocol (see the Min/Max additions to histogram).   This is the most important aspect of this PR and it's blocked on gogo-proto.

opentelemetry/proto/metrics/v1/metrics.proto
anuraaga: Yeah I figured it's something like that - but since SDKs currently don't know if their data will end up in Prometheus, I think either 1) they always fill it or 2) they are able to negotiate what they are required to fill is needed, isn't it?

tigrannajaryan:
Does this need to be distinguishable in the Collector (e.g. converting from OTLP to Prometheus)? We won't be able to implement it since Gogoproto does not support it.

>The important aspect is whether or not we CAN rely on field-presence to evolve our protocol (see the Min/Max additions to histogram). This is the most important aspect of this PR and it's blocked on gogo-proto.

Short-term I think we can't since we rely on Gogoproto in the Collector. Long term, probably we can, assuming we can move away from Gogoproto.

jsuereth:
@tigrannajaryan You missed the collector SiG where @bogdandrutu  and I discussed this.  I'm not willing to make this change until we have a soution for gogoproto.  I'm investigating updating goproto itself (this change actually isn't too hard to add, it's just impossible to submit + merge a PR to gogoproto, so we'd need to fork/vendor)

tigrannajaryan:
@jsuereth SGTM. I removed my block. Let's see if there is a solution with gogoproto.

jmacd:
See #279.
Is it worth waiting so that 0.11 contains both the exponential histogram and min/max? Requires this change first.

jsuereth:
@jmacd Might be worth discussing offline.   Any solution for Go (likely) requires vendoring GoGo proto for a patch allowing optional *or* creating our own proto generation plugin with go-specific optimsiations from GoGo proto.  I'm investigating how big that would be right now.

codeboten:
Update on this, I believe we have a working solution for adding optional fields support in the collector. Please take a look at https://github.com/open-telemetry/opentelemetry-collector/pull/4750 and add comments there if there are any concerns. 

**Note**: that PR was using the proto from https://github.com/open-telemetry/opentelemetry-proto/pull/279, but the same work can be used to support this PR.

bogdandrutu:
@codeboten we really have a good solution now correct?

codeboten:
@bogdandrutu yes the current fix in https://github.com/open-telemetry/opentelemetry-collector/pull/4929 allows the collector to continue using the proto generation moving forward. This is good to move forward 👍

jmacd:
Close in favor of #366.

PR #337, opened 2021-10-04
Add native kotlin support

vincentfree:
Kotlin support has been added to protobuf since version v3.17.0. 
The generated kotlin files have kotlin dal support as an added bonus over direct usage of the java versions of the protobuf .java files


Makefile
```diff
@@ -74,6 +75,14 @@ gen-java:
 	mkdir -p ./$(PROTO_GEN_JAVA_DIR)
 	$(foreach file,$(PROTO_FILES),$(call exec-command, $(PROTOC) --java_out=./$(PROTO_GEN_JAVA_DIR) $(file)))
 
+# Generate gRPC/Protobuf implementation for Kotlin.
+.PHONY: gen-kotlin
+gen-kotlin:
```
arminru, line 80: Thank you! Please also add it here: https://github.com/open-telemetry/opentelemetry-proto/blob/main/.github/workflows/build-check.yaml

vincentfree: Thanks for the suggestion, I have added a job to the workflow for GitHub actions 👍🏽

vincentfree:
I have also updated the `otel/build-protobuf` image from version 0.4.0 to 0.7.0

Makefile
```diff
@@ -15,16 +15,17 @@ endef
 
 # Generate all implementations
 .PHONY: gen-all
-gen-all: gen-cpp gen-csharp gen-go gen-java gen-objc gen-openapi gen-php gen-python gen-ruby
+gen-all: gen-cpp gen-csharp gen-go gen-java gen-objc gen-openapi gen-php gen-python gen-ruby gen-kotlin
 
-OTEL_DOCKER_PROTOBUF ?= otel/build-protobuf:0.4.0
+OTEL_DOCKER_PROTOBUF ?= otel/build-protobuf:0.7.0
```
arminru, line 20: As per https://github.com/open-telemetry/build-tools/releases this should be a no-op change anyway but just to double-check:
@bogdandrutu @tigrannajaryan are we fine updating this? The next version (0.8.0) will also bump the alpine image and Go version (1.15->1.17).

vincentfree: The build failed again so I looked into the image an I see that the version used is quite old: https://github.com/open-telemetry/build-tools/blob/cccfc34393ba3b71ca2956060b7d351b38564945/protobuf/Dockerfile#L5
This version needs to be updated to v3.17.0 at least for this to be able to create the native Kotlin classes.
I can also update the Dockerfile though it might have some implications

bogdandrutu: It is ok to upgrade the image.

bogdandrutu (approved):

Makefile
```diff
@@ -15,16 +15,17 @@ endef
...
 PROTOC := docker run --rm -u ${shell id -u} -v${PWD}:${PWD} -w${PWD} ${OTEL_DOCKER_PROTOBUF} --proto_path=${PWD}
 PROTO_INCLUDES := -I/usr/include/github.com/gogo/protobuf
 
 PROTO_GEN_CPP_DIR ?= $(GENDIR)/cpp
 PROTO_GEN_CSHARP_DIR ?= $(GENDIR)/csharp
 PROTO_GEN_GO_DIR ?= $(GENDIR)/go
 PROTO_GEN_JAVA_DIR ?= $(GENDIR)/java
+PROTO_GEN_KOTLIN_DIR ?= $(GENDIR)/kotlin
```
SergeyKanzhelev, line 28: nit: the list seemed to be alphabetically ordered before. Perhaps we can keep it as such?

tigrannajaryan:
The new github action failed

Makefile
alolita: +1 on keeping the list alphabetical.

vincentfree:
> The new github action failed

I will have to update the `otel/build-protobuf` image before this will pass since the current version of protobuf is quite old and does not support kotlin. I had a PR open for this so I'll fix those issues and hope It will make version 0.9.0. Afterwards  this will also pass.

So this PR is dependend on: https://github.com/open-telemetry/build-tools/pull/76

vincentfree:
I have fixed the issue with the otel image locally. The builds are running and I hope this new functionality can be published soon so this PR can be finished and closed.

bogdandrutu:
@vincentfree dependency merged, what else can I help with?

vincentfree:
@bogdandrutu
Thanks, when the image is published then kotlin support should work and this PR can be updated with the new base image.

I've tested it on my laptop so I'm quite sure that this PR is almost done.

Makefile
```diff
@@ -74,6 +75,14 @@ gen-java:
 	mkdir -p ./$(PROTO_GEN_JAVA_DIR)
 	$(foreach file,$(PROTO_FILES),$(call exec-command, $(PROTOC) --java_out=./$(PROTO_GEN_JAVA_DIR) $(file)))
 
+# Generate gRPC/Protobuf implementation for Kotlin.
+.PHONY: gen-kotlin
+gen-kotlin:
+	rm -rf ./$(PROTO_GEN_KOTLIN_DIR)
+	mkdir -p ./$(PROTO_GEN_KOTLIN_DIR)
+	$(foreach file,$(PROTO_FILES),$(call exec-command, $(PROTOC) --java_out=./$(PROTO_GEN_JAVA_DIR) --kotlin_out=./$(PROTO_GEN_KOTLIN_DIR) $(file)))
```
bogdandrutu, line 83: Can this depend on gen-java instead of re-generating the java files in their dir?

vincentfree: It could but there is a relation between the java and kotlin generated classes. The kotlin file 's refer to those from Java so it might work but I've never seen kotlin generation without the `java_out` option.

vincentfree: The reason why the build failed is because it has a dependency on the newer otel image with my other PR.

bogdandrutu: @vincentfree I understand the build failure, but still not sure about the regenerating of the java files in this rule. I would give a try without to see what happens :D

vincentfree: Doesn't hurt to try

bogdandrutu (commented):
With the latest update you need to change `gen-openapi` as well, no more swagger.

vincentfree:
@bogdandrutu thanks for the new release of `otel/build-protobuf:0.9.0`
Updated the codebase with the new version, I'm now referencing `gen-java` in `gen-kotlin` so proper reuse ;)
And updated the `swagger_out` to `openapiv2_out` everything should be fine now.

I've tested `gen-kotlin` and `gen-openapi` locally. Both work.

Makefile
bogdandrutu: +1

bogdandrutu (approved):

vincentfree:
@arminru your review is required to pass this PR. It's fully ready to merge back to main

CHANGELOG.md
```diff
@@ -10,11 +10,12 @@ Full list of differences found in [this compare](https://github.com/open-telemet
 
 ### Changed
 
-* Remove if no changes for this section before release.
+* `Swagger` generation updated to `openapiv2` due to the use of an updated version of protoc in `otel/build-protobuf`
+* Update the `otel/build-protobuf` version from `0.8.0` to `0.9.0`
```
tigrannajaryan, line 14:
```suggestion
* Update the `otel/build-protobuf` version from `0.4.0` to `0.9.0`
```

Makefile
```diff
@@ -94,13 +103,13 @@ gen-objc:
...
 .PHONY: gen-openapi
 gen-openapi:
 	mkdir -p $(PROTO_GEN_OPENAPI_DIR)
-	$(PROTOC) --swagger_out=logtostderr=true,grpc_api_configuration=opentelemetry/proto/collector/trace/v1/trace_service_http.yaml:$(PROTO_GEN_OPENAPI_DIR) opentelemetry/proto/collector/trace/v1/trace_service.proto
-	$(PROTOC) --swagger_out=logtostderr=true,grpc_api_configuration=opentelemetry/proto/collector/metrics/v1/metrics_service_http.yaml:$(PROTO_GEN_OPENAPI_DIR) opentelemetry/proto/collector/metrics/v1/metrics_service.proto
-	$(PROTOC) --swagger_out=logtostderr=true,grpc_api_configuration=opentelemetry/proto/collector/logs/v1/logs_service_http.yaml:$(PROTO_GEN_OPENAPI_DIR) opentelemetry/proto/collector/logs/v1/logs_service.proto
+	$(PROTOC) --openapiv2_out=logtostderr=true,grpc_api_configuration=opentelemetry/proto/collector/trace/v1/trace_service_http.yaml:$(PROTO_GEN_OPENAPI_DIR) opentelemetry/proto/collector/trace/v1/trace_service.proto
+	$(PROTOC) --openapiv2_out=logtostderr=true,grpc_api_configuration=opentelemetry/proto/collector/metrics/v1/metrics_service_http.yaml:$(PROTO_GEN_OPENAPI_DIR) opentelemetry/proto/collector/metrics/v1/metrics_service.proto
```
tigrannajaryan, line 111: Why are these changes in the PR about Kotlin support? Can they be a separate PR?

vincentfree: No they can't since the version upgrade of grpc changes the implementation from swagger to openapiv2. The base Image needed an update for kotlin support, that forced this change.

vincentfree: I could do it with first only the base Image change and then the kotlin stuff as a separate pr. Though there is a dependency between the two. 
My goal was adding the kotlin generation, a byproduct was this grpc api doc generation change

tigrannajaryan (approved):

PR #342, opened 2021-10-29
Remove deprecated messages and fields from metrics

bogdandrutu:

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -158,6 +158,8 @@ message InstrumentationLibraryMetrics {
 // when the start time is truly unknown, setting StartTimeUnixNano is
 // strongly encouraged.
 message Metric {
+  reserved 4, 6, 8;
+
   // name of the metric, including its DNS name prefix. It must be unique.
   string name = 1;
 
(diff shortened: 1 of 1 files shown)


tigrannajaryan:
Do we have a green flag on the Collector side? Any components still referencing these fields that need to be fixed (other than pdata package)?

jsuereth:
I'd like to be a little more cautious before delivering this.  I think once we have most languages up-to-date on the latest metrics specification.   Specifically, since we asked SiGs to hold off on metrics until spec was more stable, I think we should also delay the deprecation removal policy on these just a bit.

bogdandrutu:
@jsuereth what is the problem if they have not moved?

MrAlias (commented):
From [the stability definitions](https://github.com/open-telemetry/community/blob/47813530864b9fe5a5146f466a58bd2bb94edc72/maturity-matrix.yaml#L79-L82):

> "stable" maturity APIs should not introduce backwards-incompatible changes more than once every twelve months, and will make every effort to provide compatibility bridges if at all possible.

These fields were marked deprecated 8 months ago. Is this change compatible with our stability guarantees?

bogdandrutu:
@MrAlias yes since it was already "deprecated" when we marked the project stable. We can wait 12m if you think that helps, but I doubt.

jmacd (approved):

tigrannajaryan (approved):
LGTM. I don't think we need to wait 12months since deprecation was done before we declared the proto stable (so Beta rules apply in this case, not Stable rules).
I assume we are good on Collector-side since proposed by Bogdan (I didn't check).

MrAlias:
> @MrAlias yes since it was already "deprecated" when we marked the project stable.

SGTM

jsuereth (approved):

PR #343, opened 2021-11-29
Use gogofaster to generate the go code

dmathieu:
This switches the Go code generation from `gogo_out` to `gogofaster_out`.
With this, we also generate methods to unmarshal otlp data.

Getting those unmarshal methods is required to be able to implement an HTTP endpoint which would receive OTLP data transmitted by a library.
The collector already uses it for example: https://github.com/open-telemetry/opentelemetry-collector/blob/main/Makefile#L280

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ gen-csharp:
 gen-go:
 	rm -rf ./$(PROTO_GEN_GO_DIR)
 	mkdir -p ./$(PROTO_GEN_GO_DIR)
-	$(foreach file,$(PROTO_FILES),$(call exec-command,$(PROTOC) $(PROTO_INCLUDES) --gogo_out=plugins=grpc:./$(PROTO_GEN_GO_DIR) $(file)))
+	$(foreach file,$(PROTO_FILES),$(call exec-command,$(PROTOC) $(PROTO_INCLUDES) --gogofaster_out=plugins=grpc:./$(PROTO_GEN_GO_DIR) $(file)))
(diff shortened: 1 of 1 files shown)


tigrannajaryan (commented):
What's the purpose of this change?
The `gen-go` target is run for the purpose of verifying that the generation for Go is successful. We don't keep the generated code, all generated code is discarded after the run. Is there any particular reason to believe that it is more important to verify gogofaster generation than the regular one?
Note that the Collector does its own generation and stores the generated files. It does not depend on generation here.

dmathieu:
Even though this project doesn't use the generated code itself, it can be easier for any project that wishes to build the code to rely on this command instead of having to duplicate code.

bogdandrutu:
> Getting those unmarshal methods is required to be able to implement an HTTP endpoint which would receive OTLP data transmitted by a library.

This is wrong because even the official Google protobuf implementation allows you to Marshal/Unmarshal, but you need to call `proto.Marshal/Unmarshal` see https://pkg.go.dev/github.com/golang/protobuf/proto#Unmarshal and https://pkg.go.dev/github.com/golang/protobuf/proto#Marshal

dmathieu:
It doesn't, at least for services, which don't implement `protoreflect.ProtoMessage`.

> server/http_handler.go:35:31: cannot use &orig (type *"github.com/dmathieu/opentelemetry-proto-go/collector/trace/v1".ExportTraceServiceRequest) as type protoreflect.ProtoMessage in argument to "google.golang.org/protobuf/proto".Unmarshal:
        *"github.com/dmathieu/opentelemetry-proto-go/collector/trace/v1".ExportTraceServiceRequest does not implement protoreflect.ProtoMessage (missing ProtoReflect method)

PR #346, opened 2021-12-13
Add Parquet support

atoulme:
Depends on https://github.com/open-telemetry/build-tools/pull/78 being merged.

This adds Parquet schema output from the protobuf model of logs, metrics traces.

Example of log Parquet schema output:
```
message logsdata {
  optional group resource_logs (LIST) {
    repeated group list {
      optional group element {
        optional group resource {
          optional group attributes (LIST) {
            repeated group list {
              optional group element {
                optional binary key (UTF8);
                optional group value {
                  optional binary string_value (UTF8);
                  optional boolean bool_value;
                  optional int64 int_value;
                  optional double double_value;
                  optional binary array_value (JSON);
                  optional binary kvlist_value (JSON);
                  optional binary bytes_value;
                }
              }
            }
          }
          optional uint32 dropped_attributes_count;
        }
        optional group instrumentation_library_logs (LIST) {
          repeated group list {
            optional group element {
              optional group instrumentation_library {
                optional binary name (UTF8);
                optional binary version (UTF8);
              }
              optional group log_records (LIST) {
                repeated group list {
                  optional group element {
                    optional uint64 time_unix_nano;
                    optional binary severity_number (UTF8);
                    optional binary severity_text (UTF8);
                    optional binary name (UTF8);
                    optional group body {
                      optional binary string_value (UTF8);
                      optional boolean bool_value;
                      optional int64 int_value;
                      optional double double_value;
                      optional binary array_value (JSON);
                      optional binary kvlist_value (JSON);
                      optional binary bytes_value;
                    }
                    optional group attributes (LIST) {
                      repeated group list {
                        optional group element {
                          optional binary key (UTF8);
                          optional group value {
                            optional binary string_value (UTF8);
                            optional boolean bool_value;
                            optional int64 int_value;
                            optional double double_value;
                            optional binary array_value (JSON);
                            optional binary kvlist_value (JSON);
                            optional binary bytes_value;
                          }
                        }
                      }
                    }
                    optional uint32 dropped_attributes_count;
                    optional uint32 flags;
                    optional binary trace_id;
                    optional binary span_id;
                  }
                }
              }
              optional binary schema_url (UTF8);
            }
          }
        }
        optional binary schema_url (UTF8);
      }
    }
  }
}


```


Makefile
```diff
@@ -102,6 +105,13 @@ gen-openapi:
 	$(PROTOC) --swagger_out=logtostderr=true,grpc_api_configuration=opentelemetry/proto/collector/metrics/v1/metrics_service_http.yaml:$(PROTO_GEN_OPENAPI_DIR) opentelemetry/proto/collector/metrics/v1/metrics_service.proto
 	$(PROTOC) --swagger_out=logtostderr=true,grpc_api_configuration=opentelemetry/proto/collector/logs/v1/logs_service_http.yaml:$(PROTO_GEN_OPENAPI_DIR) opentelemetry/proto/collector/logs/v1/logs_service.proto
 
+# Generate gRPC/Protobuf implementation for Parquet.
+.PHONY: gen-parquet
+gen-parquet:
+	rm -rf ./$(PROTO_GEN_PARQUET_DIR)
+	mkdir -p ./$(PROTO_GEN_PARQUET_DIR)
+	$(foreach file,$(PROTO_FILES),$(call exec-command, $(PROTOC) --parquet_out=go_file=true$(comma)output_messages=TracesData$(semicolon)LogsData$(semicolon)MetricsData$(comma)max_recursion=3:./$(PROTO_GEN_PARQUET_DIR) $(file)))
```
bogdandrutu, line 113: This "comma" and "semicolon" trick can be done consistent across all gen and in a separate PR.

atoulme:
I have commented on this issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/2835

Makefile
atoulme: That's just needed as a pretty bad crutch for this foreach instruction. There might be a better way around.

atoulme: this crutch is needed right now or the generation fails, so leaving this in for now.

achille-roussel:
Hello @atoulme

I am chiming in because we are probably working on similar problems, and I wanted to better understand how you're thinking of the end solution; we are experimenting with opentelemetry as a backbone for a lot of our observability solutions and I would like to make sure we don't diverge too far off from what will eventually be the opentelemetry standard.

From what I grasp of the discussions in [#2835](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/2835), https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6903, and here, it seems the proposed addition of a parquet format would be the schema used in files written to object storage like S3, with the intent being to support using tools like Athena to explore the data set.

While I see the value of using a direct protobuf-to-parquet translation (tooling and documentation probably exists, it creates opportunities for interoperability, etc...), I am also wondering how to gauge the cost of delegating most of the complexity of the direct translation to the end user: the deeply nested structure will make queries extremely complex to write, likely requiring another layer of tooling to enable users to leverage the dataset (e.g. Athena views or custom connectors).

To illustrate, in the parquet definition for `logsdata` that you shared, the `attribute` field is a translation of the protobuf union type [AnyValue](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/common/v1/common.proto#L24-L39). A query that would filter by attributes would need to know the underlying type that was used (or types if the application evolved its schema). Normalizing the values into string representations or using a different structure could be a design decision that would result in having data sets that are a lot more practical to query.

I am also wondering whether the use of `optional` on every field might create unnecessary complexity and yield inefficiencies. Some fields in the protobuf definition are not optional (e.g. trace ids), marking them as such in parquet would require the exporter to encode unnecessary definition levels in the parquet file, which have to be decoded and interpreted by query engines. The deeply nested structure also results in high max repetition and definition levels, the depth of 29 fields in the `logsdata` example means the parquet representation of definition and repetition levels will probably be twice larger than what could be achieved with a more compact representation.

The question I would like to discuss is whether some of the complexity could be absorbed by the exporter, designing translation rules between the protobuf and parquet representation that optimize for a "best-of-both-worlds" scenario. I recognize we should avoid early optimizations, tho when petabytes of this data have been written to the cloud, changing the format will be a very costly operation. Maybe there is prior art that we can draw parallels to in order to strike the right balance?

Also, if this is not the right place for this discussion, I'm happy to engage on a different channel, let me know!

atoulme:
@achille-roussel it's a good time to discuss this and I was hoping to precisely generate this type of discussion with this well meaning, but certainly ham-handed effort. Thank you for writing this well researched, thorough comment.

I am learning about Apache Parquet myself and I googled a bit around - you're right, given that the protobuf model is recursive, the Parquet model looks like a beast and is probably not the most effective.

To get started with this effort, I googled around how to best handle recursion with Parquet - seems like allowing for a max recursion depth is the solution of least effort. So I went with that.

I am not sure how to go about it otherwise. It looks like we could potentially flatten the structure of the data when we write it, but that would also create mass duplication. I am keen and interested in hearing the best approach on mapping those models.

It's very likely that what we decide here goes to live in the opentelemetry-specification repository, so the mapping is documented. While the conversation is early and we're trying to piece together what to do, we can talk here, and then move to draft in the spec based off the requirements.

atoulme:
BTW, the data model of OpenTelemetry is not final yet.
Whatever mapping we come up with may change based off changes to the data model.
If we design an explicit mapping, it will need to update based off changes to Protobuf.
We will need to version the Parquet data so we can read back data and map it to pdata - Protobuf according to the mapping version.

achille-roussel:
> I am not sure how to go about it otherwise. It looks like we could potentially flatten the structure of the data when we write it, but that would also create mass duplication. I am keen and interested in hearing the best approach on mapping those models.

I agree that we should stay as close to the OpenTelemetry format as practical, the attribute list first came to mind as a potential area for improvement because the spec allows some [flexibility with the representation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attribute):

> An Attribute is a key-value pair, which MUST have the following properties:
>
> The attribute key MUST be a non-null and non-empty string.
> The attribute value is either:
> * A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer.
> * An array of primitive type values. The array MUST be homogeneous, i.e., it MUST NOT contain values of different types. For protocols that do not natively support array values such values SHOULD be represented as JSON strings.

As for next steps, I can make a few suggestions, let me know what you think about it:
* default to `required` fields in parquet for OpenTelemetry fields unless they are documented as optional in the spec
* represent array values in attributes as binary parquet values annotated with the JSON logical type, this way we can play along with engines that have support for querying JSON values in parquet, but don't have to deal with the complexity of representing repeating nested values of heterogenous types

Assuming the suggestions above, this would be the schema of attributes:

```
required group attributes (LIST) {
  repeated group list {
    required group element {
      required binary key (UTF8);
      required group value {
        optional binary string_value (UTF8);
        optional boolean bool_value;
        optional int64 int_value;
        optional double double_value;
        optional binary array_value (JSON);
      }
    }
  }
}
```

We could also consider an alternative implementation which uses the `MAP` annotation and may be more semantically meaningful; it is also more compact:

```
required group attributes (MAP) {
  repeated group key_value {
    required binary key (UTF8);
    required group value {
      optional binary string_value (UTF8);
      optional boolean bool_value;
      optional int64 int_value;
      optional double double_value;
      optional binary array_value (JSON);
    }
  }
}
```

I am having trouble finding details about whether OpenTelemetry attribute lists are supposed to support keys being present multiple times, the [protobuf definition](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L138) appears to indicate it would be OK since attributes are represented as a `KeyValueList`, however the layout is also wire-compatible with the way maps are encoded in protobuf so decoders could assume uniqueness of attribute keys and successfully decode OpenTelemetry objects without knowing they have dropped data.

On the parquet side, while the nested map format technically allows for multiple keys to exist, [the guidelines](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#maps) are suggesting it should not be used to encode such data structures, and describes how processors should handle multiple occurrences of the same key:

> If there are multiple key-value pairs for the same key, then the final value for that key must be the last value. Other values may be ignored or may be added with replacement to the map container in the order that they are encoded. The MAP annotation should not be used to encode multi-maps using duplicate keys.

Thankfully the protobuf spec [requires the same behavior](https://developers.google.com/protocol-buffers/docs/proto3#maps):

> When parsing from the wire or when merging, if there are duplicate map keys the last key seen is used. When parsing a map from text format, parsing may fail if there are duplicate keys.

On the OpenTelemetry side, there is a mention of _"unique attribute keys"_ in the [Attribute Limits](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attribute-limits) specification, but it is unclear if _"unique"_ refers to keys not appearing multiple times or just the number of keys in the attribute list.

I have the intuition that it's not worth the complexity to support repeating keys and we would be better off assuming/requiring unique keys. An update to the OpenTelemetry specification to explicitly declare whether attribute keys must be unique could be useful to help guide the design decision here.

atoulme:
That's helpful, thank you.

Right now, the Parquet schema is derived directly from the Protobuf schema. I think that's a strength and something to try to preserve if possible. It looks like from your analysis that specific structures would need to be mapped out differently. It might be worth trying to see how to annotate the model to generate exactly the right Parquet schema.

I believe it would be worthwhile to follow up with a question around key unicity, would you like to file the issue?

achille-roussel:
> Right now, the Parquet schema is derived directly from the Protobuf schema. I think that's a strength and something to try to preserve if possible.

I agree, I think what I was trying to highlight is the current tool used for translation may be creating a schema that ends up being sub-optimal to query; there are trade offs we can make that could yield more value out of parquet schema for OpenTelemetry, especially optimizing for querying or data exploration in my opinion. This can be solved by writing custom code to translate from protobuf to parquet, or modifying/improving the tool that performs the conversion so it better supports the use case.

All that being said, it can also be fine to take a first step to learn what works well or not with the existing tooling, then iterate from there.

> I believe it would be worthwhile to follow up with a question around key unicity, would you like to file the issue?

Will do 👍

atoulme:
Funny enough, I actually have the ability to modify the tool. I modified it to handle recursion. So we can fix up the tool as we go.

bogdandrutu:
@atoulme I think the dependent PR was merged, what does this depend on now? Is everything ready?

atoulme:
@bogdandrutu it depends on changing the way we create the schema so it no longer is using a deeply nested structure, like @achille-roussel pointed out. This might require us to annotate the protobuf model so the generator knows how to interpret structures.

atoulme:
@achille-roussel please review this change https://github.com/atoulme/protoc-gen-parquet/pull/1

It will add JSON support. If/when merged, we can make a new release of the generator, and use it in build-tools to generate the schema. Two fields in the `AnyValue` message will need to be annotated:

```
@@ -32,8 +33,8 @@ message AnyValue {
     bool bool_value = 2;
     int64 int_value = 3;
     double double_value = 4;
-    ArrayValue array_value = 5;
-    KeyValueList kvlist_value = 6;
+    ArrayValue array_value = 5 [(parquet_options.field_opts).is_json = true];
+    KeyValueList kvlist_value = 6 [(parquet_options.field_opts).is_json = true];
     bytes bytes_value = 7;
   }
 }
```

atoulme:
I have updated the PR ^^ see at the top for the new schema generated.

atoulme:
This PR now depends on https://github.com/open-telemetry/build-tools/pull/81 to be merged first.

achille-roussel:
I like the addition of `[(parquet_options.field_opts).is_json = true]`, seems like a great way to have the tool automatically generate the conversion back and forth, while also encoding the data in a way that can better serve other consumers.

We continued experimenting with this internally as well, and what we are landing on to represent attributes is actually this structure:

```
  optional group attributes (MAP) {
    repeated group key_value {
      required binary key (UTF8);
      required binary value (JSON);
    }
  }
```

In our case, we wanted to avoid having to encode null values for 5 of the 6 columns that represent the union, while also avoiding to push the complexity of knowing which column the value is encoded in. JSON can represent all types of the union here (primitives + array of primitives) and we have only a single `value` column, it suits the requirements.

I'm not necessarily suggesting that we follow the same route here, just wanted to contribute more info to the conversation in case it gives new perspectives.

atoulme:
OK, how about this - we have something that satisfies you right now, and we can iterate over it to match your improvement. For now, we can merge what we have and open a new PR to discuss your idea further. Does that work for you?

atoulme:
This is almost there. My PR to build-tools was merged, we need a new release of the build-tools docker image, and then we can make this PR pass here. Then it will be time to merge and iterate.

opentelemetry/proto/common/v1/common.proto
```diff
@@ -15,6 +15,7 @@
 syntax = "proto3";
 
 package opentelemetry.proto.common.v1;
+import "options/parquet_options.proto";
```
bogdandrutu, line 18: I think this will affect everyone who is not using the docker container to build the protos (using something like a build tool specific for the language, e.g. gradle). Can we ask the maintainers if this will cause them troubles?

atoulme: True, how do we ask maintainers? Is there an alternative we could explore here?

anuraaga: Yeah this will break the Java build, and I guess it should be a goal for this repo to at least be able to build with common proto tools, not requiring build-tools just for a simple build of the OTLP protocol.

Since the parquet protoc only implements this functionality in a personal fork, does it make sense to fork it into opentelemetry itself, and let it recognize a `.proto` file within this repo?

Alternatively, `sed` on the outputing schema files to add `(JSON)` to these two lines may be the most practical if avoiding such a fork, hacky but would work I guess

atoulme: @anuraaga happy to push this generator to opentelemetry. It's not a personal fork anymore, the original maintainer stopped helping and I took it on and made releases. It could use (more) tests.

No, you can't just add JSON to the schema lines, this is also stopping the generator from going into a recursive infinite loop.

But we could add the annotations to the proto files via sed if you like, that's a fine way to go around it.

Could we inline the options package into opentelemetry-proto otherwise? No harm if we have that checked in?

anuraaga: @atoulme Yeah if the referenced proto file is in this repo then there shouldn't be any issues.

atoulme: It might look kludgey but if you’re ok with it that should be a quick fix.

atoulme: done!

opentelemetry/proto/common/v1/common.proto
```diff
@@ -21,6 +21,36 @@ option java_package = "io.opentelemetry.proto.common.v1";
 option java_outer_classname = "CommonProto";
 option go_package = "go.opentelemetry.io/proto/otlp/common/v1";
 
+// inlined parquet-gen options
+import "google/protobuf/descriptor.proto";
+
+message MessageOptions {
```
anuraaga, line 27: Chatted with @atoulme offline and thought that since whether a field is encoded as JSON or not is relatively generic it could be put in this file, for potential use in other signal-specific protos too.

But I hadn't looked well at the proto file, there is more than I was expecting and this table_name especially does seem quite parquet specific. @atoulme do you intend to use the features other than `is_json`? If not we should scope down to just that

atoulme: Definitely

atoulme: @anuraaga ptal should be good now.

anuraaga (approved):

atoulme:
Latest issue depending on https://github.com/open-telemetry/build-tools/pull/88 to be merged.

opentelemetry/proto/common/v1/common.proto
```diff
@@ -21,6 +21,36 @@ option java_package = "io.opentelemetry.proto.common.v1";
...
+  // As long as table_name is not blank,
+  // a schema is generated for top-level messages in each file.
+  string table_name = 1;
+}
+
+extend google.protobuf.MessageOptions {
+  MessageOptions message_opts = 1023;
+}
+
+enum TimestampType {
```
jmacd, line 37: How will `TimestampType` be used? It seems to be not used in this PR.

atoulme: it's not used right now but it's the first option inherited by the parquet generator (`1` below) and therefore I figured it would make sense to bring it along so `is_json` is not strangely assigned `2`.

atoulme: I would assume the timestampType will come handy pretty quick too.

atoulme:
Release of build-tools is out, this PR is ready for review and merge.

tigrannajaryan:
@atoulme I don't know anything about Parquet and I don't see any explanation why this change is necessary. Please add the explanation in the PR description or create an issue that explains the problem being solved and link from this PR.

atoulme:
@tigrannajaryan I have linked this PR to this ticket https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/2835.

Apache Parquet is a columnar data format well suited to drop data into S3, where it becomes easily queryable by different tools. The idea is to create a Parquet exporter and receiver (exporter is here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6903) that will allow to push data into Parquet format.

tigrannajaryan:
If I understand correctly this introduces a new format for representing OTLP data (in addition to Protobufs and JSON). If this is the case we should state in README that this is experimental and the format is subject to breaking changes.

atoulme:
Done, added mention.

opentelemetry/proto/common/v1/common.proto
tigrannajaryan, line 27: Please add a comment to explain the purpose of this message or other new messages introduced in this PR.

opentelemetry/proto/common/v1/common.proto
```diff
@@ -21,6 +21,36 @@ option java_package = "io.opentelemetry.proto.common.v1";
...
+// inlined parquet-gen options
+import "google/protobuf/descriptor.proto";
+
+message MessageOptions {
+  // As long as table_name is not blank,
+  // a schema is generated for top-level messages in each file.
+  string table_name = 1;
+}
+
+extend google.protobuf.MessageOptions {
```
tigrannajaryan, line 33: Add a comment explain the purpose of the extending the message and of the field inside.

opentelemetry/proto/common/v1/common.proto
tigrannajaryan, line 37: Why do we need different types? OTLP only uses nanos in its proto. Please clarify the purpose.

opentelemetry/proto/common/v1/common.proto
```diff
@@ -32,8 +62,8 @@ message AnyValue {
     bool bool_value = 2;
     int64 int_value = 3;
     double double_value = 4;
-    ArrayValue array_value = 5;
-    KeyValueList kvlist_value = 6;
+    ArrayValue array_value = 5 [(field_opts).is_json = true];
```
tigrannajaryan, line 65: What does this annotation do?

atoulme: This annotation tells the parquet schema generator to stop recursing and treat the field as a JSON data blob, to be serialized as one column. Otherwise, we are stuck in a recursive loop with parquet, which requires a flat schema.

opentelemetry/proto/common/v1/common.proto
atoulme: it's a holdover from bringing in the parquet gen field options. The generator defines a timestamp type to define how to serialize fields with parquet.

opentelemetry/proto/common/v1/common.proto
atoulme: Well, there's a comment line 24 -- I guess what you're saying is that it's not cutting it

tigrannajaryan: I think it would be great to have more comments to explain what is going on here. I am not familiar with Parquet and probably some other contributors to this repos also aren't. We need some help to understand the change and also be able to maintain it in the future.

atoulme: OK I have added more information.

opentelemetry/proto/common/v1/common.proto
tigrannajaryan: Can you also add this explanation to the code? More details will be event better, since I am not sure why it is supposed to recurse and why it needs to stop.

opentelemetry/proto/common/v1/common.proto
tigrannajaryan: Why do we need a "holdover" here? I don't understand.

tigrannajaryan:
@atoulme sorry, I know it took a long time, but I don't see how this can be merged in the current form. This requires _a lot more_ explanations before I can approve it. It is still far from being clear to me.

However, I realize it may be just me, with my lack of experience with Parquet. If there are others who understand this change better and are willing to maintain this part of the proto themselves then they are free to review/approve and take the responsibility.

atoulme:
@achille-roussel would you feel like helping out? Alternatively, I'll try and bring this up in a WG conf call so we can talk out what's unclear.

achille-roussel:
I don't currently have time to dedicate to this, tho I am continuing to pursue a similar work internally which I hope will be useful data points into this OpenTelemetry initiative.

I believe Grafana's proposal also walks in this direction and would provide useful learnings https://github.com/grafana/tempo/pull/1407

jmacd:
@atoulme I appreciate your help driving this.

 'm eager to understand how OTLP and Parquet integration could work because I would like a queryable long-term storage format. I wanted to create a Parquet metrics exporter just to try this out.

I followed the discussion above and it leads to questions I couldn't answer. I'm using the github.com/apache/arrow/go package since that looks like a direct route to an OTel collector integration. However, it's not clear this library is mature compared with the Rust and Java ecosystem around Arrow. For example, I couldn't find a way to _parse_ an arrow schema. I think this means I couldn't yet use an Arrow integration from Go without coding an OTLP metrics schema by hand, which seems like a small task compared with the effort to create a standard Parquet format via the protocol compiler. As the discussion in this PR shows, there are lots of ways to organize a Parquet schema. 

I liked the point made by @achille-roussel about simplifying the attribute structure, https://github.com/open-telemetry/opentelemetry-proto/pull/346#issuecomment-1034042580. I tried to follow this recipe in my coded-by-hand Arrow schema, but again could not discover the required support in the Go library. How do I mark a BINARY data object as JSON, for example? I am considering that for most applications, metric and resource attributes are string valued, so having a BINARY (JSON) in this position adds a lot of headache (especially if the library doesn't support it yet).

In short, I think I'm able to answer the questions raised by other reviewers above, but couldn't answer my own questions. If we accept this PR or one like it, what's the next step toward an OTel Collector / Apache Arrow integration? Should there be _multiple_ representations in Parquet, since applications are different? Lastly, we need to compare and contrast with the work done by @lquerel in https://github.com/open-telemetry/oteps/pull/171.

lquerel:
@jmacd and @atoulme I also think that OTLP attributes are a bit too complicated but I decided in OTEP 0156 to support them in their entirety even if it may require some improvements to the Arrow Go library.  A new version of the [specification](https://github.com/lquerel/oteps/blob/main/text/0156-columnar-encoding.md) can be found below as well as the corresponding [PR](https://github.com/open-telemetry/oteps/pull/171). 

I'm working on a library to translate automatically OTEL entity batches into Arrow RecordBatch in the context of this OTEP. I hope to make it public this summer. This will not only open the door to a complete implementation of OTEP 0156, but will also provide access to the entire Arrow ecosystem to save data in the parquet format, for example.

atoulme:
This PR seems to have a smaller scope of what is presented in https://github.com/open-telemetry/oteps/pull/171 - I would be in favor of closing it and working on the OTEP to see how it maps out.

PR #348, opened 2022-01-04
Add support for buf

remyleone:
Hello 👋  this add support for pushing the [Buf Schema Registry](https://buf.build/explore). Could you also reserve the namespace `buf.build/opentelemetry` ? This would be useful to have a simple import such as `buf.build/opentelemetry/opentelemetry` available for all buf users.


remyleone:
https://github.com/open-telemetry/opentelemetry-proto/issues/349

johanbrandhorst:
FYI, to support https://buf.build/envoyproxy/envoy, Buf has made https://buf.build/opentelemetry/opentelemetry a managed third party library on the BSR, so there's no need for this unless you're interested in adopting the maintenance of the library. I think this PR can be closed.

PR #351, opened 2022-01-07
Add ObservedTimestamp to LogRecord

tigrannajaryan:
This implements the log data model specification change https://github.com/open-telemetry/opentelemetry-specification/pull/2184


opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -116,6 +116,23 @@ message LogRecord {
   // Value of 0 indicates unknown or missing timestamp.
   fixed64 time_unix_nano = 1;
 
+  // Time when the event was observed by the collection system.
+  // For events that originate in OpenTelemetry (e.g. using OpenTelemetry Logging SDK)
+  // this timestamp is typically set at the generation time and is equal to Timestamp.
+  // For events originating externally and collected by OpenTelemetry (e.g. using
+  // Collector) this is the time when OpenTelemetry's code observed the event measured
+  // by the clock of the OpenTelemetry code. This field SHOULD be set once the event is
```
djaglowski, line 124: Can we strengthen the requirement here? My understanding is that this field is intended to be required.
```suggestion
  // by the clock of the OpenTelemetry code. This field MUST be set once the event is
```

tigrannajaryan: Yes, I think that's a good suggestion. Changed.

djaglowski (approved):

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -116,6 +116,23 @@ message LogRecord {
...
+  // observed by OpenTelemetry.
+  //
+  // For converting OpenTelemetry log data to formats that support only one timestamp or
+  // when receiving OpenTelemetry log data by recipients that support only one timestamp
+  // internally the following logic is recommended:
+  //   - Use time_unix_nano if it is present, otherwise use observed_time_unix_nano.
+  //
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
+  // Value of 0 indicates unknown or missing timestamp.
+  fixed64 observed_time_unix_nano = 11;
```
bogdandrutu, line 134: Questions:
* Should this be optional when same with timestamp?
* Why is this only important for logs and not for spans/metrics data points? Idea is, should this be added to all signals?

tigrannajaryan: >Should this be optional when same with timestamp?

We could do that, but I think it is simpler to just mandate it. It is fine to have both timestamps to be the same. It will compress nicely with gzip, so not a problem for on the wire size.

>Why is this only important for logs and not for spans/metrics data points? Idea is, should this be added to all signals?

It is important for logs because logs often come from legacy sources and are parsed and parsing is not always a very robust and reliable way to extract the timestamp (since parsers may be user-defined and mistakes are possible). This second timestamp helps detect these bad parsing situations. There is a bit more details here https://github.com/open-telemetry/opentelemetry-specification/issues/1875

I do not think it is applicable for metrics and traces because metrics and traces typically have far better defined formats and protocols and converting timestamps from such formats to OTLP is a fairly reliable process.

bogdandrutu: @tigrannajaryan for metrics is kind of a same, when you use a "pull/scrape" protocol correct? Probably for every signal in case of scraping/pulling.

tigrannajaryan: @bogdandrutu Do you refer to pull/scrape using something like Prometheus text format protocol? In that case I think the answer is no, it is not the same since Prometheus wire format is well-defined with the timestamp field having precise parsing rules in text format that do not need to be user-specified and be subjected to a mistake.

Perhaps there are other pull/scrape protocols which I am not aware of which have ambiguous timestamps and are susceptible to the same timestamp parsing problem as text log formats? In that case you may be right.

djaglowski: > This second timestamp helps detect these bad parsing situations.

To expand on the use case, there are at least two other scenarios where we can expect not to see a timestamp extracted from a log event.
1. Quite commonly users will ingest logs without attempting to parse a timestamp from them.
2. In some rare cases, legacy log formats [do not contain a timestamp](https://bugzilla.redhat.com/show_bug.cgi?id=1223792).

zenmoto (approved):

tigrannajaryan:

yurishkuro (approved):

kumoroku (approved):

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: @bogdandrutu any other concerns or we can merge?

tigrannajaryan:
@open-telemetry/opamp-spec-approvers final call. This has enough approvals, I will merge this unless there are objections.

pmm-sumo (approved):

Issue #355, opened 2022-02-01
Generating C# from protos results in uncompilable code


alanwest:
The following results in a property name `Equals` which conflicts with the [`Equals`](https://docs.microsoft.com/en-us/dotnet/api/system.object.equals) method present on the base type that all types in C# derive from.

https://github.com/open-telemetry/opentelemetry-proto/blob/92d15b79b5f24c711cf2c00fee366904069bcb6b/opentelemetry/proto/metrics/experimental/metrics_config_service.proto#L71-L76

Can `equals` be renamed to `equal_to` or something to that affect?

tigrannajaryan:
I think it is OK to do it. This is in the experimental part, we are free to make changes there.

alanwest:
Thanks, resolved by #359

PR #356, opened 2022-02-01
Rename equals to equals_to

tigrannajaryan:
Fixes https://github.com/open-telemetry/opentelemetry-proto/issues/355

In code generated for C# the property name Equals conflicts with the Equals
method present on the base type that all types in C# derive from. This
results in uncompilable code.

### DIFF
--- a/opentelemetry/proto/metrics/experimental/metrics_config_service.proto
+++ b/opentelemetry/proto/metrics/experimental/metrics_config_service.proto
@@ -70,7 +70,7 @@ message MetricConfigResponse {
     // match against metric names. It should not exceed 100k characters.
     message Pattern {
       oneof match {
-        string equals = 1;       // matches the metric name exactly
+        string equals_to = 1;       // matches the metric name exactly
         string starts_with = 2;  // prefix-matches the metric name
       }
     }


anuraaga (commented):
This file doesn't seem to have an owner to drive it through - is it a good time to consider deleting the file instead of having maintenance complications like this one?

cijothomas (approved):

opentelemetry/proto/metrics/experimental/metrics_config_service.proto
```diff
@@ -70,7 +70,7 @@ message MetricConfigResponse {
     // match against metric names. It should not exceed 100k characters.
     message Pattern {
       oneof match {
-        string equals = 1;       // matches the metric name exactly
+        string equals_to = 1;       // matches the metric name exactly
```
arminru, line 73:
```suggestion
        string equal_to = 1;     // matches the metric name exactly
```
"equal to" would be grammatically correct, right?

tigrannajaryan:
> This file doesn't seem to have an owner to drive it through - is it a good time to consider deleting the file instead of having maintenance complications like this one?

Sounds good to me. I am not aware of any usages of these experimental messages by anyone.

tigrannajaryan:
If https://github.com/open-telemetry/opentelemetry-proto/pull/359 is merged this PR can be discarded.

tigrannajaryan:
https://github.com/open-telemetry/opentelemetry-proto/pull/359 is merged. This is no longer necessary.

PR #362, opened 2022-02-22
Rename InstrumentationLibrary to InstrumentationScope

tigrannajaryan:
This applies changes described in spec PR https://github.com/open-telemetry/opentelemetry-specification/pull/2276

This is not a breaking change for OTLP. The wire format does not change.

The change is done in a way that the transition can be handled gracefully for OTLP/JSON and for the generated code. See the comments for explanation on how to do it in the senders and receivers.

### DIFF
--- a/opentelemetry/proto/common/v1/common.proto
+++ b/opentelemetry/proto/common/v1/common.proto
@@ -65,10 +65,10 @@ message KeyValue {
   AnyValue value = 2;
 }
 
-// InstrumentationLibrary is a message representing the instrumentation library information
+// InstrumentationScope is a message representing the instrumentation scope information
 // such as the fully qualified name and version. 
-message InstrumentationLibrary {
-  // An empty instrumentation library name means the name is unknown. 
+message InstrumentationScope {
+  // An empty instrumentation scope name means the name is unknown.
   string name = 1;
   string version = 2;
 }
--- a/opentelemetry/proto/common/v1/common.proto
+++ b/opentelemetry/proto/common/v1/common.proto
@@ -72,3 +72,14 @@ message InstrumentationScope {
   string name = 1;
   string version = 2;
 }
+
+// InstrumentationLibrary is a message representing the instrumentation library information
+// such as the fully qualified name and version.
+// InstrumentationLibrary is wire-compatible with InstrumentationScope for binary
+// Protobuf format.
+// This message is deprecated and will be removed on June 15, 2022.
+message InstrumentationLibrary {
+  // An empty instrumentation library name means the name is unknown.
+  string name = 1;
+  string version = 2;
+}
--- a/opentelemetry/proto/common/v1/common.proto
+++ b/opentelemetry/proto/common/v1/common.proto
@@ -65,14 +65,6 @@ message KeyValue {
   AnyValue value = 2;
 }
 
-// InstrumentationScope is a message representing the instrumentation scope information
-// such as the fully qualified name and version. 
-message InstrumentationScope {
-  // An empty instrumentation scope name means the name is unknown.
-  string name = 1;
-  string version = 2;
-}
-
 // InstrumentationLibrary is a message representing the instrumentation library information
 // such as the fully qualified name and version.
 // InstrumentationLibrary is wire-compatible with InstrumentationScope for binary
@@ -80,6 +72,14 @@ message InstrumentationScope {
 // This message is deprecated and will be removed on June 15, 2022.
 message InstrumentationLibrary {
   // An empty instrumentation library name means the name is unknown.
+  string name = 1;
+  string version = 2;
+}
+
+// InstrumentationScope is a message representing the instrumentation scope information
+// such as the fully qualified name and version. 
+message InstrumentationScope {
+  // An empty instrumentation scope name means the name is unknown.
   string name = 1;
   string version = 2;
 }
--- a/opentelemetry/proto/common/v1/common.proto
+++ b/opentelemetry/proto/common/v1/common.proto
@@ -71,6 +71,8 @@ message KeyValue {
 // Protobuf format.
 // This message is deprecated and will be removed on June 15, 2022.
 message InstrumentationLibrary {
+  option deprecated = true;
+
   // An empty instrumentation library name means the name is unknown.
   string name = 1;
   string version = 2;
--- a/opentelemetry/proto/logs/v1/logs.proto
+++ b/opentelemetry/proto/logs/v1/logs.proto
@@ -43,27 +43,26 @@ message LogsData {
   repeated ResourceLogs resource_logs = 1;
 }
 
-// A collection of InstrumentationLibraryLogs from a Resource.
+// A collection of ScopeLogs from a Resource.
 message ResourceLogs {
   // The resource for the logs in this message.
   // If this field is not set then resource info is unknown.
   opentelemetry.proto.resource.v1.Resource resource = 1;
 
-  // A list of InstrumentationLibraryLogs that originate from a resource.
-  repeated InstrumentationLibraryLogs instrumentation_library_logs = 2;
+  // A list of ScopeLogs that originate from a resource.
+  repeated ScopeLogs scope_logs = 2;
 
   // This schema_url applies to the data in the "resource" field. It does not apply
-  // to the data in the "instrumentation_library_logs" field which have their own
-  // schema_url field.
+  // to the data in the "scope_logs" field which have their own schema_url field.
   string schema_url = 3;
 }
 
-// A collection of Logs produced by an InstrumentationLibrary.
-message InstrumentationLibraryLogs {
-  // The instrumentation library information for the logs in this message.
-  // Semantically when InstrumentationLibrary isn't set, it is equivalent with
-  // an empty instrumentation library name (unknown).
-  opentelemetry.proto.common.v1.InstrumentationLibrary instrumentation_library = 1;
+// A collection of Logs produced by a Scope.
+message ScopeLogs {
+  // The instrumentation scope information for the logs in this message.
+  // Semantically when InstrumentationScope isn't set, it is equivalent with
+  // an empty instrumentation scope name (unknown).
+  opentelemetry.proto.common.v1.InstrumentationScope scope = 1;
 
   // A list of log records.
   repeated LogRecord log_records = 2;
--- a/opentelemetry/proto/logs/v1/logs.proto
+++ b/opentelemetry/proto/logs/v1/logs.proto
@@ -52,6 +52,35 @@ message ResourceLogs {
   // A list of ScopeLogs that originate from a resource.
   repeated ScopeLogs scope_logs = 2;
 
+  // A list of InstrumentationLibraryLogs that originate from a resource.
+  // This field is deprecated and will be removed after grace period expires on June 15, 2022.
+  //
+  // During the grace period the following rules SHOULD be followed:
+  //
+  // For Binary Protobufs
+  // ====================
+  // Binary Protobuf senders SHOULD NOT set instrumentation_library_logs. Instead
+  // scope_logs SHOULD be set.
+  //
+  // Binary Protobuf receivers SHOULD check if instrumentation_library_logs is set
+  // and scope_logs is not set then the value in instrumentation_library_logs
+  // SHOULD be used instead by converting InstrumentationLibraryLogs into ScopeLogs.
+  // If scope_logs is set then instrumentation_library_logs SHOULD be ignored.
+  //
+  // For JSON
+  // ========
+  // JSON senders that set instrumentation_library_logs field MAY also set
+  // scope_logs to carry the same logs, essentially double-publishing the same data.
+  // Such double-publishing MAY be controlled by a user-settable option.
+  // If double-publishing is not used then the senders SHOULD set scope_logs and
+  // SHOULD NOT set instrumentation_library_logs.
+  //
+  // JSON receivers SHOULD check if instrumentation_library_logs is set and
+  // scope_logs is not set then the value in instrumentation_library_logs
+  // SHOULD be used instead by converting InstrumentationLibraryLogs into ScopeLogs.
+  // If scope_logs is set then instrumentation_library_logs field SHOULD be ignored.
+  repeated InstrumentationLibraryLogs instrumentation_library_logs = 4 [deprecated = true];
+
   // This schema_url applies to the data in the "resource" field. It does not apply
   // to the data in the "scope_logs" field which have their own schema_url field.
   string schema_url = 3;
@@ -65,6 +94,23 @@ message ScopeLogs {
   opentelemetry.proto.common.v1.InstrumentationScope scope = 1;
 
   // A list of log records.
+  repeated LogRecord log_records = 2;
+
+  // This schema_url applies to all logs in the "logs" field.
+  string schema_url = 3;
+}
+
+// A collection of Logs produced by an InstrumentationLibrary.
+// InstrumentationLibraryLogs is wire-compatible with ScopeLogs for binary
+// Protobuf format.
+// This message is deprecated and will be removed on June 15, 2022.
+message InstrumentationLibraryLogs {
+  // The instrumentation library information for the logs in this message.
+  // Semantically when InstrumentationLibrary isn't set, it is equivalent with
+  // an empty instrumentation library name (unknown).
+  opentelemetry.proto.common.v1.InstrumentationLibrary instrumentation_library = 1;
+
+  // A list of logs that originate from an instrumentation library.
   repeated LogRecord log_records = 2;
 
   // This schema_url applies to all logs in the "logs" field.
--- a/opentelemetry/proto/logs/v1/logs.proto
+++ b/opentelemetry/proto/logs/v1/logs.proto
@@ -79,7 +79,7 @@ message ResourceLogs {
   // scope_logs is not set then the value in instrumentation_library_logs
   // SHOULD be used instead by converting InstrumentationLibraryLogs into ScopeLogs.
   // If scope_logs is set then instrumentation_library_logs field SHOULD be ignored.
-  repeated InstrumentationLibraryLogs instrumentation_library_logs = 4 [deprecated = true];
+  repeated InstrumentationLibraryLogs instrumentation_library_logs = 1000 [deprecated = true];
 
   // This schema_url applies to the data in the "resource" field. It does not apply
   // to the data in the "scope_logs" field which have their own schema_url field.
@@ -105,6 +105,8 @@ message ScopeLogs {
 // Protobuf format.
 // This message is deprecated and will be removed on June 15, 2022.
 message InstrumentationLibraryLogs {
+  option deprecated = true;
+
   // The instrumentation library information for the logs in this message.
   // Semantically when InstrumentationLibrary isn't set, it is equivalent with
   // an empty instrumentation library name (unknown).
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -79,7 +79,7 @@ message ResourceSpans {
   // scope_spans is not set then the value in instrumentation_library_spans
   // SHOULD be used instead by converting InstrumentationLibrarySpans into ScopeSpans.
   // If scope_spans is set then instrumentation_library_spans field SHOULD be ignored.
-  repeated InstrumentationLibrarySpans instrumentation_library_spans = 4 [deprecated = true];
+  repeated InstrumentationLibrarySpans instrumentation_library_spans = 1000 [deprecated = true];
 
   // This schema_url applies to the data in the "resource" field. It does not apply
   // to the data in the "scope_spans" field which have their own schema_url field.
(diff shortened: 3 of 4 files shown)


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -43,29 +43,28 @@ message TracesData {
...
+// A collection of ScopeSpans from a Resource.
 message ResourceSpans {
   // The resource for the spans in this message.
   // If this field is not set then no resource info is known.
   opentelemetry.proto.resource.v1.Resource resource = 1;
 
-  // A list of InstrumentationLibrarySpans that originate from a resource.
-  repeated InstrumentationLibrarySpans instrumentation_library_spans = 2;
+  // A list of ScopeSpans that originate from a resource.
+  repeated ScopeSpans scope_spans = 2;
```
MadVikingGod, line 53: Wouldn't this be a breaking change?  E.g. I couldn't send a message generated with the previous version to a receiver with this version, and the same if the versions were swapped.

tigrannajaryan: Yes, what @Aneurysm9 said.

It is not a breaking change for Protobuf binary wire format. It is a breaking change for JSON format but that is OK, JSON format is not stable.

I believe it is also OK to break generated code since we do not provide any guarantees for that. We probably should start providing some guarantees for that at some point, probably when we decide that JSON is stable, since these 2 are somewhat interconnected (field names are visible in JSON).

Aneurysm9: I think it will have ripple effects on things that use that generated code.  For instance, when https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/otlp/otlptrace/clients.go#L51 is updated to a version with this change it will force that upgrade on any users providing their own proto objects to the client.

tigrannajaryan: @Aneurysm9 is that Client interface part of the API that is already promised to be stable?

Aneurysm9: Yes.  It is [at v1.4.1](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlptrace@v1.4.1#Client).  We were comfortable taking this dependency as it was our understanding that the OTLP data model was stable and only backward compatible changes would be made.  In hindsight, we probably should have insisted on a v1.0.0 release and clarification regarding stability for just the bits on-the-wire or for the generated code we provide for interop purposes.

tigrannajaryan: OK, so we have a problem here. Some possible approaches:
1. Don't make this change. The message names and field names are somewhat weird though. We can keep the names and explain that they really are a "scope" in the comments. I am not a big fan of this approach.
2. Otel Go SDK pays the price of leaking an unstable dependency in its stable interface. We break the dependency and the Go SDK interface breaks its Client interface for exporter implementors.
3. We make a major version increase in https://github.com/open-telemetry/opentelemetry-proto-go for this breaking change. Otel Go SDK implements a new ClientV2 interface and uses the new major version dependency in this interface. Both old and new Client interface will be supported in parallel.
4. Something else?

dyladan: +1 to not breaking JSON. The JSON exporter in JS is in fairly wide use and I think a break like this would have quite some impact on our users.

tigrannajaryan: FYI, JSON is explicitly marked "Alpha" in this repository and in the spec and we should be free to break it:
>"alpha" maturity APIs can change in incompatible ways at any time.

Aneurysm9: I'm not sure what the right thing to do here is.  Based on the stability description in the README, it seems like we have an expectation that we should be able to change field names at any time, but only because JSON is unstable and once JSON is stable then changing field names would impact compatibility for users.  

I think the same needs to be extended to the generated identifiers for stable signals.  It doesn't seem right to say that traces are stable but we can still change the message format in ways that will break anything using code generated from it.

tigrannajaryan: >It doesn't seem right to say that traces are stable but we can still change the message format in ways that will break anything using code generated from it.

I tend to agree with you, but until now this was the thinking (that it is OK to break generated code). If we knew this was supposed to a limitation we would be a lot more careful with names and with declaring OTLP stable.

I do not think we can retroactively apply such stability limitation right now. If we don't want to break existing code then the possible approach is to honour semver and increase the major version number in the generated code (approach 3 above).

After this change we should consider having formal stability guarantees for generated code.

bogdandrutu: So I believe `opentelemetry-go` made a mistake exposing publicly a type from a version `0.x` module and not sure if that is expected or normal. @tigrannajaryan opentelemetry-proto-go is at version `0.12.0` so it does not need a major version bump, because that will make it `1.0` :)

bogdandrutu: @tigrannajaryan I think we can make the change to the name, but also say that in JSON receivers should still accept the old name. And we will do this in the collector.

tigrannajaryan: @bogdandrutu I understand that `opentelemetry-proto-go` doesn't need a major version bump itself because it is on 0.x. However with proposal number 3 in order for `opentelemetry-go` to avoid breaking their clients it needs to be able to consume 2 different versions of `opentelemetry-proto-go` simultaneously. This requires a major version bump or for `opentelemetry-proto-go` to expose the changed generated code using a different import path (which would be even more awkward).

I don't know if this is a good way, I am just trying to find a way to allow us to change the proto here and also at the same time avoid breaking `opentelemetry-go`.

Maybe there is another way.

bogdandrutu: The other thing is to fix the `oteltrace` package. Deprecate that package with a fix dependency on the v0.12.0 and design a new package that does not expose this public.

MadVikingGod: Will there ever be a point where the entirety of the API (including names) is considered stable?  I feel like this needs to be discussed because projects that internally generate code will have to change.  That seems to be the conventional definition of unstable.

The other option is to go through the process of updating a field in a backward and forward compatible way:

1. Add a new field, and deprecate the old.
2. Wait 12 months by the guidelines that we have set out.
3. Remove the old field.  (This is still technically a breaking change, but one that we have allowed ourselves)

tigrannajaryan: >The other option is to go through the process of updating a field in a backward and forward compatible way

This would mean we are applying the most strict version of stability guarantees. This seems excessive to me for something that was not explicitly promised to be stable and implicitly was marked to be unstable (one reason JSON is marked unstable is because we said we don't want to guarantee the current message/field names).

dyladan: >FYI, JSON is explicitly marked "Alpha" in this repository and in the spec and we should be free to break it:

While I agree technically, I would hope that we take these breaking changes seriously anyways in the interest of our users.

There are also other problems not completely related to JSON. If a field can always be renamed, the logic to transform an SDK object into an OTLP object can never be shared between two different exporters even when JSON is completely out of the picture (HTTP/proto and gRPC for example). Even if generated code is not used, the code that transforms the SDK object to the OTLP object needs to pick a field name to use in its output which is then used as input to the final format encoder (protobuf in this case).

tigrannajaryan: >implicitly was marked to be unstable (one reason JSON is marked unstable is because we said we don't want to guarantee the current message/field names).

I need to correct myself. The message/field names are actually explicitly declared unstable. This is written in the README of this repository:
>Note that maturity guarantees apply only to wire-level compatibility for the binary Protobuf serialization. Neither message, field, nor enum names of Protobuf messages are visible on the wire and are not considered part of the guarantees. We are free to make a change to the names.

tigrannajaryan: >the logic to transform an SDK object into an OTLP object can never be shared between two different exporters even when JSON is completely out of the picture (HTTP/proto and gRPC for example). Even if generated code is not used, the code that transforms the SDK object to the OTLP object needs to pick a field name to use in its output which is then used as input to the final format encoder (protobuf in this case).

I don't understand the problem that this describes. Please clarify.

anuraaga: I suspect JSON is actually being used in production - changing the format IMO is acceptable, but it's a lot more tricky if it's a hard break. Ideally we have at least one version of the collector that supports both names in the OTLP receiver. Can the collector be updated to do a backwards-compatibility pass on the incoming JSON to allow it to support both names temporarily?

MadVikingGod: To clarify what, I think, @dyladan is saying is the reason that go has this problem is because of shared code.  We have a tiered system of SDK type -> `otlptrace` (which marshals to protobuf) -> Client (which sends over the wire, HTTP or grpc).  The part that we can't share is the marshaling of SDK types into protobuf types because that would expose the protobufs as part of a public API.

dyladan: @MadVikingGod got it exactly. If I write code that transforms the SDK types into types that would be serializable to protobuf or JSON, I can't share that code between multiple exporter packages if field names might change. The only way to resolve that would be for the transformation code to return an object with both field names forever (and mark deprecated the old one). The JSON exporter would then be responsible for stripping all deprecated fields, also forever, in order to avoid bloating the payload size.

Additionally @anuraaga is correct JSON is for sure being used in production in browser environments.

dyladan: It also affects the types between the SDK and exporter. The interface for the exporter already has a type which uses the old name, and the SDK type already contains the old names. Should these types also be extended to have both names indefinitely?

Aneurysm9: We're going to have to carry the [`instrumentation.Library` type](https://pkg.go.dev/go.opentelemetry.io/otel/sdk@v1.4.1/instrumentation#Library) indefinitely, and I expect we'll have to carry it in our SDK export structures alongside a new type for the instrumentation scope that duplicates the data.

tigrannajaryan: >Can the collector be updated to do a backwards-compatibility pass on the incoming JSON to allow it to support both names temporarily?

Probably. I am not sure how exactly we can do it since JSON unmarshaling is a Protobuf library function, but perhaps it is doable.

anuraaga: FWIW, my idea is brute force - process an untyped JSON tree before passing it to the protobuf library, which should naturally be doable. The main concern could be performance, but we expect JSON to be slower than binary anyways so while the format is considered alpha, having a temporary performance hit for handling this compatibility issue seems OK. We generally except the burden of compatibility to be on a receiver rather than an exporter, though of course the other option is for the JSON exporters to populate the same data into both naming schemes, using a transformation on the untyped JSON on export for exporters using the generated protobuf code. For reference, since Java doesn't use generated protos, it would be relatively trivial with no significant performance penalty (other than transmitting the additional data) to export both structs if that were deemed the way forward.

tigrannajaryan: Let's assume that receivers (such as Collector) can gracefully handle the old and new names simultaneously in JSON payloads. I think it is doable one way or another.

@dyladan @Aneurysm9 @MadVikingGod The remaining question is are you able to handle this change in your SDKs gracefully? If so we can move forward and merge the PR and you can upgrade to the new proto version when you are ready.

We can also bring this up in the maintainers meeting to make sure other language maintainers also have a chance to review.

tigrannajaryan: > The other option is to go through the process of updating a field in a backward and forward compatible way:
> 
> 1. Add a new field, and deprecate the old.
> 2. Wait 12 months by the guidelines that we have set out.
> 3. Remove the old field.  (This is still technically a breaking change, but one that we have allowed ourselves)

If we decide that we do want to gracefully handle old and new names in JSON receivers the above option may be the easiest from implementation perspective. (We can discuss whether it needs to be 12 months or less of grace period).

dyladan: > Let's assume that receivers (such as Collector) can gracefully handle the old and new names simultaneously in JSON payloads. I think it is doable one way or another.

This assumes our customers can update their collectors. If the SDK is updated with a minor version number, the user could reasonably expect it would continue to work with the collector version they're already using.

> @dyladan @Aneurysm9 @MadVikingGod The remaining question is are you able to handle this change in your SDKs gracefully? If so we can move forward and merge the PR and you can upgrade to the new proto version when you are ready.

In the SDK type we can simply duplicate the field with a new name and deprecate the old name. It is a small overhead, but I think not that big of a deal as it is just local and not over the wire.

tigrannajaryan: > This assumes our customers can update their collectors. If the SDK is updated with a minor version number, the user could reasonably expect it would continue to work with the collector version they're already using.

If we want to be extra careful for some grace period the senders can also double-publish the data under both names in JSON.

alolita:
Naming changes are hard. They should be considered as breaking changes for any developers (as an end-user). We should be treating what we call "stable" as "stable". Going forward, can we treat the names of fields and types are unbreakable (i.e. stable).  We need to stabilize message, fields, enum names.

Imo, having a name change as in this PR should not be happening at this stage of the project.

Re: https://github.com/open-telemetry/opentelemetry-proto/blob/main/README.md 

Let's discuss in the spec SIG tomorrow.

tigrannajaryan:
> Naming changes are hard. They should be considered as breaking changes for any developers (as an end-user). We should be treating what we call "stable" as "stable".

I agree. We do not call names "stable" though. They are clearly and explicitly specified to be unstable.


> Going forward, can we treat the names of fields and types are unbreakable (i.e. stable). We need to stabilize message, fields, enum names.

I agree and I think we are very close to achieving this. Logs proto are about to be declared stable, races and metrics already are. With all 3 signals stable we should be more confident to say that we are done making breaking changes, including the names, etc.

opentelemetry/proto/trace/v1/trace.proto
dyladan: >If we want to be extra careful for some grace period the senders can also double-publish the data under both names in JSON

That is what we will likely have to do. The field will have to be duplicated essentially everywhere.

- SDK export pipeline
- JSON serialization
- SDK -> OTLP transformer

MSNev:
Adding comment from the Spec SIG.
I believe we should be versioning the spec rather than double publishing / consuming as double publishing / consuming introduces a bunch of unnecessary complexity and is going to introduce more code to handle "duplicated" fields.

And I agree with @alolita comment
> Imo, having a name change as in this PR should not be happening at this stage of the project.

tigrannajaryan:
> Adding comment from the Spec SIG. I believe we should be versioning the spec rather than double publishing / consuming as double publishing / consuming introduces a bunch of unnecessary complexity and is going to introduce more code to handle "duplicated" fields.

Versioning like that is extremely undesirable because it means breaking interoperability between versions. We will not do that. We promised wire compatibility for binary Protobuf format and in the past we publicly recognized that breaking interoperability between network nodes is one of the most painful form of causing user trouble, we should avoid that, even at the high cost of pains to us, developers. Versioning like that may be the last resort, which is not the case right now, we have more graceful options.

> 
> And I agree with @alolita comment
> 
> > Imo, having a name change as in this PR should not be happening at this stage of the project.

I agree it is not undesirable, so going forward we need to rethink our guarantees.

tigrannajaryan:
All, I made changes to the PR to make the transition more graceful. Please read the [comments here](https://github.com/open-telemetry/opentelemetry-proto/pull/362/commits/1d10ca05dcba7166440d7c2a08f5747dc48ff5f1#diff-5c407ac9c675a60170f94ab620f5329f25e5388d6790864b7799abce66f79332) and let me know what you think.

dyladan:
The compromise seems acceptable to me. So JSON senders should send only the new name by default but offer an old collector compatibility option which sends both?

tigrannajaryan:
> The compromise seems acceptable to me. So JSON senders should send only the new name by default but offer an old collector compatibility option which sends both?

Yes, that's what I was aiming for, but open for alternate suggestions.

In the Collector we will also implementing gracefully accepting either the old or the new name, so it should work fine with old or new senders.

Essentially we are ensuring interoperability from both sides. If at least one of the sides correctly implements the new recommendations it should be able to work with an old peer.

tigrannajaryan:
> All, I made changes to the PR to make the transition more graceful. Please read the [comments here](https://github.com/open-telemetry/opentelemetry-proto/pull/362/commits/1d10ca05dcba7166440d7c2a08f5747dc48ff5f1#diff-5c407ac9c675a60170f94ab620f5329f25e5388d6790864b7799abce66f79332) and let me know what you think.

@MadVikingGod @Aneurysm9 what do you think?

Aneurysm9:
I think this is a good way to transition.  Do we also need to do the same for `ResourceMetrics`?

anuraaga:
> The compromise seems acceptable to me. So JSON senders should send only the new name by default but offer an old collector compatibility option which sends both?

For reference, I suspect we probably wouldn't update the Java JSON exporter to send both old and new as JSON export from Java should be unpopular enough that gating users on updating their collector is fine IMO.

If we were to send both in JSON, we would send both in binary too though as deciding the behavior for one or the other would complicate our export code quite a bit.

tigrannajaryan:
> I think this is a good way to transition. Do we also need to do the same for `ResourceMetrics`?

Yes, I will add that if we are happy with this approach.

tigrannajaryan:
> > The compromise seems acceptable to me. So JSON senders should send only the new name by default but offer an old collector compatibility option which sends both?
> 
> For reference, I suspect we probably wouldn't update the Java JSON exporter to send both old and new as JSON export from Java should be unpopular enough that gating users on updating their collector is fine IMO.
> 
> If we were to send both in JSON, we would send both in binary too though as deciding the behavior for one or the other would complicate our export code quite a bit.

I think this is OK. The proposed approach exceeds the formal guarantees that we have to follow and is not an absolute requirement (I deliberately used "SHOULD" clauses instead of "MUST"). If the particular SIG feels JSON is not an important case for them then they just need to follow the formal requirements of the "Alpha" stability, which says "you can break it anytime".

For the record: Collector doesn't even have an implementation of OTLP/JSON exporter, so nothing to do there. The Collector has an OTLP/JSON receiver, which we should be able to modify to accept both old and new formats gracefully.

tigrannajaryan:
> I think this is a good way to transition. Do we also need to do the same for `ResourceMetrics`?

Done now for metrics and logs too.

tigrannajaryan:

opentelemetry/proto/common/v1/common.proto
```diff
@@ -66,9 +66,20 @@ message KeyValue {
 }
 
 // InstrumentationLibrary is a message representing the instrumentation library information
-// such as the fully qualified name and version. 
+// such as the fully qualified name and version.
+// InstrumentationLibrary is wire-compatible with InstrumentationScope for binary
+// Protobuf format.
+// This message is deprecated and will be removed on June 15, 2022.
 message InstrumentationLibrary {
```
bogdandrutu, line 73: FYI: There is a proto deprecated annotation.

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -43,29 +43,74 @@ message LogsData {
...
+  // scope_logs to carry the same logs, essentially double-publishing the same data.
+  // Such double-publishing MAY be controlled by a user-settable option.
+  // If double-publishing is not used then the senders SHOULD set scope_logs and
+  // SHOULD NOT set instrumentation_library_logs.
+  //
+  // JSON receivers SHOULD check if instrumentation_library_logs is set and
+  // scope_logs is not set then the value in instrumentation_library_logs
+  // SHOULD be used instead by converting InstrumentationLibraryLogs into ScopeLogs.
+  // If scope_logs is set then instrumentation_library_logs field SHOULD be ignored.
+  repeated InstrumentationLibraryLogs instrumentation_library_logs = 4 [deprecated = true];
```
bogdandrutu, line 82: Do we want to waste a small ID for this? I would suggest to use one over 1000 for this, since this is only to support deprecated JSON.

opentelemetry/proto/common/v1/common.proto
bogdandrutu: Add this as first line:

```proto
message InstrumentationLibrary {
  option deprecated = true;

   ...
}
```

tigrannajaryan: Added.

bogdandrutu (approved):
For me this looks good.

jmacd (approved):

pmm-sumo (approved):

tigrannajaryan:
@open-telemetry/specs-logs-approvers please review.

tigrannajaryan:
We have enough approvals. @dyladan @MadVikingGod @Aneurysm9 I will merge unless you object.

dyladan (approved):

MadVikingGod:
I have not had a chance to actually confirm that this won't be a breaking change.  That said I do think that this should work. I don't want to hold this up any further.

PR #366, opened 2022-03-04
Add field of present flag to proto build

### LINKED ISSUE #314: Cannot distinguish Histogram's Sum field between 0 vs not assigned.

codeboten:
This is a clone of #336

Update the build to allow field presence to be used in OTLP.

* Enable the optional field-presence flag.

Fixes #314

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ gen-all: gen-cpp gen-csharp gen-go gen-java gen-kotlin gen-objc gen-openapi gen-
 OTEL_DOCKER_PROTOBUF ?= otel/build-protobuf:0.9.0
 PROTOC := docker run --rm -u ${shell id -u} -v${PWD}:${PWD} -w${PWD} ${OTEL_DOCKER_PROTOBUF} --proto_path=${PWD}
 PROTO_INCLUDES := -I/usr/include/github.com/gogo/protobuf
+PROTO_FLAGS := --experimental_allow_proto3_optional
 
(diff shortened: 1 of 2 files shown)


jmacd (approved):

Makefile
```diff
@@ -19,7 +19,7 @@ gen-all: gen-cpp gen-csharp gen-go gen-java gen-kotlin gen-objc gen-openapi gen-
 
 OTEL_DOCKER_PROTOBUF ?= otel/build-protobuf:0.9.0
 PROTOC := docker run --rm -u ${shell id -u} -v${PWD}:${PWD} -w${PWD} ${OTEL_DOCKER_PROTOBUF} --proto_path=${PWD}
-PROTO_INCLUDES := -I/usr/include/github.com/gogo/protobuf
+PROTO_FLAGS := --experimental_allow_proto3_optional
```
anuraaga, line 22: Is the flag still needed?

https://github.com/protocolbuffers/protobuf/releases/tag/v3.15.0

codeboten: Good point, I've removed it.

anuraaga (approved):

bogdandrutu (approved):

Issue #166, opened 2020-07-10
Add sampled flag to Span type


kbrockhoff:
Lack of a sampled flag in the Span type results in ambiguity. It implies that non-sampled traces are dropped at the instrumentation level and not forwarded through the system. If using tail-sampling decisions, this means all traces need to be marked sampled up front. In addition, it leaves no flexibility for different routing of traces. For example, routing all traces to a logging exporter and only sampled ones to a full-scale backend.

In addition, other protocols such as Zipkin have a sampled flag. The lack of this data makes it impossible to preserve full fidelity when converting between protocols.

Oberon00:
Isn't an unsampled but exported span an oxymoron? How could this happen in practice?

kbrockhoff:
There is work going on to do tail-based sampling in the collector. For this to work, all spans need to be exported to the collector.

bogdandrutu:
I am +1 on this

anuraaga:
@kbrockhoff Do you mind describing more concretely the use case you have in mind? I think with pure tail sampling, we would expect the SDK to do 100% sampling and that wouldn't need to be exported. Is the goal to support multiple sampling decisions, e.g., not sampled for trace, but sampled for logs or for secondary sampling of a subset of the trace? I guess instead of a sampling flag, we'd want a list of `SamplingDecision`, each can be namespaced to a backend or system by a user.

kbrockhoff:
There are several use cases for having a sampled flag.

1. Using tail-based sampling where the team controlling the tail-based sampling and monitoring does not have control over upfront sampling configuration. This can occur in a supply chain with multiple organizations or even a siloed single organization. In this use case having a list of SamplingDecision as suggested by @anuraaga is an even better idea.

2. Using protobuf as a general storage mechanism for tracing data. An example would be compact storage of a test data set.

In either case, adequate information can potentially be stored in the TraceState field but would then need to be parsed. I can be talked into either this approach or the SamplingDecision approach.

andrewhsu:
@arminru from the bulk move of `P3` `required-for-ga` issues to `after-ga` earlier today during the issue triage mtg with TC, i noticed this issue was skipped. should the labels for tracking this issue be updated to something else?

arminru:
This can be added after-ga, if still desired, yes.
@bogdandrutu you were +1 on ths, WDYT?

jsuereth:
This was fixed in #503

PR #384, opened 2022-04-27
Add missing trace flags to Span and Link messages

### LINKED ISSUE #382: encoding of context flags in otlp messages
[The spec for tracing](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spancontext) includes a trace flags in the context:

> TraceFlags contain details about the trace. Unlike TraceState values, TraceFlags are present in all traces. The current version of the specification only supports a single flag called [sampled](https://www.w3.org/TR/trace-context/#sampled-flag).

However, the otlp proto message has no appropriate proto field to store the flags for `message Span` and `message Link`.

I think the flags are important to be encoded in the otlp format for the following reasons:
- It is handy that when holding a link, to be able to tell if it points on a sampled or non sampled span. For example in messaging specification, a consumer may link to a producer span, but it is possible and common that the consumer is sampled and the producer is not. Currently, consumer links just store `trace_id`, `span_id` and `trace_state`. If someone tries to follow the link and cannot find the source span, there is no way to tell if it was not sampled (valid) or simply missing (lost - invalid). It is can also be expensive to follow a link (query to db) - and by knowing if a link is sampled or not one can skip this operation for processing and visualization purposes. There can also be UI value in presenting a link when it is not sampled vs sampled link.
- OTLP can be handy for storing spans, not only for transmitting them. For transmitting spans from SDKs to collector, we do not anticipate any non-sampled spans. However, someone might need to serialize and deserialize non-sampled spans (for example - to debug) in which case this data will currently be lost. I think protocol should support this kind of use-cases
- Even though flags currently only support single "sampled" flag, they can be extended in the future to hold new values. We cannot anticipate the use of these future values, and having them in the otlp model will allow using it always.

Does it make sense to add "trace_flags" field to Span and Link messages in proto? or are there any good reasons why no to include them?

blumamir:
fixes #382 and #166

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -291,6 +291,15 @@ message Span {
     // dropped_attributes_count is the number of dropped attributes. If the value is 0,
     // then no attributes were dropped.
     uint32 dropped_attributes_count = 5;
+
+    // Flags, a bit field. 8 least significant bits are the trace flags as
+    // defined in W3C Trace Context specification. 24 most significant bits are reserved
+    // and must be set to 0. Readers must not assume that 24 most significant bits
+    // will be zero and must correctly mask the bits when reading 8-bit trace flag (use
+    // flags & TRACE_FLAGS_MASK).
+    // This value is required, as not setting it will be interpeted as 0 which is 
+    // a valid representation of all flags being present and set to 0.
+    fixed32 flags = 6;
   }
 
   // links is a collection of Links, which are references from this span to a span
@@ -304,6 +313,16 @@ message Span {
   // An optional final status for this span. Semantically when Status isn't set, it means
   // span's status code is unset, i.e. assume STATUS_CODE_UNSET (code = 0).
   Status status = 15;
+
+  // Flags, a bit field. 8 least significant bits are the trace flags as
+  // defined in W3C Trace Context specification. 24 most significant bits are reserved
+  // and must be set to 0. Readers must not assume that 24 most significant bits
+  // will be zero and must correctly mask the bits when reading 8-bit trace flag (use
+  // flags & TRACE_FLAGS_MASK).
+  // This value is required, as not setting it will be interpeted as 0 which is 
+  // a valid representation of all flags being present and set to 0.
+  fixed32 flags = 16;
+
 }
 
 // The Status type defines a logical error model that is suitable for different


blumamir:
I now wonder if the otlp messages should also contain the [w3c version](https://www.w3.org/TR/trace-context-1/#version) so when more flags are added in the future, the receiver can differentiate between them being set to "0" vs "not set" by examining the version.
Any thoughts?
@dyladan

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -291,6 +291,15 @@ message Span {
...
     uint32 dropped_attributes_count = 5;
+
+    // Flags, a bit field. 8 least significant bits are the trace flags as
+    // defined in W3C Trace Context specification. 24 most significant bits are reserved
+    // and must be set to 0. Readers must not assume that 24 most significant bits
+    // will be zero and must correctly mask the bits when reading 8-bit trace flag (use
+    // flags & TRACE_FLAGS_MASK).
+    // This value is required, as not setting it will be interpeted as 0 which is 
+    // a valid representation of all flags being present and set to 0.
+    fixed32 flags = 6;
```
bogdandrutu, line 302: Can you move this closer to the ids?

bogdandrutu:
@blumamir I think we can add the version when w3c changes to v2, before that, the absence of the version means version 1.

opentelemetry/proto/trace/v1/trace.proto
blumamir: No problem, but it looks like the messages are currently sorted by "field number", and moving it up will break this pattern. Are we OK with that?

blumamir:
> @blumamir I think we can add the version when w3c changes to v2, before that, the absence of the version means version 1.

If we add it when w3c bump the version, then all existing exporters out there will not report it which will be interpreted as `0` and we are back to the original backward support problem that we are now facing.

It means that a processor will not be able to use a new flag until *all* the exporters in production are upgraded to a version that supports the new `proto` release with the w3c version field, which can be a long and painful process.

bogdandrutu:
> It means that a processor will not be able to use a new flag until all the exporters in production are upgraded to a version that supports the new proto release with the w3c version field, which can be a long and painful process.

Isn't that true anyway? Let's assume we add the field and everyone just exports "v=1", what is the difference?

Oberon00:
Should we maybe wait a bit with this and do it together with IsRemote once https://github.com/open-telemetry/oteps/pull/182 is merged, to minimize the chance of inconsistencies?

tigrannajaryan:
> Should we maybe wait a bit with this and do it together with IsRemote once [open-telemetry/oteps#182](https://github.com/open-telemetry/oteps/pull/182) is merged, to minimize the chance of inconsistencies?

We can wait. If it is delayed for a long time we can also move forward with this since I think we can add new flags in backward compatible manner. For IsRemote we likely will need 2 bits to represent (Unspecified=00, IsRemote=01, IsNotRemote=01) state. Unspecified will be the default state matching the zero state of the field (which will make it backward compatible with existing OTLP versions).

tigrannajaryan:
https://github.com/open-telemetry/oteps/pull/182 has not seen much progress. Do we want to move forward with this PR?

jmacd:
Closing this in favor of #503. The purpose is identical, but we believe @blumamir is no longer focused on OTel so I will carry this out.

dyladan:
@blumamir was on a break for a while but is back and active in otel again. I'm sure he probably doesn't mind work being consolidated into #503 though

PR #390, opened 2022-05-09
Make it possible to indicate partial success in an OTLP export response

joaopgrassi:
This PR is a first attempt in solving/addressing https://github.com/open-telemetry/opentelemetry-specification/issues/2454.

At first, I tried having "top-level" properties under the `Export*ServiceResponse` types like so:

```proto
message ExportMetricsServiceResponse {
  optional int64 accepted_data_points = 1;
  optional string error_message = 2;
}
```

But in my view, this has a drawback. OTLP exporters/senders using the new protos would have to always check the [explicit presence](https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md?rgh-link-date=2022-04-27T13%3A13%3A21Z#presence-in-proto3-apis) of the new fields (e.g. `HasAcceptedDatapoints`) since they don't know which proto version the server is using. Also, since there's more than one field, would it need to check both? What if we add new fields later?

Then I thought having a new type that combines these two new fields is a better compromise. This way, it only needs to check if the new field is not null. If it's not null, we can have some guarantee that the other fields are there (we indicate it with a MUST in the spec, for ex)

I did some tests with sample apps to confirm the cases for backwards compatibility: (code here [joaopgrassi/otlp-partial-success-experimental](https://github.com/joaopgrassi/otlp-partial-success-experimental)).

## 1. Old exporters working with receivers/servers using the new protos

All works, as the exporters were not expecting the new fields. They will not have the fields, so they can't inform the user of a partial success - status quo.

## 2. New exporters working with receivers/servers using the old protos

Exporters MUST check if the `Details` field is not null to be able to log a message with the details for troubleshooting.

```csharp
var client = new MetricsService.MetricsServiceClient(channel);
var req = new ExportMetricsServiceRequest();
var response = client.Export(req);

if (response.Details != null)
{
	Console.WriteLine("Number of lines accepted: {0}", response.Details.AcceptedDataPoints);
	Console.WriteLine("Error message: {0}", response.Details.ErrorMessage);
}
```

## 3. New exporters working with receivers/servers using the new protos

Since exporters don't know the proto version servers are sending back, it always have to perform the check on point 2 above.

### DIFF
--- a/opentelemetry/proto/collector/logs/v1/logs_service.proto
+++ b/opentelemetry/proto/collector/logs/v1/logs_service.proto
@@ -56,7 +56,16 @@ message ExportLogsServiceResponse {
 
 message ExportLogsPartialSuccess {
   // The number of accepted log records
-  int64 accepted_log_records = 1;
+  //
+  // If the request is partially successful and the number of
+  // accepted log records is known, the server MUST set
+  // the accepted_log_records field.
+  // 
+  // The value '0' indicates that all log records were dropped by the server.
+  //
+  // Since the value '0' is meaningful, the field is marked as 'optional'
+  // so it is possible to distinguish between the not set value and the default.
+  optional int64 accepted_log_records = 1;
 
   // A developer-facing human-readable error message in English. It should
   // explain why the server rejected parts of the data,
--- a/opentelemetry/proto/collector/logs/v1/logs_service.proto
+++ b/opentelemetry/proto/collector/logs/v1/logs_service.proto
@@ -44,9 +44,13 @@ message ExportLogsServiceRequest {
 
 message ExportLogsServiceResponse {
   // The details of a partially successful export request.
-  // When the request is only partially successful (due to e.g. invalid/bad data),
-  // the server MUST populate the partial_success property with a
-  // developer-facing error message along with the number of accepted log records.
+  // If the processing of the request is partially successful
+  // (i.e. when the server accepts only parts of the data and rejects the rest)
+  // and the number of accepted log records is known, the server MUST initialize
+  // the partial_success field and MUST set the accepted_log_records.
+
+  // When the export request is processed successfully, the server MUST leave
+  // the partial_success field unset.
   ExportLogsPartialSuccess partial_success = 1;
 }
 
@@ -55,6 +59,7 @@ message ExportLogsPartialSuccess {
   int64 accepted_log_records = 1;
 
   // A developer-facing human-readable error message in English. It should
-  // both explain the error and offer an actionable resolution to it.
-  string error_message = 2;
+  // explain why the server rejected parts of the data,
+  // and might offer guidance on how users can address the issues.
+  optional string error_message = 2;
 }
--- a/opentelemetry/proto/collector/logs/v1/logs_service.proto
+++ b/opentelemetry/proto/collector/logs/v1/logs_service.proto
@@ -42,4 +42,18 @@ message ExportLogsServiceRequest {
 }
 
 message ExportLogsServiceResponse {
+  // The details of a partially successfull export request.
+  // When the request is only partially successfull (due to e.g. invalid/bad data),
+  // the server MUST populate the details property with a developer-facing error
+  // message along with the number of accepted log records.
+  ExportLogsPartialSuccess details = 1;
+}
+
+message ExportLogsPartialSuccess {
+  // The number of accepted log records
+  int64 accepted_log_records = 1;
+
+  // A developer-facing human-readable error message in English. It should
+  // both explain the error and offer an actionable resolution to it.
+  string error_message = 2;
 }
--- a/opentelemetry/proto/collector/logs/v1/logs_service.proto
+++ b/opentelemetry/proto/collector/logs/v1/logs_service.proto
@@ -61,5 +61,8 @@ message ExportLogsPartialSuccess {
   // A developer-facing human-readable error message in English. It should
   // explain why the server rejected parts of the data,
   // and might offer guidance on how users can address the issues.
-  optional string error_message = 2;
+  //
+  // error_message is an optional field. An error_message with an empty value
+  // is equivalent of it not being set.
+  string error_message = 2;
 }
--- a/opentelemetry/proto/collector/logs/v1/logs_service.proto
+++ b/opentelemetry/proto/collector/logs/v1/logs_service.proto
@@ -42,11 +42,11 @@ message ExportLogsServiceRequest {
 }
 
 message ExportLogsServiceResponse {
-  // The details of a partially successfull export request.
-  // When the request is only partially successfull (due to e.g. invalid/bad data),
-  // the server MUST populate the details property with a developer-facing error
-  // message along with the number of accepted log records.
-  ExportLogsPartialSuccess details = 1;
+  // The details of a partially successful export request.
+  // When the request is only partially successful (due to e.g. invalid/bad data),
+  // the server MUST populate the partial_success property with a
+  // developer-facing error message along with the number of accepted log records.
+  ExportLogsPartialSuccess partial_success = 1;
 }
 
 message ExportLogsPartialSuccess {
--- a/opentelemetry/proto/collector/metrics/v1/metrics_service.proto
+++ b/opentelemetry/proto/collector/metrics/v1/metrics_service.proto
@@ -61,5 +61,8 @@ message ExportMetricsPartialSuccess {
   // A developer-facing human-readable error message in English. It should
   // explain why the server rejected parts of the data,
   // and might offer guidance on how users can address the issues.
-  optional string error_message = 2;
+  //
+  // error_message is an optional field. An error_message with an empty value
+  // is equivalent of it not being set.
+  string error_message = 2;
 }
--- a/opentelemetry/proto/collector/metrics/v1/metrics_service.proto
+++ b/opentelemetry/proto/collector/metrics/v1/metrics_service.proto
@@ -42,4 +42,18 @@ message ExportMetricsServiceRequest {
 }
 
 message ExportMetricsServiceResponse {
+  // The details of a partially successfull export request.
+  // When the request is only partially successfull (due to e.g. invalid/bad data),
+  // the server MUST populate the details property with a developer-facing error
+  // message along with the number of accepted data points.
+  ExportMetricsPartialSuccess details = 1;
+}
+
+message ExportMetricsPartialSuccess {
+  // The number of accepted data points
+  int64 accepted_data_points = 1;
+
+  // A developer-facing human-readable error message in English. It should
+  // both explain the error and offer an actionable resolution to it.
+  string error_message = 2;
 }
--- a/opentelemetry/proto/collector/metrics/v1/metrics_service.proto
+++ b/opentelemetry/proto/collector/metrics/v1/metrics_service.proto
@@ -56,7 +56,16 @@ message ExportMetricsServiceResponse {
 
 message ExportMetricsPartialSuccess {
   // The number of accepted data points
-  int64 accepted_data_points = 1;
+  //
+  // If the request is partially successful and the number of
+  // accepted data points is known, the server MUST set
+  // the accepted_data_points field.
+  // 
+  // The value '0' indicates that all data points were dropped by the server.
+  //
+  // Since the value '0' is meaningful, the field is marked as 'optional'
+  // so it is possible to distinguish between the not set value and the default.
+  optional int64 accepted_data_points = 1;
 
   // A developer-facing human-readable error message in English. It should
   // explain why the server rejected parts of the data,
--- a/opentelemetry/proto/collector/metrics/v1/metrics_service.proto
+++ b/opentelemetry/proto/collector/metrics/v1/metrics_service.proto
@@ -44,9 +44,13 @@ message ExportMetricsServiceRequest {
 
 message ExportMetricsServiceResponse {
   // The details of a partially successful export request.
-  // When the request is only partially successful (due to e.g. invalid/bad data),
-  // the server MUST populate the partial_success property with a
-  // developer-facing error message along with the number of accepted data points.
+  // If the processing of the request is partially successful
+  // (i.e. when the server accepts only parts of the data and rejects the rest)
+  // and the number of accepted data points is known, the server MUST initialize
+  // the partial_success field and MUST set the accepted_data_points.
+
+  // When the export request is processed successfully, the server MUST leave
+  // the partial_success field unset.
   ExportMetricsPartialSuccess partial_success = 1;
 }
 
@@ -55,6 +59,7 @@ message ExportMetricsPartialSuccess {
   int64 accepted_data_points = 1;
 
   // A developer-facing human-readable error message in English. It should
-  // both explain the error and offer an actionable resolution to it.
-  string error_message = 2;
+  // explain why the server rejected parts of the data,
+  // and might offer guidance on how users can address the issues.
+  optional string error_message = 2;
 }
--- a/opentelemetry/proto/collector/metrics/v1/metrics_service.proto
+++ b/opentelemetry/proto/collector/metrics/v1/metrics_service.proto
@@ -42,11 +42,11 @@ message ExportMetricsServiceRequest {
 }
 
 message ExportMetricsServiceResponse {
-  // The details of a partially successfull export request.
-  // When the request is only partially successfull (due to e.g. invalid/bad data),
-  // the server MUST populate the details property with a developer-facing error
-  // message along with the number of accepted data points.
-  ExportMetricsPartialSuccess details = 1;
+  // The details of a partially successful export request.
+  // When the request is only partially successful (due to e.g. invalid/bad data),
+  // the server MUST populate the partial_success property with a
+  // developer-facing error message along with the number of accepted data points.
+  ExportMetricsPartialSuccess partial_success = 1;
 }
 
 message ExportMetricsPartialSuccess {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ Full list of differences found in [this compare](https://github.com/open-telemet
 ### Added
 
 * Introduce Scope Attributes. [#395](https://github.com/open-telemetry/opentelemetry-proto/pull/395)
+* Introduce partial success fields in `Export*ServiceResponse`.
+ [#390](https://github.com/open-telemetry/opentelemetry-proto/pull/390)
 
 ### Removed
 
--- a/opentelemetry/proto/collector/trace/v1/trace_service.proto
+++ b/opentelemetry/proto/collector/trace/v1/trace_service.proto
@@ -61,5 +61,8 @@ message ExportTracePartialSuccess {
   // A developer-facing human-readable error message in English. It should
   // explain why the server rejected parts of the data,
   // and might offer guidance on how users can address the issues.
-  optional string error_message = 2;
+  //
+  // error_message is an optional field. An error_message with an empty value
+  // is equivalent of it not being set.
+  string error_message = 2;
 }
--- a/opentelemetry/proto/collector/trace/v1/trace_service.proto
+++ b/opentelemetry/proto/collector/trace/v1/trace_service.proto
@@ -44,9 +44,13 @@ message ExportTraceServiceRequest {
 
 message ExportTraceServiceResponse {
   // The details of a partially successful export request.
-  // When the request is only partially successful (due to e.g. invalid/bad data),
-  // the server MUST populate the partial_success property with a
-  // developer-facing error message along with the number of accepted spans.
+  // If the processing of the request is partially successful
+  // (i.e. when the server accepts only parts of the data and rejects the rest)
+  // and the number of accepted spans is known, the server MUST initialize
+  // the partial_success field and MUST set the accepted_spans.
+
+  // When the export request is processed successfully, the server MUST leave
+  // the partial_success field unset.
   ExportTracePartialSuccess partial_success = 1;
 }
 
@@ -55,6 +59,7 @@ message ExportTracePartialSuccess {
   int64 accepted_spans = 1;
 
   // A developer-facing human-readable error message in English. It should
-  // both explain the error and offer an actionable resolution to it.
-  string error_message = 2;
+  // explain why the server rejected parts of the data,
+  // and might offer guidance on how users can address the issues.
+  optional string error_message = 2;
 }
--- a/opentelemetry/proto/collector/trace/v1/trace_service.proto
+++ b/opentelemetry/proto/collector/trace/v1/trace_service.proto
@@ -56,7 +56,16 @@ message ExportTraceServiceResponse {
 
 message ExportTracePartialSuccess {
   // The number of accepted spans
-  int64 accepted_spans = 1;
+  //
+  // If the request is partially successful and the number of
+  // accepted spans is known, the server MUST set
+  // the accepted_spans field.
+  // 
+  // The value '0' indicates that all spans were dropped by the server.
+  //
+  // Since the value '0' is meaningful, the field is marked as 'optional'
+  // so it is possible to distinguish between the not set value and the default.
+  optional int64 accepted_spans = 1;
 
   // A developer-facing human-readable error message in English. It should
   // explain why the server rejected parts of the data,
--- a/opentelemetry/proto/collector/trace/v1/trace_service.proto
+++ b/opentelemetry/proto/collector/trace/v1/trace_service.proto
@@ -42,11 +42,11 @@ message ExportTraceServiceRequest {
 }
 
 message ExportTraceServiceResponse {
-  // The details of a partially successfull export request.
-  // When the request is only partially successfull (due to e.g. invalid/bad data),
-  // the server MUST populate the details property with a developer-facing error
-  // message along with the number of accepted spans.
-  ExportTracePartialSuccess details = 1;
+  // The details of a partially successful export request.
+  // When the request is only partially successful (due to e.g. invalid/bad data),
+  // the server MUST populate the partial_success property with a
+  // developer-facing error message along with the number of accepted spans.
+  ExportTracePartialSuccess partial_success = 1;
 }
 
 message ExportTracePartialSuccess {
--- a/opentelemetry/proto/collector/trace/v1/trace_service.proto
+++ b/opentelemetry/proto/collector/trace/v1/trace_service.proto
@@ -42,4 +42,18 @@ message ExportTraceServiceRequest {
 }
 
 message ExportTraceServiceResponse {
+  // The details of a partially successfull export request.
+  // When the request is only partially successfull (due to e.g. invalid/bad data),
+  // the server MUST populate the details property with a developer-facing error
+  // message along with the number of accepted spans.
+  ExportTracePartialSuccess details = 1;
+}
+
+message ExportTracePartialSuccess {
+  // The number of accepted spans
+  int64 accepted_spans = 1;
+
+  // A developer-facing human-readable error message in English. It should
+  // both explain the error and offer an actionable resolution to it.
+  string error_message = 2;
 }


opentelemetry/proto/collector/logs/v1/logs_service.proto
```diff
@@ -42,4 +42,18 @@ message ExportLogsServiceRequest {
 }
 
 message ExportLogsServiceResponse {
+  // The details of a partially successfull export request.
+  // When the request is only partially successfull (due to e.g. invalid/bad data),
+  // the server MUST populate the details property with a developer-facing error
+  // message along with the number of accepted log records.
+  ExportLogsPartialSuccess details = 1;
```
jmacd, line 49:
```suggestion
  // The details of a partially successful export request.
  // When the request is only partially successful (due to e.g. invalid/bad data),
  // the server MUST populate the details property with a developer-facing error
  // message along with the number of accepted log records.
  ExportLogsPartialSuccess partial_success = 1;
```

I would prefer to name this field "partial_success" than "details", since the user who sees this data may not read the type name. I want the reader, seeing one of these, to read "I have a partial success", instead of "I have details".

jmacd:
Thank you @joaopgrassi, this feature is sorely needed. I have two related requests, which I'll ask for in connection with these protocol changes to see what you think.

## 1. Structured error details

[Lightstep's Prometheus sidecar converts Prometheus WAL logs into OTLP, and Lightstep has fairly strict metric type checking which results in partial successes.](https://github.com/lightstep/opentelemetry-prometheus-sidecar#validation-errors). Because we had no way to convey these, I used gRPC/HTTP response headers. The Lightstep backend produces a _**sample**_ of the metric names (max 1 per request) that are dropping data as a result; this works because the headers contain structured information, e.g. a header like:

```
otlp-invalid-${failed_metric_name}: failure explanation
```

The recipient of these headers turns them into metrics about failed metrics, which aggregated over time probabilistically contain all the failing metrics even though each response is limited in size. The feedback is that a single, unstructured error details string is possibly less useful than it could be. On the other hand, OTel has a data model for error logs: could this be a structured error message in the form of an OTLP log message returned by the server? Can we have semantic conventions for conveying sample errors that support aggregation as described above? 

## 2. Client-side telemetry consumed/dropped metrics for OTLP exporters

It is expensive to imagine printing the error details to the console, especially if they're repetitive. This leaves me wondering what the user is expected to do with the error details here, otherwise. Assuming the SDK supports logging, it's likely that the user is simply going to drop the error details back in a payload to the same service or a connected backend service. I would like us to discourage OTel users from echoing error details from server-to-client-to-server again, meaning: OTLP consumers SHOULD consider delivering the error_message to the user of their system directly and OTLP producers SHOULD NOT echo error details back through a default-configured logs SDK for the same process. If the error details log is failing itself, we do not want to create a cycle, but practically speaking that is all that we can expect from an unstructured message.

Users who monitor their OTLP exporters want to know how many spans/metrics/logs are being written and how many are being dropped. Partial successes indicate how many points are dropped when the request succeeds, but we need semantic conventions for counting points that are dropped for all reasons, which include total failures too. Again referring to the OTel Prometheus sidecar linked above as an example, I think we should declare metrics with names like `otlp.metrics.consumed` (1 per successful point), `otlp.metrics.dropped` (1 per failed point), `otlp.spans.consumed` (1 per successful span), `otlp.spans.dropped` (1 per failed span), and so on. The OTLP exporter will be expected to use the `partial_success` information about how many spans/metric points/logs succeeded as well as knowledge of how many were sent to generate these metrics. These metrics would be the first semantic conventions for OTLP SDKs to generate metrics about themselves; since these are meta-telemetry metrics I think they should be always use cumulative temporality, that way telemetry produced/dropped can be accounted for during network failures, for example. Unlike with logging, metrics cannot create a cycle because the signal is aggregated--if these metrics are dropped they will be accounted for as ordinary dropped metric points.

opentelemetry/proto/collector/logs/v1/logs_service.proto
tigrannajaryan: Is this field supposed to be unset if the request is fully successful? Or it can be set but accepted_log_records must match the number of records in the request?

tigrannajaryan:
Do ExportLogsPartialSuccess, ExportMetricsPartialSuccess and ExportTracesPartialSuccess need to be 3 separate message or we can have one ExportPartialSuccess message with `accepted_records` field?

Do we anticipate that these 3 may need to evolve in the future in different ways?

jack-berg:
> The recipient of these headers turns them into metrics about failed metrics, which aggregated over time probabilistically contain all the failing metrics even though each response is limited in size. The feedback is that a single, unstructured error details string is possibly less useful than it could be.

Can the client actually act on that information? And if it can, is conveying it to the client in a response such that it gets logged to standard out the best way to convey the information? As an alternative, the backend itself could track metrics about rejected data, and surface alert the user to it in the UI. 

> we need semantic conventions for counting points that are dropped for all reasons, which include total failures too

👍 Otel Java already [records](https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/ExporterMetrics.java) some experimental metrics about the counts of spans, metrics, and logs successfully / unsuccessfully exported via OTLP. Also have similar metrics we record about batch span / log processor successes and failures. 

> I think they should be always use cumulative temporality, that way telemetry produced/dropped can be accounted for during network failures, for example.

Disagree, since that implies that instrumentation has some say over the temporality. Could do something tricky like use an up down counter instead of a counter though 😁. This is a bit off topic though.

opentelemetry/proto/collector/logs/v1/logs_service.proto
```diff
@@ -42,4 +42,18 @@ message ExportLogsServiceRequest {
...
+}
+
+message ExportLogsPartialSuccess {
+  // The number of accepted log records
+  int64 accepted_log_records = 1;
```
jack-berg, line 54: If we invert this to be `rejected_log_records` it would be 0 under normal circumstances. I'm not a protobuf expert, but I imagine that could result in a small payload most of the time.

opentelemetry/proto/collector/metrics/v1/metrics_service.proto
```diff
@@ -42,4 +42,18 @@ message ExportMetricsServiceRequest {
...
+  ExportMetricsPartialSuccess details = 1;
+}
+
+message ExportMetricsPartialSuccess {
+  // The number of accepted data points
+  int64 accepted_data_points = 1;
+
+  // A developer-facing human-readable error message in English. It should
+  // both explain the error and offer an actionable resolution to it.
+  string error_message = 2;
```
reyang, line 58: This makes it hard for the client to perform any systematic handling besides putting the error message to a log file or somewhere else.

Consider something like:

```json
{"accepted_data_points": 5, "error": [
  {"index": 0, "status_code": 400, "message": "the timestamp xyz is invalid"},
  {"index": 3, "status_code": 502, "message": "the tenant does not exist"}
]}
```

opentelemetry/proto/collector/logs/v1/logs_service.proto
reyang: I vote for accepted instead of rejected.

opentelemetry/proto/collector/metrics/v1/metrics_service.proto
gsmirnov-splk: Very much agreed. The response should be very clear in indicating:
* which records were accepted successfully
* which records should be re-sent (e.g. after a throttling timeout)
* which should be dropped forever (e.g. due to the user having no entitlement or data being invalid).

opentelemetry/proto/collector/logs/v1/logs_service.proto
joaopgrassi: About the naming: I don't really have a preference, I struggled to find one, so `partial_success` is good for me. I was not happy with `details` as well. 

>Is this field supposed to be unset if the request is fully successful?
Yes, that's what I was going for. If all is well, I would assume it is not useful to dig into any details, since they can be inferred.

joaopgrassi:
@jmacd thanks for the review! I will try replying to your points:

> It is expensive to imagine printing the error details to the console, especially if they're repetitive. This leaves me wondering what the user is expected to do with the error details here, otherwise. Assuming the SDK supports logging, it's likely that the user is simply going to drop the error details back in a payload to the same service or a connected backend service. [...]

That's a good point, that didn't cross my mind.

The initial discussions in the issue lead us to think on rather simplistic cases. Having the exporter log the error message to stdout helps to quickly address things without having to go to my vendor UI and browser for logs, but sure, it's probably not a good practice and not a behavior we should encourage.

> Users who monitor their OTLP exporters want to know how many spans/metrics/logs are being written and how many are being dropped. Partial successes indicate how many points are dropped when the request succeeds, but we need semantic conventions for counting points that are dropped for all reasons, which include total failures too [...]

My view is that back-ends probably should also keep track of these metrics, since they are the ones actually doing the "parsing" of OTLP data, in whatever ways they do, but we are not the ones to tell them what to do I guess.

About introducing these "self monitoring" metrics to OTLP exporters, I think it makes sense, and we probably should do that (most likely not as part of this though). But I have one concern about

> Unlike with logging, metrics cannot create a cycle 

If we say exporters produce self monitoring metrics using whatever is returned in `partial_success`, couldn't it become also a cycle, if the produced self monitoring metrics also cause datapoints to be dropped? I'm just thinking out loud now, but what if I reached some sort of quota in my back-end, that causes the datapoints to be dropped (or any other condition that would lead to this)? Then further producing more, will create a cycle, no? Also, if we go this route, we need to make sure users can opt-out of this, since ingesting these sort of data costs money. Potentially even having granularity to opt-out for certain metrics only rather than a "all-or-nothing" switch.

About the topic around having a more structured response and not simply strings: This was briefly touched in the issue, and I agree with you that it would make sense, at least for metrics. For example, for metrics we could have something like:


```proto

message ExportMetricsPartialSuccess {
  // The number of accepted data points
  int64 accepted_data_points = 1;

  // A developer-facing human-readable error message in English. It should
  // both explain the error and offer an actionable resolution to it.
  string error_message = 2;

  repeated RejectedLogRecord rejected_datapoints  = 3;
}


message RejectedMetric {
  string error_message = 1; // human-readable reason/message about why it was rejected
  string metric_key = 2;
  uint32 index = 3
  ...
}
```

That would work I think, since we have the metric key + index of the datapoint it's easy to pin-point/identify it, also easy for users since they probably know their metric keys right away. But I'm not sure how could we achieve the same for Spans and Logs. How do we identify a span/log then? In a myriad of them, even an index might not be enough..

opentelemetry/proto/collector/logs/v1/logs_service.proto
tigrannajaryan: I also like "accepted" a bit more since it explicitly can be compared with the number of the records sent.
If all records are accepted then I think the expectation is that the entire ExportLogsPartialSuccess os omitted so the payload size issue is irrelevant.

opentelemetry/proto/collector/metrics/v1/metrics_service.proto
tigrannajaryan: I do not think that the detailed, structured error response is always desirable. If my receiver is under stress and needs to drop a significant number of records I probably do not want to spend time constructing a large payload with potentially thousands of elements in the error field in the response.

We may want the additional, more detailed "error" field that include this information, but I think it must be optional. I think a simple error_message needs to stay and is the one that is always required.

I would also prefer to start small, with just error_message first and only add the more detailed response later, if we really see the need and clearly understand the use cases.

joaopgrassi:
> Do ExportLogsPartialSuccess, ExportMetricsPartialSuccess and ExportTracesPartialSuccess need to be 3 separate message or we can have one ExportPartialSuccess message with accepted_records field?

> Do we anticipate that these 3 may need to evolve in the future in different ways?

From just the thread here around having more structure, I'm leaning more in having separated types. It makes things more obvious and if we later need different things for each Signal, we are prepared.  We already had different types anyway so it makes sense to me :)

opentelemetry/proto/collector/metrics/v1/metrics_service.proto
joaopgrassi: > We also need to consider client side retry - need to let the client know which part of data should be retried (e.g. in a partial success, there could be some retriable error AND some non-retriable error).


I'm not sure if this is even possible on the client. I feel there's too many variables involved and would be quite tricky to get retry in place. For example, back-ends have different requirements on how they deal with metrics and what they allow (e.g. special chars in metric names, number of attributes, etc) how would we make this generic enough that clients could figure it out what is "retry able" and what is not? Then there's things like quota exceeded, authorization (my subscription does not allow me to ingest certain types of logs).. it's just a lot, and to me, very close to back-end things and not OTel.

The [current spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#bad-data) says that for "Bad Data" the client must not retry and I feel for partial success is the same.  I'd vote to keep retry out of the equation here.

reyang: > I do not think that the detailed, structured error response is always desirable.

+1, it should be optional instead of required.

tigrannajaryan (approved):
This looks good to me as the first step to introduce partial success capabilities. I think we can accept it and adopt in OTLP implementations as is.
If we see the need for more detailed description of partial success it can be added as backward-compatible change in the future.

opentelemetry/proto/collector/logs/v1/logs_service.proto
joaopgrassi: Changed to `partial_success` (+ grammar) in [4bd818f](https://github.com/open-telemetry/opentelemetry-proto/pull/390/commits/4bd818fb5f743b1c2716a28ada57d242b0193e36)

opentelemetry/proto/collector/logs/v1/logs_service.proto
joaopgrassi: I will resolve this as `accepted` is what we are leaning towards. And like @tigrannajaryan mentioned, in case a export is successful (all logs, spans or metrics) are ingested this property will be null, so no change in the payload size compared to what we have now.

jmacd:
@tigrannajaryan asked:

> Do ExportLogsPartialSuccess, ExportMetricsPartialSuccess and ExportTracesPartialSuccess need to be 3 separate message or we can have one ExportPartialSuccess message with accepted_records field?

I like this idea. Can we have one type, instead of three?

joaopgrassi:
@jmacd hum.. I'm not sure. If the goal is to have structure and be able to achieve something with it, isn't relying on `AnyValue` and a list of `KeyValue` counterintuitive? How would I know what to expect in those blobs of data? Wouldn't that lead to code having to perform checks to find out what it's inside? I'm not sure if that's better than a simply string with the error. 

Just to be clear, I'm not against having more structure and I think  all your use cases are valid. I'm just not sure if we have enough knowledge as of know to be able to define such structure, so in this part I agree with @tigrannajaryan where maybe a simple first version is enough. I might be wrong as well. 

I just think that if we really want more structure, introducing proper types with well-defined properties makes more sense. I think this makes the work of deriving whatever we want from the partial success response easier instead of having to check in a map of things. WDYT?

joaopgrassi:
> If we return a simple string, I think we're not helping the user with their telemetry problem -- the error message just turns one telemetry problem into a new telemetry problem, which has to be reported as an exception for a human to parse.

@jmacd yes, I completely agree with you on this. As I mentioned earlier, I'm all in for a well structured response other than a simple string. 

What I'm not sure about is (maybe I'm not getting it) if `LogRecord` is a good candidate for it. I kinda see the value in the `AnyValue` body + the attributes list idea. It is definitely flexible since we can just add anything there (reason, error codes, etc) but what I'm trying to challenge here is: Isn't too much flexibility also bad? Doesn't this approach complicate things on the receivers side? Since it can basically hold anything, we would have to define semantics and everything around that to make sure receivers that want to handle it better can do so in a consistent way.

That's why I'm thinking if we shouldn't be considering having a type with well defined properties that will allow receivers derive whatever we decide it's important/relevant? e.g. your example with self monitoring metrics.

Also: `LogRecord` contains other fields like `trace id`, `span id` and `timestamp` which I'm not sure are relevant for this. Unless you were not suggesting to actually use this type but just copy those two fields? (`body` and `attributes`)?

>However, for a recipient that does not care to perform special checks (or an older version), the structured message is just a simple message. Having structure lets us define semantic conventions later while still conveying error messages now; it's backwards compatible.

Yes, this is true for "both approachs" we are discussing - using the `LogRecord` or adding well-defined fields in the new types introduced in this PR. Recievers that don't care (or old ones) will simply not care for the message.

opentelemetry/proto/collector/logs/v1/logs_service.proto
```diff
@@ -42,4 +42,18 @@ message ExportLogsServiceRequest {
...
 
 message ExportLogsServiceResponse {
+  // The details of a partially successful export request.
+  // When the request is only partially successful (due to e.g. invalid/bad data),
+  // the server MUST populate the partial_success property with a
+  // developer-facing error message along with the number of accepted log records.
+  ExportLogsPartialSuccess partial_success = 1;
+}
+
+message ExportLogsPartialSuccess {
```
bogdandrutu, line 52: Why another message instead of embedding it in the parent message?

joaopgrassi: You mean why not add the new fields directly on the existing `Export*ServiceResponse` types? I mentioned that in the PR description with more detail, but in a nutshell it's that senders will have to check if the new fields exist before accessing them (for backwards compatibility) as they don't know which proto version the receivers are returning. Having all new fields in a new message, makes the check easier, as in if the message is `null` it means either no problem or the receiver is using an older proto version. Whereas if we have the fields on the parent, existing message, which fields should receivers use to check if a "partial success" was returned?

jmacd:
@joaopgrassi 

> Since it can basically hold anything, we would have to define semantics and everything around that to make sure senders that want to handle it better can do so in a consistent way.

The same is true for a bare string called "error_message". It can hold anything and it can be printed. I think you are reasoning that the client-side logic for say reporting which metrics are experiencing errors will be simpler if we spell it out in `.proto` structure, as opposed to semantic-conventional structure. If we take the explicit-structure route, it means that old clients sending to new servers won't see any benefit until they upgrade. If we structure an error message with semantic conventions, it prints as an error message to the old client while it serves as both an error message and structured detail to the new client.

Either way, I don't want the structure of the error message to block this work. I care about monitoring the number of dropped metric points much more than the client being able to identify which metric names are being dropped. (After all, we are able to report which metric names are being dropped directly through our SaaS, client's don't need to track this detail.)

joaopgrassi:
>If the exporter is being traced itself, there is a well-defined span/trace associated with the request and it could be used in the error message.

Hum I think I'm a bit confused by this. The server/receiver is the one that will ingest and produce the `partial_success` response (if necessary) to senders, how would it include a spanid in the response? The exporter might be traced, yes, but I'm failing to see how that resonates with the response the server is producing? 


> The same is true for a bare string called "error_message". It can hold anything and it can be printed. I think you are reasoning that the client-side logic for say reporting which metrics are experiencing errors will be simpler if we spell it out in .proto structure, as opposed to semantic-conventional structure.

Yes. I personally prefer to have explictly defined fields in `.protos` instead of relying on conventions ,be it on a string, or via attributes we may define. I think it just makes the code harder and more complicated + open for mistakes as implementors have to follow the conventions correctly. If we offer concrete fields, there's no way to "mess it up".

>If we take the explicit-structure route, it means that old clients sending to new servers won't see any benefit until they upgrade. If we structure an error message with semantic conventions, it prints as an error message to the old client while it serves as both an error message and structured detail to the new client.

The way I see it, old clients/senders cannot use anything we are defining here anyway. Both approaches (having explicit-structure vs string-error message with conventions) requires adding new fields to the `Export*ServiceResponse` proto messages. Old clients will not have the new fields so, nothing we can do. As I mentioned in the OP along with the test app, only new clients/receivers can parse it, but they also always need to check if it exists, as they might be talking with old servers/receivers. 

>Either way, I don't want the structure of the error message to block this work. I care about monitoring the number of dropped metric points much more than the client being able to identify which metric names are being dropped. (After all, we are able to report which metric names are being dropped directly through our SaaS, client's don't need to track this detail.)

Yes, agreed. That is possible with the current proposal, as we can derive the number of rejected datapoints, spans, logs. As mentioned, I think that is a sensible first start, and we can discuss about the structure in separated issues as I suggested.

joaopgrassi:
@jmacd @bogdandrutu, @reyang @jack-berg  et al. pinging to see if we have reached a consensus here. My understand of the current state of discussions is: 

The current PR is a good start and improves the current status quo because:

- Allows receivers to partially accept data
- Allows receivers to return _some_ information to senders about the partial success - Today, the only possibility is returning a 400 where _something_ can be added to the response body
- Allows senders to potentially derive metrics around number of dropped logs/spans/datapoints

But it is not where we want to stay because:

- Only an string with a error message is not so useful - will become a log in the receivers and that's it
- We can't do more complex handling of a partial success - derive more metrics, maybe even retrying where we can

As I suggested above, we could go with this as is as it's already a great improvement from the status quo. Then I could create separated issues to discuss _what_ more complex scenarios we want to handle _per signal_ and then go from there and add a more complex structure to the messages introduced here. 

What do you all think?

jack-berg (approved):

jmacd (approved):
I don't care for the error details field, but the partial success are important.

opentelemetry/proto/collector/metrics/v1/metrics_service.proto
joaopgrassi: I opened an issue to discuss the retry + more structure #404 to not block this. I will mark this as resolved.

opentelemetry/proto/collector/logs/v1/logs_service.proto
joaopgrassi: Hi @bogdandrutu does ^ makes sense to you?

joaopgrassi:
I created an issue to continue the discussion around more structure and next steps for metrics #404. This is so this is not blocked and we have a place to continue the discussions from here.

carlosalberto:
Ping @bogdandrutu - waiting for one more review from you ;)

reyang (approved):

bogdandrutu (approved):

opentelemetry/proto/collector/logs/v1/logs_service.proto
```diff
@@ -43,4 +43,18 @@ message ExportLogsServiceRequest {
 }
 
 message ExportLogsServiceResponse {
+  // The details of a partially successful export request.
+  // When the request is only partially successful (due to e.g. invalid/bad data),
+  // the server MUST populate the partial_success property with a
+  // developer-facing error message along with the number of accepted log records.
+  ExportLogsPartialSuccess partial_success = 1;
```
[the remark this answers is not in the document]
bogdandrutu: If that is not the case then let's make it optional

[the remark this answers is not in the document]
jmacd: I agree. @joaopgrassi made the point in in https://github.com/open-telemetry/opentelemetry-proto/pull/390#discussion_r885831423, so we should call these optional since presence/absence is considered meaningful.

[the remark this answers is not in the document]
joaopgrassi: Hi all, I'm not entirely sure to which field you are referring to? The `partial_success` one or the numeric field with the number of accepted records? 

If it's for the non-primitive `partial_success` field, I'm not sure if there's any gain in marking it optional? Since it's not a primitive type, we will not have the _explicit presence_ semantics on it, right? I'm not a proto expert but that's how I understand it. 

If you are talking about the numeric field, I also think we should not mark it as optional. The idea of the new `partial_success` field is to **only** be populated in cases where there was a partial success AND the server is able to report it (processing of the request is not done at a later point for ex). 

In that case, the idea I believe is to make it a MUST in the spec. Meaning, `partial_success` will always be initialized and the numeric field containing the accepted records is always there. So `0` there would really mean 0 records were accepted. 

For the "total success case" e.g. when ALL records were accepted by the server, OR the server does the processing later thus can't know how many records were accepted, the `partial_success` should NOT be set, so it will be null.

Does this make sense? I put these requirements in the spec change here https://github.com/open-telemetry/opentelemetry-specification/pull/2636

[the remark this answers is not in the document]
joaopgrassi: For example, I can change the comments here to reflect the spec text more or less:

```
  // The details of a partially successful export request.
  // If the processing of the request is partially successful
  // (i.e. when the server accepts only parts of the data and rejects the rest)
  // and the number of accepted log records is known, the server MUST initialize
  // the partial_success field and MUST set the accepted_log_records.

  // When the export request is processed successfully, the server MUST leave
  // the partial_success field unset.
```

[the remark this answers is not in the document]
joaopgrassi: @bogdandrutu as we talked, I added some doc on `error_message` stating that an empty string is equivalent of it not being set at all, to avoid the overhead. I don't think it's important to distinguish between it "ever being set" from "empty string". Thanks for the pointer. [f299099](https://github.com/open-telemetry/opentelemetry-proto/pull/390/commits/f29909964eaf2885c2a8a8c831c97dc80aa5a861)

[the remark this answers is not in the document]
joaopgrassi: @jmacd is that also what you meant? Or did I miss something else?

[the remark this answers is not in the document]
jmacd: I believe the question here is whether when `error_message` is set to a non-empty string AND `accepted_things` equals 0, what does it mean? Since the error message is set and number of accepted things is not optional, the implication is that 0 is meaningful, i.e., no data was accepted. Thus if you are going to set an error message, you MUST set the number of accepted things. Is this true?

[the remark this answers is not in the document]
jmacd: Or, @joaopgrassi is the implication that if no "things" were accepted, there is no "partial" success, and an RPC error should be returned instead? That seems awkward to me: if every point drops, I'd like to return a partial success with 0 accepted things.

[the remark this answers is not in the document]
joaopgrassi: @jmacd 

>if every point drops, I'd like to return a partial success with 0 accepted things.

Yes, that is the approach we are going for here.

The idea is: If there's a partial success (meaning, the server knows before responding to clients how many "things" it accepted and how many it rejected) then the server MUST initialize the `partial_success` message and it MUST set the number in the `accepted_xxx` field. Here then `0` is meaningful. The server dropped all spans/points/log records. I think it's important that this case is supported like this instead of relying on RPC error.

The server MAY also set the `error_message` but that's not required. If `error_message` is `""`, it means it was either empty by intention or never set. There's no distinction in this case. Since we don't offer guidance on what to put in the message, making it a MAY makes sense to me.

I wrote all of these nuances in the spec. I would appreciate if you all could take a look at it. Hopefully it clarifies all the questions brought up here: https://github.com/open-telemetry/opentelemetry-specification/pull/2636

bogdandrutu (approved):

opentelemetry/proto/collector/logs/v1/logs_service.proto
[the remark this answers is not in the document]
jmacd: Three possibilities:
1. Make each `*PartialSuccess` field optional
2. Make each `accepted_*` field optional
3. Require error_details to be non-empty when `accepted_*` is meaningfully zero

I prefer number 3.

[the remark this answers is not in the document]
joaopgrassi: @jmacd I believe the original intention was: 

If there was a "complete success" OR the server does the processing later, thus can't know how many things it will accept, then the `PartialSucess` **must** be left unset.

The `PartialSuccess` **must only be set** when there's in fact a partial success on the server. I think this way it's easier to reason about than making x field required IF y field is present with some value. `0` then means nothing was accepted and there's no other alternative interpretation. 

All of this is explained in the spec PR I have open: https://github.com/open-telemetry/opentelemetry-specification/pull/2636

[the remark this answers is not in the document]
jmacd: Google has written an explainer on field presence for proto2 and proto3, https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md,

The implication (under proto3) is that you cannot set an `error_details` string without causing the whole message to be serialized, at which point it will look like the `accepted_things` field is present. So behavior (3) described by me is equivalent to number (1) as long as google doesn't go changing the definition of field presence again.

By these rules, you cannot encode a meaningful 0-valued `accepted_things` inside a message with an empty `error_details`, because it is equivalent to the serialized form of a not-present field; the deserializer seeing a `0` and an empty string will leave you a null and your meaningful-0 disappears.

To ensure that we get a `null` when the field is not set, we need to use the optional keyword. I think we should go with option 1 (functionally equivalent to 3).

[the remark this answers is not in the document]
jmacd: @bogdandrutu do you agree?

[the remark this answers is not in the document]
joaopgrassi: @jmacd 

Let's make sure we are in the same page for the "total success" case. In this, the new message here will not be set, as per spec so we are good. This is how we will distinguish a "total success" case from a "partial success". Also, it's not a primitive type, so there's no point in making that `optional`.

Now, if I got your point across, your concern is: There's a partial success. The server _can_ instantiate the `partial_success` sub message and leave everything as is with their default values. With this, it will not be possible to tell if `0` was actually set or if it was just the default?

If this is the case, then I think making `int64 accepted_*` as `optional` would solve it, no? The spec will say that when `partial_success` is initialized, the server MUST set a value to `accepted_*` so, it will be clear then that `0` actually means nothing was accepted. 

If the server doesn't follow this, and initializes the `partial_success` anyway, at least we will have `accepted_` as not set/null, as marking it as `optional` enables that in proto3.

With this I think we don't need to touch the `error_message`. We discussed here a lot that this message is probably not that useful to receivers so I feel we shouldn't use it's state to any logic definition of what actually happened. 

I also feel we might be talking pass each other and it's tricky to convey all this with words 😅. I'm up for a live discussion, be it in the next spec call or via Slack. I feel that will be more productive. What do you think?

opentelemetry/proto/collector/logs/v1/logs_service.proto
```diff
@@ -43,4 +43,26 @@ message ExportLogsServiceRequest {
...
+  // (i.e. when the server accepts only parts of the data and rejects the rest)
+  // and the number of accepted log records is known, the server MUST initialize
+  // the partial_success field and MUST set the accepted_log_records.
+
+  // When the export request is processed successfully, the server MUST leave
+  // the partial_success field unset.
+  ExportLogsPartialSuccess partial_success = 1;
+}
+
+message ExportLogsPartialSuccess {
```
jsuereth, line 57: Are we attempting to report partial failure so... we can retry problematic logs?

Or is this more about NOT retrying e.g. a "log of death"?

Specifically, should this include some notion of WHICH logs were rejected?

joaopgrassi: > Or is this more about NOT retrying e.g. a "log of death"?

More or less on this I'd say. 

It's about informing that a server accepted only parts of the data. The why it happened can be due to N reasons and depends highly on the server, so no attempt of retry is defined for now. The current state of this PR can also allow senders to "selfmonitor" and report the number of accepted things as a metric for example.

Also no attempt to indicate which logs/datapoints/spans the server rejected, because that is still not clear how it will be modeled and also could be a burden on receivers under pressure for ex. There was a discussion around this here https://github.com/open-telemetry/opentelemetry-proto/pull/390#discussion_r871989856

> Specifically, should this include some notion of WHICH logs were rejected?

This is also related to the discussions I linked above. We agreed that including more info to determine which logs/spans/data points were rejected would make sense but we need to discuss it better since it comes with its own set of challenges. I suggested creating separated issues to discuss it per sinal to not block this. There's one for metrics here https://github.com/open-telemetry/opentelemetry-proto/issues/404

opentelemetry/proto/collector/logs/v1/logs_service.proto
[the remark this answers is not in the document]
joaopgrassi: I went with @jmacd option (2)

> Make each accepted_* field optional

This way, if the server doesn't set it, it will be possible to distinguish the 'not set' value from the default one, `0`. I also extended the documentation on the field to make this clear. 

I will mark this conversation as resolved.

MrAlias (approved):

jmacd:
Looks good! Let's get https://github.com/open-telemetry/opentelemetry-specification/pull/2636 approved and merged first.

opentelemetry/proto/collector/logs/v1/logs_service.proto
```diff
@@ -43,4 +43,35 @@ message ExportLogsServiceRequest {
...
+  // The number of accepted log records
+  //
+  // If the request is partially successful and the number of
+  // accepted log records is known, the server MUST set
```
tigrannajaryan, line 61: What is the situation when the number of accepted log records is unknown and why is it useful for the sender to know about this situation specifically? If there is no such situation I suggest that we get rid of this and require the server always sets this number (if it doesn't know then set it either to 0 or to total number - either way is fine). Doing so will eliminate the need to have optional accepted_log_records (optional fields are poorly supported in some Protobuf libraries and are a pain to deal with).

joaopgrassi: >What is the situation when the number of accepted log records is unknown and why is it useful for the sender to know about this situation specifically

I also tend to think this shouldn't happen. If the server doesn't know the number of accepted "things", it should return an empty response, without populating the `partial_success` field, just as it is done today.

The discussion around `optional` was without it, a server could initialize the `partial_success` property, set the `error_message` and leave the `accepted_<signal>` field with the default value, which is `0`. In this situation, it would not be possible to differentiate a `0` being the the default/unset from the server actually dropping everything.

@jmacd had some other alternatives to this here: https://github.com/open-telemetry/opentelemetry-proto/pull/390#discussion_r916322737

joaopgrassi: >require that the accepted_<signal> field value is always set (if partial_success is set)

Yes, I agree and that's what I wrote in the spec now. Although we still have the issue where a sender receives a `partial_success` with something in the error message and `0` for `accepted_<signal>` and it then must interpret it as "nothing was accepted" which is clear to me, but could be open to other interpretations, since `0` is also the not set/default value.

tigrannajaryan: I still see `optional` qualifier for `accepted_log_records` field. Do you plan to remove it or you disagree with my opinion that we don't need it?

joaopgrassi: I didn't remove the `optional` in any for now. I wanted to see what others think before changing it, specially since there is this "nuance" with the default/not set value being the same thing. I know from talking with @jmacd that he had concerns with this double meaning. I think I agree with you that if the spec says to always populate it, if you get a `0` it means nothing was accepted so I'm in favor of removing it.

jmacd: > IMO, we should: get rid of the optional for accepted_<signal> field require that the accepted_<signal> field value is always set (if partial_success is set). There is no such thing as ...

The problem is that "partial success is set" cannot be answered because without `optional` there is no such thing as field presence for messages. If the server responds with a partial_success message having 0 for the accepted_things field and an empty message, the receiver will see an unset partial_success. It should be possible to meaningfully indicate that 0 points were accepted, and we can't rely on field presence without `optional`. Otherwise, we need to infer that partial_success was "set" because it has a non-default field. 

My goal behind supporting this PR is to get to a place where clients have exact information about how many points are being successfully sent vs. dropped. There should be no ambiguity. Moreover, we should not have to fill in a struct when 100% of points are accepted, because it creates an extra burden on the sender. In order to have a meaningful zero and gain the performance benefit without `optional`, we need to change again this PR to count "dropped" things instead of "accepted" things.

I support removing `optional`, but all the "accepted_" fields should revert back to "dropped_" fields. The 0 value will always be meaningful, then.

jmacd:
The review thread for this PR is too long to effectively read. Please close and re-open after more progress in https://github.com/open-telemetry/opentelemetry-specification/pull/2636.

opentelemetry/proto/collector/logs/v1/logs_service.proto
joaopgrassi: I can do the change and submit a new PR if that's what everyone agrees, no problem. 

But I wonder, if we will not have the same dilemma again. If the server responds with a `partial_success` with `rejected_<signal>` = `0` and `error_message` = `"` what does that mean? Nothing was dropped? Then why did the server even initialized the `partial_success`, which it shouldn't?  Then we can argue that we will write in the spec that this shouldn't happen but isn't it what is already written. just the other way around?

> It should be possible to meaningfully indicate that 0 points were accepted

The current spec change says if everything was accepted, `partial_success` should **not** be set, so I feel it's already enough if you get a `partial_success` and `accepted_<signal>` is `0` then nothing was accepted. Sure, with `optional` there's _more_ certainty but without it, I feel we are also covered, like @tigrannajaryan mentioned. 

Moreover, we can't enforce what servers will do, only recommend what it _should_ do. I feel with what we have written in the spec, we already have it covered. If it's important to me that clients get this right,

jmacd: > If the server responds with a partial_success with rejected_<signal> = 0 and error_message = " what does that mean? Nothing was dropped? Then why did the server even initialized the partial_success, which it shouldn't?

There is effectively no test that distinguishes "server initialized the partial_success" vs not when the default-value is used, short of using `optional`.

Without `optional`, there are two choices:
- We either can infer that the number-value is meaningful because the string-value is not empty
- We can declare that the number-value is always meaningful. 

We cannot differentiate a field that is set to the default value from a field that was not set. To your question, when the partial success field is not set, as it will not be for older OTLP versions, I will assume nothing dropped. I do not believe there is any other valid interpretation, since we have not yet specified any form of partial success. Until we specify partial success, you have all-or-none semantics.

tigrannajaryan: To me this is reasonable semantics:
1. If `partial_success` is not set then it means the request is processed fully successfully. This also happens to be the old behavior, so no changes in the semantics, which is good - we are backward compatible.
2. If `partial_success` is set then it means something went wrong. The `accepted_<signal>` field inside `Export*PartialSuccess` message MUST be set. 0 is a valid value for `accepted_<signal>` field and means all records failed to process. `error_message` MUST be also set to indicate what the problem was. Empty string happens to be a valid value, meaning that there is no _additional_ error text to provide, however the entire situation is still erroneous.

I think this is all we need, it does not need to be more complicated than this.

How this is used in the Server:
1. If the Server processed everything successfully then don't set `partial_success`.
2. If processing of any record failed the set `partial_success` and calculate `accepted_<signal>` field value as `total records in the record`-`failed to process record`. Again, 0 is perfectly fine value to end up with. Also set `error_message` to indicate the failure reason (can be as simple as "some metric data points are invalid" or more complex if desirable, e.g. one line per failed record with precise failure reason).

How this is used in the Client:
1. If `partial_success` in the response is not set, good, nothing else to do.
2. If `partial_success` in the response is set then print an error message in the log and include `error_message`. Also log `accepted_<signal>` value. Can also keep a metric based on `accepted_<signal>` value if desirable.

I may be missing something here. Perhaps there is a scenario that is not covered here and where the Server or the Client need to make a different decision and behave differently?

PR #396, opened 2022-05-26
add protoc-gen-swift to Makefile to update proto files for opentelemetry-swift

vvydier:
I would like to add a gen-swift task in the Makefile, so we can update the protobuf files for opentelemetry-swift when a change happens in the proto repo:
How do I add the swift's protoc (https://github.com/apple/swift-protobuf) in the build process. Looks like its using the docker image otel/build-protobuf . I get this error when I run `make gen-swift`

```
Please specify a program using absolute path or make sure the program is available in your PATH system variable
--swift_out: protoc-gen-swift: Plugin failed with status code 1.
make: *** [gen-swift] Error 1
```

The build is using a docker image otel/build-protobuf:0.9.0 which has protoc for other languages but not one for swift. How do I add swift's protoc (https://github.com/apple/swift-protobuf) so we can use it to build the protobuf files for swift as well?

Thanks

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ endef
 
 # Generate all implementations
 .PHONY: gen-all
-gen-all: gen-cpp gen-csharp gen-go gen-java gen-kotlin gen-objc gen-openapi gen-php gen-python gen-ruby
+gen-all: gen-cpp gen-csharp gen-go gen-java gen-kotlin gen-objc gen-openapi gen-php gen-python gen-ruby gen-swift
 
 OTEL_DOCKER_PROTOBUF ?= otel/build-protobuf:0.9.0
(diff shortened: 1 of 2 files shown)


bogdandrutu:
@vvydier the docker image that we use is generated from https://github.com/open-telemetry/build-tools/tree/main/protobuf, so go ahead and change the Dockerfile to include swift support.

tigrannajaryan:
This looks stale. Is this still in progress?

vvydier:
> ress?

vvydier:
The docker image for build is based on alpine linux https://github.com/open-telemetry/build-tools/blob/main/protobuf/Dockerfile#L1 and there is no distribution of swift on alpine linux. There is swift for ubuntu, centos and amazon linux https://www.swift.org/blog/additional-linux-distros/ so this method will not work for using the grpc-plugin for swift, unbless we switch to a docker that uses a base image that swift is supported on.

Issue #363, opened 2022-02-23
Revisit enum value names


bogdandrutu:
## Background 

Currently we do not have a consistent way of naming values in the enums:

```proto
  enum StatusCode {
    STATUS_CODE_UNSET               = 0;
    STATUS_CODE_OK                  = 1;
    STATUS_CODE_ERROR               = 2;
  };
```

```proto
  enum SpanKind {
    SPAN_KIND_UNSPECIFIED = 0;
    SPAN_KIND_INTERNAL = 1;
    SPAN_KIND_SERVER = 2;
    SPAN_KIND_CLIENT = 3;
    SPAN_KIND_PRODUCER = 4;
    SPAN_KIND_CONSUMER = 5;
  }
```

```proto
enum AggregationTemporality {
  AGGREGATION_TEMPORALITY_UNSPECIFIED = 0;
  AGGREGATION_TEMPORALITY_DELTA = 1;
  AGGREGATION_TEMPORALITY_CUMULATIVE = 2;
}
```

```proto
enum DataPointFlags {
  FLAG_NONE = 0;
  FLAG_NO_RECORDED_VALUE = 1;
}
```

```proto
// Possible values for LogRecord.SeverityNumber.
enum SeverityNumber {
  // UNSPECIFIED is the default SeverityNumber, it MUST NOT be used.
  SEVERITY_NUMBER_UNSPECIFIED = 0;
  SEVERITY_NUMBER_TRACE  = 1;
  SEVERITY_NUMBER_TRACE2 = 2;
  SEVERITY_NUMBER_TRACE3 = 3;
  SEVERITY_NUMBER_TRACE4 = 4;
  SEVERITY_NUMBER_DEBUG  = 5;
  SEVERITY_NUMBER_DEBUG2 = 6;
  SEVERITY_NUMBER_DEBUG3 = 7;
  SEVERITY_NUMBER_DEBUG4 = 8;
  SEVERITY_NUMBER_INFO   = 9;
  SEVERITY_NUMBER_INFO2  = 10;
  SEVERITY_NUMBER_INFO3  = 11;
  SEVERITY_NUMBER_INFO4  = 12;
  SEVERITY_NUMBER_WARN   = 13;
  SEVERITY_NUMBER_WARN2  = 14;
  SEVERITY_NUMBER_WARN3  = 15;
  SEVERITY_NUMBER_WARN4  = 16;
  SEVERITY_NUMBER_ERROR  = 17;
[...]

PR #397, opened 2022-05-27
Ensure enums are consistently declared

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-proto/issues/363

The rules are:
- Place enums outside the messages.
- Prefix the enum value names with enum name, using underscores with uppercase.

The changes in this PR should be backwards compatible (or allowed because
they are in experimental parts):
- For Binary Protobuf: the encoding does not use enum names anywhere on the wire.
  There are no changes on the wire at all.
- For JSON Protobuf:
  - DataPointFlags and LogRecordFlags are not visible on the wire since they are
    just helper enums to be used in the code.
  - Moving SpanKind and StatusCode out of the message should not result in any
    changes on the wire.
  - ConstantDecision is experimental and we are free to break it.

Note: all of these changes are breaking for the code that consumes the generated
proto files. We consider this acceptable since this repository is not yet declared
Stable.

I would like someone to independently confirm my analysis to make sure
this PR does not break anything on the wire (neither for binary nor for JSON).

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -364,12 +364,12 @@ enum AggregationTemporality {
 //   (point.flags & FLAG_NO_RECORDED_VALUE) == FLAG_NO_RECORDED_VALUE
 //
 enum DataPointFlags {
-  FLAG_NONE = 0;
+  DATA_POINTS_FLAG_NONE = 0;
 
   // This DataPoint is valid but has no recorded value.  This value
   // SHOULD be used to reflect explicitly missing data in a series, as
   // for an equivalent to the Prometheus "staleness marker".
-  FLAG_NO_RECORDED_VALUE = 1;
+  DATA_POINTS_FLAG_NO_RECORDED_VALUE = 1;
 
   // Bits 2-31 are reserved for future use.
 }
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -171,36 +171,6 @@ message Span {
   // This field is required.
   string name = 5;
 
-  // SpanKind is the type of span. Can be used to specify additional relationships between spans
-  // in addition to a parent/child relationship.
-  enum SpanKind {
-    // Unspecified. Do NOT use as default.
-    // Implementations MAY assume SpanKind to be INTERNAL when receiving UNSPECIFIED.
-    SPAN_KIND_UNSPECIFIED = 0;
-
-    // Indicates that the span represents an internal operation within an application,
-    // as opposed to an operation happening at the boundaries. Default value.
-    SPAN_KIND_INTERNAL = 1;
-
-    // Indicates that the span covers server-side handling of an RPC or other
-    // remote network request.
-    SPAN_KIND_SERVER = 2;
-
-    // Indicates that the span describes a request to some remote service.
-    SPAN_KIND_CLIENT = 3;
-
-    // Indicates that the span describes a producer sending a message to a broker.
-    // Unlike CLIENT and SERVER, there is often no direct critical path latency relationship
-    // between producer and consumer spans. A PRODUCER span ends when the message was accepted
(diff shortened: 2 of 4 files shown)


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -314,18 +314,18 @@ message Status {
...
-    STATUS_CODE_ERROR               = 2;
-  };
-
   // The status code.
   StatusCode code = 3;
 }
+
+// For the semantics of status codes see
+// https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status
+enum StatusCode {
```
bogdandrutu, line 323: SpanStatusCode? (this is a real question, not sure if it is better)

opentelemetry/proto/trace/v1/trace_config.proto
```diff
@@ -56,14 +56,15 @@ message ConstantSampler {
...
-    ALWAYS_PARENT = 2;
-  }
   ConstantDecision decision = 1;
 }
 
+enum ConstantDecision {
+  CONSTANT_DECISION_ALWAYS_OFF = 0;
+  CONSTANT_DECISION_ALWAYS_ON = 1;
+  CONSTANT_DECISION_ALWAYS_PARENT = 2;
+}
```
bogdandrutu, line 66: `ConstantSamplerDecision`?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -364,12 +364,12 @@ enum AggregationTemporality {
 //   (point.flags & FLAG_NO_RECORDED_VALUE) == FLAG_NO_RECORDED_VALUE
 //
 enum DataPointFlags {
```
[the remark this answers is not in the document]
tigrannajaryan: I used plural for flags to indicate that it is a bitset. Not a strong opinion. If you think `s` is unnecessary I can remove it.

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: If we do this it will break JSON format (since we also have to rename enum values). I wanted to avoid doing it. Last time we made changes people objected to breaking and we had to do complicated interoperability steps. I'd prefer to avoid doing this again unless we think it is important.

opentelemetry/proto/trace/v1/trace_config.proto
tigrannajaryan: Good call. Done.

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
bogdandrutu: Another option is to remove this, and say like we do for w3c trace context that last bit is used for that without any constant value.

tigrannajaryan:
> Do we need to do this? What value does it provide?

Consistency in enum names and how they are defined. I don't know if this is enough to justify the pains, but I think it is now or never since the names are also going to be hard-coded in JSON format.

> We saw when we renamed `InstrumentationLibrary` to `InstrumentationScope` that even though this repository has attempted to define "stable" in a way that would allow this change, the reality is that the generated identifiers are used in ways that expect them to be truly stable. We have many languages with stable trace and metrics systems that use these definitions because they were told that they were stable. We should have a very high bar for making disruptive name changes at this point and I don't think that this meets that bar.

To ease the transition we can try to keep the old enums/names in deprecated form. Would that help?

Also, after this change we can also decide if we want to make our stability guarantees to apply to the generated code. I think we can do it together with JSON stability (since it forces stability of many generated symbols anyway).

tigrannajaryan:
I submitted only helper enum changes as a separate PR https://github.com/open-telemetry/opentelemetry-proto/pull/405 to see if we can at least agree on that part.

tigrannajaryan:
Closing since we can't find consensus and there are no strong arguments and favour of the PR.

PR #399, opened 2022-05-31
Define csharp_namespace option

alanwest:
Fixes generated C# namespaces to have the prefix `OpenTelemetry` instead of `Opentelemetry`

### DIFF
--- a/opentelemetry/proto/collector/logs/v1/logs_service.proto
+++ b/opentelemetry/proto/collector/logs/v1/logs_service.proto
@@ -18,6 +18,7 @@ package opentelemetry.proto.collector.logs.v1;
 
 import "opentelemetry/proto/logs/v1/logs.proto";
 
+option csharp_namespace = "OpenTelemetry.Proto.Collector.Logs.V1";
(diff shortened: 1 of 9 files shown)


cijothomas (approved):

Kielek (approved):

arminru (approved):

tigrannajaryan:
Is this a breaking change for existing C# code? The existing code will no longer compile, right? Just want to confirm we are OK with this. 

Formally we are allowed to do this, since the code is explicitly excluded from our stability guarantees.

alanwest:
> Is this a breaking change for existing C# code? The existing code will no longer compile, right? Just want to confirm we are OK with this.

From the standpoint of the .NET SDK, we will be ok making this change as we do not expose any of the generated code to end users.

However, this will also be a breaking change for any end users that are generating code directly from these protos. Personally, I suspect the impact will be low. We should be ok citing that the protos have not reached 1.0 yet.

SergeyKanzhelev (approved):

bogdandrutu (approved):

Issue #400, opened 2022-06-07
Decide Conditions on Symbolic Stability to Declare OTLP 1.0


tigrannajaryan:
Data models and binary Protobuf for all 3 signals (traces,metrics,logs) is now declared Stable in OTLP.

We need to decide what else is remaining to declare OTLP as version 1.0.

We had a discussion in the Spec SIG meeting, here is the summary:

1. OTLP/JSON. We believe OTLP/JSON must be also declared Stable before OTLP is marked 1.0.0. This work is happening here https://github.com/open-telemetry/opentelemetry-specification/issues/1957

2. Symbolic name stability in the proto files. This has a few components:
- (a) Field name stability. I believe these must be declared stable and be part of our stability guarantees. The reason is that these names are visible on the wire for OTLP/JSON when using the default Protobuf/JSON implementation (while it is possible to decouple JSON wire and symbolic names in proto files using custom implementations, we do not think we want to go this route).
- (b) Enum value names for enums which are field types (e.g. SPAN_KIND_CLIENT). These are also visible on the wire for OTLP/JSON and I believe these must be declared stable.
- (c) Enum names (e.g. SpanKind). These are not visible on the wire anywhere (is this correct?).
- (d) Helper enum names and enum value names (e.g. LogRecordFlags). These are not visible on the wire anywhere.
- (e) The location of messages and enums, i.e. whether they are declared at the top lexical scope or nested inside another message. 
- (f) Message names, package names and directory structure.
- Note that (c),(d),(e),(f) have no impact on the wire format. However, they impact the generated code. There is currently no agreement whether these should be part of our stability guarantees.

3. Generated code stability, assuming no changes in proto files. This is dependent on the generator and we are NOT going to provide guarantees about generator behavior.

I believe 1, 2a and 2b are the absolute minimum before declaring OTLP 1.0. 

We need to decide what we do about 2c-f. This decision needs to be made before declaring OTLP 1.0.

tigrannajaryan:
@bogdandrutu @Aneurysm9 @dyladan @MrAlias @carlosalberto @ocelotl please see above and correct me if I missed anything.

Aneurysm9:
> 3\. Generated code stability, assuming no changes in proto files. This is dependent on the generator and we are NOT going to provide guarantees about generator behavior.

[...]

PR #405, opened 2022-06-21
Ensure helper enums are consistently declared

tigrannajaryan:
The changes in this PR do not affect wire formats in any way
since these are helper enums. This is not a breaking change
for the protocol.

This is a breaking change for the generated code.

This change is the smaller subset of https://github.com/open-telemetry/opentelemetry-proto/pull/397

### DIFF
--- a/opentelemetry/proto/logs/v1/logs.proto
+++ b/opentelemetry/proto/logs/v1/logs.proto
@@ -152,8 +152,8 @@ enum SeverityNumber {
 
 // Masks for LogRecord.flags field.
 enum LogRecordFlags {
-  LOG_RECORD_FLAG_UNSPECIFIED = 0;
-  LOG_RECORD_FLAG_TRACE_FLAGS_MASK = 0x000000FF;
+  LOG_RECORD_FLAGS_UNSPECIFIED = 0;
+  LOG_RECORD_FLAGS_TRACE_FLAGS_MASK = 0x000000FF;
 }
 
 // A log record according to OpenTelemetry Log Data Model:
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -362,15 +362,15 @@ enum AggregationTemporality {
 // enum is a bit-mask.  To test the presence of a single flag in the flags of
 // a data point, for example, use an expression like:
 //
-//   (point.flags & FLAG_NO_RECORDED_VALUE) == FLAG_NO_RECORDED_VALUE
+//   (point.flags & DATA_POINTS_FLAGS_NO_RECORDED_VALUE) == DATA_POINTS_FLAGS_NO_RECORDED_VALUE
 //
 enum DataPointFlags {
-  FLAG_NONE = 0;
+  DATA_POINTS_FLAGS_NONE = 0;
 
   // This DataPoint is valid but has no recorded value.  This value
   // SHOULD be used to reflect explicitly missing data in a series, as
   // for an equivalent to the Prometheus "staleness marker".
-  FLAG_NO_RECORDED_VALUE = 1;
+  DATA_POINTS_FLAGS_NO_RECORDED_VALUE = 1;
 
   // Bits 2-31 are reserved for future use.
 }


opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -152,8 +152,8 @@ enum SeverityNumber {
 
 // Masks for LogRecord.flags field.
 enum LogRecordFlags {
-  LOG_RECORD_FLAG_UNSPECIFIED = 0;
-  LOG_RECORD_FLAG_TRACE_FLAGS_MASK = 0x000000FF;
+  LOG_RECORD_FLAGS_UNSPECIFIED = 0;
+  LOG_RECORD_FLAGS_TRACE_FLAGS_MASK = 0x000000FF;
 }
```
bogdandrutu, line 157: I would rather remove them, since this repo is for the protocol definition and not for the "helpers" around the protocol.

tigrannajaryan: These help to implement the protocol correctly. Without these enums we have to explain the flags using English prose, which is more fragile and error prone. 

I understand the desire to minimize the contract surface but I believe to help increase OTLP popularity we need to take some pains ourselves in order to reduce the pains for others who adopt OTLP. See my comment about the proposed approach here: https://github.com/open-telemetry/opentelemetry-proto/issues/400#issuecomment-1161848977

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -365,12 +365,12 @@ enum AggregationTemporality {
 //   (point.flags & FLAG_NO_RECORDED_VALUE) == FLAG_NO_RECORDED_VALUE
 //
 enum DataPointFlags {
-  FLAG_NONE = 0;
+  DATA_POINTS_FLAG_NONE = 0;
```
Aneurysm9, line 368:
```suggestion
  DATA_POINTS_FLAGS_NONE = 0;
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -365,12 +365,12 @@ enum AggregationTemporality {
...
 
   // This DataPoint is valid but has no recorded value.  This value
   // SHOULD be used to reflect explicitly missing data in a series, as
   // for an equivalent to the Prometheus "staleness marker".
-  FLAG_NO_RECORDED_VALUE = 1;
+  DATA_POINTS_FLAG_NO_RECORDED_VALUE = 1;
```
Aneurysm9, line 373:
```suggestion
  DATA_POINTS_FLAGS_NO_RECORDED_VALUE = 1;
```

If we're going to change these we should be consistent with the `LOG_RECORD_FLAGS_*` values.

opentelemetry/proto/logs/v1/logs.proto
bogdandrutu: I disagree since a standard like w3c does not define "helpers" for trace flags for example. Also it is very bad design right now that helpers related to trace flags are defined in "logs" package.

I strongly suggest to remove for the moment, and determine if you are right or not about the "wrong" implementation, once we first hear about that we can have a small design for the helpers: where do they live, what are the guarantees since in this repo we mention that we guarantee wire not generated code (then how do you use this?), etc.

opentelemetry/proto/metrics/v1/metrics.proto
bogdandrutu: @tigrannajaryan my point on removing seems to be proven here. I `DATA_POINTS_FLAGS_NO_RECORDED_VALUE` sounds like the value of the "flags" not the value of the "mask" that you need to "bit and/or". I do suggest even more to remove them.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: >I disagree since a standard like w3c does not define "helpers" for trace flags for example.

I am not sure the analogy with w3c standard that you bring helps. The entire w3c trace context spec is just English prose, so there is no mechanism available for formal, compiler-enforced enum declarations there. I think the fact that w3c standard doesn't define helpers is not an argument that helps us to make a decision in our case, where such Protobuf compiler-enforced mechanism is luckily available.

>Also it is very bad design right now that helpers related to trace flags are defined in "logs" package.

I think there is some confusion about what this enum is. This helper is not trace flags definitions. This helper enum is the definition of `LogRecord.flags` field. Yes, 8 bits of the `flags` field will contain the value of w3c trace flags, but note how we don't define what is contained in those 8 bits here - the fact that only 1 bit is actually used here is irrelevant here, is not OTLP's business and is intentionally omitted. There may be other LogRecord flag definitions added here in the future, totally unrelated to the w3c trace flags. So, this is really `LogRecord.flags` field definition. Can you clarify why you think it is a bad design?

To be clear, I am not strongly opposed to removing the helpers. I do think the helpers are useful and have a slight preference to keep them, but if you see strong arguments in favour of removing we can remove them.

opentelemetry/proto/metrics/v1/metrics.proto

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -362,15 +362,15 @@ enum AggregationTemporality {
 // enum is a bit-mask.  To test the presence of a single flag in the flags of
 // a data point, for example, use an expression like:
 //
-//   (point.flags & FLAG_NO_RECORDED_VALUE) == FLAG_NO_RECORDED_VALUE
+//   (point.flags & DATA_POINTS_FLAGS_NO_RECORDED_VALUE) == DATA_POINTS_FLAGS_NO_RECORDED_VALUE
 //
 enum DataPointFlags {
-  FLAG_NONE = 0;
+  DATA_POINTS_FLAGS_NONE = 0;
```
jsuereth, line 368: This name change seems 100% inline w/ C++ generated code (vs. other language).  Is that correct?  Do we have a naming convention for enums that requires the value name to include the enum name?

opentelemetry/proto/logs/v1/logs.proto
jsuereth: I'm with @tigrannajaryan.   Soft-yes to include them, mostly because I don't see strong arguments for removing them.

jsuereth (approved):

tigrannajaryan:
Closing since there is no consensus on how to move forward.

PR #406, opened 2022-06-21
Clarify current stability guarantees

tigrannajaryan:
This PR captures the _current_ understanding of stability guarnatees.
I want to make sure it is explicitly visible until we make a decision
on https://github.com/open-telemetry/opentelemetry-proto/issues/400

The most likely outcome of https://github.com/open-telemetry/opentelemetry-proto/issues/400
is going to be stronger than current definition of "Stable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,13 +39,35 @@ All messages                         | Alpha    |
 (See [maturity-matrix.yaml](https://github.com/open-telemetry/community/blob/47813530864b9fe5a5146f466a58bd2bb94edc72/maturity-matrix.yaml#L57)
 for definition of maturity levels).
 
-Note that maturity guarantees apply only to wire-level compatibility for the binary
-Protobuf serialization. Neither message, field, nor enum names of Protobuf messages
-are visible on the wire and are not considered part of the guarantees. We are free
-to make a change to the names.
+## Stability Definition
 
-In the future when OTLP/JSON is declared stable, field names will also become part of
-the maturity guarantees, since field names are visible on the wire for JSON encoding.
+Components marked `Stable` provide the following guarantees:
+
+- Field types will not change.
+- Field numbers will not change.
+- Numbers assigned to enum choices will not change.
+- Service names and service package names will not change.
+- Service operation names, parameter and return types will not change.
+
+The following changes are allowed:
+
+- Message names may change.
+- Field names may change.
+- Enum names may change.
+- Enum choice names may change.
+- The location of messages and enums, i.e. whether they are declared at the top
+  lexical scope or nested inside another message may change.
+- Package names may change.
+- Directory structure, location and the name of the files may change.
+
+Note that none of the above allowed changes affects the binary wire representation.
+
+No guarantees are provided whatsoever about the stability of the code that
+is generated from the .proto files by any particular code generator.
+
+In the future when OTLP/JSON is declared stable, several of the changes that
+are currently allowed will become disallowed since they are visible on the wire
+for JSON encoding.
 
 ## Experiments
 


bogdandrutu (approved):
Be clear that by "now" it means the time when "JSON" is alpha.

README.md
```diff
@@ -39,13 +39,33 @@ All messages                         | Alpha    |
...
 
-In the future when OTLP/JSON is declared stable, field names will also become part of
-the maturity guarantees, since field names are visible on the wire for JSON encoding.
+Components marked `Stable` provide the following guarantees:
+
+- Field types will not change.
+- Field numbers will not change.
+- Numbers assigned to enum choices will not change.
+- Service names and service package name will not change.
+- Service operation names, parameters names and types will not change.
```
bogdandrutu, line 50: Why params name (actually what is the param name)?

```proto
service LogsService {
  // For performance reasons, it is recommended to keep this RPC
  // alive for the entire life of the application.
  rpc Export(ExportLogsServiceRequest) returns (ExportLogsServiceResponse) {}
}
```

tigrannajaryan: Hmm, not names. It should be parameter and return types. Fixed.

tigrannajaryan:
> Be clear that by "now" it means the time when "JSON" is alpha.

@bogdandrutu I have this in the text:
```
In the future when OTLP/JSON is declared stable, several of the changes that
are currently allowed will become disallowed since they are visible on the wire
for JSON encoding.
```

Are you looking for more?

README.md
```diff
@@ -39,13 +39,33 @@ All messages                         | Alpha    |
...
-Protobuf serialization. Neither message, field, nor enum names of Protobuf messages
-are visible on the wire and are not considered part of the guarantees. We are free
-to make a change to the names.
+## Stability Definition
 
-In the future when OTLP/JSON is declared stable, field names will also become part of
-the maturity guarantees, since field names are visible on the wire for JSON encoding.
+Components marked `Stable` provide the following guarantees:
+
+- Field types will not change.
```
arminru, line 46: Is the "optional" attribute considered part of the type?

README.md
```diff
@@ -39,13 +39,33 @@ All messages                         | Alpha    |
...
+- Field numbers will not change.
+- Numbers assigned to enum choices will not change.
+- Service names and service package names will not change.
+- Service operation names, parameter and return types will not change.
+
+The following changes are allowed:
```
arminru, line 52:
```suggestion
The following changes are allowed as they do not affect the binary wire representation:
```
To clarify that this does not break anything (for readers not closely familiar with protobuf).

arminru (approved):

README.md
```diff
@@ -39,13 +39,33 @@ All messages                         | Alpha    |
...
+
+- Message names may change.
+- Field names may change.
+- Enum names may change.
+- Enum choice names may change.
```
pirgeo, line 57: Is it possible to add more Enum values? Above it just says `- Numbers assigned to enum choices will not change.`, which would still be fulfilled if adding more choices, right?

README.md
bogdandrutu: I would say no.

jack-berg (approved):

README.md
```diff
@@ -39,13 +39,33 @@ All messages                         | Alpha    |
...
+## Stability Definition
 
-In the future when OTLP/JSON is declared stable, field names will also become part of
-the maturity guarantees, since field names are visible on the wire for JSON encoding.
+Components marked `Stable` provide the following guarantees:
+
+- Field types will not change.
+- Field numbers will not change.
+- Numbers assigned to enum choices will not change.
+- Service names and service package names will not change.
```
reyang, line 49: What does "service package name" mean?

README.md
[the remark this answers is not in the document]
tigrannajaryan: I think we should go with UNDEFINED. That's by default how I think about our specs. If it is not specified then it is undefined and can become defined later.

README.md
pirgeo: Sounds good, do we need to call it out in the list of allowed changes? This is also somewhat related to @reyang's [comment](https://github.com/open-telemetry/opentelemetry-proto/pull/406#discussion_r908634739), as it would be UNDEFINED if we don't add it.

README.md
tigrannajaryan: Done.

tigrannajaryan:
I am merging this since I believe it is an improvement over the current state. If further improvements are desirable please submit new PRs.

Issue #412, opened 2022-07-04
Consider adding short keys for JSON encoding


tigrannajaryan:
`json_name` option can be used to shorten keys for JSON encoding. This can result in significant size savings for uncompressed JSON, which appears to be important for e.g. https://github.com/open-telemetry/oteps/pull/208#discussion_r904191239

tigrannajaryan:
Here is size comparison using one or two letter JSON keys:
```
===== Encoded sizes
Encoding                        Uncomp/json[Improved]   zlib/json[Improved]
OTLP/Logs                       55005 by   [1.000]      4627 by [1.000]
ShortKeys/Logs                  41768 by   [1.317]      4386 by [1.055]

Encoding                        Uncomp/json[Improved]   zlib/json[Improved]
OTLP/Trace/Attribs              41738 by   [1.000]      3513 by [1.000]
ShortKeys/Trace/Attribs         31998 by   [1.304]      3356 by [1.047]

Encoding                        Uncomp/json[Improved]   zlib/json[Improved]
OTLP/Trace/Events               49336 by   [1.000]      2572 by [1.000]
ShortKeys/Trace/Events          34396 by   [1.434]      2435 by [1.056]

Encoding                        Uncomp/json[Improved]   zlib/json[Improved]
OTLP/Metric/Histogram           37993 by   [1.000]      1809 by [1.000]
ShortKeys/Metric/Histogram      25171 by   [1.509]      1641 by [1.102]

Encoding                        Uncomp/json[Improved]   zlib/json[Improved]
OTLP/Metric/MixOne             149353 by   [1.000]      4185 by [1.000]
ShortKeys/Metric/MixOne        104231 by   [1.433]      3894 by [1.075]

Encoding                        Uncomp/json[Improved]   zlib/json[Improved]
OTLP/Metric/MixSeries          549540 by   [1.000]     13819 by [1.000]
ShortKeys/Metric/MixSeries     395682 by   [1.389]     13103 by [1.055]
```

The uncompressed JSON improvement with short keys is around 1.3-1.5 times compared to the regular keys, which use full field names.

Oberon00:
We could even use the numeric values (stringized) as JSON keys (decimal or hex or base64). Then we would also get more stable / renaming-resistant keys.

tigrannajaryan:
To summarize the findings in [the PR](https://github.com/open-telemetry/opentelemetry-proto/pull/413):
[...]

PR #413, opened 2022-07-05
Use short keys for OTLP/JSON

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-proto/issues/412

This change sets short one or two letter keys for all fields when JSON
encoding is used. This results in about 1.3-1.5 times smaller uncompressed
payload.

Here is size comparison using some sample data from exp-otelproto test bench:

```
===== Encoded sizes                                                       
Encoding                       Uncomp/json[Improved]   zlib/json[Improved]
OTLP 0.18/Logs                 52577 by   [1.000]      4601 by [1.000]    
ShortKeys/Logs                 39668 by   [1.325]      4344 by [1.059]    
                                                                          
Encoding                       Uncomp/json[Improved]   zlib/json[Improved]
OTLP 0.18/Trace/Attribs        41704 by   [1.000]      3189 by [1.000]    
ShortKeys/Trace/Attribs        31998 by   [1.303]      3060 by [1.042]    
                                                                          
Encoding                       Uncomp/json[Improved]   zlib/json[Improved]
OTLP 0.18/Trace/Events         49302 by   [1.000]      1917 by [1.000]    
ShortKeys/Trace/Events         34396 by   [1.433]      1806 by [1.061]    
                                                                          
Encoding                       Uncomp/json[Improved]   zlib/json[Improved]
OTLP 0.18/Metric/Histogram     42376 by   [1.000]      1067 by [1.000]    
ShortKeys/Metric/Histogram     27071 by   [1.565]       839 by [1.272]    
                                                                          
Encoding                       Uncomp/json[Improved]   zlib/json[Improved]
OTLP 0.18/Metric/MixOne       184836 by   [1.000]      2778 by [1.000]    
ShortKeys/Metric/MixOne       119031 by   [1.553]      2143 by [1.296]    
                                                                          
Encoding                       Uncomp/json[Improved]   zlib/json[Improved]
OTLP 0.18/Metric/MixSeries    707615 by   [1.000]     11482 by [1.000]    
ShortKeys/Metric/MixSeries    457010 by   [1.548]      9829 by [1.168]    
```

Unfortunately **this is a breaking change** for default configuration of
Protobuf/JSON marshaler, which marshals field names in lowerCamelCase.
This is not a breaking change for marshalers which use the "OrigName=true"
JSON marshalling option. Nothing changes in the output when "OrigName=true"
is used.

I do not see an easy way to make this change gracefully. It will require
duplicating the entire proto, give the duplicate messages different names,
then handle the duplicates in the receivers. It is quite a lot of work
that can be also error prone. I think in this particular case we should
not attempt to handle it gracefully and simply stick to our formal
stability guarantees, which for JSON are "Alpha" and allow any changes
any time.

### Example Outputs

Current JSON output before this change, using default lowerCamelCase marshaler:

```json
{
    "resourceSpans": [
        {
            "resource": {
                "attributes": [
                    {
                        "key": "StartTimeUnixnano",
                        "value": {
                            "intValue": "12345678"
                        }
                    },
                    {
                        "key": "Pid",
                        "value": {
                            "intValue": "1234"
                        }
                    },
                    {
                        "key": "HostName",
                        "value": {
                            "stringValue": "fakehost"
                        }
                    },
                    {
                        "key": "ServiceName",
                        "value": {
                            "stringValue": "generator"
                        }
                    }
                ]
            },
            "scopeSpans": [
                {
                    "scope": {
                        "name": "io.opentelemetry"
                    },
                    "spans": [
                        {
                            "traceId": "AQAAAAAAAADw3rwKeFY0Eg==",
                            "spanId": "AQAAAAAAAAA=",
                            "name": "load-generator-span",
                            "kind": "SPAN_KIND_CLIENT",
                            "startTimeUnixNano": "1572516672000000013",
                            "endTimeUnixNano": "1572516672000000013",
                            "attributes": [
                                {
                                    "key": "db.mongodb.collection",
                                    "value": {
                                        "stringValue": "!##$"
                                    }
                                }
                            ],
                            "events": [
                                {
                                    "timeUnixNano": "1572516672000000013",
                                    "attributes": [
                                        {
                                            "key": "te",
                                            "value": {
                                                "intValue": "1"
                                            }
                                        }
                                    ]
                                }
                            ]
                        }
                    ]
                }
            ]
        }
    ]
}
```

JSON output before and after this change, using OrigName=true marshaler.

```json
{
    "resource_spans": [
        {
            "resource": {
                "attributes": [
                    {
                        "key": "StartTimeUnixnano",
                        "value": {
                            "int_value": "12345678"
                        }
                    },
                    {
                        "key": "Pid",
                        "value": {
                            "int_value": "1234"
                        }
                    },
                    {
                        "key": "HostName",
                        "value": {
                            "string_value": "fakehost"
                        }
                    },
                    {
                        "key": "ServiceName",
                        "value": {
                            "string_value": "generator"
                        }
                    }
                ]
            },
            "scope_spans": [
                {
                    "scope": {
                        "name": "io.opentelemetry"
                    },
                    "spans": [
                        {
                            "trace_id": "AQAAAAAAAADw3rwKeFY0Eg==",
                            "span_id": "AQAAAAAAAAA=",
                            "name": "load-generator-span",
                            "kind": "SPAN_KIND_CLIENT",
                            "start_time_unix_nano": "1572516672000000013",
                            "end_time_unix_nano": "1572516672000000013",
                            "attributes": [
                                {
                                    "key": "db.mongodb.collection",
                                    "value": {
                                        "string_value": "!##$"
                                    }
                                }
                            ],
                            "events": [
                                {
                                    "time_unix_nano": "1572516672000000013",
                                    "attributes": [
                                        {
                                            "key": "te",
                                            "value": {
                                                "int_value": "1"
                                            }
                                        }
                                    ]
                                }
                            ]
                        }
                    ]
                }
            ]
        }
    ]
}
```

JSON output after this change, using proposed short keys and default
lowerCamelCase marshaler:

```json
{
    "s": [
        {
            "r": {
                "a": [
                    {
                        "k": "StartTimeUnixnano",
                        "v": {
                            "i": "12345678"
                        }
                    },
                    {
                        "k": "Pid",
                        "v": {
                            "i": "1234"
                        }
                    },
                    {
                        "k": "HostName",
                        "v": {
                            "s": "fakehost"
                        }
                    },
                    {
                        "k": "ServiceName",
                        "v": {
                            "s": "generator"
                        }
                    }
                ]
            },
            "s": [
                {
                    "i": {
                        "n": "io.opentelemetry"
                    },
                    "s": [
                        {
                            "ti": "AQAAAAAAAADw3rwKeFY0Eg==",
                            "si": "AQAAAAAAAAA=",
                            "n": "load-generator-span",
                            "k": "SPAN_KIND_CLIENT",
                            "s": "1572516672000000013",
                            "e": "1572516672000000013",
                            "a": [
                                {
                                    "k": "db.mongodb.collection",
                                    "v": {
                                        "s": "!##$"
                                    }
                                }
                            ],
                            "ev": [
                                {
                                    "t": "1572516672000000013",
                                    "a": [
                                        {
                                            "k": "te",
                                            "v": {
                                                "i": "1"
                                            }
                                        }
                                    ]
                                }
                            ]
                        }
                    ]
                }
            ]
        }
    ]
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Full list of differences found in [this compare](https://github.com/open-telemet
   InstrumentationLibraryMetrics messages. Delete deprecated
   instrumentation_library_logs, instrumentation_library_spans and
   instrumentation_library_metrics fields.
+* :stop_sign: [BREAKING] Use short keys for OTLP/JSON.
 
 ### Added
 
--- a/opentelemetry/proto/collector/logs/v1/logs_service.proto
+++ b/opentelemetry/proto/collector/logs/v1/logs_service.proto
@@ -39,7 +39,7 @@ message ExportLogsServiceRequest {
   // element. Intermediary nodes (such as OpenTelemetry Collector) that receive
   // data from multiple origins typically batch the data before forwarding further and
   // in that case this array will contain multiple elements.
-  repeated opentelemetry.proto.logs.v1.ResourceLogs resource_logs = 1;
+  repeated opentelemetry.proto.logs.v1.ResourceLogs resource_logs = 1 [json_name="l"];
 }
 
 message ExportLogsServiceResponse {
--- a/opentelemetry/proto/collector/metrics/v1/metrics_service.proto
+++ b/opentelemetry/proto/collector/metrics/v1/metrics_service.proto
@@ -39,7 +39,7 @@ message ExportMetricsServiceRequest {
   // element. Intermediary nodes (such as OpenTelemetry Collector) that receive
   // data from multiple origins typically batch the data before forwarding further and
   // in that case this array will contain multiple elements.
-  repeated opentelemetry.proto.metrics.v1.ResourceMetrics resource_metrics = 1;
+  repeated opentelemetry.proto.metrics.v1.ResourceMetrics resource_metrics = 1 [json_name="m"];
 }
 
 message ExportMetricsServiceResponse {
--- a/opentelemetry/proto/collector/trace/v1/trace_service.proto
+++ b/opentelemetry/proto/collector/trace/v1/trace_service.proto
@@ -39,7 +39,7 @@ message ExportTraceServiceRequest {
   // element. Intermediary nodes (such as OpenTelemetry Collector) that receive
   // data from multiple origins typically batch the data before forwarding further and
   // in that case this array will contain multiple elements.
-  repeated opentelemetry.proto.trace.v1.ResourceSpans resource_spans = 1;
+  repeated opentelemetry.proto.trace.v1.ResourceSpans resource_spans = 1 [json_name="s"];
 }
 
 message ExportTraceServiceResponse {
--- a/opentelemetry/proto/common/v1/common.proto
+++ b/opentelemetry/proto/common/v1/common.proto
@@ -29,13 +29,13 @@ message AnyValue {
   // The value is one of the listed fields. It is valid for all values to be unspecified
   // in which case this AnyValue is considered to be "empty".
   oneof value {
-    string string_value = 1;
-    bool bool_value = 2;
-    int64 int_value = 3;
-    double double_value = 4;
-    ArrayValue array_value = 5;
-    KeyValueList kvlist_value = 6;
-    bytes bytes_value = 7;
+    string string_value = 1 [json_name="s"];
+    bool bool_value = 2 [json_name="b"];
+    int64 int_value = 3 [json_name="i"];
+    double double_value = 4 [json_name="d"];
+    ArrayValue array_value = 5 [json_name="a"];
+    KeyValueList kvlist_value = 6 [json_name="l"];
+    bytes bytes_value = 7 [json_name="by"];
   }
 }
 
@@ -43,7 +43,7 @@ message AnyValue {
 // since oneof in AnyValue does not allow repeated fields.
 message ArrayValue {
   // Array of values. The array may be empty (contain 0 elements).
-  repeated AnyValue values = 1;
+  repeated AnyValue values = 1 [json_name="v"];
 }
 
 // KeyValueList is a list of KeyValue messages. We need KeyValueList as a message
@@ -56,22 +56,22 @@ message KeyValueList {
   // contain 0 elements).
   // The keys MUST be unique (it is not allowed to have more than one
   // value with the same key).
-  repeated KeyValue values = 1;
+  repeated KeyValue values = 1 [json_name="v"];
 }
 
 // KeyValue is a key-value pair that is used to store Span attributes, Link
 // attributes, etc.
 message KeyValue {
-  string key = 1;
-  AnyValue value = 2;
+  string key = 1 [json_name="k"];
+  AnyValue value = 2 [json_name="v"];
 }
 
 // InstrumentationScope is a message representing the instrumentation scope information
 // such as the fully qualified name and version. 
 message InstrumentationScope {
   // An empty instrumentation scope name means the name is unknown.
-  string name = 1;
-  string version = 2;
-  repeated KeyValue attributes = 3;
-  uint32 dropped_attributes_count = 4;
+  string name = 1 [json_name="n"];
+  string version = 2 [json_name="v"];
+  repeated KeyValue attributes = 3 [json_name="a"];
+  uint32 dropped_attributes_count = 4 [json_name="da"];
 }
--- a/opentelemetry/proto/logs/v1/logs.proto
+++ b/opentelemetry/proto/logs/v1/logs.proto
@@ -41,7 +41,7 @@ message LogsData {
   // one element. Intermediary nodes that receive data from multiple origins
   // typically batch the data before forwarding further and in that case this
   // array will contain multiple elements.
-  repeated ResourceLogs resource_logs = 1;
+  repeated ResourceLogs resource_logs = 1 [json_name="l"];
 }
 
 // A collection of ScopeLogs from a Resource.
@@ -50,14 +50,14 @@ message ResourceLogs {
 
   // The resource for the logs in this message.
   // If this field is not set then resource info is unknown.
-  opentelemetry.proto.resource.v1.Resource resource = 1;
+  opentelemetry.proto.resource.v1.Resource resource = 1 [json_name="r"];
 
   // A list of ScopeLogs that originate from a resource.
-  repeated ScopeLogs scope_logs = 2;
+  repeated ScopeLogs scope_logs = 2 [json_name="l"];
 
   // This schema_url applies to the data in the "resource" field. It does not apply
   // to the data in the "scope_logs" field which have their own schema_url field.
-  string schema_url = 3;
+  string schema_url = 3 [json_name="u"];
 }
 
 // A collection of Logs produced by a Scope.
@@ -65,13 +65,13 @@ message ScopeLogs {
   // The instrumentation scope information for the logs in this message.
   // Semantically when InstrumentationScope isn't set, it is equivalent with
   // an empty instrumentation scope name (unknown).
-  opentelemetry.proto.common.v1.InstrumentationScope scope = 1;
+  opentelemetry.proto.common.v1.InstrumentationScope scope = 1 [json_name="i"];
 
   // A list of log records.
-  repeated LogRecord log_records = 2;
+  repeated LogRecord log_records = 2 [json_name="l"];
 
   // This schema_url applies to all logs in the "logs" field.
-  string schema_url = 3;
+  string schema_url = 3 [json_name="u"];
 }
 
 // Possible values for LogRecord.SeverityNumber.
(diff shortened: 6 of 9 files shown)


tigrannajaryan:
@open-telemetry/javascript-approvers please review.

tigrannajaryan:
@t2t2 please review.

tigrannajaryan:
> Wouldn't we be better off just recommending that receivers accept compressed payloads?

Receivers [are already required](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#protocol-details) to accept compressed payloads.

This change is a performance optimization for JS sources which don't have an an easy/cheap way to compress the payload.

alanwest:
I can't help but think it's hilarious that this is PR #413. https://www.rfc-editor.org/rfc/rfc9110.html#name-413-content-too-large

Aneurysm9:
I would second the sentiment that compression at the transport layer is the answer here.  This impairs human readability and is likely to result in duplicate effort for receivers as noted by @kentquirk.  The [`CompressionStreams` API](https://wicg.github.io/compression/) would provide a native solution for this and is [supported in Chromium-based browsers](https://developer.mozilla.org/en-US/docs/Web/API/Compression_Streams_API#browser_compatibility) already.  For browsers where that is not an option there are [pure JS implementations](https://www.npmjs.com/package/gzip-js) that can be used.  

This seems like a significant change for questionable benefit and I am not in favor.

Oberon00:
Also copying my comment from the issue here:

> We could even use the numeric tag values from the binary protocol (stringized) as JSON keys (decimal or hex or base64). Then we would also get more stable / renaming-resistant keys.

That would be a significant benefit (would further decrease readability of course, you have to guess by the values then)

kentquirk:
If I understand it, the argument is "some combinations of lightweight browsers and small applications want to avoid the performance hit from implementing compression in the browser".

But those same applications are going to be so heavily instrumented that it's worth obfuscating all the JSON for all the clients worldwide (and doing it with a breaking change)? I feel like that's a significant stretch. 

I think there might be room for a lightweight client-side spec that is intended for cheap implementation in JSON, and also intended to be sent to a specific receiver in the collector that could expand it into the standard protocol. After all, if your client is that thin, you almost certainly don't want it sending data direct to third party sites anyway.

jsuereth:
It's things like this where I wish [transit](https://github.com/cognitect/transit-format) had gained more popularity.

Specifically, in transit, they found out that encoding a symbol table of JSON keys, replacing with unique 1-liner strings and deconstructing actually led to *better* performance than raw JSON, as it's a rather efficient encoding practice and a simple idea.   It *seems* like it's the same idea you have here for JSON just the symbol-table is "offline".

From my perspective:

- We SHOULD provide the symbol-table as an option for performance sensitive users.  I think this is a good feature.
- We SHOULD NOT require this for our JSON format.  There's an aspect of human-readable JSON that's important.

Aneurysm9:
> From my perspective:
> 
>     * We SHOULD provide the symbol-table as an option for performance sensitive users.  I think this is a good feature.
>     * We SHOULD NOT require this for our JSON format.  There's an aspect of human-readable JSON that's important.

I don't see how we reconcile these two points.  If we provide the option for producers then we effectively require that consumers implement it.

dyladan:
> > From my perspective:
> > ```
> > * We SHOULD provide the symbol-table as an option for performance sensitive users.  I think this is a good feature.
> > * We SHOULD NOT require this for our JSON format.  There's an aspect of human-readable JSON that's important.
> > ```
> 
> I don't see how we reconcile these two points. If we provide the option for producers then we effectively require that consumers implement it.

I think the idea is to require it for consumers but not producers. The requirement is just written from a client/producer perspective. Similar to how gzip is implemented today.

Aneurysm9:
> I think the idea is to require it for consumers but not producers. The requirement is just written from a client/producer perspective. Similar to how gzip is implemented today.

That makes it required.  Even if producers don't need to implement it consumers will because they can't choose their partners.  For something like gzip that is effectively transparent and can be implemented via existing, off-the-shelf proxies that's a non-issue.  This is much more significant.

MSNev:
From a browser perspective, bringing in a compression library and the CPU to require GZIP or to construct this (even with short codes) is problematic.

Using the numeric value (which has been mentioned several times already) of the field rather than an additional "shortcode" would be MUCH more preferable as having the short code would require an additional lookup table. And I'm personally not interested in the human readable argument, my focus is getting the events packaged and off the box as quickly and efficiently as possible without affecting the user experience.

Using GZIP has several problems
- The package size for including the compression library
- The time required to perform the compression, as this (will likely, by default) be occurring on the main thread the UI will "pause" while the payload is being compressed. Some applications have extremely tight requirements when it comes to CPU usage and even a "pause" of 1ms is noticeable by the user. Yes there are web workers to help with this, but that has other issues in relation to final page navigation and ensuring that all "events" are sent.

A MUCH better approach would be to have the Attributes encoded WAY more efficiently from a JSON perspective as something like the following with the caveat that when using this format some of the types just are not supported (specifically the difference between int and double and bytes would always need to be encoded as a string or just not supported). As the decoder would need to make assumptions about the type.

```
"key": true / false => boolean
"key": 0, 0.1, -1 => just a numeric so use double??
"key": "xxx" => string
"key": [] => Array of something
```

```JSON
{
    "resourceSpans": [
        {
            "resource": {
                "attributes": {
                    "StartTimeUnixnano": "12345678",
                    "Pid": "1234",
                    "HostName": "fakehost",
                    "ServiceName": "generator"
                }
            },
            "scopeSpans": [
                {
                    "scope": {
                        "name": "io.opentelemetry"
                    },
                    "spans": [
                        {
                            "traceId": "AQAAAAAAAADw3rwKeFY0Eg==",
                            "spanId": "AQAAAAAAAAA=",
                            "name": "load-generator-span",
                            "kind": "SPAN_KIND_CLIENT",
                            "startTimeUnixNano": "1572516672000000013",
                            "endTimeUnixNano": "1572516672000000013",
                            "attributes": {
                                "db.mongodb.collection": "!##$"
                            },
                            "events": [
                                {
                                    "timeUnixNano": "1572516672000000013",
                                    "attributes": {
                                        "te": "1"
                                    }
                                }
                            ]
                        }
                    ]
                }
            ]
        }
    ]
}
```

yurishkuro:
Just throwing this out there, but why can't we just implement a proto encoding in Javascript? I remember, back in the days, Zipkin Java SDK, that is used on Android and is also subject to size limitations, simply ditched the dependency on Thrift library and implemented serialization manually, which was like a 100 lines of code.

Aneurysm9:
> A MUCH better approach would be to have the Attributes encoded WAY more efficiently from a JSON perspective

This would be a different protocol.  If we want to do that and treat it separately with its own content type, decoupling it from OTLP and removing this issue as a blocker for OTLP 1.0, I'd have no problems with moving in that direction.  I think ultimately it might be a bad idea as it then means we need to maintain two protocols and many consumers would still have to implement two protocols, but my primary concern right now is stabilizing OTLP so that it can be taken as a dependency by systems that have stability requirements.

Honestly, I think we're making a mountain out of a molehill and I'm strongly dubious of claims that a 1ms latency is user perceptible.  Even at 60FPS each frame is 16ms and detecting single frame drops is not something I experience.  Package size for including a pure-JS gzip implementation is a concern, but given the standardization effort for including compressed streams in browsers with native implementations is already well underway this is a temporary concern at best.  We're designing a protocol that will be used for years, do we really want to upend it at the last minute for an issue that may disappear well before the end of the protocol's useful life?

ladd:
Specifically, it looks like `CompressionStream` already has fairly good uptake: https://caniuse.com/?search=compressionstream

Aneurysm9:
> Just throwing this out there, but why can't we just implement a proto encoding in Javascript?

I'm not even sure we'd need to implement it ourselves.  https://github.com/protobufjs/protobuf.js/ exists and appears to bu under active development, with a `v7.0.0` release just ten days ago.  It claims [excellent performance](https://github.com/protobufjs/protobuf.js/#performance) compared to JSON serialization.

yurishkuro:
@Aneurysm9 nice! `"protobufjs/minimal"` is 6.5Kb.

tigrannajaryan:
Let me try to summarize the important findings:
- There is an option to do gzip compression using [CompressionStreams API](https://wicg.github.io/compression/) available in Chromium only.
- There is an option to use binary Protobuf, and https://github.com/protobufjs/protobuf.js/ is 6.5KB and claims performance faster than native JSON. Protobuf is going to be more compact than JSON with short keys (about 1.6-2.2 in my benchmarks).

We also discussed this in the Spec meeting yesterday and no-one in the call believed that we need the JSON with short keys given that other options are available. JSON with short keys loses its readability, while not gaining as much in the performance as the other options we have.

Given the above, I am closing this PR and will marked the linked issue as "won't do". If there are new arguments the issue can be re-opened (but it must happen before we declare JSON format Stable).

Issue #404, opened 2022-06-21
OTLP metrics export partial success - structure, semantic conventions and beyond


joaopgrassi:
In #390 we introduced the concept of partial success in OTLP export responses. The initial version only contained fields to indicate the # of accepted datapoints + a string error message. During the PR review it was brought up that only these fields might not be enough and that we may need more structure to be able to achieve more things with it. 

This issue is to continue the discussion, focusing on **what we want to achieve** and agree on **how we want to do it**.

A couple of things that came up was:

### Derive telemetry for consumed/dropped metrics for OTLP exporters

Exporters would record metrics using the `partial_success` response. Things like `otlp.metrics.consumed`, `otlp.metrics.dropped` can be derived from the # of accepted datapoints. With the current structure we defined in #390 this is already possible, but we need to define semantic conventions for these metrics so OTLP exporters are consistent in how they report it (if desired).

**Action items:** 

- Define use-cases and which metrics we want/need
- Define semantic conventions for such metrics
- Define how to turn this behavior off on exporters?

### Possibility to retry

With only the # of accepted datapoints + a string error message we currently have, it's not possible to achieve any level of retry. It might be possible to perform a retry in some cases but to do that, we need more structure on the `partial_success` result.

From the PR, an example of such response would be:

```
{"accepted_data_points": 5, "error": [
  {"index": 0, "status_code": 400, "message": "the timestamp xyz is invalid"},
  {"index": 3, "status_code": 502, "message": "the tenant does not exist"}
]}
```
Another important topic was that any extra details **should be optional** to not overload receivers/senders that are under pressure.

**Action items:** 

- Discuss if retries really make sense for a partial success response.
- If we want _some_ level of retry, identify which and come up with a structure to represent it
- Make sure such extra fields are **optional**

PR #414, opened 2022-07-22
Make it possible to indicate partial success in an OTLP export response [2]

joaopgrassi:
This PR is a second attempt in solving/addressing https://github.com/open-telemetry/opentelemetry-specification/issues/2454. This uses the `rejected` semantics instead of `accepted` as discussed in the initial PR https://github.com/open-telemetry/opentelemetry-proto/pull/390. 

Related spec PR: https://github.com/open-telemetry/opentelemetry-specification/pull/2696

The need for this new PR was that with `accepted` we were running into problems with the `0` value being the same as the default/not set, plus general intention to avoid having `optional` on the fields.

I did some tests with sample apps to confirm the cases for backwards compatibility. Those can be checked in: [joaopgrassi/otlp-partial-success-experimental](https://github.com/joaopgrassi/otlp-partial-success-experimental). **The TL;DR:** Since it's a new field no breaking change is introduced. New sender will always have to check if the `partial_success` is set because they can be working with servers using the old protos.

Some context on why this feature is desired:

This feature **IS** intended to help in the following scenarios:

- Enable servers to provide information to users about a partial success without having to return a `400 - BadRequest` (status quo today)
- Enable senders to generate metrics about themselves (`otlp.exporter.rejected.data_points=x`)
- Enable senders to log messages when a partial success is encountered (very useful to troubleshoot things quickly, without having to reach/search for your vendor's logs for ex)
- Capture situations that are not necessarily client or server errors. 
  - Change the type of metric instrument after an initial submission - Gauge > Counter
  - Customers reaching their quotas in the middle of an export request
  - Data not conforming with vendor's constraints - metric keys with not accepted characters coming from a library outside the application's owner control (with the errors they can rename them with a view for ex)
  - OTLP data arriving with timestamp in the past - no explicit contract for this in OTel. Might be tricky for users to debug what's going on

  
This feature **IS NOT** intended to help in the following scenarios:

- Enable clients to retry a partial success request
   - Complicated to handle. Needs more structure, semantic convention on typical errors. Was discussed [in the initial proto PR](https://github.com/open-telemetry/opentelemetry-proto/pull/390#discussion_r871989856)
   - This can always be added later on in a non-breaking manner. I started a separated discussion on this topic for metrics [here](https://github.com/open-telemetry/opentelemetry-proto/issues/404)
- Identify which exact log line/data point/span was rejected. Similar to the point above, it needs more discussion on how this can be done and which structure we should have. Always can be done/improved later

cc @tigrannajaryan @jmacd @jsuereth @jack-berg @reyang

### DIFF
--- a/opentelemetry/proto/collector/logs/v1/logs_service.proto
+++ b/opentelemetry/proto/collector/logs/v1/logs_service.proto
@@ -57,13 +57,10 @@ message ExportLogsServiceResponse {
 
 message ExportLogsPartialSuccess {
   // The number of rejected log records.
-  //
-  // If the request is only partially accepted, the server MUST set
-  // the `rejected_log_records` field.
   // 
-  // Given the server should only initialize the `partial_success` field
+  // Given the server MUST only initialize the `partial_success` field
   // when it rejects parts of the data, a `rejected_log_records` with
-  // value = '0' is considered an invalid result. Senders should
+  // value = '0' is considered an invalid result. Senders MUST
   // interpret such responses from a server as if all log records were accepted.
   int64 rejected_log_records = 1;
 
--- a/opentelemetry/proto/collector/logs/v1/logs_service.proto
+++ b/opentelemetry/proto/collector/logs/v1/logs_service.proto
@@ -48,25 +48,26 @@ message ExportLogsServiceResponse {
   // If the request is only partially accepted
   // (i.e. when the server accepts only parts of the data and rejects the rest)
   // the server MUST initialize the `partial_success` field and MUST
-  // set the `rejected_log_records` with the number of log records it rejected.
+  // set the `rejected_<signal>` with the number of items it rejected.
   //
-  // The server MUST NOT set the `partial_success` field when the
-  // request was accepted in its entirety.
+  // Servers MAY also make use of the `partial_success` field to convey
+  // warnings/suggestions to senders when the request was fully accepted.
+  // In such cases, the `rejected_<signal>` MUST have a value of `0` and
+  // the `error_message` MUST be non-empty.
   ExportLogsPartialSuccess partial_success = 1;
 }
 
 message ExportLogsPartialSuccess {
   // The number of rejected log records.
   // 
-  // Given the server MUST only initialize the `partial_success` field
-  // when it rejects parts of the data, a `rejected_log_records` with
-  // value = '0' is considered an invalid result. Senders MUST
-  // interpret such responses from a server as if all log records were accepted.
+  // A `rejected_<signal>` field holding a `0` value indicates that the
+  // request was fully accepted.
   int64 rejected_log_records = 1;
 
-  // A developer-facing human-readable error message in English. It should
-  // explain why the server rejected parts of the data,
-  // and might offer guidance on how users can address the issues.
+  // A developer-facing human-readable message in English. It should be used
+  // either to explain why the server rejected parts of the data during a partial
+  // success or to convey warnings/suggestions during a full success. The message
+  // should offer guidance on how users can address such issues.
   //
   // error_message is an optional field. An error_message with an empty value
   // is equivalent of it not being set.
--- a/opentelemetry/proto/collector/logs/v1/logs_service.proto
+++ b/opentelemetry/proto/collector/logs/v1/logs_service.proto
@@ -43,4 +43,35 @@ message ExportLogsServiceRequest {
 }
 
 message ExportLogsServiceResponse {
+  // The details of a partially successful export request.
+  //
+  // If the request is only partially accepted
+  // (i.e. when the server accepts only parts of the data and rejects the rest)
+  // the server MUST initialize the `partial_success` field and MUST
+  // set the `rejected_log_records` with the number of log records it rejected.
+  //
+  // The server MUST NOT set the `partial_success` field when the
+  // request was accepted in its entirety.
+  ExportLogsPartialSuccess partial_success = 1;
+}
+
+message ExportLogsPartialSuccess {
+  // The number of rejected log records.
+  //
+  // If the request is only partially accepted, the server MUST set
+  // the `rejected_log_records` field.
+  // 
+  // Given the server should only initialize the `partial_success` field
+  // when it rejects parts of the data, a `rejected_log_records` with
+  // value = '0' is considered an invalid result. Senders should
+  // interpret such responses from a server as if all log records were accepted.
+  int64 rejected_log_records = 1;
+
+  // A developer-facing human-readable error message in English. It should
+  // explain why the server rejected parts of the data,
+  // and might offer guidance on how users can address the issues.
+  //
+  // error_message is an optional field. An error_message with an empty value
+  // is equivalent of it not being set.
+  string error_message = 2;
 }
--- a/opentelemetry/proto/collector/logs/v1/logs_service.proto
+++ b/opentelemetry/proto/collector/logs/v1/logs_service.proto
@@ -51,7 +51,7 @@ message ExportLogsServiceResponse {
   // set the `rejected_<signal>` with the number of items it rejected.
   //
   // Servers MAY also make use of the `partial_success` field to convey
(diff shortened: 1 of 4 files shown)


opentelemetry/proto/collector/logs/v1/logs_service.proto
```diff
@@ -43,4 +43,35 @@ message ExportLogsServiceRequest {
...
+  ExportLogsPartialSuccess partial_success = 1;
+}
+
+message ExportLogsPartialSuccess {
+  // The number of rejected log records.
+  //
+  // If the request is only partially accepted, the server MUST set
+  // the `rejected_log_records` field.
+  // 
+  // Given the server should only initialize the `partial_success` field
```
arminru, line 64:
```suggestion
  // Given the server MUST only initialize the `partial_success` field
```
I think you can just use the same level as in line 53 above.

jmacd: :+1: and the same change in the two appearances below.

jmacd (approved):
Great, thank you!

opentelemetry/proto/collector/logs/v1/logs_service.proto
```diff
@@ -43,4 +43,35 @@ message ExportLogsServiceRequest {
...
+  //
+  // The server MUST NOT set the `partial_success` field when the
+  // request was accepted in its entirety.
+  ExportLogsPartialSuccess partial_success = 1;
+}
+
+message ExportLogsPartialSuccess {
+  // The number of rejected log records.
+  //
+  // If the request is only partially accepted, the server MUST set
```
tigrannajaryan, line 61: >If the request is only partially accepted

Isn't this part superfluous? If the request is fully accepted the entire ExportLogsPartialSuccess should not exist. This is what the line 48 says.

joaopgrassi: I had the same thought but ended up leaving just to be "extra clear" on the expectations. But I can remove, I agree it's duplicated.

opentelemetry/proto/collector/logs/v1/logs_service.proto
joaopgrassi: @jmacd  you mean the next "should" as in here change to MUST?

```
// A developer-facing human-readable error message in English. It should..
// explain why the server rejected parts of the data,
```

I ask because it's different from what I have currently in the related specification PR. Do we want to enforce servers to _always_ populate the message?

opentelemetry/proto/collector/logs/v1/logs_service.proto
joaopgrassi: Removed in [32ab10e](https://github.com/open-telemetry/opentelemetry-proto/pull/414/commits/32ab10ec580906780e6cedb47f362c8d35be83f1)

opentelemetry/proto/collector/logs/v1/logs_service.proto
joaopgrassi: Changed parts of it in [32ab10e](https://github.com/open-telemetry/opentelemetry-proto/pull/414/commits/32ab10ec580906780e6cedb47f362c8d35be83f1).

tigrannajaryan (approved):

opentelemetry/proto/collector/logs/v1/logs_service.proto
jmacd: Sorry for being unclear. Your changes are good, I just meant to apply the same suggestion for traces and metrics sections.

opentelemetry/proto/collector/logs/v1/logs_service.proto
```diff
@@ -43,4 +43,32 @@ message ExportLogsServiceRequest {
...
+  ExportLogsPartialSuccess partial_success = 1;
+}
+
+message ExportLogsPartialSuccess {
+  // The number of rejected log records.
+  // 
+  // Given the server MUST only initialize the `partial_success` field
+  // when it rejects parts of the data, a `rejected_log_records` with
+  // value = '0' is considered an invalid result. Senders MUST
+  // interpret such responses from a server as if all log records were accepted.
```
jmacd, line 64: I would prefer to discuss this detail in https://github.com/open-telemetry/opentelemetry-specification/pull/2696#discussion_r930249217

joaopgrassi: As discussed in the thread linked by @jmacd , we will allow `rejected_<signal>` = `0` so the server can use a `partial_success` to convey warnings/suggestions to clients via the `error_message` field. I adapted the text around it to match the spec change in [3461d1f](https://github.com/open-telemetry/opentelemetry-proto/pull/414/commits/3461d1f15a4a12b0bab5c567f22b0030b72e070e)

arminru (approved):

opentelemetry/proto/collector/logs/v1/logs_service.proto
```diff
@@ -43,4 +43,33 @@ message ExportLogsServiceRequest {
...
+  // If the request is only partially accepted
+  // (i.e. when the server accepts only parts of the data and rejects the rest)
+  // the server MUST initialize the `partial_success` field and MUST
+  // set the `rejected_<signal>` with the number of items it rejected.
+  //
+  // Servers MAY also make use of the `partial_success` field to convey
+  // warnings/suggestions to senders even when the request was fully accepted.
+  // In such cases, the `rejected_<signal>` MUST have a value of `0` and
+  // the `error_message` MUST be non-empty.
+  ExportLogsPartialSuccess partial_success = 1;
```
bogdandrutu, line 57: Please explain if there is a difference between `ExportLogsPartialSuccess` not present (proto supports not present for messages) and an empty `ExportLogsPartialSuccess` (where rejected is 0 and error_message is empty).

jmacd: We debated whether to make this field optional (at length). The protocol compiler defines this equivalence, between non-optional-field-not-present and field-present-with-all-zeros. I feel like we answered your question by deciding to make this message not optional, which leaves me unsure what else needs to be written here. Can you suggest an edit?

joaopgrassi: Done in [49bcec0](https://github.com/open-telemetry/opentelemetry-proto/pull/414/commits/49bcec06f20a67f28b77894ebd23591b2432360f). Let me know if this clarify things!

jmacd: @joaopgrassi added an explicit statement (in three places). @bogdandrutu does this look good?

bogdandrutu: This looks good.

@jmacd "optional" has no effect on "messages" which by default are "optional" and cannot be changed (unless you use a different version of the compiler like the gogo proto). Optional keyword is important only for primitives.

jmacd: @bogdandrutu Thanks. I had misunderstood the rule for optional with message fields.  In fact, I stated the opposite during a SIG call in favor of counting `rejected_<thing>` instead of `accepted_<thing>`, because it lets the zero-state match legacy behavior. It also means we can ignore field presence (and/or means we can continue to use Gogo).

Given this information, I still prefer to count rejected things. @bogdandrutu @joaopgrassi does this change your opinion of this PR and https://github.com/open-telemetry/opentelemetry-specification/pull/2696?

arminru (approved):

opentelemetry/proto/collector/logs/v1/logs_service.proto
joaopgrassi: @jmacd in my view we are good, I also like the rejected approach. 

I think with the instructions we have in the spec + the extra text I added as requested by @bogdandrutu make things clear.

Pointing out that the "empty" value of `partial_success` is equivalent to it not being set and that is also equivalent of a "total success" makes sense to me.

PR #415, opened 2022-07-28
Use buf to detect for wire breaking changes

aabmass:
Use [Buf breaking](https://docs.buf.build/breaking/overview) to check PRs for wire and json breaking changes. [Here is an example PR](https://github.com/aabmass/opentelemetry-proto/pull/5/files#diff-62ba46dcb97aa81748b276b6eb42ae8e6700cd4b467515542539435eda06d01f) showing this in action. There was some interest for this in https://github.com/open-telemetry/opentelemetry-proto/issues/400#issuecomment-1156617896.

The config can be tweaked to add/remove specific rules if this is too noisy, see https://docs.buf.build/breaking/configuration.


buf.yaml
```diff
@@ -0,0 +1,7 @@
+version: v1
+
+# See https://docs.buf.build/breaking/configuration
+breaking:
+  use:
+    - WIRE
+    - WIRE_JSON
```
tigrannajaryan, line 7: I think we don't need WIRE_JSON for now. JSON is not considered stable yet.

tigrannajaryan (commented):
Thanks for the PR, this is very useful.

.github/workflows/build-check.yaml
```diff
@@ -82,3 +82,22 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    # This is needed because checkout action does a shallow clone
+    - name: Fetch main branch for comparison
+      run: |
+        git remote set-branches origin '*'
+        git fetch --depth=1 origin main
+        git branch main origin/main
+    - name: Run make breaking-change with json output to annotate PR
```
arminru, line 97: I agree that this is very helpful, thank you!

It would be nice if you could add a breaking change in a new (dummy) PR based off this branch to illustrate how this would work.

Makefile
```diff
@@ -18,13 +18,16 @@ endef
...
+
 PROTOC := docker run --rm -u ${shell id -u} -v${PWD}:${PWD} -w${PWD} ${OTEL_DOCKER_PROTOBUF} --proto_path=${PWD}
+BUF := docker run --rm -v "${PWD}:/workspace" -w /workspace ${BUF_DOCKER}
+BUF_AGAINST ?= ".git\#branch=main"
 
 PROTO_GEN_CPP_DIR ?= $(GENDIR)/cpp
 PROTO_GEN_CSHARP_DIR ?= $(GENDIR)/csharp
 PROTO_GEN_GO_DIR ?= $(GENDIR)/go
 PROTO_GEN_JAVA_DIR ?= $(GENDIR)/java
-PROTO_GEN_JS_DIR ?= $(GENDIR)/js
```
arminru, line 27: Removed accidentally?

arminru (approved):

Makefile
```diff
@@ -18,7 +18,11 @@ endef
 gen-all: gen-cpp gen-csharp gen-go gen-java gen-kotlin gen-objc gen-openapi gen-php gen-python gen-ruby
 
 OTEL_DOCKER_PROTOBUF ?= otel/build-protobuf:0.9.0
+BUF_DOCKER ?= bufbuild/buf:1.7.0
+
 PROTOC := docker run --rm -u ${shell id -u} -v${PWD}:${PWD} -w${PWD} ${OTEL_DOCKER_PROTOBUF} --proto_path=${PWD}
+BUF := docker run --rm -v "${PWD}:/workspace" -w /workspace ${BUF_DOCKER}
+BUF_AGAINST ?= ".git\#branch=main"
```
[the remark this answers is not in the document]
tigrannajaryan: Can you add some comments here to explain what is going on and how the comparison works? "buf" is relatively unknown tool and we don't want every reader of the makefile to read buf docs to understand how it works here.

.github/workflows/build-check.yaml
aabmass: https://github.com/aabmass/opentelemetry-proto/pull/5/files#diff-62ba46dcb97aa81748b276b6eb42ae8e6700cd4b467515542539435eda06d01f

Makefile
aabmass: Oops, yes

Makefile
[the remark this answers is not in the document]
aabmass: @tigrannajaryan I simplified things a bit by putting the upstream repo's URL here instead which I think is self explanatory. Added a comment anyway tho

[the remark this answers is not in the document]
tigrannajaryan: Yes, I think it is clearer now.

tigrannajaryan (approved):

.github/workflows/build-check.yaml
arminru: Looks good, thank you!

arminru (approved):

PR #423, opened 2022-08-17
Remove helper masks, not part of the protocol

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-proto/issues/363

The reason is to keep the proto interface as minimal as possible, and the helper masks are just helpers, and not part of the protocol. They can be defined by the language SIGs in their usages, but don't need to be part of the proto stability.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -311,24 +311,6 @@ enum AggregationTemporality {
   AGGREGATION_TEMPORALITY_CUMULATIVE = 2;
 }
 
-// DataPointFlags is defined as a protobuf 'uint32' type and is to be used as a
-// bit-field representing 32 distinct boolean flags.  Each flag defined in this
-// enum is a bit-mask.  To test the presence of a single flag in the flags of
-// a data point, for example, use an expression like:
-//
-//   (point.flags & FLAG_NO_RECORDED_VALUE) == FLAG_NO_RECORDED_VALUE
-//
-enum DataPointFlags {
-  FLAG_NONE = 0;
-
-  // This DataPoint is valid but has no recorded value.  This value
-  // SHOULD be used to reflect explicitly missing data in a series, as
-  // for an equivalent to the Prometheus "staleness marker".
-  FLAG_NO_RECORDED_VALUE = 1;
-
-  // Bits 2-31 are reserved for future use.
-}
-
 // NumberDataPoint is a single data point in a timeseries that describes the
 // time-varying scalar value of a metric.
 message NumberDataPoint {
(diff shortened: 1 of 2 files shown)


tigrannajaryan (commented):
The definitions of the bits in the `flags` field is an essential part of OTLP protocol. In my opinion, having these definitions in the form helper enums is valuable and helps prevent implementation mistakes. Because of this I would prefer to keep the helper enums.

MikeGoldsmith:
Yeah, I agree - these helper numbers help ensure there isn't language implementation drift. I would vote to keep them.

bogdandrutu:
May help, but not part of the protocol and guarantees. I am definitely not going to accept this as part of our guarantees of stability.

jmacd:
How can `FLAG_NO_RECORDED_VALUE` not be considered part of a stability guarantee? Are you saying that you'd like to reserve the right to change the name of the symbol so that NO_RECORDED_VALUE could be renamed? The _meaning_ of this field has to be considered stable IMO.

bogdandrutu:
>The meaning of this field has to be considered stable IMO. 

100% agree

> Are you saying that you'd like to reserve the right to change the name of the symbol so that NO_RECORDED_VALUE could be renamed?

YES

tigrannajaryan:
I wonder what other @open-telemetry/specs-approvers think about usefulness of these enums.

My preference would be to keep them, but it is not a strong opinion.

Also, we should get rid of the zero values: LOG_RECORD_FLAG_UNSPECIFIED and FLAG_NONE since they encourage poor practice. The correct usage is to always do bitwise AND with the right bit, never compare to the 0 value.

jsuereth:
I'm w/ @tigrannajaryan in that I think having them here is generally helpful and helps us avoid (obvious) breaking changes in their definition as well as gives folks "one pane of glass" to look at.

bogdandrutu:
> I'm w/ @tigrannajaryan in that I think having them here is generally helpful and helps us avoid (obvious) breaking changes in their definition as well as gives folks "one pane of glass" to look at.

This is a protocol not a "helper" library. I don't think these constants belong to the protocol, similar to a standard like w3c trace context, does not require/define/encourage a constant for `0x1` for the sampling bit, it is who is generating the code responsibility if they want to provide helpers.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -311,24 +311,6 @@ enum AggregationTemporality {
...
-//
-//   (point.flags & FLAG_NO_RECORDED_VALUE) == FLAG_NO_RECORDED_VALUE
-//
-enum DataPointFlags {
-  FLAG_NONE = 0;
-
-  // This DataPoint is valid but has no recorded value.  This value
-  // SHOULD be used to reflect explicitly missing data in a series, as
-  // for an equivalent to the Prometheus "staleness marker".
-  FLAG_NO_RECORDED_VALUE = 1;
```
jmacd, line 327: Where did this flag go?

jmacd:
I disagree with the effort to protect users from being foolish. I support the idea of removing zero-valued enums, but not removing non-zero values. The FLAG_NO_RECORDED_VALUE has disappeared, which I do not support. Users who use bitmasks should know what they are doing and/or read documents.

opentelemetry/proto/metrics/v1/metrics.proto
Aneurysm9: I believe the idea is to remove it in favor of comments describing values that can be used in data point flags.  I believe that is a mistake.  Having an enumeration allows for clearly named consistent values and separate documentation of each potential flag.

tigrannajaryan:
Discussed in TC meeting and decided that https://github.com/open-telemetry/opentelemetry-proto/pull/461 should be sufficient and we can close this issue.

PR #432, opened 2022-10-18
[DON'T MERGE] Define OTLP 1.0 Stability guarantees

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-proto/issues/400

This is a draft that is not supposed to be merged yet. Let's agree on what the guarantees for OTLP 1.0 will be once we decide that it is time.


tigrannajaryan:
@bogdandrutu @Aneurysm9 @dyladan @MrAlias @carlosalberto @ocelotl @aabmass @jsuereth please review.

README.md
```diff
@@ -43,32 +43,34 @@ for definition of maturity levels).
 
 Components marked `Stable` provide the following guarantees:
 
-- Field types will not change.
-- Field numbers will not change.
-- Numbers assigned to enum choices will not change.
-- Service names and service package names will not change.
-- Service operation names, parameter and return types will not change.
+- Field types, numbers and names will not change.
```
bogdandrutu, line 46: Why not allowing to remove a field? Since we allow sending "unknown" fields, removing a field is not a wire breaking change.

Not that I want to do this, but I want to clarify the "boundaries" and this is only related to generated code.

README.md
```diff
@@ -43,32 +43,34 @@ for definition of maturity levels).
...
+- Names of messages and enums will not change.
```
bogdandrutu, line 47: Why? Not important at all for wire compatibility? Only generated code. I think we should say that "SIGs" must not expose this externally.

Not that I want to do this, but I want to clarify the "boundaries" and this is only related to generated code.

README.md
```diff
@@ -43,32 +43,34 @@ for definition of maturity levels).
...
+- Names of enum choices and numbers assigned to enum choices will not change.
```
bogdandrutu, line 48: Same questions for enum names.

bogdandrutu (commented):
You seem to try to offer "generated" code compatibility, which may be fine, but I think we should agree on that if this is what we want.

README.md
```diff
@@ -43,32 +43,34 @@ for definition of maturity levels).
...
+- The location of messages and enums, i.e. whether they are declared at the top lexical
+  scope or nested inside another message will not change.
+- Package names and directory structure will not change.
```
bogdandrutu, line 51: Definitely package names for the gRPC services MUST not change, for the other packages is just relevant for the generated code.

README.md
```diff
@@ -43,32 +43,34 @@ for definition of maturity levels).
...
 
-- Field types will not change.
-- Field numbers will not change.
-- Numbers assigned to enum choices will not change.
-- Service names and service package names will not change.
-- Service operation names, parameter and return types will not change.
+- Field types, numbers and names will not change.
+- Names of messages and enums will not change.
+- Names of enum choices and numbers assigned to enum choices will not change.
+- The location of messages and enums, i.e. whether they are declared at the top lexical
```
bogdandrutu, line 49: Relevant only for the generated code.

README.md
```diff
@@ -43,32 +43,34 @@ for definition of maturity levels).
...
+  scope or nested inside another message will not change.
+- Package names and directory structure will not change.
+- gRPC service names, service method names, parameters and return types will not change.
+- Service method parameter names will not change.
```
bogdandrutu, line 53: Relevant only for the generated code.

README.md
```diff
@@ -43,32 +43,34 @@ for definition of maturity levels).
...
+- Service method kind (unary vs streaming) will not change.
+- No existing symbol will be deleted.
```
bogdandrutu, line 55: Why?

README.md
```diff
@@ -43,32 +43,34 @@ for definition of maturity levels).
...
+- `optional` and `repeated` declarators of existing fields will not change.
```
bogdandrutu, line 56: Why for `optional`? Changing from non-optional to optional should be fine, even the other way if the default value is ok.

README.md
tigrannajaryan: From the wire perspective you are right, in theory we could do this and have an explanation that for receivers it is equivalent to the previous state when sender didn't include the field. In practice it would be very difficult to do without breaking some functionality that depends on the field. We don't really have unnecessary fields :-)
  
However because we also aim to maintain symbolic compatibility remove a field would breaking any existing code that uses it. So from this perspective it definitely should be prohibited.

README.md
tigrannajaryan: Yes. Not important from wire perspective.

However, in this PR I am going for full symbolical compatibility of the generate code (except stability of the generator itself, which we don't guarantee of course).

README.md
tigrannajaryan: Changing from optional to non-optional can break the generated code since it affects the presence check functions.

tigrannajaryan:
> You seem to try to offer "generated" code compatibility, which may be fine, but I think we should agree on that if this is what we want.

Yes. I updated the PR description to make the goal clear.

README.md
dyladan: Removing a field without reserving the ID brings the possibility the ID would be used later with a non-compatible data. If we reserve the field, there is no real difference between keeping it (marked deprecated) and removing it.

tigrannajaryan: >If we reserve the field, there is no real difference between keeping it (marked deprecated) and removing it.

There is a difference for the generated code. It is a breaking change to remove a field.

Aneurysm9:
I support this definition.  I believe it reflects the consensus resulting from discussion on #400 and also what the user community expects from OTLP.

tigrannajaryan:
Here is the PR that only adds guarantees necessary for OTLP/JSON: https://github.com/open-telemetry/opentelemetry-proto/pull/435

Aneurysm9 (approved):

mx-psi (approved):

jack-berg (approved):

README.md
```diff
@@ -48,23 +48,18 @@ for definition of maturity levels).
...
-  the wire.
-- The location of messages and enums, i.e. whether they are declared at the top
-  lexical scope or nested inside another message may change.
-- Package names may change.
-- Directory structure, location and the name of the files may change.
-
-Note that none of the above allowed changes affects the binary wire representation or the
-JSON wire representation.
+- Service method names will not change.
+- Service method parameter names will not change.
```
jack-berg, line 53: Is generated code susceptible to breaking because of this?

jmacd: I think it would be a service request that failed in grpc due to a breaking naming change.

jmacd (approved):

README.md
tigrannajaryan: Service names and method names are on the wire, so absolutely necessary for stability: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

I am not sure about parameter names. I will check.

tigrannajaryan: So, looks like parameter names are nowhere on the wire. However, I am not sure how to confirm they are not part of the generated API (and thus may be subject to symbolic stability). I could try going through all supported languages and see what the generated code looks like and whether parameter names matter, but that is going to be very difficult (and hard to confirm especially for dynamic languages where you can access parameter names at runime).

I suggest to play safe and just keep it here, unless someone has a better idea on how to confirm that we don't need this stability.

carlosalberto (approved):

tedsuo (approved):

arminru (approved):

dyladan (approved):

tigrannajaryan:
A follow-up PR to make it clear that symbolic guarantees go into effect only when 1.0 is released: https://github.com/open-telemetry/opentelemetry-proto/pull/467

PR #442, opened 2023-01-03
Clarify behavior for empty/not present/invalid trace_id and span_id fields

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/3040

This is not a breaking change:

- For Span it now defines more precisely the receiver behavior that was previously defined vaguely (e.g. it was unclear what "empty" means for bytes field).

- For LogRecord it now defines the receiver behavior that was previously unspecified. This ensures that the wording is consistent with what we have for the Span.

### DIFF
--- a/opentelemetry/proto/logs/v1/logs.proto
+++ b/opentelemetry/proto/logs/v1/logs.proto
@@ -164,14 +164,29 @@ message LogRecord {
   fixed32 flags = 8;
 
   // A unique identifier for a trace. All logs from the same trace share
-  // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes
-  // is considered invalid. Can be set for logs that are part of request processing
-  // and have an assigned trace id. [Optional].
+  // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes OR
+  // of length other than 16 bytes is considered invalid (empty string in OTLP/JSON
+  // is zero-length and thus is also invalid).
+  //
+  // This field is optional.
+  //
+  // The receivers SHOULD assume that the log record is not associated with a
+  // trace if any of the following is true:
+  //   - the field is not present,
+  //   - the field contains an invalid value.
   bytes trace_id = 9;
 
   // A unique identifier for a span within a trace, assigned when the span
-  // is created. The ID is an 8-byte array. An ID with all zeroes is considered
-  // invalid. Can be set for logs that are part of a particular processing span.
-  // If span_id is present trace_id SHOULD be also present. [Optional].
+  // is created. The ID is an 8-byte array. An ID with all zeroes OR of length
+  // other than 8 bytes is considered invalid (empty string in OTLP/JSON
+  // is zero-length and thus is also invalid).
+  //
+  // This field is optional. If the sender specifies a valid span_id then it SHOULD also
+  // specify a valid trace_id.
+  //
+  // The receivers SHOULD assume that the log record is not associated with a
+  // span if any of the following is true:
+  //   - the field is not present,
+  //   - the field contains an invalid value.
   bytes span_id = 10;
 }
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -79,15 +79,17 @@ message ScopeSpans {
 // The next available field id is 17.
 message Span {
   // A unique identifier for a trace. All spans from the same trace share
-  // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes
-  // is considered invalid.
+  // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes OR
+  // of length other than 16 bytes is considered invalid (empty string in OTLP/JSON
+  // is zero-length and thus is also invalid).
   //
   // This field is required.
   bytes trace_id = 1;
 
   // A unique identifier for a span within a trace, assigned when the span
-  // is created. The ID is an 8-byte array. An ID with all zeroes is considered
-  // invalid.
+  // is created. The ID is an 8-byte array. An ID with all zeroes OR of length
+  // other than 8 bytes is considered invalid (empty string in OTLP/JSON
+  // is zero-length and thus is also invalid).
   //
   // This field is required.
   bytes span_id = 2;


djaglowski (approved):

Aneurysm9 (approved):

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -82,20 +82,24 @@ message Span {
   // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes
   // is considered invalid.
   //
-  // This field is semantically required. Receiver should generate new
-  // random trace_id if empty or invalid trace_id was received.
-  //
   // This field is required.
+  //
+  // The receivers SHOULD generate a new random trace_id if any of the following is true:
```
jmacd, line 87: Why use a random ID? Why not reject the data as invalid?

tigrannajaryan: That's the way it is currently. I can't change the current behavior, it would be a breaking change.
Why it was specified that way originally, I don't know.

jmacd: Where is this specified? It's specified for OTLP transport as well as OTLP/JSON transport? Thanks.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -82,20 +82,24 @@ message Span {
   // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes
   // is considered invalid.
   //
-  // This field is semantically required. Receiver should generate new
-  // random trace_id if empty or invalid trace_id was received.
-  //
```
jmacd, line 87: Sorry, this is not a specification document. I would discard this language and treat spans w/o valid trace IDs and span IDs as invalid spans. Why change this now?

opentelemetry/proto/trace/v1/trace.proto
jmacd: > current behavior, it would be a breaking

Current behavior is broken, which I think allows us to change this behavior. How could anyone know the difference between a randomly-generated ID and an invalid data?

opentelemetry/proto/trace/v1/trace.proto
Aneurysm9: Agreed.  By the time span data are represented as OTLP we're out of the SDK and the data are either valid or not.

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -165,13 +165,27 @@ message LogRecord {
...
-  // and have an assigned trace id. [Optional].
+  // is considered invalid.
+  //
+  // This field is optional.
+  //
+  // The receivers SHOULD assume that the log record is not associated with a
+  // trace if any of the following is true:
+  //   - the field is not present,
+  //   - the field is equal to zero-length bytes sequence (or to empty string in OTLP/JSON),
+  //   - the field contains an invalid value.
```
yurishkuro, line 176: L175 seems to be subsumed by L176, since zero-length is also invalid. 

Also, rather than explaining invalid here I would pull it into L168: "An ID with all zeroes OR of length other than 16 bytes is considered invalid."

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -165,13 +165,27 @@ message LogRecord {
...
   bytes trace_id = 9;
 
   // A unique identifier for a span within a trace, assigned when the span
   // is created. The ID is an 8-byte array. An ID with all zeroes is considered
-  // invalid. Can be set for logs that are part of a particular processing span.
-  // If span_id is present trace_id SHOULD be also present. [Optional].
+  // invalid.
+  //
+  // This field is optional. If valid span_id is present then trace_id SHOULD be also present.
```
yurishkuro, line 183:
```suggestion
  // This field is optional. If a producer supplies a valid span_id then trace_id SHOULD be also present.
```
(because as written the phrase is meaningless to a consumer)

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: This is currently specified in this repo: https://github.com/open-telemetry/opentelemetry-proto/blob/d8729d40f629dba12694b44c4c32c1eab109b00a/opentelemetry/proto/trace/v1/trace.proto#L85

Quote:
> Receiver should generate new random trace_id if empty or invalid trace_id was received.

Again, the aim of this PR is not to change any existing behavior, merely to clarify it.

If we believe what is currently specified is a bug that needs to be fixed then that's a different story. That was not the goal of my PR.

I agree with you that generating a random id when an invalid one is received doesn't sounds right. However, I don't want to change this without understanding why it was done so in the first place (and also because I think it would be a breaking change).

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: >Why change this now?

I am not changing it. It is already like that, see https://github.com/open-telemetry/opentelemetry-proto/blob/d8729d40f629dba12694b44c4c32c1eab109b00a/opentelemetry/proto/trace/v1/trace.proto#L85

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: Reworded, PTAL.

tigrannajaryan: Reworded, PTAL.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: Reworded, PTAL.

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -164,14 +164,29 @@ message LogRecord {
   fixed32 flags = 8;
 
   // A unique identifier for a trace. All logs from the same trace share
-  // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes
-  // is considered invalid. Can be set for logs that are part of request processing
-  // and have an assigned trace id. [Optional].
+  // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes OR of length other
+  // than 16 bytes is considered invalid is considered invalid (empty string in OTLP/JSON
```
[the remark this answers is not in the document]
tigrannajaryan: Oops. Fixed.

yurishkuro (approved):

jmacd:
Still, to me these statements do not belong in the specification anywhere; they look like accidents. Does the collector relaly implement this (if so where)?

I would call this a defect covered by https://github.com/open-telemetry/opentelemetry-specification/issues/1929. We are aware that there are ill-formed trace and span IDs, but we are referring to a `.proto` file instead of a data model specification. The metrics data model & SDK specifications address a number of potential ill-formed features in the data, such as duplicate-conflicting instrument definitions. The data model and SDK specifications explain that processors should pass through such invalid data and let consumers decide what to do with it. The question here is, what do we gain by filling in random fields with no other indication that the IDs are not real to cover for invalid data?

opentelemetry/proto/trace/v1/trace.proto
jsuereth: I think this should definitely treat the whole span as invalid if the trace_id is invalid.

Aneurysm9 (changes requested):
@jmacd has convinced me that this is not the correct change to make.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -79,23 +79,27 @@ message ScopeSpans {
...
-  // This field is semantically required. Receiver should generate new
-  // random span_id if empty or invalid span_id was received.
+  // is created. The ID is an 8-byte array. An ID with all zeroes OR of length
+  // other than 8 bytes is considered invalid (empty string in OTLP/JSON
+  // is zero-length and thus is also invalid).
   //
   // This field is required.
+  //
+  // The receivers SHOULD generate a new random span_id if any of the following is true:
```
jsuereth, line 100: Also disagree with generated spans here.  This would break log<->span and span<->exemplar correlation.  THis should just treat the span as invalid if the id is invalid.

tigrannajaryan:
I did some spelunking and the "generate" recommendation comes from the very first commit: https://github.com/open-telemetry/opentelemetry-proto/commit/b5bcfffce09c956b1419b2b56a9aaff384797517#diff-ef5f80fbf835dd57e14cb9264944f03d80cf6b04cc7671b0e7fb33167c67efcc where they were copied from Java repo, to which they were copied from OpenCensus https://github.com/open-telemetry/opentelemetry-java/pull/134 and in OpenCensus the wording first time appeared here https://github.com/census-instrumentation/opencensus-proto/pull/160 (authored by @SergeyKanzhelev, merged by @bogdandrutu ). I don't see any discussion on why "generating" is the right or wrong approach.

@bogdandrutu @SergeyKanzhelev do you by any chance remember why this wording is used?

tigrannajaryan:
I also checked the Collector's OTLP receiver and I don't see any code that implements the "generate on invalid" part.

Unless we hear some arguments that explain why generating is important I agree to treat it as a bug and delete the requirement.

tigrannajaryan:
I created a separate PR to delete the "generate" requirement: https://github.com/open-telemetry/opentelemetry-proto/pull/444

If that gets accepted I will rewrite this PR as necessary.

opentelemetry/proto/trace/v1/trace.proto
abitrolly: So what to do with invalid spans?

tigrannajaryan: https://github.com/open-telemetry/opentelemetry-proto/pull/444 is now merged and we no longer generate random ids.

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: https://github.com/open-telemetry/opentelemetry-proto/pull/444 is now merged and we no longer generate random ids.

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: https://github.com/open-telemetry/opentelemetry-proto/pull/444 is now merged and we no longer generate random ids.

tigrannajaryan:
All, please take another look. https://github.com/open-telemetry/opentelemetry-proto/pull/444 is now merged and this is rebased from main.

Aneurysm9 (approved):

mx-psi (approved):

tigrannajaryan:
@jsuereth @jmacd please take another look.

carlosalberto (approved):

bogdandrutu (approved):

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -164,14 +164,29 @@ message LogRecord {
...
+  // other than 8 bytes is considered invalid (empty string in OTLP/JSON
+  // is zero-length and thus is also invalid).
+  //
+  // This field is optional. If the sender specifies a valid span_id then it SHOULD also
+  // specify a valid trace_id.
+  //
+  // The receivers SHOULD assume that the log record is not associated with a
+  // span if any of the following is true:
+  //   - the field is not present,
+  //   - the field contains an invalid value.
```
bogdandrutu, line 190: For me to understand, if I am a client (sender) and want to send JSON encoded logs, and for one `LogRecord` I cannot determine the correlation with the traces in my library, can I:
1. Not send trace_id and span_id values.
2. Send both trace_id and span_id as empty strings.
3. Send both trace_id and span_id as invalid (correct sizes, all 0s).

tigrannajaryan: Yes, all 3 situations SHOULD result in receiver to assume the LogRecord is not associated with a trace/span, although of course we don't encourage sending empty strings or all 0 (which are both considered "invalid").

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -79,15 +79,17 @@ message ScopeSpans {
...
-  // is considered invalid.
+  // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes OR
+  // of length other than 16 bytes is considered invalid (empty string in OTLP/JSON
+  // is zero-length and thus is also invalid).
   //
   // This field is required.
   bytes trace_id = 1;
 
   // A unique identifier for a span within a trace, assigned when the span
-  // is created. The ID is an 8-byte array. An ID with all zeroes is considered
```
akasyagroup, line 89: // yaratıldı. Kimlik, 8 baytlık bir dizidir. Tamamı sıfır olan bir kimlik kabul edilir

akasyagroup (approved):

tigrannajaryan:
Has enough approvals, 2 days passed since last change, merging.

abitrolly:
Is it possible to have human readable documentation? Not everybody is .proto literate. There are many of us DevOps who are skilled in YAML only.

tigrannajaryan:
> Is it possible to have human readable documentation? Not everybody is .proto literate. There are many of us DevOps who are skilled in YAML only.

@abitrolly The target audience of this repository are developers who implement OTLP protocol. The topics for devops engineers are here https://opentelemetry.io/ and if you would like something to be added there please create an issue at https://github.com/open-telemetry/opentelemetry.io/issues

abitrolly:
@tigrannajaryan missed the repo name. Opened a PR in https://github.com/open-telemetry/opentelemetry-specification/pull/3152 to fill missing docs in protocol limitation.

PR #445, opened 2023-01-30
Split common package into value and scope

bogdandrutu:
This is a no-op for generated code, but helps within the collector pdata where we want to replace the generated code for the value (remove the entire value file essentially).

### DIFF
--- a/opentelemetry/proto/common/v1/scope.proto
+++ b/opentelemetry/proto/common/v1/scope.proto
@@ -1,0 +1,35 @@
+// Copyright 2019, OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package opentelemetry.proto.common.v1;
+
+import "opentelemetry/proto/common/v1/value.proto";
+
+option csharp_namespace = "OpenTelemetry.Proto.Common.V1";
+option java_multiple_files = true;
+option java_package = "io.opentelemetry.proto.common.v1";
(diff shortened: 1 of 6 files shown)


opentelemetry/proto/common/v1/scope.proto
```diff
@@ -0,0 +1,33 @@
...
+option java_outer_classname = "CommonProto";
+option go_package = "go.opentelemetry.io/proto/otlp/common/v1";
+
+// InstrumentationScope is a message representing the instrumentation scope information
+// such as the fully qualified name and version.
+message InstrumentationScope {
+  // An empty instrumentation scope name means the name is unknown.
+  string name = 1;
+  string version = 2;
+  repeated KeyValue attributes = 3;
```
tigrannajaryan, line 31: I think this needs an `import "opentelemetry/proto/common/v1/value.proto";` to compile.

tigrannajaryan (commented):
Please add a link to Collector pdata issue that requires this change.

yurishkuro:
> This is a no-op for generated code

Are you sure? May have different effect in different languages, e.g. produce scope.py/value.py in Python that must be imported separately (not saying it would be a reason not to do this).

bogdandrutu:
> Are you sure? May have different effect in different languages, e.g. produce scope.py/value.py in Python that must be imported separately (not saying it would be a reason not to do this).

You are right, did not think about this. Probably not worth it then. I will do a hack on the collector.

bogdandrutu:
@jsuereth @tigrannajaryan since you want to declare proto 1.0, I want to document if this change is considered a breaking change or not.

jsuereth:
I would like to NOT consider this a breaking change, but I'd also like to avoid changes like this post 1.0.

The cost of cosmetic changes like this can often outweigh the benefits, even in the long run.  However, at this time, I think this may be the last point in time where I'd be supportive of a change like this.

tigrannajaryan:
> > Are you sure? May have different effect in different languages, e.g. produce scope.py/value.py in Python that must be imported separately (not saying it would be a reason not to do this).
> 
> You are right, did not think about this. Probably not worth it then. I will do a hack on the collector.

Closing this since @bogdandrutu no longer needs it.

> @jsuereth @tigrannajaryan since you want to declare proto 1.0, I want to document if this change is considered a breaking change or not.

I think this is considered a breaking change in the proposed PR: https://github.com/open-telemetry/opentelemetry-proto/pull/432/

We can further discuss on that PR.

Issue #451, opened 2023-03-03
metrics.proto will cause compilation errors when the protobuf version is less than 3.15.0


ninghejun:
When I compile with protobuf v3.8.0, some compile errors occurred.
The compilation errors is as follows：

```bash
opentelemetry/proto/metrics/v1/metrics.proto:417:12: Explicit 'optional' labels are disallowed in the Proto3 syntax. To define 'optional' fields in Proto3, simply remove the 'optional' label, as fields are 'optional' by default.
opentelemetry/proto/metrics/v1/metrics.proto:452:12: Explicit 'optional' labels are disallowed in the Proto3 syntax. To define 'optional' fields in Proto3, simply remove the 'optional' label, as fields are 'optional' by default.
......
opentelemetry/proto/collector/metrics/v1/metrics_service.proto:19:1: Import "opentelemetry/proto/metrics/v1/metrics.proto" was not found or had errors.
opentelemetry/proto/collector/metrics/v1/metrics_service.proto:42:12: "opentelemetry.proto.metrics.v1.ResourceMetrics" is not defined.
```
Then I found that versions below protobuf v3.15.0 do not support optional labels. This problem needs to be fixed because protobuf v3.8.0 is heavily used in our business and it is difficult to upgrade. This prevents us from using opentelemetry.

PR #452, opened 2023-03-03
Remove the 'optional' label in metrics.proto, because it will cause compilation errors when the protobuf version is less than 3.15.0

ninghejun:
Remove the 'optional' label in metrics.proto, because it will cause compilation errors when the protobuf version is less than 3.15.0

The issue: https://github.com/open-telemetry/opentelemetry-proto/issues/451


jack-berg (changes requested):
The addition of optional fields was only done after extensive debate and we should not undo that decision. #279 contains some of this history.

PR #458, opened 2023-04-27
Eliminate nested "protocol" directory and fix all links

tigrannajaryan:
The "protocol" sub-directory is superfluous, I just moved all the files to the "specification" directory.

Fixed all links, used relative links to files in this repo.

TODO: add link checker in another PR.

PR #459, opened 2023-04-27
Add link checking

tigrannajaryan:
Borrowed from https://github.com/open-telemetry/opentelemetry-specification


tigrannajaryan:
markdown-link-check is expected to fail since we have some broken links.

jsuereth:
Can you fix the dead link?

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "devDependencies": {
+    "markdown-link-check": "3.10.3"
```
jack-berg, line 3: In the other repos we add `package-lock.json` to `.gitignore`

tigrannajaryan:
> Can you fix the dead link?

Yes, will be fixed by this: https://github.com/open-telemetry/opentelemetry-proto/pull/458

package.json
tigrannajaryan: I think we are doing it wrong in other repos :)

See https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json:
>This file is intended to be committed into source repositories

package.json
Kielek, line 1: IMO it is worth to configure dependabot for this repository to keep it all checks up to date.
It should include at least this file and all workflows).

tigrannajaryan: Good point. That can be a separate PR.

package.json
jack-berg: I'm not an expert, but in my experience package-lock.json contents are fairly dependent on a particular version of node / npm. Perhaps package-lock.json is ignored elsewhere to avoid pesky diffs when folks are using different versions?

tigrannajaryan: I am not an expert either, a couple relevant SO questions:
- https://stackoverflow.com/questions/48524417/should-the-package-lock-json-file-be-added-to-gitignore (300 upvotes for accepted answer)
- https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5 (2520 upvotes for accepted answer)

I can't find significant evidence for the opposite (that package-lock.json should be gitignored).

jack-berg: @open-telemetry/javascript-maintainers Maybe you could offer some guidance here.. Repos like this and and `opentelemetry-specification` use node tools for various `make` steps. Any thoughts on whether we should check `package-lock.json` into version control? Would doing so change the [make install-tools](https://github.com/open-telemetry/opentelemetry-specification/blob/main/Makefile#L101-L103) step or the advice we give for [contributors](https://github.com/open-telemetry/opentelemetry-specification/blob/main/CONTRIBUTING.md#consistency-checks)?

legendecas: As a JS library, the https://github.com/open-telemetry/opentelemetry-js repository doesn't check in the `packge-lock.json` to stay updated with dependencies and test them in a timely manner. The `package-lock.json` is not published to the npm registry along with the product.

However, it can be a good practice to check-in `package-lock.json` for toolings as it would help avoid unexpected breakages. `package-lock.json` locks the dependency trees and needs to be upgraded regularly with tools like renovate to address security updates (e.g. https://github.com/open-telemetry/opentelemetry-js/blob/main/renovate.json).

tigrannajaryan:
@open-telemetry/specs-approvers  please review.

tigrannajaryan:

package.json
jack-berg: Yup thanks @legendecas and @dyladan. Will try to remember to apply this advice elsewhere.

jack-berg (approved):

jsuereth (approved):

Issue #433, opened 2022-10-18
Helper enums ZERO value is incorrectly used


bogdandrutu:
In the collector before we removed access to the helper enums like `DataPointFlags` (https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L321), I've seen code comparing with `FLAG_NONE` or initializing as `FLAG_NONE` the value of the flags to check if a value is present. 

If the helper enums contain "masks" then ZERO value does not make sense, if the enums contain "absolute" values, then zero value makes sense but the other values don't make sense.

My personal suggestion is still to remove them (see https://github.com/open-telemetry/opentelemetry-proto/pull/423) but if that is not ok, this inconsistency between masks/absolute value has to be fixed.

Assigning to @tigrannajaryan since he is a big supported of these values.

tigrannajaryan:
>If the helper enums contain "masks" then ZERO value does not make sense

@bogdandrutu I think you are right. I support removing the zero value from enums for bitmasks.

tigrannajaryan:
I think this should be done before 1.0 so adding to the milestone.

Aneurysm9:
I'm not sure what the problem is with the way things are currently.  If a helper enum contains bitmask values then a `NONE` value still makes sense.

```go
const (
	FLAG_NONE = 0
	FLAG_RED = 1
	FLAG_BLUE = 2
	FLAG_GREEN = 4
)

// 7
value := FLAG_NONE | FLAG_RED | FLAG_BLUE | FLAG_GREEN

// false
isNone := value == FLAG_NONE // or, value == value & FLAG_NONE

// true
isRed := FLAG_RED == value & FLAG_RED

//true
isGreen := FLAG_GREEN == value & FLAG_GREEN

//true
isBlue := FLAG_BLUE == value & FLAG_BLUE
```

While it can allow for nonsensical combinations, this is likely the case for many bitmask-type flag sets and there are still sensible ways to interact with them.  Is a user required to utilize the values correctly? Yes.  That's the case regardless whether we include helper enums or not, though.  

Are there concrete examples of these values being used incorrectly that can be pointed out?  I think in their absence it will be difficult to progress here.

bogdandrutu:
What is the meaning of "None" in a bit map? Why do you not offer a "All" (or other subset)?

tigrannajaryan:
[...]

PR #461, opened 2023-05-02
Clarify that zero values bitfield helper enums should not be used

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-proto/issues/433

## Problem

The zero values defined in helper enums encourage incorrect usage of the bitfields.

The typical incorrect code looks like this:

```proto
enum DataPointFlags {
  FLAG_NONE = 0;
  FLAG_NO_RECORDED_VALUE = 1;
  // Bits 2-31 are reserved for future use.
}
```

```go
if datapoint.Flags == FLAG_NONE {
  // do something when there is no recorded value
}
```

This is incorrect because it does not take into account that the Flags field can be extended in the future to use the reserved bits. The `if` statement above will be incorrect if any other bit is set.

The correct code looks like this:
```go
if (datapoint.Flags & FLAG_NO_RECORDED_VALUE) == 0 {
  // do something when there is no recorded value
}
```

## Solution

To prevent this mistake some additional comments are added and the zero value in the enum is prefixed with an underscore to discourage its use.

## Alternates Tried

I also tried to remove the zero values from enums altogether, but it turned out to be impossible. I tried a few possible approaches and none worked.

Protobufs [require that enums start with a 0 value](https://protobuf.dev/programming-guides/proto3/#enum).

If you try to omit it you get a compilation error:
```
opentelemetry/proto/metrics/v1/metrics.proto:325:28: The first enum value must be zero in proto3.
```

Alternatively, trying to use a dummy name, e.g.:
```
enum DataPointFlags {
  _ = 0;
 FLAG_NO_RECORDED_VALUE = 1;
}
```

Fails Objective-C generator:
```
error: got empty string for making TextFormat data, input: "", desired: "_".
```

Also tried declaring it reserved as the doc says should be possible:
```
enum DataPointFlags {
  reserved 0;
  FLAG_NO_RECORDED_VALUE = 1;
}
```

but get an error again:
```
opentelemetry/proto/metrics/v1/metrics.proto:327:28: The first enum value must be zero in proto3.
```

### DIFF
--- a/opentelemetry/proto/logs/v1/logs.proto
+++ b/opentelemetry/proto/logs/v1/logs.proto
@@ -104,10 +104,21 @@ enum SeverityNumber {
   SEVERITY_NUMBER_FATAL4 = 24;
 }
 
-// Masks for LogRecord.flags field.
+// LogRecordFlags is defined as a protobuf 'uint32' type and is to be used as
+// bit-fields. Each non-zero value defined in this enum is a bit-mask.
+// To extract the bit-field, for example, use an expression like:
+//
+//   (logRecord.flags & LOG_RECORD_FLAG_TRACE_FLAGS_MASK)
+//
 enum LogRecordFlags {
-  LOG_RECORD_FLAG_UNSPECIFIED = 0;
+  // The zero value for the enum. Should not be used for comparisons.
+  // Instead use bitwise "and" with the appropriate mask as shown above.
+  LOG_RECORD_FLAG_DO_NOT_USE = 0;
+
+  // Bits 0-7 are used for trace flags.
   LOG_RECORD_FLAG_TRACE_FLAGS_MASK = 0x000000FF;
+
+  // Bits 8-31 are reserved for future use.
 }
 
 // A log record according to OpenTelemetry Log Data Model:
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -316,15 +316,17 @@ enum AggregationTemporality {
 // enum is a bit-mask.  To test the presence of a single flag in the flags of
 // a data point, for example, use an expression like:
 //
-//   (point.flags & FLAG_NO_RECORDED_VALUE) == FLAG_NO_RECORDED_VALUE
+//   (point.flags & FLAG_NO_RECORDED_VALUE_MASK) == FLAG_NO_RECORDED_VALUE_MASK
 //
 enum DataPointFlags {
-  FLAG_NONE = 0;
+  // The zero value for the enum. Should not be used for comparisons.
+  // Instead use bitwise "and" with the appropriate mask as shown above.
+  FLAG_DO_NOT_USE = 0;
 
   // This DataPoint is valid but has no recorded value.  This value
   // SHOULD be used to reflect explicitly missing data in a series, as
   // for an equivalent to the Prometheus "staleness marker".
-  FLAG_NO_RECORDED_VALUE = 1;
+  FLAG_NO_RECORDED_VALUE_MASK = 1;
 
   // Bits 2-31 are reserved for future use.
 }


opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -104,10 +104,21 @@ enum SeverityNumber {
...
 }
 
-// Masks for LogRecord.flags field.
+// LogRecordFlags is defined as a protobuf 'uint32' type and is to be used as
+// bit-fields. Each non-zero value defined in this enum is a bit-mask.
+// To extract the bit-field, for example, use an expression like:
+//
+//   (logRecord.flags & LOG_RECORD_FLAG_TRACE_FLAGS_MASK)
+//
 enum LogRecordFlags {
```
bogdandrutu, line 113: Having a "Mask" in all the values (by having it in the enum and prefix all values with enum name as we do everywhere else).

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -319,7 +319,9 @@ enum AggregationTemporality {
 //   (point.flags & FLAG_NO_RECORDED_VALUE) == FLAG_NO_RECORDED_VALUE
 //
 enum DataPointFlags {
-  FLAG_NONE = 0;
```
bogdandrutu, line 322: For me this is the worst name, I saw at least 5 times people using equality with this value, so make sure we change this.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: I renamed FLAG_NO_RECORDED_VALUE -> FLAG_NO_RECORDED_VALUE_MASK. Now all enum elements have the _MASK suffix.

Do you also want to have the Mask suffix for the enum names (in addition to enum element names)?

opentelemetry/proto/metrics/v1/metrics.proto
tigrannajaryan: Does renaming FLAG_NONE -> _FLAG_UNSPECIFIED and addition of comments help?

Aneurysm9:
> ```go
> if datapoint.Flags == FLAG_NONE {
>   // do something when there is no recorded value
> }
> ```
>
> This is incorrect because it does not take into account that the Flags field can be extended in the future to use the reserved bits. The if statement above will be incorrect if any other bit is set.

This is incorrect because it assumes that no value in the flag means no recorded value in the data point, which is ignoring the flag altogether.

> The correct code looks like this:
> 
> ```go
> if (datapoint.Flags & FLAG_NO_RECORDED_VALUE) == 0 {
>   // do something when there is no recorded value
> }
> ```

This is also wrong.  This predicate will be true when the no recorded value flag is not set.  This is  the opposite of what is expected.  The correct check for no recorded value would be this:

```go
if (datapoint.Flags & FLAG_NO_RECORDED_VALUE) == FLAG_NO_RECORDED_VALUE { // or != 0
  // do something when there is no recorded value
}
```

If you really want a mask value to indicate that no non-reserved flags are set then you would want something like this:

```go
const NO_RESERVED_FLAGS_MASK = 0 ^ FLAG_NO_RECORDED_VALUE // also xor any other valid flags
```

See this demonstration on the playground: https://go.dev/play/p/Rf0DU9CYvmi

Aneurysm9 (changes requested):
This change violates the OTLP stability guarantees.

> Names of enum choices and numbers assigned to enum choices will not change.

carlosalberto:
> This change violates the OTLP stability guarantees.

We haven't published [this change](https://github.com/open-telemetry/opentelemetry-proto/pull/432/files) yet (that is, not release includes this part yet). I'm guessing the plan is to do a release once 1.0 is reached (very soon hopefully).

tigrannajaryan:
> > This change violates the OTLP stability guarantees.
> 
> We haven't published [this change](https://github.com/open-telemetry/opentelemetry-proto/pull/432/files) yet (that is, not release includes this part yet). I'm guessing the plan is to do a release once 1.0 is reached (very soon hopefully).

Yeah, I think we should have made it clear that Stability guarantees are for 1.0. That was my intent, I had no intent to lock the current repo state as is, while it is still 0.x.

Aneurysm9:
> > This change violates the OTLP stability guarantees.
> 
> We haven't published [this change](https://github.com/open-telemetry/opentelemetry-proto/pull/432/files) yet (that is, not release includes this part yet). I'm guessing the plan is to do a release once 1.0 is reached (very soon hopefully).

We didn't publish a release when we stabilized OTLP/JSON either.  The relevant spec is stable and we have changed our public communication regarding what "stable" means in a way that makes this proposed change incompatible with the stability guarantees we say we're making.

> Yeah, I think we should have made it clear that Stability guarantees are for 1.0. That was my intent, I had no intent to lock the current repo state as is, while it is still 0.x.

If that was the intent then merging the stability guarantees change should have been the last thing done and should have been explicitly coupled to the 1.0 version.  We have long said that parts of this spec are "stable" even without a 1.0 version number and I expect to be able to rely on that.

carlosalberto:
> We didn't publish a release when we stabilized OTLP/JSON either. 

That was a mistake, but we are doing the same for the Spec - if it's not released, it's not accepted. We have even reverted changes (permanently or temporarily). So let's try to do it right this time and do a prompt release once all the 1.0 items are done.

tigrannajaryan:
Here is a PR to make it clear that symbolic guarantees go into effect only when 1.0 is released: https://github.com/open-telemetry/opentelemetry-proto/pull/467

jsuereth (approved):
Agree w/ @tigrannajaryan's rationale for these changes.

bogdandrutu (approved):

bogdandrutu (approved):

opentelemetry/proto/metrics/v1/metrics.proto
tigrannajaryan: I can get rid of the prefix. Was just trying to make them look special. @open-telemetry/specs-approvers any thoughts?

tigrannajaryan:
Now that https://github.com/open-telemetry/opentelemetry-proto/pull/467 is merged it is clear that this PR is allowed.

arminru (approved):

carlosalberto (approved):

opentelemetry/proto/logs/v1/logs.proto
bogdandrutu, line 113: LogRecordFlags -> LogRecordFlag or the values get the "S"?

opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -104,10 +104,21 @@ enum SeverityNumber {
...
-  LOG_RECORD_FLAG_UNSPECIFIED = 0;
+  // The zero value for the enum. Should not be used for comparisons.
+  // Instead use bitwise "and" with the appropriate mask as shown above.
+  _LOG_RECORD_FLAG_UNSPECIFIED = 0;
```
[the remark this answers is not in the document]
bogdandrutu: `LOG_RECORD_FLAGS_DO_NOT_USE`?

opentelemetry/proto/metrics/v1/metrics.proto
tigrannajaryan: OK, remove the _ prefix and renamed to use _DO_NOT_USE suffix instead.

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: We use plural for enum name (it contains many flags) and singular for each item in the enum. I think it is reasonable, so let's keep it that way and avoid more breaking changes than we need to.

tigrannajaryan:
All, I  removed the `_` prefix from the 0 elements and renamed to use `_DO_NOT_USE` suffix instead.

opentelemetry/proto/logs/v1/logs.proto
bogdandrutu: Then what about `LOG_RECORD_FLAG[S]_TRACE_FLAGS_MASK` to be consistent with other enums?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -316,15 +316,17 @@ enum AggregationTemporality {
...
 // a data point, for example, use an expression like:
 //
-//   (point.flags & FLAG_NO_RECORDED_VALUE) == FLAG_NO_RECORDED_VALUE
+//   (point.flags & FLAG_NO_RECORDED_VALUE_MASK) == FLAG_NO_RECORDED_VALUE_MASK
 //
 enum DataPointFlags {
-  FLAG_NONE = 0;
+  // The zero value for the enum. Should not be used for comparisons.
+  // Instead use bitwise "and" with the appropriate mask as shown above.
+  FLAG_DO_NOT_USE = 0;
```
bogdandrutu, line 324: Should this be `DATA_POINT_FLAG[S]_`

opentelemetry/proto/logs/v1/logs.proto
tigrannajaryan: Filed an issue to look at it: https://github.com/open-telemetry/opentelemetry-proto/issues/473

opentelemetry/proto/metrics/v1/metrics.proto
tigrannajaryan: Filed an issue to look at it https://github.com/open-telemetry/opentelemetry-proto/issues/473

tigrannajaryan:
@open-telemetry/specs-approvers this has enough approvals. Please object, otherwise I will merge.

PR #463, opened 2023-05-04
Add OTLP JSON payload examples

### LINKED ISSUE #462: Add examples to OTLP/JSON
OTLP/JSON is a human readable format. It would be nice to have some example payloads in the OTLP specification to demonstrate what it can look like. Examples sometimes are very useful when you want to understand the structure of the payload.

See for example a user asking for it: https://github.com/open-telemetry/opentelemetry-specification/issues/2773#issuecomment-1245573473

joaopgrassi:
Fixes #462

I was not sure where to add the files, so I just went with some initial idea. I'm open to suggestions on where to put them/refer to them. 

I tested it with a local collector and all works. 

cc @tigrannajaryan

### DIFF
--- a/specification/otlp-json-examples/README.md
+++ b/specification/otlp-json-examples/README.md
@@ -1,0 +1,43 @@
+# OTLP JSON request examples
+
+This folder contains a collection of OTLP JSON files for all signals
+that can be used as request payloads.
+
+- Trace [trace.json](trace.json)
+- Metrics [metrics.json](metrics.json)
+- Logs [logs.json](logs.json)
+
+## Trying it out
+
+First run a OpenTelemetry collector with the following configuration:
+
+```yaml
+receivers:
+  otlp:
+    protocols:
+      http:
+
+exporters:
+  logging:
+    verbosity: detailed
+
+service:
+  pipelines:
+    traces:
+      receivers: [otlp]
+      exporters: [logging]
+    metrics:
+      receivers: [otlp]
+      exporters: [logging]
+    logs:
+      receivers: [otlp]
+      exporters: [logging]
+```
+
+Then send a curl request to the collector (e.g. for Logs):
+
+```shell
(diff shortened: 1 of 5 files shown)


arminru (approved):

specification/otlp-json-examples/README.md
```diff
@@ -0,0 +1,43 @@
+# OTLP JSON request examples
+
+This folder contains a collection of OTLP JSON files for all signals
```
tigrannajaryan, line 3:
```suggestion
This folder contains a collection of example OTLP JSON files for all signals
```

bogdandrutu (commented):
Because I know that we will do manual work here withe these examples. should we have at least a check that they are valid JSONs?

bogdandrutu (commented):
Because I know that we will do manual work here withe these examples. should we have at least a check that they are valid JSONs?

bogdandrutu (commented):
Because I know that we will do manual work here withe these examples. should we have at least a check that they are valid JSONs?

tigrannajaryan:
Yes more checks would be useful. Related: we have a PR coming that adds link checking: https://github.com/open-telemetry/opentelemetry-proto/pull/459

tigrannajaryan:
> I was also thinking if we shouldn't publish a json schema for OTLP. That way the guess work of writing OTLP JSON is removed and we could also check it against the schema in a gh action. I saw some tool to generate JSON schema from proto files but didn't have time to check. If you think it's valid, I can create issues and maybe work on it.

I would not want to maintain another schema definition. The Protobuf definitions are the source of truth. If we could generate JSON schema from Protobufs that would be great.

jsuereth (approved):

tigrannajaryan:
Merging. @joaopgrassi it would be great if you could follow up with a PR to add JSON checks.

joaopgrassi:
Will do!

chalin:
Hi all, I've posted a suggestion to:

- #471

Comments are welcome.

PR #469, opened 2023-05-09
Rename ./specification to ./docs

chalin:
Hi @tigrannajaryan @svrnm @cartermp et al.,

Why this PR? Because the documented OTLP specification is strictly contained within `./specification/otlp.md`. The remaining pages under `./specification` are supporting documentation. So, in this PR I propose the following:

- Rename `./specification` to `./docs`
- Rename `specification/otlp.md` to `./docs/specification.md`
- Adjusts links in other .md pages

Thoughts?

Contributes to:
 
- https://github.com/open-telemetry/opentelemetry.io/issues/2642

/cc @open-telemetry/specs-approvers

PR #472, opened 2023-05-11
Move examples into an example folder

### LINKED ISSUE #471: Move examples out of spec into an examples folder
This is a followup to:

- #463

@tigrannajaryan @joaopgrassi, et al.: In #463, the examples were added to the spec, is that because the examples are they meant to be binding / enforce some sort of specification-related conformance? If not, they should be moved out of the spec folder IMHO, into, say, an `examples` folder.

Keep in mind that everything in the spec folder (hopefully to be renamed to `docs` via #469), will soon be published to the OTel website.

/cc @svrnm @cartermp

chalin:
- Closes #471 by moving the recently added examples into a top-level `examples` folder.
- An alternative would be to put the JSON examples under `examples/json`, but using just `examples` seemed ok for now. Thoughts welcome.

/cc @joaopgrassi @tigrannajaryan


tigrannajaryan (approved):

arminru (approved):

arminru:
Thanks, @chalin! I'll merge this one so your other PR #469 is unblocked.

Issue #473, opened 2023-05-11
Enum element naming consistency


tigrannajaryan:
Most of the enum element names are prefixed by enum name in CAPITAL_CASE, e.g.:

enum StatusCode -> element STATUS_CODE_OK

We have 2 enums that don't follow this rule: LogRecordFlags and DataPointFlags.

Should we fix this?

Need to resolve this for 1.0.

tigrannajaryan:
@open-telemetry/specs-approvers If we fix this we will break some existing code. We are formally allowed to do that but it is still an annoyance. What do you think, is it worth it?

PR #474, opened 2023-05-15
Fix LogRecordFlags and DataPointFlags enum element name inconsistency

tigrannajaryan:
Other enum element names in this repo are prefixed by enum name in CAPITAL_CASE.

This change makes naming of LogRecordFlags and DataPointFlags elements consistent with the rest of the enums.

Resolves https://github.com/open-telemetry/opentelemetry-proto/issues/473

@open-telemetry/specs-approvers this breaks generated code. We are allowed to make such breaking changes, but I want us to be aware that we are causing some pain and only make this change if we think it is worth it. I would want to see a majority of TC approving this before I merge. I can personally argue both ways and do not see the definite argument to make up my mind.


dyladan:
> I would want to see a majority of TC approving this before I merge

Since this is such a high bar should we set some deadline so it doesn't sit forever? There may be TC members who don't feel strongly enough to block it but also don't want to approve it.

tedsuo (approved):
We want OTel to be a long lasting standard. It's best to fix these inconsistencies when we have a chance.

jack-berg (approved):

carlosalberto (approved):

bogdandrutu (approved):

jmacd (approved):

tigrannajaryan:
We have enough approvals. I am going to keep it open for another day (2 day rule) and then will merge.

PR #107, opened 2020-02-18
remove local_child_span_count

SergeyKanzhelev:
Fixes #105

PR #484, opened 2023-06-13
Indicate if a span's parent or link is remote using 2 bit flag as described in OTEP 0182

estolfo:
~~This adds `parent_span_is_remote` field to Span message~~ introduced in OTEP 0182: https://github.com/open-telemetry/oteps/blob/main/text/0182-otlp-remote-parent.md

Update: The PR now uses 2 bits from `flags` to represent the 3 states of parent or link is remote: unknown, not remote, is remote.

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -274,6 +274,10 @@ message Span {
     // SPAN_FLAGS_TRACE_FLAGS_MASK).  [Optional].
     //
     // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
+    //
+    // Bits 8 and 9 represent the 3 states of whether the link
+    // is remote. The states are unknown, is not remote, is remote.
+    // To read the 2-bit flag, use `flags & SPAN_FLAGS_LINK_IS_REMOTE_MASK`.
     fixed32 flags = 6;
   }
 
@@ -340,6 +344,10 @@ enum SpanFlags {
   // See SpanFlagsParentIsRemote for the valid values.
   SPAN_FLAGS_PARENT_IS_REMOTE_MASK = 0x00000300;
 
+  // Bits 8 and 9 are used for for the link is remote flag.
+  // See SpanFlagsLinkIsRemote for the valid values.
+  SPAN_FLAGS_LINK_IS_REMOTE_MASK = 0x00000300;
+
   // Bits 10-31 are reserved for future use.
 }
 
@@ -357,4 +365,20 @@ enum SpanFlagsParentIsRemote {
   SPAN_FLAGS_PARENT_IS_REMOTE = 0x00000300;
   // The parent is not remote: 01
   SPAN_FLAGS_PARENT_IS_NOT_REMOTE = 0x00000100;
+}
+
+// SpanFlagsLinkIsRemote represents constants used to interpret
+// bits 8 and 9 of the Link.flags field. These two bits represent whether
+// the link is remote. Bit 8 represents whether the field has been
+// set and bit 9 represents whether the link is remote.
+// Once the bits have been read using SPAN_FLAGS_LINK_IS_REMOTE_MASK,
+// the following constants can be used to interpret them.
+enum SpanFlagsLinkIsRemote {
+  // Older clients may not set this field, so 00 represents that this value
+  // has not been set.
+  SPAN_FLAGS_LINK_IS_REMOTE_UNKNOWN = 0;
+  // The linked span is remote: 11
+  SPAN_FLAGS_LINK_IS_REMOTE = 0x00000300;
+  // The linked span is not remote: 01
+  SPAN_FLAGS_LINK_IS_NOT_REMOTE = 0x00000100;
 }
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -353,7 +353,7 @@ enum SpanFlags {
 // remote.
 // Once the bits have been read using SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_MASK,
 // the following constants can be used to interpret them.
-enum SpanFlagsParentOrLinkIsRemote {
+enum SpanFlagsContextIsRemote {
   // Older clients may not set this field, so 00 represents that this value
   // has not been set.
   SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_UNKNOWN = 0;
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -109,7 +109,9 @@ message Span {
   // field must be empty. The ID is an 8-byte array.
   bytes parent_span_id = 4;
 
-  // Flags, a bit field. Bits 0-7 (8 least significant bits) are the trace
+  // Flags, a bit field. 
+  //
+  // Bits 0-7 (8 least significant bits) are the trace
   // flags as defined in W3C Trace Context specification.
   // To read the 8-bit W3C trace flag, use `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
   //
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -285,7 +285,8 @@ message Span {
   // ParentSpanIsRemote indicates whether the span's parent is remote (true / false) or
   // if the field is unset.
   enum ParentSpanIsRemote {
-    // Unset.
+    // Unset. This value is expected from older clients. In which case, ParentSpanIsRemote
+    // can be derived by looking at the parent span and determining if it has a different Resource.
     PARENT_SPAN_IS_REMOTE_UNSET = 0;
 
     // The parent span is not remote.
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -358,8 +358,8 @@ enum SpanFlags {
 // Once the bits have been read using SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK,
 // the following constants can be used to interpret them.
 enum SpanFlagsContextIsRemote {
-  // Older clients may not set this field, so 00 represents that this value
-  // has not been set.
+  // Older client not supporting this flag, or not known whether parent or
+  // linked span is remote: 00
   SPAN_FLAGS_CONTEXT_IS_REMOTE_UNKNOWN = 0;
   // The parent or linked span is remote: 11
   SPAN_FLAGS_CONTEXT_IS_REMOTE = 0x00000300;
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -284,7 +284,7 @@ message Span {
 
   // ParentSpanIsRemote indicates whether the span's parent is remote (true / false) or
   // if the field is unset.
-  enum ParentIsRemote {
+  enum ParentSpanIsRemote {
     // Unset.
     PARENT_SPAN_IS_REMOTE_UNSET = 0;
 
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -123,6 +123,10 @@ message Span {
   // [Optional].
   //
   // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
+  //
+  // Bits 8 and 9 are reserved for representing the 3 states of whether a span's parent
+  // is remote. The states are unknown, is not remote, is remote.
+  // To read the 2-bit flag, use `flags & SPAN_FLAGS_PARENT_IS_REMOTE_MASK`.
   fixed32 flags = 16;
 
   // A description of the span's operation.
@@ -329,5 +333,8 @@ enum SpanFlags {
   // Bits 0-7 are used for trace flags.
   SPAN_FLAGS_TRACE_FLAGS_MASK = 0x000000FF;
 
-  // Bits 8-31 are reserved for future use.
+  // Bits 8 and 9 are used for for the parent is remote flag.
+  SPAN_FLAGS_PARENT_IS_REMOTE_MASK = 0x00000300;
+
+  // Bits 10-31 are reserved for future use.
 }
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -341,7 +341,7 @@ enum SpanFlags {
 
   // Bits 8 and 9 are used to indicate that the parent span or link span is remote.
   // See SpanFlagsParentOrLinkIsRemote for the valid values.
-  SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_MASK = 0x00000300;
+  SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK = 0x00000300;
 
   // Bits 10-31 are reserved for future use.
 }
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -119,7 +119,8 @@ message Span {
   //
   // Bits 8 and 9 represent the 3 states of whether a span's parent
   // is remote. The states are (unknown, is not remote, is remote).
-  // To read the 2-bit flag, use `flags & SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK`.
+  // To read whether the value is known, use `flags & SPAN_FLAGS_CONTEXT_HAS_IS_REMOTE_MASK`.
+  // To read whether the span is remote, use `flags & SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK`.
   //
   // When creating span messages, if the message is logically forwarded from another source
   // with an equivalent flags fields (i.e., usually another OTLP span message), the field SHOULD
@@ -275,7 +276,8 @@ message Span {
     //
     // Bits 8 and 9 represent the 3 states of whether the link is remote. The states
     // are unknown, is not remote, is remote.
-    // To read the 2-bit flag, use `flags & SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK`.
+    // To read whether the value is known, use `flags & SPAN_FLAGS_CONTEXT_HAS_IS_REMOTE_MASK`.
+    // To read whether the link is remote, use `flags & SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK`.
     //
     // Readers MUST NOT assume that bits 10-31 (22 most significant bits) will be zero.
     // When creating new spans, bits 10-31 (most-significant 22-bits) MUST be zero.
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -344,25 +344,9 @@ enum SpanFlags {
   SPAN_FLAGS_TRACE_FLAGS_MASK = 0x000000FF;
 
   // Bits 8 and 9 are used to indicate that the parent span or link span is remote.
-  // See SpanFlagsContextIsRemote for the valid values.
+  // Bit 8 indicates whether the value is known. Bit 9 indicates whether the span
+  // or link is remote.
   SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK = 0x00000300;
 
   // Bits 10-31 are reserved for future use.
-}
-
-// SpanFlagsContextIsRemote represents constants used to interpret
-// bits 8 and 9 of the Spans.flags or Link.flags field. These two bits
-// represent whether the parent or link is remote. Bit 8 represents whether
-// the field has been set and bit 9 represents whether the parent or link is
-// remote.
-// Once the bits have been read using SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK,
-// the following constants can be used to interpret them.
-enum SpanFlagsContextIsRemote {
-  // Older client not supporting this flag, or not known whether parent or
-  // linked span is remote: 00
-  SPAN_FLAGS_CONTEXT_IS_REMOTE_UNKNOWN = 0;
-  // The parent or linked span is remote: 11
-  SPAN_FLAGS_CONTEXT_IS_REMOTE = 0x00000300;
-  // The parent or linked span is not remote: 01
-  SPAN_FLAGS_CONTEXT_IS_NOT_REMOTE = 0x00000100;
 }
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -340,13 +340,13 @@ enum SpanFlags {
   SPAN_FLAGS_TRACE_FLAGS_MASK = 0x000000FF;
 
   // Bits 8 and 9 are used to indicate that the parent span or link span is remote.
-  // See SpanFlagsParentOrLinkIsRemote for the valid values.
+  // See SpanFlagsContextIsRemote for the valid values.
   SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK = 0x00000300;
 
   // Bits 10-31 are reserved for future use.
 }
 
-// SpanFlagsParentOrLinkIsRemote represents constants used to interpret
+// SpanFlagsContextIsRemote represents constants used to interpret
 // bits 8 and 9 of the Spans.flags or Link.flags field. These two bits
 // represent whether the parent or link is remote. Bit 8 represents whether
 // the field has been set and bit 9 represents whether the parent or link is
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -282,11 +282,24 @@ message Span {
   // span's status code is unset, i.e. assume STATUS_CODE_UNSET (code = 0).
   Status status = 15;
 
+  // ParentSpanIsRemote indicates whether the span's parent is remote (true / false) or
+  // if the field is unset.
+  enum ParentIsRemote {
+    // Unset.
+    PARENT_SPAN_IS_REMOTE_UNSET = 0;
+
+    // The parent span is not remote.
+    PARENT_SPAN_IS_REMOTE_FALSE = 1;
+
+    // The parent span is remote.
+    PARENT_SPAN_IS_REMOTE_TRUE = 2;
+  }
+
   // parent_span_is_remote identifies whether a span's parent is remote or not.
   // This helps determine whether a span can be considered an entry-point span.
   // A span is an entry-point span if it has no parent (parent_span_id is empty),
-  // or if parent_span_is_remote is true.
-  bool parent_span_is_remote = 16;
+  // or if parent_is_remote is true.
+  ParentSpanIsRemote parent_span_is_remote = 16;
 }
 
 // The Status type defines a logical error model that is suitable for different
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -351,7 +351,7 @@ enum SpanFlags {
 // represent whether the parent or link is remote. Bit 8 represents whether
 // the field has been set and bit 9 represents whether the parent or link is
 // remote.
-// Once the bits have been read using SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_MASK,
+// Once the bits have been read using SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK,
 // the following constants can be used to interpret them.
 enum SpanFlagsContextIsRemote {
   // Older clients may not set this field, so 00 represents that this value
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -356,9 +356,9 @@ enum SpanFlags {
 enum SpanFlagsContextIsRemote {
   // Older clients may not set this field, so 00 represents that this value
   // has not been set.
-  SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_UNKNOWN = 0;
+  SPAN_FLAGS_CONTEXT_IS_REMOTE_UNKNOWN = 0;
   // The parent or linked span is remote: 11
-  SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE = 0x00000300;
+  SPAN_FLAGS_CONTEXT_IS_REMOTE = 0x00000300;
   // The parent or linked span is not remote: 01
-  SPAN_FLAGS_PARENT_OR_LINK_IS_NOT_REMOTE = 0x00000100;
+  SPAN_FLAGS_CONTEXT_IS_NOT_REMOTE = 0x00000100;
 }
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -109,11 +109,11 @@ message Span {
   // field must be empty. The ID is an 8-byte array.
   bytes parent_span_id = 4;
 
-  // Flags, a bit field. 
+  // Flags, a bit field.
   //
-  // Bits 0-7 (8 least significant bits) are the trace
-  // flags as defined in W3C Trace Context specification.
-  // To read the 8-bit W3C trace flag, use `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
+  // Bits 0-7 (8 least significant bits) are the trace flags as defined in W3C Trace
+  // Context specification. To read the 8-bit W3C trace flag, use
+  // `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
   //
   // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
   //
@@ -265,9 +265,11 @@ message Span {
     // then no attributes were dropped.
     uint32 dropped_attributes_count = 5;
 
-    // Flags, a bit field. Bits 0-7 (8 least significant bits) are the trace
-    // flags as defined in W3C Trace Context specification. To read the 8-bit W3C trace
-    // flag, use `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
+    // Flags, a bit field.
+    //
+    // Bits 0-7 (8 least significant bits) are the trace flags as defined in W3C Trace
+    // Context specification. To read the 8-bit W3C trace flag, use
+    // `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
     //
     // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
     //
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -344,9 +344,10 @@ enum SpanFlags {
   SPAN_FLAGS_TRACE_FLAGS_MASK = 0x000000FF;
 
   // Bits 8 and 9 are used to indicate that the parent span or link span is remote.
-  // Bit 8 indicates whether the value is known. Bit 9 indicates whether the span
-  // or link is remote.
-  SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK = 0x00000300;
+  // Bit 8 (`HAS_IS_REMOTE`) indicates whether the value is known.
+  // Bit 9 (`IS_REMOTE`) indicates whether the span or link is remote.
+  SPAN_FLAGS_CONTEXT_HAS_IS_REMOTE_MASK = 0x00000100;
+  SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK = 0x00000200;
 
   // Bits 10-31 are reserved for future use.
 }
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -286,13 +286,13 @@ message Span {
   // if the field is unset.
   enum ParentSpanIsRemote {
     // Unset. This value is expected from older clients.
-    PARENT_SPAN_IS_REMOTE_UNSET = 0;
+    PARENT_SPAN_UNKNOWN_REMOTE = 0;
 
     // The parent span is not remote.
-    PARENT_SPAN_IS_REMOTE_FALSE = 1;
+    PARENT_SPAN_NOT_REMOTE = 1;
 
     // The parent span is remote.
-    PARENT_SPAN_IS_REMOTE_TRUE = 2;
+    PARENT_SPAN_REMOTE = 2;
   }
 
   // parent_span_is_remote identifies whether a span's parent is remote or not.
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -118,7 +118,7 @@ message Span {
   // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
   //
   // Bits 8 and 9 represent the 3 states of whether a span's parent
-  // is remote. The states are unknown, is not remote, is remote.
+  // is remote. The states are (unknown, is not remote, is remote).
   // To read the 2-bit flag, use `flags & SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK`.
   //
   // When creating span messages, if the message is logically forwarded from another source
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -281,6 +281,12 @@ message Span {
   // An optional final status for this span. Semantically when Status isn't set, it means
   // span's status code is unset, i.e. assume STATUS_CODE_UNSET (code = 0).
   Status status = 15;
+
+  // parent_span_is_remote identifies whether a span's parent is remote or not.
+  // This helps determine whether a span can be considered an entry-point span.
+  // A span is an entry-point span if it has no parent (parent_span_id is empty),
+  // or if parent_span_is_remote is true.
+  bool parent_span_is_remote = 16;
 }
 
 // The Status type defines a logical error model that is suitable for different
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -127,9 +127,9 @@ message Span {
   // is remote. The states are unknown, is not remote, is remote.
   // To read the 2-bit flag, use `flags & SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_MASK`.
   //
-  // [Optional].
+  // Readers MUST NOT assume that bits 10-31 (22 most significant bits) will be zero.
   //
-  // Readers MUST not assume that bits 10-31 (22 most significant bits) will be zero.
+  // [Optional].
   fixed32 flags = 16;
 
   // A description of the span's operation.
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -113,20 +113,17 @@ message Span {
   // flags as defined in W3C Trace Context specification.
   // To read the 8-bit W3C trace flag, use `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
   //
-  // When creating span messages, if the message is logically forwarded from another source
-  // with an equivalent flags fields (i.e., usually another OTLP span message), the field SHOULD
-  // be copied as-is. If creating from a source that does not have an equivalent flags field
-  // (such as a runtime representation of an OpenTelemetry span), the high 22 bits MUST
-  // be set to zero.
-  //
-  // [Optional].
-  //
   // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
   //
   // Bits 8 and 9 represent the 3 states of whether a span's parent
   // is remote. The states are unknown, is not remote, is remote.
   // To read the 2-bit flag, use `flags & SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_MASK`.
   //
+  // When creating span messages, if the message is logically forwarded from another source
+  // with an equivalent flags fields (i.e., usually another OTLP span message), the field SHOULD
+  // be copied as-is. If creating from a source that does not have an equivalent flags field
+  // (such as a runtime representation of an OpenTelemetry span), the high 22 bits MUST
+  // be set to zero.
   // Readers MUST NOT assume that bits 10-31 (22 most significant bits) will be zero.
   //
   // [Optional].
@@ -266,18 +263,20 @@ message Span {
     // then no attributes were dropped.
     uint32 dropped_attributes_count = 5;
 
-    // Flags, a bit field. Bit 0-7 (8 least significant bits) are the trace
-    // flags as defined in W3C Trace Context specification. Readers
-    // MUST not assume that bits 10-31 (22 most significant bits) will be zero.
-    // When creating new spans, bits 10-31 (most-significant 22-bits) MUST be
-    // zero. To read the 8-bit W3C trace flag (use flags &
-    // SPAN_FLAGS_TRACE_FLAGS_MASK).  [Optional].
+    // Flags, a bit field. Bits 0-7 (8 least significant bits) are the trace
+    // flags as defined in W3C Trace Context specification. To read the 8-bit W3C trace
+    // flag, use `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
     //
     // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
     //
-    // Bits 8 and 9 represent the 3 states of whether the link
-    // is remote. The states are unknown, is not remote, is remote.
+    // Bits 8 and 9 represent the 3 states of whether the link is remote. The states
+    // are unknown, is not remote, is remote.
     // To read the 2-bit flag, use `flags & SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_MASK`.
+    //
+    // Readers MUST NOT assume that bits 10-31 (22 most significant bits) will be zero.
+    // When creating new spans, bits 10-31 (most-significant 22-bits) MUST be zero.
+    //
+    // [Optional].
     fixed32 flags = 6;
   }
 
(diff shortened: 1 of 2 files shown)


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -249,6 +249,12 @@ message Span {
   // An optional final status for this span. Semantically when Status isn't set, it means
   // span's status code is unset, i.e. assume STATUS_CODE_UNSET (code = 0).
   Status status = 15;
+
+  // parent_span_is_remote identifies whether a span's parent is remote or not.
+  // This helps determine whether a span can be considered an entry-point span.
```
Oberon00, line 254: I have never heard "entry-point span". The term I heard before for this was "local root span" (e.g. opentelemetry-specification#366).

estolfo: The term "entry-point span" was used in the OTEP; does it matter if this change matches the terminology used in the OTEP?

Oberon00: In general the OTEP wording is preferable, but here I would definitely prefer "local root span" as you will find many other issues around the OTel repos with that keyword. Alternatively, I would also be OK if you just added

```suggestion
  // This helps determine whether a span can be considered an entry-point span (also known as "local root span").
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -249,6 +249,12 @@ message Span {
   // An optional final status for this span. Semantically when Status isn't set, it means
   // span's status code is unset, i.e. assume STATUS_CODE_UNSET (code = 0).
   Status status = 15;
+
+  // parent_span_is_remote identifies whether a span's parent is remote or not.
+  // This helps determine whether a span can be considered an entry-point span (also known as "local root span").
+  // A span is an entry-point span if it has no parent (parent_span_id is empty),
+  // or if parent_span_is_remote is true.
```
tigrannajaryan, line 256: What's the interoperability story here? All old OTLP sources that are unaware of this field will emit data such that parent_span_is_remote will arrive as false to all new OTLP receivers that are expecting the field. Is this what we want? What is the behavior of the receivers going to be when every single span that arrives has parent_span_is_remote=false?

tigrannajaryan: I also suggest to include a link to https://github.com/open-telemetry/oteps/blob/main/text/0182-otlp-remote-parent.md in the comments for readers who want to know more but are not familiar with the concept.

tigrannajaryan (changes requested):
I am blocking this until we make 1.0 release. Ideally we don't want any changes other than bug fixes between the RC that we just did and the final release.

opentelemetry/proto/trace/v1/trace.proto
Oberon00: Good point.
Since proto3 now supports optional, probably we should mark the field as such.

tigrannajaryan: Not all Protobuf generators generate presence functions for optional fields. IIRC, Gogoproto used by Collector doesn't. This means the data passed through the Collector will be destroyed (a present parent_span_is_remote=false will be forwarded with the field removed). Because of this in the past we avoided relying on optional fields. Things may have changed recently since I am not longer involved in the Collector actively.

@open-telemetry/collector-maintainers do we support optional fields in pdata currently?

Oberon00: This is what the OTEP says BTW https://github.com/open-telemetry/oteps/blob/main/text/0182-otlp-remote-parent.md#internal-details:

> For backwards compatibility with older OTLP versions, the protobuf field should be nullable (true, false, or unspecified) and the opentelemetry-collector protogen code should provide an API that enables backend exporters to identify whether the field is set.

tigrannajaryan: I just checked how Collector supports the optionals and I think at least from Collector's perspective it was suboptimal to use them in OTLP:
- The Collector patches and replaces the optionals by oneof. (optional and oneof are wire compatible, which makes the patching possible).
- In memory oneof fields in Gogoproto are represented as an interface pointing to a struct. On 64bit arch that takes 16 bytes in the message struct plus an allocation for the struct containing the bool value (another 8 bytes on heap). This uses a ton of memory unnecessarily (at least 24 bytes) and has particularly bad performance due to the extra allocation.

Compare to enum:
- Same size as optional bool on the wire (for values of enum <127 and we only need 0..2). 
- In memory 4 bytes in the message struct (or at worst 8 bytes if int32 is padded to 64bits) and 0 allocations (so 3x or 6x less memory). Also much better performance because there is no extra allocation.

I am not familiar with how Protobuf Java code is generated. @Oberon00 you mentioned enum is an object. I wonder what the impact of that is on performance compared to optional boolean (which I imagine Protobuf would still need to model as something other than a primitive Boolean field in message's POD). If it is a wash between enum and optional in Java then I suggest we choose what is best for Go: stick to enums.

estolfo:
Hi @tigrannajaryan @Oberon00 @bogdandrutu can we reopen this discussion now that v1.0 is out?

carlosalberto:
(added it to the Spec call agenda next week)

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: I re-read the thread again and my recommendation is still to go with an enum. Have an "unspecified" value of 0 in the enum. If you need to differentiate between "false" and "unspecified" then make "false" another value. An obviously "true" needs to be another value of the enum.

One argument I see that can make me change my mind is if someone can demonstrate enums are very inefficient in languages other than Go (compared to optional boolean).

jack-berg: @tigrannajaryan in java we manually serialize our data from a custom in memory representation (i.e. one not based off generated classes) to the protobuf wire format. Either is fine for us.

I'm fine with enum, but want to call out that this will create a precedent for using enums to represent a tri-states instead of optional booleans. That's fine with me but I think there are valid use cases for optional (such as the min / max fields) and I hope we continue to use them when appropriate.

Oberon00: >  I think there are valid use cases for optional (such as the min / max fields) and I hope we continue to use them when appropriate.

If we already use optionals elsewhere, wouldn't that be an argument for using it here as well?

jack-berg: The other places are floating point numbers, so not possible to represent them with enums. A boolean tri-state is unique because it can be represented in multiple ways.

carolabadeer:
Hi @tigrannajaryan @Oberon00 @carlosalberto, any updates on this discussion?

opentelemetry/proto/trace/v1/trace.proto
carlosalberto: > I'm fine with enum, but want to call out that this will create a precedent for using enums to represent a tri-states instead of optional booleans.

Would be that bad? I'm wondering whether documenting this somewhere (pros and cons) would be enough?

AlexanderWert: @tigrannajaryan We just discussed this in the Specification SIG and would like to get this PR in soon. What's your opinion on the type question? Are you fine with the above proposal of using `enum` in this case?

Oberon00: It was @tigrannajaryan who suggested the enum, I think? IIUC, nobody objected against the enum, there was just a discussion if an optional boolean wouldn't be slightly better but that seems to be resolved now in favor of the enum.

Oberon00: @estolfo Can you please change the PR to use an enum instead of "optional"? I think we have an agreement to move forward with the enum approach.

tigrannajaryan:
Sorry for slow response, I was out for a couple weeks.

Re-reading the thread I don't see any new arguments, so I think we should go ahead with a tristate enum.

opentelemetry/proto/trace/v1/trace.proto
estolfo: Hi @Oberon00 I was also out for a few weeks. I'm back now and will make the change, thanks!

estolfo: I've defined an enum with
0 - unset
1 - false
2 - true

Is there a convention for representing tristate booleans as enums in protobuf definitions? Let me know if the enum should be adjusted.

jsuereth (approved):
Recommendation:

Can you add documentation for `Unset` denoting that older clients may not set this field, in which case how would one figure out whether ParentSpanIsRemote?

Something like: 

 "In this case, ParentSpanIsRemote can be derived by looking at the parent span and determining if it has a different `Resource`.  This value is expected from older clients."

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -249,6 +249,26 @@ message Span {
...
+
+    // The parent span is remote.
+    PARENT_SPAN_IS_REMOTE_TRUE = 2;
+  }
+
+  // parent_span_is_remote identifies whether a span's parent is remote or not.
+  // This helps determine whether a span can be considered an entry-point span.
+  // A span is an entry-point span if it has no parent (parent_span_id is empty),
+  // or if parent_is_remote is true.
+  ParentSpanIsRemote parent_span_is_remote = 16;
```
jack-berg, line 271: [This comment](https://github.com/open-telemetry/opentelemetry-proto/blob/81a296f9dba23e32d77f46d58c8ea4244a2157a6/opentelemetry/proto/trace/v1/trace.proto#L79) on span indicates the next available field is 17, which suggests 16 has previously been allocated. Should investigate this and use 17 / update the comment if necessary.

jack-berg: [It looks like](https://github.com/open-telemetry/opentelemetry-proto/commit/a9a9d064d9592e8a1454ed0d8026c0da7427468f) field 16 was previously used for `local_child_span_count`.

tigrannajaryan:
Another approach is to represent this as a 2-bitfield of a new int32 field, part of which will be also used by https://github.com/open-telemetry/opentelemetry-proto/pull/503

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -249,6 +249,26 @@ message Span {
   // An optional final status for this span. Semantically when Status isn't set, it means
   // span's status code is unset, i.e. assume STATUS_CODE_UNSET (code = 0).
   Status status = 15;
+
+  // ParentSpanIsRemote indicates whether the span's parent is remote (true / false) or
+  // if the field is unset.
+  enum ParentSpanIsRemote {
```
Oberon00, line 255: Either this should be made generic (OptionalBool with values UNSET, FALSE, TRUE) or it should get more concrete names for the enumerators (PARENT_SPAN_UNKNOWN_REMOTE, PARENT_SPAN_NOT_REMOTE, PARENT_SPAN_REMOTE?). Seeing that the more concrete names are a bit awkward to come up with, I'd slightly prefer the generic OptionalBool, but am fine with either.

Oberon00:
@tigrannajaryan I think we should not block this PR with this, but consider this refactoring _before_ the next release, which should probably wait until after both this PR and the flags PR(s) have been merged.

tigrannajaryan:
> @tigrannajaryan I think we should not block this PR with this, but consider this refactoring _before_ the next release, which should probably wait until after both this PR and the flags PR(s) have been merged.

I am not sure it is a good idea. This is a repository that is declared stable. We should minimize any churn here. An accidental release in an intermediate incorrect state can become a huge problem. It may be worth looking into using branches to prepare work for next releases instead of doing it on main branch.

Oberon00:
> An accidental release in an intermediate incorrect state can become a huge problem. It may be worth looking into using branches to prepare work for next releases instead of doing it on main branch.

The state would not be incorrect, unless you view that optimization as vital.
Stable applies only to releases, not any intermediate main branch state, right?

opentelemetry/proto/trace/v1/trace.proto
jack-berg: We discussed in the 8/22/23 spec SIG and concluded that 16 is the correct field number. Several field numbers including 16 and 17 where used for other purposes prior to the 0.7.0 version. However, 0.7.0 introduced several breaking changes that make it not possible to produce with any version prior to 0.7.0. The thinking is that we should effectively ignore that versions prior to 0.7.0 existed for the purposes of field numbers.

Relevant links:
- [remove local_child_span_count](https://github.com/open-telemetry/opentelemetry-proto/pull/107/files) - merged on  2020/02/19
- [v0.7.0 release](https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v0.7.0) - released on 2021/01/27

This means that the [comment](https://github.com/open-telemetry/opentelemetry-proto/blob/81a296f9dba23e32d77f46d58c8ea4244a2157a6/opentelemetry/proto/trace/v1/trace.proto#L79) indicating that 17 is the next available field is incorrect. The next available field is in fact 16.

jack-berg (approved):

tigrannajaryan (changes requested):
Blocking to wait for https://github.com/open-telemetry/opentelemetry-proto/pull/503 and use the introduced bit field.

carlosalberto:
Now that #503 has been merged, we can discuss this PR again. IIRC @tigrannajaryan wanted this information to exist in the newly added `flags` field in the mentioned PR, rather than exist as an independent `parent_is_remote` field. Could you confirm/clarify this please?

tigrannajaryan:
> Now that #503 has been merged, we can discuss this PR again. IIRC @tigrannajaryan wanted this information to exist in the newly added `flags` field in the mentioned PR, rather than exist as an independent `parent_is_remote` field. Could you confirm/clarify this please?

Yes, that was my suggestion. We can take 2 more bits from `flags` to represent the 3 states we need for `parent_is_remote`.

estolfo:
@tigrannajaryan I've updated this PR to define a 2-bit flag that can be used to indicate if the span's parent is remote. Let me know if something need to be adjusted about the definition. Thanks!

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -123,6 +123,10 @@ message Span {
   // [Optional].
   //
   // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
+  //
+  // Bits 8 and 9 are reserved for representing the 3 states of whether a span's parent
+  // is remote. The states are unknown, is not remote, is remote.
```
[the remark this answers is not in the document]
Oberon00: since this is a bit field, shouldn't we use 1 and 3 or 2 and 3, so that one bit represents "it is known whether remote or not" and the other represents the remoteness?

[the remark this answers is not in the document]
tigrannajaryan: @Oberon00 sure, that works.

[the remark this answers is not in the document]
estolfo: I've included an enum using bit 8 as unset/set and bit 9 as the remoteness.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -259,11 +266,11 @@ message Span {
...
     // flags as defined in W3C Trace Context specification. Readers
-    // MUST not assume that 24 most significant bits will be zero.
-    // When creating new spans, the most-significant 24-bits MUST be
-    // zero.  To read the 8-bit W3C trace flag (use flags &
+    // MUST not assume that bits 10-31 (22 most significant bits) will be zero.
+    // When creating new spans, bits 10-31 (most-significant 22-bits) MUST be
+    // zero. To read the 8-bit W3C trace flag (use flags &
     // SPAN_FLAGS_TRACE_FLAGS_MASK).  [Optional].
     //
     // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
```
estolfo, line 276: @tigrannajaryan Does a comment about the parent is remote flag go here too? I'm not so familiar with the span Link use case.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -109,20 +109,27 @@ message Span {
...
   // To read the 8-bit W3C trace flag, use `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
   //
   // When creating span messages, if the message is logically forwarded from another source
   // with an equivalent flags fields (i.e., usually another OTLP span message), the field SHOULD
   // be copied as-is. If creating from a source that does not have an equivalent flags field
-  // (such as a runtime representation of an OpenTelemetry span), the high 24 bits MUST
+  // (such as a runtime representation of an OpenTelemetry span), the high 22 bits MUST
   // be set to zero.
   //
   // [Optional].
```
estolfo, line 122: @tigrannajaryan, does this `[Optional]` reference refer to the flags overall or the trace flag?

opentelemetry/proto/trace/v1/trace.proto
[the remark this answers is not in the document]
Oberon00: After thinking about it, I think the flags field is currently underdefined or maybe even defined misleadingly: https://github.com/open-telemetry/opentelemetry-proto/pull/503

Depending on how it's meant to be defined, and also considering usage on Links, it might be confusing to use "parent" in the name of the flag, i.e. instead of SPAN_FLAGS_PARENT_IS_REMOTE we might need to call it SPAN_FLAGS_IS_REMOTE, if the flags generally come from the parent (since it's also SPAN_FLAGS_TRACE_FLAGS_MASK, and not SPAN_FLAGS_PARENT_TRACE_FLAGS_MASK)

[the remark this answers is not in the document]
tigrannajaryan: >it might be confusing to use "parent" in the name of the flag

This is about recording the fact in the _current_ span that its _parent_ span is _remote_ (the definition of "remote" AFAIK is that it describes an operation in a different service). I think using the word "parent" is important. It is the parent that is remote, not the current span.

opentelemetry/proto/trace/v1/trace.proto
yurishkuro: How would the code that creates a Link be able to indicate "has remote parent" information in `Link.flags`? Is that part of the AddLink API?

opentelemetry/proto/trace/v1/trace.proto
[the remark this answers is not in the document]
Oberon00: But the same is the case e.g. for the W3C sampled flag ... or is it not?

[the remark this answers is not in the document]
Oberon00: @yurishkuro That does not really make sense, and also contradicts the description of the PR where the flag was added, see https://github.com/open-telemetry/opentelemetry-proto/issues/507

opentelemetry/proto/trace/v1/trace.proto
AlexanderWert: @yurishkuro 
For a Link these flags would indicate whether the linked span is remote (which is strictly speaking not a `remote parent`). The `AddLink` API takes the `SpanContext` (of the span to link to). And the `SpanContext` has the `isRemote` API which indicates whether that SpanContext represents a remote span or not. And this information should be accessible from the exporters, right?

opentelemetry/proto/trace/v1/trace.proto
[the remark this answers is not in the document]
AlexanderWert: @Oberon00 see my comment in https://github.com/open-telemetry/opentelemetry-proto/issues/507#issuecomment-1727202313

opentelemetry/proto/trace/v1/trace.proto
carlosalberto: cc @Oberon00

opentelemetry/proto/trace/v1/trace.proto
estolfo: I would propose then for clarity that we have a separate enum for use for a Link. Otherwise, we could try to find a term that could refer to a parent and linked span and use one enum, though finding that term might be difficult.
```
enum SpanFlagsLinkedSpanIsRemote {
  // Older clients may not set this field, so 00 represents that this value
  // has not been set.
  SPAN_FLAGS_LINKED_SPAN_IS_REMOTE_UNKNOWN = 0;
  // The linked span is remote: 11
  SPAN_FLAGS_LINKED_SPAN_IS_REMOTE = 0x00000300;
  // The linked span is not remote: 01
  SPAN_FLAGS_LINKED_SPAN_IS_NOT_REMOTE = 0x00000100;
}
```

opentelemetry/proto/trace/v1/trace.proto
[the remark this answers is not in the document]
AlexanderWert: > But the same is the case e.g. for the W3C sampled flag ... or is it not?

@Oberon00 Do you agree that the discussion in #507 (mainly around the W3C trace flags) is not directly related to this PR? Here the two newly introduced bits really refer to describing the current span from which we want to know whether the `PARENT` is remote. So it MUST be `SPAN_FLAGS_PARENT_IS_REMOTE` (not SPAN_FLAGS_IS_REMOTE). Otherwise it would defeat the initial intent of what we are encoding into these two flags. 

Can we resolve this here?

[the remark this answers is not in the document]
Oberon00: I disagree that this is unrelated to #507, but I think there is now enough agreement on #507 about the semantics of this field that we can conclude that PARENT should be used in the name.

opentelemetry/proto/trace/v1/trace.proto
Oberon00: I just moved this around, I thought it referred to the field as a whole.

estolfo:
I've just made a change to add a mask and enum for interpreting bits 8 and 9 of a Link's flags instead of reusing the masks for the span's flags. Let me know if this is the preferred approach.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -109,20 +109,27 @@ message Span {
...
   //
   // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
+  //
+  // Bits 8 and 9 represent the 3 states of whether a span's parent
+  // is remote. The states are unknown, is not remote, is remote.
+  // To read the 2-bit flag, use `flags & SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_MASK`.
+  //
+  // [Optional].
+  //
+  // Readers MUST not assume that bits 10-31 (22 most significant bits) will be zero.
```
Oberon00, line 132:
```suggestion
  // Readers MUST NOT assume that bits 10-31 (22 most significant bits) will be zero.
  //
  // [Optional].
```

(for https://github.com/open-telemetry/opentelemetry-proto/pull/484/files#r1329956387, also uppercase the NOT in MUST NOT)

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -329,5 +340,42 @@ enum SpanFlags {
...
+  // Bits 10-31 are reserved for future use.
+}
+
+// SpanFlagsParentIsRemote represents constants used to interpret
+// bits 8 and 9 of the Span.flags field. These two bits represent whether
+// the span's parent is remote. Bit 8 represents whether the field has been
+// set and bit 9 represents whether the parent is remote.
+// Once the bits have been read using SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_MASK,
+// the following constants can be used to interpret them.
+enum SpanFlagsParentIsRemote {
```
Oberon00, line 357:
```suggestion
enum NonW3cSpanFlags {
```
How about this? Or do we want one enum per flag group? Could also be sensible.

Oberon00 (approved):

Oberon00:
@tigrannajaryan Your blocking review https://github.com/open-telemetry/opentelemetry-proto/pull/484#pullrequestreview-1591380897 is still active. If you want to dismiss it, you need to do it through the box at the bottom of the PR conversation that says "1 Change requested". Just submitting another review as "Comment" won't dismiss it (although submitting an Approve will)

tigrannajaryan:
> @tigrannajaryan Your blocking review [#484 (review)](https://github.com/open-telemetry/opentelemetry-proto/pull/484#pullrequestreview-1591380897) is still active. If you want to dismiss it, you need to do it through the box at the bottom of the PR conversation that says "1 Change requested". Just submitting another review as "Comment" won't dismiss it (although submitting an Approve will)

I keep the block deliberately to make sure we don't accidentally merge it since technically there is enough approvals.

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: I am not sure I like this approach. I think I prefer every group of bits to have its own enum. If we do what you suggest then future bits that we will carve out from the reserved bits will share the enum name and it will be easier to make a mistake and use the wrong flag.

opentelemetry/proto/trace/v1/trace.proto
jmacd, line 122:
```suggestion
  //
  // [Optional].
```

There are two `[Optional]` markers.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -329,5 +340,42 @@ enum SpanFlags {
...
+// the following constants can be used to interpret them.
+enum SpanFlagsLinkIsRemote {
+  // Older clients may not set this field, so 00 represents that this value
+  // has not been set.
+  SPAN_FLAGS_LINK_IS_REMOTE_UNKNOWN = 0;
+  // The linked span is remote: 11
+  SPAN_FLAGS_LINK_IS_REMOTE = 0x00000300;
+  // The linked span is not remote: 01
+  SPAN_FLAGS_LINK_IS_NOT_REMOTE = 0x00000100;
 }
```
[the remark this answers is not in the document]
jmacd: I agree. We can take this approach, I think, by changing the name from span flags to "context flags".
I would suggest names:

```
CONTEXT_FLAGS_REFERENCE_UNKNOWN = 0
CONTEXT_FLAGS_REFERENCE_IS_REMOTE = 1
CONTEXT_FLAGS_REFERENCE_IS_LOCAL = 2
```

With language to say that the Span.context_flags describe the span's context, where a reference must be to the parent. For Link.context_flags, the reference must be the link itself.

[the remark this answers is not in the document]
estolfo: My latest updates assumed we wanted to keep two enums.
I'm happy to merge them into one. Do we agree on the use of the terms "context" and "reference"?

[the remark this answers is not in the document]
jmacd: In the context of the discussion in https://github.com/open-telemetry/opentelemetry-proto/issues/507, where @Oberon00 asked about representing the parent-is-sampled property (e.g., w/ bit 10), I think we would:

- For spans, the trace-flags-sampled bit can be different from the parent-is-sampled bit
- For links, the trace-flags-sampled bit is meaningful, and the parent-is-sampled bit would not be meaningful.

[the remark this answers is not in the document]
tigrannajaryan: >Do we agree on the use of the terms "context" and "reference"?

I like a bit more the "PARENT_OR_LINK" prefixed used above. So we can maybe have:
```proto
enum SpanParentOrLinkIsRemote {
  SPAN_PARENT_OR_LINK_IS_REMOTE_UNKNOWN = 0;
  SPAN_PARENT_OR_LINK_IS_REMOTE = 0x00000300;
  SPAN_PARENT_OR_LINK_IS_NOT_REMOTE = 0x00000100;
}
```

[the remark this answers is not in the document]
estolfo: I used `SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_*` (with the additional `FLAGS` in the name for clarity)
Let me know if you think that's redundant.

jmacd:
Structurally, I am happy with this change. I would suggest naming changes as hinted above, making the new flags into context flags whether they are part of the span or the link, including:

`Span.flags` -> `Span.context_flags`
`Link.flags` -> `Link.context_flags`

and

```
enum ContextFlagsReferenceIsRemote {
  CONTEXT_FLAGS_REFERENCE_IS_REMOTE_UNKNOWN = 0;
  CONTEXT_FLAGS_REFERENCE_IS_REMOTE_TRUE = ...;
  CONTEXT_FLAGS_REFERENCE_IS_REMOTE_FALSE = ...;
}
```

tigrannajaryan:
> Structurally, I am happy with this change.

+1.


> making the new flags into context flags whether they are part of the span or the link, including:
> 
> `Span.flags` -> `Span.context_flags` `Link.flags` -> `Link.context_flags`

This will limit the bit field to be useable only for context-related information. What if we need to record something about the Span that has nothing to do with the context? The name also deviates from what we do for logs where it is called [flags](https://github.com/open-telemetry/opentelemetry-proto/blob/ac3242b03157295e4ee9e616af53b81517b06559/opentelemetry/proto/logs/v1/logs.proto#L183).

I would prefer to stay consistent with logs, keep the current name and make it useful for other future bitfields.

As for the the enum name suggestion I don't mind using the term "context reference" if that's what we prefer. So perhaps we can do this:

```proto
enum SpanFlagsContextReferenceIsRemote {
   SPAN_FLAGS_CONTEXT_REFERENCE_IS_REMOTE_UNKNOWN = 0;
   SPAN_FLAGS_CONTEXT_REFERENCE_IS_REMOTE_TRUE = ...;
   SPAN_FLAGS_CONTEXT_REFERENCE_IS_REMOTE_FALSE = ...;
}
```

And also rename `SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_MASK` to `SPAN_FLAGS_CONTEXT_REFERENCE_IS_REMOTE_MASK`.

If that's too long we can shorten `CONTEXT_REFERENCE` to `CONTEXT_REF` or simply to `CONTEXT`.

jmacd:
@tigrannajaryan thank you. I agree with your proposal, to keep the name "flags" and for names like `SPAN_FLAGS_CONTEXT_REFERENCE_IS_REMOTE_MASK`.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -109,20 +109,24 @@ message Span {
...
-  // MUST not assume that 24 most significant bits will be zero.
+  // Flags, a bit field. Bits 0-7 (8 least significant bits) are the trace
+  // flags as defined in W3C Trace Context specification.
   // To read the 8-bit W3C trace flag, use `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
   //
+  // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
+  //
+  // Bits 8 and 9 represent the 3 states of whether a span's parent
+  // is remote. The states are unknown, is not remote, is remote.
+  // To read the 2-bit flag, use `flags & SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_MASK`.
```
jmacd, line 120:
```suggestion
  // To read the 2-bit flag, use `flags & SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK`.
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -259,14 +263,20 @@ message Span {
...
-    // SPAN_FLAGS_TRACE_FLAGS_MASK).  [Optional].
+    // Flags, a bit field. Bits 0-7 (8 least significant bits) are the trace
+    // flags as defined in W3C Trace Context specification. To read the 8-bit W3C trace
+    // flag, use `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
     //
     // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
+    //
+    // Bits 8 and 9 represent the 3 states of whether the link is remote. The states
+    // are unknown, is not remote, is remote.
+    // To read the 2-bit flag, use `flags & SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_MASK`.
```
jmacd, line 274:
```suggestion
    // To read the 2-bit flag, use `flags & SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK`.
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -329,5 +339,26 @@ enum SpanFlags {
   // Bits 0-7 are used for trace flags.
   SPAN_FLAGS_TRACE_FLAGS_MASK = 0x000000FF;
 
-  // Bits 8-31 are reserved for future use.
+  // Bits 8 and 9 are used to indicate that the parent span or link span is remote.
+  // See SpanFlagsParentOrLinkIsRemote for the valid values.
+  SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_MASK = 0x00000300;
```
jmacd, line 344:
```suggestion
  SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK = 0x00000300;
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -329,5 +339,26 @@ enum SpanFlags {
...
+// Once the bits have been read using SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_MASK,
+// the following constants can be used to interpret them.
+enum SpanFlagsParentOrLinkIsRemote {
+  // Older clients may not set this field, so 00 represents that this value
+  // has not been set.
+  SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_UNKNOWN = 0;
+  // The parent or linked span is remote: 11
+  SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE = 0x00000300;
+  // The parent or linked span is not remote: 01
+  SPAN_FLAGS_PARENT_OR_LINK_IS_NOT_REMOTE = 0x00000100;
```
jmacd, line 363:
```suggestion
  SPAN_FLAGS_CONTEXT_IS_REMOTE_UNKNOWN = 0;
  // The parent or linked span is remote: 11
  SPAN_FLAGS_CONTEXT_IS_REMOTE = 0x00000300;
  // The parent or linked span is not remote: 01
  SPAN_FLAGS_CONTEXT_IS_NOT_REMOTE = 0x00000100;
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -329,5 +339,26 @@ enum SpanFlags {
...
+
+  // Bits 10-31 are reserved for future use.
+}
+
+// SpanFlagsParentOrLinkIsRemote represents constants used to interpret
+// bits 8 and 9 of the Spans.flags or Link.flags field. These two bits
+// represent whether the parent or link is remote. Bit 8 represents whether
+// the field has been set and bit 9 represents whether the parent or link is
+// remote.
+// Once the bits have been read using SPAN_FLAGS_PARENT_OR_LINK_IS_REMOTE_MASK,
```
jmacd, line 354:
```suggestion
// Once the bits have been read using SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK,
```

jmacd (commented):
Suggested naming changes.

estolfo:
@jmacd I was about to accept the suggestions but I don't see anywhere in the thread that we decided on `context` instead of `context_reference`. Was that decided in the spec meeting this week?

tigrannajaryan:
> @jmacd I was about to accept the suggestions but I don't see anywhere in the thread that we decided on `context` instead of `context_reference`. Was that decided in the spec meeting this week?

I was out last spec SIG meeting. Unless an explicit decision was made in the spec meeting, let's go with `context`.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -329,5 +339,26 @@ enum SpanFlags {
...
+// the following constants can be used to interpret them.
+enum SpanFlagsParentOrLinkIsRemote {
```
tigrannajaryan, line 356:
```suggestion
enum SpanFlagsContextIsRemote {
```

jmacd:
@estolfo and @tigrannajaryan 
The meeting had light attendance, and no explicit decisions were made. I was suggesting my preference, which is that we keep the names short. Thank you @estolfo for following up.

estolfo:
@jmacd and @tigrannajaryan I've completed all the changes based on the latest decisions. Please take a look when you have a chance, thanks!

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -109,20 +109,24 @@ message Span {
...
-  // flags as defined in W3C Trace Context specification. Readers
-  // MUST not assume that 24 most significant bits will be zero.
+  // Flags, a bit field. Bits 0-7 (8 least significant bits) are the trace
+  // flags as defined in W3C Trace Context specification.
   // To read the 8-bit W3C trace flag, use `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
   //
+  // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
+  //
+  // Bits 8 and 9 represent the 3 states of whether a span's parent
+  // is remote. The states are unknown, is not remote, is remote.
```
tigrannajaryan, line 119:
```suggestion
  // is remote. The states are (unknown, is not remote, is remote).
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -109,20 +109,24 @@ message Span {
   // field must be empty. The ID is an 8-byte array.
   bytes parent_span_id = 4;
 
-  // Flags, a bit field. 8 least significant bits are the trace
-  // flags as defined in W3C Trace Context specification. Readers
-  // MUST not assume that 24 most significant bits will be zero.
+  // Flags, a bit field. Bits 0-7 (8 least significant bits) are the trace
```
tigrannajaryan, line 112:
```suggestion
  // Flags, a bit field. 
  //
  // Bits 0-7 (8 least significant bits) are the trace
```

tigrannajaryan: Please also format the `Link` comments in a similar way.

jmacd (approved):

tigrannajaryan (approved):
LGTM.

I would like move reviews/approvals before we merge this.

tigrannajaryan:
@jack-berg @jsuereth @Oberon00 please take another look, this changed since you approved the first time.

jack-berg (approved):

AlexanderWert (approved):

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -329,5 +343,26 @@ enum SpanFlags {
...
+// bits 8 and 9 of the Spans.flags or Link.flags field. These two bits
+// represent whether the parent or link is remote. Bit 8 represents whether
+// the field has been set and bit 9 represents whether the parent or link is
+// remote.
+// Once the bits have been read using SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK,
+// the following constants can be used to interpret them.
+enum SpanFlagsContextIsRemote {
+  // Older clients may not set this field, so 00 represents that this value
+  // has not been set.
+  SPAN_FLAGS_CONTEXT_IS_REMOTE_UNKNOWN = 0;
```
pyohannes, line 363: Nit: This comment seems to suggest that the field is only set to `00` in case of older clients, however, I think it can also be `00` when it's not known whether the parent/link is remote?

```suggestion
  // Older client not supporting this flag, or not known whether parent or
  // linked span is remote: 00
  SPAN_FLAGS_CONTEXT_IS_REMOTE_UNKNOWN = 0;
```

pyohannes (approved):

carlosalberto (approved):

jsuereth (approved):

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -329,5 +343,26 @@ enum SpanFlags {
...
+}
+
+// SpanFlagsContextIsRemote represents constants used to interpret
+// bits 8 and 9 of the Spans.flags or Link.flags field. These two bits
+// represent whether the parent or link is remote. Bit 8 represents whether
+// the field has been set and bit 9 represents whether the parent or link is
+// remote.
+// Once the bits have been read using SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK,
+// the following constants can be used to interpret them.
+enum SpanFlagsContextIsRemote {
```
sfc-gh-bdrutu, line 360: Confused why duplicate the `SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK` value? Can we only have mask values in the `SpanFlags` enum?

sfc-gh-bdrutu (changes requested):

opentelemetry/proto/trace/v1/trace.proto
bogdandrutu, line 360: Do we need this extra enum? We agreed to offer only masks values, this is getting messier to have all possible values here and again my comment about mis-using them gets back to this.

bogdandrutu (dismissed):
I would like to revisit all the helpers we provide, can we stick with only the mask values?

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -329,5 +343,26 @@ enum SpanFlags {
...
+  // Older client not supporting this flag, or not known whether parent or
+  // linked span is remote: 00
+  SPAN_FLAGS_CONTEXT_IS_REMOTE_UNKNOWN = 0;
+  // The parent or linked span is remote: 11
+  SPAN_FLAGS_CONTEXT_IS_REMOTE = 0x00000300;
+  // The parent or linked span is not remote: 01
+  SPAN_FLAGS_CONTEXT_IS_NOT_REMOTE = 0x00000100;
```
bogdandrutu, line 367: Bit random choice, why not `00` - unknown, `01` - false, `10` - true?

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: @bogdandrutu We agreed that 0 values for enums should not be used since it is easy to misuse them. This was the PR we merged after that discussion: https://github.com/open-telemetry/opentelemetry-proto/pull/461

However, I don't think we said that other (non-zero) values of enums are also discouraged.

I am open to discussing but I think it useful to have this enum. The enum codifies what otherwise would be just comments. We need that information in some form otherwise unclear what are the values for the 2 bits.

bogdandrutu: Can we postpone adding this until we agree on that? Also this way we don't block the main functionality, and can always add the helper later.

tigrannajaryan: Sounds good to me. We can remove the enum for now and discuss it separately.

opentelemetry/proto/trace/v1/trace.proto
AlexanderWert: @bogdandrutu The idea was that the first bit indicates whether the value is set or not and the second bit represents `false`/`true`:

- `00` unknown
- `x1` 'known' 
  - `01` false
  - `11` true

> 00 - unknown, 01 - false, 10 - true

What would be the logic with that? How would that be better?

Does it really matter at the end? As long as we have three different values that follow _some_ logic it should be fine, right?

bogdandrutu: It may or may not be better, but feels more natural to have the first 3 values used. I agree that is fine, was curios about the logic behind, so just document this and it is fine.

Or actually based on this, you can have 2 bits, first that says if the second is "known". Then I feel that `00`, `10` (false) and `11`(true) makes more sense, and we can have 2 different masks `SPAN_FLAGS_CONTEXT_HAS_IS_REMOTE_MASK` and `SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK`

estolfo: @bogdandrutu see this comment https://github.com/open-telemetry/opentelemetry-proto/pull/484#discussion_r1329036599

jmacd: @bogdandrutu I think I agree. If we simply call this a 2-bit field, then we can actually use 4 values.

`00` - unknown legacy
`01` - context is not remote
`10` - context is remote
`11` - unknown, not legacy

I.e., we can distinguish unknown-remoteness from legacy-unspecified-remoteness.

For a consumer of this data inside the collector, say, I would expect to see first-class accessor methods anyway, e.g., `ContextRemoteness() RemotenessState` with no bit-fields to speak of, just the 4 potential states the two bits can have.

bogdandrutu: >For a consumer of this data inside the collector, say, I would expect to see first-class accessor methods anyway, e.g., ContextRemoteness() RemotenessState with no bit-fields to speak of, just the 4 potential states the two bits can have.

We will definitely do that and hide all the complexity to the user.

Oberon00: > unknown, not legacy

What would that mean? Do you propose to make isRemote a tristate in the OTel data model?

jmacd: @Oberon00 I was suggesting that it could be done in the future, if someone could find a reason. Do we always know if a context is or is not remote? Either way, I would like to unblock this PR and I am actually happy with it either way. I do not see a significant advantage either way, but if we're forced to look for pros/cons, this was all I could find.

opentelemetry/proto/trace/v1/trace.proto
tigrannajaryan: To make progress on this: @estolfo can you please remove the enum for now? We can then create a follow-up PR that adds the enum and discuss it separately.

estolfo: @tigrannajaryan I just watched the meeting recording. Thanks for pinging me on this. I've removed the enum and updated the comment to include a brief description.

tigrannajaryan:
@bogdandrutu PTAL.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -259,14 +265,22 @@ message Span {
...
+    // `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
     //
     // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
+    //
+    // Bits 8 and 9 represent the 3 states of whether the link is remote. The states
+    // are unknown, is not remote, is remote.
+    // To read the 2-bit flag, use `flags & SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK`.
+    //
+    // Readers MUST NOT assume that bits 10-31 (22 most significant bits) will be zero.
+    // When creating new spans, bits 10-31 (most-significant 22-bits) MUST be zero.
```
sfc-gh-bdrutu, line 281: Do we want to document which bit represents what (not by adding any enums, but just document)?

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -259,14 +265,22 @@ message Span {
...
-    // SPAN_FLAGS_TRACE_FLAGS_MASK).  [Optional].
+    // Flags, a bit field.
+    //
+    // Bits 0-7 (8 least significant bits) are the trace flags as defined in W3C Trace
+    // Context specification. To read the 8-bit W3C trace flag, use
+    // `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
     //
     // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
+    //
+    // Bits 8 and 9 represent the 3 states of whether the link is remote. The states
```
bogdandrutu, line 276: Do we want to document which bit represents what (not by adding any enums, but just document)?

tigrannajaryan: The bits are documented on line 346 and 347.

estolfo: @bogdandrutu can you please let us know if the documentation on lines 346 and 347 are sufficient? Thanks

carlosalberto: Ping @bogdandrutu

bogdandrutu: This is done, only one comment left to the flag.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -329,5 +343,10 @@ enum SpanFlags {
   // Bits 0-7 are used for trace flags.
   SPAN_FLAGS_TRACE_FLAGS_MASK = 0x000000FF;
 
-  // Bits 8-31 are reserved for future use.
+  // Bits 8 and 9 are used to indicate that the parent span or link span is remote.
+  // Bit 8 indicates whether the value is known. Bit 9 indicates whether the span
+  // or link is remote.
+  SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK = 0x00000300;
```
bogdandrutu, line 349: I would split this into 2 MASKs:

```
  // Bits 8 and 9 are used to indicate that the parent span or link span is remote.
  // Bit 8 (`HAS_IS_REMOTE`) indicates whether the value is known.
  // Bit 9 (`IS_REMOTE`) indicates whether the span or link is remote.
  SPAN_FLAGS_CONTEXT_HAS_IS_REMOTE_MASK = 0x00000100;
  SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK = 0x00000200;
```

Oberon00: I think we should merge this PR as-is. There has been far too much discussion already, and we are now (since some time already) fussing over details that are not really important. I do think the isRemote flag in the protocol has value for backends, I believe how exactly we represent the flag is not really important.

This particular detail can also be added later on (keeping the SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK  as the union of the two proposed flags and finding a different name for the proposed SPAN_FLAGS_CONTEXT_IS_REMOTE_MASK )

bogdandrutu: Per my previous comments, since we are not allowed to change enum names and these helpers are in my opinion a source of long debate, I proposed before, and repeat I am happier to not have them initially than adding them and be inconsistent with others. So because of that I would just repeat, I am ok to merge this without the helper enums and add them later, but I am not ok to have them inconsistent with other masks.

estolfo: Assuming the latest comments are advocating for, and agree on, having the mask split into two, I've made the change. Please take a look @bogdandrutu and @Oberon00

AlexanderWert:
@open-telemetry/specs-approvers I think this is ready to be merged!

tigrannajaryan:
> @open-telemetry/specs-approvers I think this is ready to be merged!

Needs at least 2 business days since last change before merging.

estolfo:
@open-telemetry/specs-approvers can we merge this please?

tigrannajaryan:
@estolfo looks like we forgot one more thing: an entry in the CHANGELOG. Please add and we can merge.

estolfo:
@tigrannajaryan ok done

tigrannajaryan (approved):

tigrannajaryan:
Thank you @estolfo and everyone for patience!

PR #488, opened 2023-06-27
Add profiles proto

petethepig:
This PR adds a new profiles signal. This is currently very much in a work-in-progress / seeking feedback kind of stage.

Relevant documents / previous discussions:
* [OTEL Profiling Vision](https://github.com/open-telemetry/oteps/blob/main/text/profiles/0212-profiling-vision.md)
* [Profiling SIG Meeting Notes](https://docs.google.com/document/d/1D-1MbT3MWpKM61JgVFx_TQi4vqWLLIZp6biK5MoHafI/edit?usp=sharing)
* [Draft PR in OTEL-collector](https://github.com/petethepig/opentelemetry-collector/pull/1)

---

The overall design is based on/heavily inspired by:
* other OTEL signals
* [Felix Geisendörfer's draft for a profiles proto](https://github.com/felixge/opentelemetry-proto/pull/1)
* [Anton Kolesnikov's draft for a profiles proto](https://gist.github.com/kolesnikovae/d61f6b668fc093eb40ee7e7a2a5afb64)
* [pprof proto](https://github.com/google/pprof/blob/main/proto/profile.proto)

This PR consists of 2 parts:
* The `profiles.proto` and `profiles_service.proto` which are mostly boilerplate and more or less finished from my perspective
* 3 alternative `Profile` implementations, still very much work-in-progress
  * `pprof` — included as a reference and as a baseline for future benchmarks
  * `denormalized` — most messages are embedded, no string table, a lot of potential for duplication => easier to work with, potentially larger payloads
  * `normalized` — most messages are linked using a foreign key, includes a string table, not much duplication => harder to work with, potentially smaller payloads

### Things that are currently supported:
* support for OTEL-compatible attributes (labels) on Profile and Samples level. Meaning you can label both whole profiles and individual stack traces
* linking with traces. TODO: do we need linking with other signals?
* support for multiple profile types in one `Profile` (similar to how it's implemented in pprof and used in go memory profiles via multiple `value` values)
* ability to specify `AggregationTemporality` for each `SampleType` (e.g something like `alloc_space` in go is cumulative)
* optional per-sample timestamps
* support for embedded original profile (`original_payload` field)

### Design TODOs / Areas where feedback would be most valuable:
* The biggest thing is we need to decide which one of these approaches (`denormalized` / `normalized` / mix of the two) we're going to take. I personally think that the optimal end product should be closer to `normalized` implementation, especially considering that on-the-wire efficiency is a [particularly important part of OTEL's profiling vision](https://github.com/open-telemetry/oteps/blob/main/text/profiles/0212-profiling-vision.md#standardize-profiling-data-model-for-industry-wide-sharing-and-reuse). But we still need to do some benchmarks to confirm that that would actually be significantly more efficient. 
* Figure out what to do with pprof messages like `Mapping` — I'm not sure if we need those
* Improve `SampleType` message. Currently `Type` and `Unit` are strings. We might want to turn them into enums, or standardize them some other way, e.g making those a part of [OTEL semantic conventions](https://opentelemetry.io/docs/concepts/semantic-conventions/)
* Things I'm forgetting / not thinking about

### Implementation TODOs / next steps
* get a corpus of some real production profiling data
* run benchmarks with this corpus of data -> fine-tune various aspects of the format
* implement a few simple OTEL processors -> fine-tune various aspects of the format
* create a reference implementation for a client -> fine-tune various aspects of the format
* create a reference implementation for a server -> fine-tune various aspects of the format
* verify that the spec works with those reference implementations -> fine-tune various aspects of the format

## Reference Diagrams

For reference, here are graph representations of other signals:

<details>

  <summary>metrics</summary>
  
metrics dot
</details>

<details>
  <summary>logs</summary>
  
logs png dot

</details>

<details>
  <summary>traces</summary>

  traces dot
</details>

---

And these are graph representations of alternatives formats under consideration:

<details>
  <summary>pprof for reference</summary>
  
pprof dot

</details>

<details>
  <summary>denormalized</summary>

denormalized dot
</details>

<details>
  <summary>normalized</summary>
  

normalized dot
</details>

---

/cc @tigrannajaryan @felixge @jhalliday @jsuereth @christos68k


tigrannajaryan:
@petethepig we have just discussed this in the TC meeting.

We would like to give you a separate repo in Otel github org, a fork of this one, where you can experiment freely with the profile protos. We will give you approving/merging rights on this new repo, so that you can also iterate faster and don't need to wait for approvals/merging from OTLP maintainers.

You can also fork Collector and easily change it to consume proto file files from your fork instead of this repo and do your experimentation in the Collector too.

When the experiments are concluded you can do a cleanup on your fork and then we will merge back the finalized version of profile protos back into this repo.

What do you think?

cc @jsuereth

opentelemetry/proto/profiles/v1/alternatives/normalized/normalized.proto
```diff
@@ -0,0 +1,141 @@
...
+  bytes span_id = 2;
+}
+
+message AttributeSet {
+  repeated opentelemetry.proto.common.v1.KeyValue attributes = 1;
+  uint32 dropped_attributes_count = 2;
+}
+
+message Sample {
+  repeated uint64 location_ids = 1;
```
cyriltovena, line 26: I wonder if those should be int32/uint32 I don't think we'll overflow this specially since they refer slice/array indexes.

This is not super important for the wire format since proto will varint encode it but for the collector this will allow it to buffer more as the memory footprint will be lower per profiles.

opentelemetry/proto/profiles/v1/profiles.proto
```diff
@@ -0,0 +1,133 @@
...
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
+  //
+  // This field is semantically required and it is expected that end_time >= start_time.
+  fixed64 start_time_unix_nano = 2;
+
+  // end_time_unix_nano is the end time of the profile.
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
+  //
+  // This field is semantically required and it is expected that end_time >= start_time.
+  fixed64 end_time_unix_nano = 3;
```
cyriltovena, line 102: nit: wouldn't be better to have a duration instead. I also wonder what happens for cumulative profile.

opentelemetry/proto/profiles/v1/profiles.proto
```diff
@@ -0,0 +1,133 @@
...
+  uint32 dropped_attributes_count = 5;
+
+  // This can be original pprof or jfr encoded profile.
+  bytes original_payload = 7;
+
+  // this is temporary until we pick one of these formats
+  // fields from one of these will be embedded in Profile message:
+  oneof alternative_profile {
+    opentelemetry.proto.profiles.v1.alternatives.pprof.Profile pprof = 8;
+    opentelemetry.proto.profiles.v1.alternatives.normalized.Profile normalized = 9;
```
cyriltovena, line 130: I'm waiting to see some benchmark, but I do think the `normalized` is the way to go.

With a lot of samples (we can expect ~10k per profiles) it avoid duplications. It will most likely have a smaller memory footprint and faster marshalling/unmarshalling.

It's also closer to pprof which is quite popular for profiling data.

opentelemetry/proto/profiles/v1/alternatives/pprof/pprof.proto
```diff
@@ -0,0 +1,221 @@
...
+  // Specifies the units of num.
+  // Use arbitrary string (for example, "requests") as a custom count unit.
+  // If no unit is specified, consumer may apply heuristic to deduce the unit.
+  // Consumers may also  interpret units like "bytes" and "kilobytes" as memory
+  // units and units like "seconds" and "nanoseconds" as time units,
+  // and apply appropriate unit conversions to these.
+  int64 num_unit = 4;  // Index into string table
+}
+
+message Mapping {
```
florianl, line 146: For postprocessing it might be great for the backend to have additional (optional) information. Additional information can be something like, if this mapping belongs to some interpreter.

Maybe something like:
```
repeated Label label = 11;
```

florianl (commented):
Thanks for the work @petethepig && @felixge  :pray:

opentelemetry/proto/profiles/v1/alternatives/pprof/pprof.proto
```diff
@@ -0,0 +1,221 @@
...
+// following conventions:
+//
+// - Consumers should treat unset optional fields as if they had been
+//   set with their default value.
+//
+// - When possible, measurements should be stored in "unsampled" form
+//   that is most useful to humans.  There should be enough
+//   information present to determine the original sampled values.
+//
+// - On-disk, the serialized proto must be gzip-compressed.
```
thomasdullien, line 27: Is there a good reason to use gzip, when pareto-improved compression methods are available? (e.g. algorithms that outperform gzip compression at less CPU usage at every level?)

opentelemetry/proto/profiles/v1/alternatives/pprof/pprof.proto
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// Each Sample records values encountered in some program
+// context. The program context is typically a stack trace, perhaps
+// augmented with auxiliary information like the thread-id, some
+// indicator of a higher level request being handled etc.
+message Sample {
+  // The ids recorded here correspond to a Profile.location.id.
+  // The leaf is at location_id[0].
+  repeated uint64 location_id = 1;
```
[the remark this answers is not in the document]
cyriltovena: I thought this ID would be to reference locations inside this profile. Are you thinking about global addressing ?

opentelemetry/proto/profiles/v1/alternatives/normalized/normalized.proto
```diff
@@ -0,0 +1,141 @@
...
+message Location {
+  // The id of the corresponding profile.Mapping for this location.
+  // It can be unset if the mapping is unknown or not applicable for
+  // this profile type.
+  uint64 mapping_id = 1;
+  // The instruction address for this location, if available.  It
+  // should be within [Mapping.memory_start...Mapping.memory_limit]
+  // for the corresponding mapping. A non-leaf address may be in the
+  // middle of a call instruction. It is up to display tools to find
+  // the beginning of the instruction if necessary.
```
thomasdullien, line 104: Display tools may not have the ability to find the beginning of the instruction unless the "in the middle" above is specified precisely.

opentelemetry/proto/profiles/v1/alternatives/denormalized/denormalized.proto
```diff
@@ -0,0 +1,131 @@
...
+  // It can be unset if the mapping is unknown or not applicable for
+  // this profile type.
+  Mapping mapping = 1;
+
+  // The instruction address for this location, if available.  It
+  // should be within [Mapping.memory_start...Mapping.memory_limit]
+  // for the corresponding mapping. A non-leaf address may be in the
+  // middle of a call instruction. It is up to display tools to find
+  // the beginning of the instruction if necessary.
+  uint64 address = 2;
```
thomasdullien, line 95: We need to specify *exactly* how deep into the call instruction the address is allowed to be, display tools do not have access to the assembly bytes and hence cannot do more than simple arithmetic to find the beginning of the instruction.

opentelemetry/proto/profiles/v1/alternatives/normalized/normalized.proto
```diff
@@ -0,0 +1,141 @@
...
+  // case the line information above represents one of the multiple
+  // symbols. This field must be recomputed when the symbolization state of the
+  // profile changes.
+  bool is_folded = 4;
+}
+
+// borrowed from pprof proto
+message Line {
+  // The id of the corresponding profile.Function for this line.
+  uint64 function_id = 1;
```
thomasdullien, line 125: Is this an index into the other array?

thomasdullien: Perhaps we should rename `_id` to `_index`?

opentelemetry/proto/profiles/v1/alternatives/pprof/pprof.proto
```diff
@@ -0,0 +1,221 @@
...
+
+  // Time of collection (UTC) represented as nanoseconds past the epoch.
+  int64 time_nanos = 9;
+  // Duration of the profile, if a duration makes sense.
+  int64 duration_nanos = 10;
+  // The kind of events between sampled occurrences.
+  // e.g [ "cpu","cycles" ] or [ "heap","bytes" ]
+  ValueType period_type = 11;
+  // The number of events between sampled occurrences.
+  int64 period = 12;
```
thomasdullien, line 85: Could the comment be clarified?

opentelemetry/proto/profiles/v1/profiles.proto
jhalliday: @cyriltovena Explain the math behind that 10k, please?

opentelemetry/proto/profiles/v1/alternatives/pprof/pprof.proto
```diff
@@ -0,0 +1,221 @@
...
+  // address, for example due to identical code folding by the linker. In that
+  // case the line information above represents one of the multiple
+  // symbols. This field must be recomputed when the symbolization state of the
+  // profile changes.
+  bool is_folded = 5;
+}
+
+message Line {
+  // The id of the corresponding profile.Function for this line.
+  uint64 function_id = 1;
```
thomasdullien, line 204: Is that an ID or an index?

opentelemetry/proto/profiles/v1/alternatives/pprof/pprof.proto
```diff
@@ -0,0 +1,221 @@
...
+  // Line number in source code.
+  int64 line = 2;
+}
+
+message Function {
+  // Unique nonzero id for the function.
+  uint64 id = 1;
```
thomasdullien, line 211: Is that an ID or an index?

opentelemetry/proto/profiles/v1/profiles.proto
```diff
@@ -0,0 +1,133 @@
+// Copyright 2019, OpenTelemetry Authors
+//
```
thomasdullien, line 2: Cut & paste issue about Copyright year.

thomasdullien (changes requested):
My instinct is also for the normalized.

I'd like to make sure that we always use the suffix `_index` for things that are array indices, and reserve `_id` for things that are likely hashes or other pseudorandom values. I'd also *strongly* push for using 128-bit values for `_id`s, always and everywhere, lest we create issues for everybody when collisions arise. 64-bit values collide randomly at data quantities that are absolutely the norm these days.

opentelemetry/proto/profiles/v1/alternatives/pprof/pprof.proto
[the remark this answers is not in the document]
florianl: Ah yes - this comment would also fit to `Location.id` just like https://github.com/open-telemetry/opentelemetry-proto/pull/488#discussion_r1246639511.

One of the motivations is really to avoid collisions in global addressing. Even if the same code is compiled, differences in the build systems/compilers are important and can make a difference.

opentelemetry/proto/profiles/v1/alternatives/pprof/pprof.proto
cyriltovena: index in that profile.

opentelemetry/proto/profiles/v1/alternatives/normalized/normalized.proto
```diff
@@ -0,0 +1,141 @@
...
+}
+
+message AttributeSet {
+  repeated opentelemetry.proto.common.v1.KeyValue attributes = 1;
+  uint32 dropped_attributes_count = 2;
+}
+
+message Sample {
+  repeated uint64 location_ids = 1;
+  repeated uint64 link_ids = 2;
```
petethepig, line 27: embed these in attributes?

opentelemetry/proto/profiles/v1/alternatives/normalized/normalized.proto
petethepig, line 26: Experiment with storing these as a tree

opentelemetry/proto/profiles/v1/alternatives/normalized/normalized.proto
```diff
@@ -0,0 +1,141 @@
...
+  repeated uint64 attribute_set_ids = 3;
+
+  // this one is repeated because there can be multiple profile kinds in one profile. Typical example is memory profiles in go that contain:
+  //   * alloc_objects
+  //   * alloc_bytes
+  //   * inuse_objects
+  //   * inuse_bytes
+  repeated fixed64 values = 4;
```
petethepig, line 35: byte-seconds might overflow when aggregated

petethepig: should not be fixed64

opentelemetry/proto/profiles/v1/alternatives/normalized/normalized.proto
```diff
@@ -0,0 +1,141 @@
...
+  repeated Location locations = 4;
+  repeated Function functions = 5;
+  repeated Link links = 6;
+  repeated AttributeSet attribute_sets = 7;
+  repeated string string_table = 8;
+}
+
+
+// borrowed from pprof proto
+message Mapping {
```
petethepig, line 70: Florian: should stay for profiles that don't have symbols

petethepig:
TODO: add a design decision log?

jsuereth:
@tigrannajaryan I created https://github.com/open-telemetry/opentelemetry-proto-profile for the profiling group to own and experiment with.

So far I only add @Rperry2174 as a maintainer.  I would add @petethepig but he's not a member of the OpenTelemetry group yet (please follow this: https://github.com/open-telemetry/community/blob/main/community-membership.md#community-membership, I can vouch for you as one of your two)

opentelemetry/proto/profiles/v1/alternatives/pprof/pprof.proto
petethepig: In pprof it is an ID. And there is some confusion right now because usually IDs and indexes are identical, but not always. The plan for the new spec is that we're going to remove ids and rely on indexes. This way there's one way to address objects and so there's going to be:
* less confusion
* better performance (array lookups are faster than search for ID)

opentelemetry/proto/profiles/v1/profiles.proto
petethepig: @christos68k Yeah, pprof was included as a reference so that we'd be able to easily compare it to other ones. Long term plan is to keep just one of these 3 formats, and based on the discussions so far it seems like it will be based on the `normalized` version.

opentelemetry/proto/profiles/v1/alternatives/normalized/normalized.proto
```diff
@@ -0,0 +1,141 @@
...
+  // The object this entry is loaded from.  This can be a filename on
+  // disk for the main binary and shared libraries, or virtual
+  // abstractions like "[vdso]".
+  int64 filename = 4;  // Index into string table
+  // A string that uniquely identifies a particular program version
+  // with high probability. E.g., for binaries generated by GNU tools,
+  // it could be the contents of the .note.gnu.build-id field.
+  int64 build_id = 5;  // Index into string table
+
+  // The following fields indicate the resolution of symbolic info.
```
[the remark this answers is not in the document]
petethepig: @aalexand Can you provide the list of combinations that are valid? I remember you mentioned that not all 16 combinations are valid, but also 4-way enum won't cut it either, so maybe we can make this an enum with more than 4 but less than 16 combinations.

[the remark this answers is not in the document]
aalexand: I think the combinations encountered in practice are these - see below.

| has_functions  | has_filenames | has_line_numbers | has_inline_frames | Description |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| true | false | false | false | A C++ / native binary with just symbol table (.symtab), no DWARF. |
| true | true | true | true | A C++ / native binary symbolized using profiler-friendly (i.e. at least `-gmlt`) DWARF. |
| true | true | true | false | Full symbol information but no distinction of inline frames - e.g. from Java. Can also correspond to an exotic case of using separate file DWARF, but not providing the *.dwp file. |
| false | false | false | false | Unsymbolized mapping |

I would appreciate if someone also familiar with the subject could verify this from their perspective.

tigrannajaryan:
@petethepig please re-create this PR in https://github.com/open-telemetry/opentelemetry-proto-profile

I am going to close this PR and you can continue discussing it in the new repo.

opentelemetry/proto/profiles/v1/alternatives/normalized/normalized.proto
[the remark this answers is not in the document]
petethepig: @aalexand thank you! I was trying to make an enum out of this, and here's what I have so far:
```proto
  enum SymbolicInfo {
    SYMBOLIC_INFO_UNSPECIFIED = 0;
    SYMBOLIC_INFO_FULL = 1;
    SYMBOLIC_INFO_FUNCTIONS_ONLY = 2;
    SYMBOLIC_INFO_NO_INLINE_FRAMES = 3;
  }
```

As I was doing this one thing I thought of is do we need these booleans / enums at all? My understanding is that these booleans are useful when combining a profile without symbols with symbols information (e.g in DWARF format). But I would assume that when you have symbols in some different format (let's say DWARF), you already know if it has functions / line numbers / etc. So it feels like maybe we don't need these booleans / enums. Unfortunately, my understanding of all of this is still pretty limited, so I would love your input on this.

---

Btw, I checked google/pprof repo, and it looks like out of these 4 booleans, the only one that's meaningfully used is [HasFunctions one](https://github.com/search?q=repo%3Agoogle%2Fpprof+HasFunctions&type=code), and even then there's only one mention (`symbolz.go`). The other fields are just copied back and forth but not actually used it seems like.

cc @korniltsev

opentelemetry/proto/profiles/v1/profiles.proto
petethepig: @cyriltovena 

I have 2 questions:
* Can you elaborate on why it would be better to have a duration? 
* Why is this going to be different for cumulative profiles? I'm not following

Btw I copied this from trace proto, so there's value in keeping like this for consistency.

opentelemetry/proto/profiles/v1/profiles.proto
```diff
@@ -0,0 +1,133 @@
...
+  // to the data in the "scope_profiles" field which have their own schema_url field.
+  string schema_url = 3;
+}
+
+// A collection of Profiles produced by an InstrumentationScope.
+message ScopeProfiles {
+  // The instrumentation scope information for the profiles in this message.
+  // Semantically when InstrumentationScope isn't set, it is equivalent with
+  // an empty instrumentation scope name (unknown).
+  opentelemetry.proto.common.v1.InstrumentationScope scope = 1;
```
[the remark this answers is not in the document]
petethepig: @florianl 

I copied this part from other signals, like traces or logs — so if this is truly a limitation then I think it should be discussed in the broader OTEL context.

I don't think it is a limitation though — [instrumentation scope describes 2 things](https://github.com/open-telemetry/opentelemetry-proto/blob/3decfa005db2885341c03843470681646d5ab239/opentelemetry/proto/common/v1/common.proto#L69-L81):
* instrumentation scope name and version
* attributes

I think it is intended that scope name and version would be the same for system-wide profile. RE attributes, the format in its current state is very flexible when it comes to assigning attributes, for example, clients can create multiple separate profiles within one `ScopeProfiles`, each with it's own set of attributes, allowing for "per application" attributes within one instrumentation scope.

opentelemetry/proto/profiles/v1/alternatives/pprof/pprof.proto
petethepig: https://github.com/petethepig/opentelemetry-proto/commit/980fc893bf5c0ebfe8bc9e77af2b36971f6ba581

I added it to `Location` as well. My logic here is that the same argument can be made for adding it to `Location`, so I figured it makes more sense to include it there. I made it a reference to an `AttributeSet`, same way the linking is done from `Sample`.

@florianl what do you think?

opentelemetry/proto/profiles/v1/alternatives/pprof/pprof.proto
petethepig: I think this is sample rate basically. I'm not going to edit this one because this is copied from pprof spec. I will make sure we document it well in the final otel proto spec though.

opentelemetry/proto/profiles/v1/alternatives/pprof/pprof.proto
petethepig: This is pprof proto and I only included pprof proto as an example.

Either way, from my perspective, compression is something that is broadly applicable to the whole OTEL / OTEL collector universe and therefore we should bring these kinds of things up to the broader OTEL collector maintainers group.

opentelemetry/proto/profiles/v1/alternatives/normalized/normalized.proto
```diff
@@ -0,0 +1,141 @@
...
+// Describes function and line table debug information.
+message Location {
+  // The id of the corresponding profile.Mapping for this location.
+  // It can be unset if the mapping is unknown or not applicable for
+  // this profile type.
+  uint64 mapping_id = 1;
+  // The instruction address for this location, if available.  It
+  // should be within [Mapping.memory_start...Mapping.memory_limit]
+  // for the corresponding mapping. A non-leaf address may be in the
+  // middle of a call instruction. It is up to display tools to find
```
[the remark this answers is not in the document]
petethepig: This comment comes from pprof proto. My understanding is that this comment refers to the limitation of some profilers where they are not always able to get exact instruction address for a given location. So it's less about what the implementer should put into this field and more of a word of caution for implementers of display tools. I might be totally wrong though — this is getting a little outside of my area of expertise, so I will need help from someone.

opentelemetry/proto/profiles/v1/alternatives/normalized/normalized.proto
petethepig: Yeah, this makes a lot more sense — changed it to `_index` and `_indices` everywhere.

PR #496, opened 2023-07-06
Add gauge subtype for prometheus unknown type

dashpole:
Part of https://github.com/open-telemetry/opentelemetry-specification/issues/3058

This an attempt at defining a "subtype" for OTel metric types, as [suggested at the 2/7 Spec SIG meeting](https://github.com/open-telemetry/opentelemetry-specification/issues/3058#issuecomment-1615998755).

This is the Proposed solution described in the proposal here: https://github.com/open-telemetry/opentelemetry-specification/issues/3058#issuecomment-1622740102.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -202,14 +202,14 @@ message Metric {
 message Gauge {
   repeated NumberDataPoint data_points = 1;
 
-  // sub_type provides additional type information for gauges.
+  // type provides additional type information for gauges.
   // 
   // When receiving from older producers of OTLP, it is safe to assume the
-  // zero-value for sub_type, GAUGE_TYPE_STANDARD. When sending OTLP to older
+  // zero-value for type, GAUGE_TYPE_STANDARD. When sending OTLP to older
   // consumers, it is safe for them to ignore the field. In both cases, the
-  // previous behavior without sub_type is maintained until producer and
+  // previous behavior without type is maintained until producer and
   // consumer are aware of the field.
-  GaugeType sub_type = 2;
+  GaugeType type = 2;
 }
 
 // Sum represents the type of a scalar metric that is calculated as a sum of all
@@ -347,19 +347,20 @@ enum DataPointFlags {
 }
 
 // GaugeType defines valid gauge "sub-types", which are used to describe
-// gauges which have additional properties. These sub-types generally apply
+// gauges which have additional properties. These types generally apply
 // to metrics which originate from protocols which support these additional
 // types, but are represented as gauges in OTLP.
 enum GaugeType {
-  // GAUGE_TYPE_STANDARD is the default GaugeType, which represents a gauge
-  // that does not have a sub_type. Since the OpenTelemetry API enforces types,
-  // this subtype is used for all metrics from OpenTelemetry instrumentation.
+  // GAUGE_TYPE_STANDARD is the default GaugeType, which represents a standard
+  // gauge. Since the OpenTelemetry API enforces types, this type is used for
+  // all metrics from OpenTelemetry instrumentation.
   GAUGE_TYPE_STANDARD = 0;
 
-  // GAUGE_TYPE_UNKNOWN is a GaugeType which represents a metric for
-  // which the type is not known, such as a Prometheus metric without a TYPE
-  // comment. For example, the metric may be a monotonically-increasing counter,
-  // or histogram, but type information was not included at the source.
+  // GAUGE_TYPE_UNKNOWN is a GaugeType which represents a metric for which the
+  // real type is not known, but is treated as a Gauge, such as a Prometheus
+  // metric without a TYPE comment. For example, the metric may be a
+  // monotonically-increasing sum, or histogram, but type information was not
+  // included at the source.
   GAUGE_TYPE_UNKNOWN = 1;
 }
 
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -201,6 +201,9 @@ message Metric {
 // "StartTimeUnixNano" is ignored for all data points.
 message Gauge {
   repeated NumberDataPoint data_points = 1;
+
+  // sub_type provides additional type information for gauges
+  GaugeSubType sub_type = 2;
 }
 
 // Sum represents the type of a scalar metric that is calculated as a sum of all
@@ -335,6 +338,22 @@ enum DataPointFlags {
   DATA_POINT_FLAGS_NO_RECORDED_VALUE_MASK = 1;
 
   // Bits 2-31 are reserved for future use.
+}
+
+// GaugeSubType defines valid gauge "sub-types", which are used to describe
+// "special" kinds of gauges which have additional properties. These sub-types
+// generally apply to metrics which originate from protocols which support
+// these additional types, but are represented as gauges in OTLP.
+enum GaugeSubType {
+  // GAUGE_SUBTYPE_NONE is the default GaugeSubType, which represents a gauge
+  // that does not have a sub_type.
+  GAUGE_SUBTYPE_NONE = 0;
+
+  // GAUGE_SUBTYPE_UNKNOWN is a GaugeSubType which represents a metric for
+  // which the type is not known, such as a Prometheus metric without a TYPE
+  // comment. For example, the metric may be a monotonically-increasing counter,
+  // or histogram, but type information was not included at the source.
+  GAUGE_SUBTYPE_UNKNOWN = 1;
 }
 
 // NumberDataPoint is a single data point in a timeseries that describes the
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -202,7 +202,13 @@ message Metric {
 message Gauge {
   repeated NumberDataPoint data_points = 1;
 
-  // sub_type provides additional type information for gauges
+  // sub_type provides additional type information for gauges.
+  // 
+  // When receiving from older producers of OTLP, it is safe to assume the
+  // zero-value for sub_type, GAUGE_SUBTYPE_NONE. When sending OTLP to older
+  // consumers, it is safe for them to ignore the field. In both cases, the
+  // previous behavior without sub_type is maintained until producer and
+  // consumer are aware of the field.
   GaugeSubType sub_type = 2;
 }
 
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -205,7 +205,7 @@ message Gauge {
   // sub_type provides additional type information for gauges.
   // 
   // When receiving from older producers of OTLP, it is safe to assume the
-  // zero-value for sub_type, GAUGE_TYPE_NONE. When sending OTLP to older
+  // zero-value for sub_type, GAUGE_TYPE_STANDARD. When sending OTLP to older
   // consumers, it is safe for them to ignore the field. In both cases, the
   // previous behavior without sub_type is maintained until producer and
   // consumer are aware of the field.
@@ -351,10 +351,10 @@ enum DataPointFlags {
 // to metrics which originate from protocols which support these additional
 // types, but are represented as gauges in OTLP.
 enum GaugeType {
-  // GAUGE_TYPE_NONE is the default GaugeType, which represents a gauge
+  // GAUGE_TYPE_STANDARD is the default GaugeType, which represents a gauge
   // that does not have a sub_type. Since the OpenTelemetry API enforces types,
   // this subtype is used for all metrics from OpenTelemetry instrumentation.
-  GAUGE_TYPE_NONE = 0;
+  GAUGE_TYPE_STANDARD = 0;
 
   // GAUGE_TYPE_UNKNOWN is a GaugeType which represents a metric for
   // which the type is not known, such as a Prometheus metric without a TYPE
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -352,7 +352,8 @@ enum DataPointFlags {
 // these additional types, but are represented as gauges in OTLP.
 enum GaugeSubType {
   // GAUGE_SUBTYPE_NONE is the default GaugeSubType, which represents a gauge
-  // that does not have a sub_type.
+  // that does not have a sub_type. Since the OpenTelemetry API enforces types,
+  // this subtype is used for all metrics from OpenTelemetry instrumentation.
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -195,6 +195,9 @@ message Metric {
 // "StartTimeUnixNano" is ignored for all data points.
 message Gauge {
   repeated NumberDataPoint data_points = 1;
+
+  // subtype provides additional type information for gauges
+  optional GaugeSubtype subtype = 2;
```
tigrannajaryan, line 200: Is `optional` necessary here? The zero-value for the enum GAUGE_SUBTYPE_NONE matches the current behavior, right? Do receivers need to be able whether `subtype` field was set on the wire?

tigrannajaryan:
```suggestion
  optional GaugeSubtype sub_type = 2;
```

tigrannajaryan:
@dashpole this also requires an "interoperability" explanation: how the new senders and old receivers should work such that nothing is broken.

opentelemetry/proto/metrics/v1/metrics.proto
dashpole: > Do receivers need to be able whether subtype field was set on the wire?

No, receivers don't need to be able to tell if it was set or not.

> Is optional necessary here? The zero-value for the enum GAUGE_SUBTYPE_NONE matches the current behavior, right? 

Thats correct.  I meant to ask for advice: Is it better to have an optional field without a NONE zero-value, or a non-optional field with a NONE zero-value?

tigrannajaryan: Optional fields are slower (at least in Go where they are pointed-to and must be heap-allocated). I would go with non-optional enum that has a zero-value of none.

dashpole:
> this also requires an "interoperability" explanation: how the new senders and old receivers should work such that nothing is broken.

If a new sender sends the new field, but the receiver does not know about it, ignoring it at the receiver is safe. This preserves the existing behavior today without the field.

If an old sender sends data without the new field to a new receiver, using the default value of NONE subtypes is safe.  It preserves existing behavior today without the field.

> what are the metric processors in the Collector supposed to do with the payloads that have these new fields set? If the processor is unaware of the new fields is existing processing logic still valid?

Processors themselves shouldn't need to do anything with this field, other than preserve it.  Pdata's [Gauge.CopyTo()](https://github.com/open-telemetry/opentelemetry-collector/blob/e1d0f265591d484e95eae942c0164c9b16046c8e/pdata/pmetric/generated_gauge.go#L49), and [Sum.CopyTo()](https://github.com/open-telemetry/opentelemetry-collector/blob/e1d0f265591d484e95eae942c0164c9b16046c8e/pdata/pmetric/generated_sum.go#L69) would need to copy this field.  Otherwise, I don't think any changes are required, and all existing processing logic should still be valid.

> When the Collector is updated to the new version of OTLP the fields will become available in the Collector and will be correctly passed through. However if a processor that knows nothing about these fields transforms a metric can it be a problem?

It won't be a problem.  One of the primary reasons why this subtype proposal is preferable to additional attributes is that it is entirely "opt-in" for existing users of OTLP, since the subtype is always safe to ignore unless you explicitly care about it.

opentelemetry/proto/metrics/v1/metrics.proto
dashpole: done, and I make all instances of "subtype" or "Subtype" two words (sub_type, sub-type, or SubType)

dashpole:
@jsuereth I updated the wording for info and stateset based on https://github.com/open-telemetry/opentelemetry-specification/issues/3058#issuecomment-1625428292 in the latest commit.  LMK what you think.  If you still have concerns, I can cut `info` and `stateset` from the proposal.

dashpole:
This now only contains the gauge subtype for unknown.  Given uncertainties around the future of OpenMetrics, it probably isn't the best idea to add those types right now.

This should resolve @jsuereth's concerns as well.

jsuereth (approved):
This LGTM.

However, as @tigrannajaryan comments on how older clients and new clients interact, can you add the details from your PR comment into the protocol as comments for interaction?


i.e. I think it's totally fine to require NEWER generated OTLP producers + receivers to understand these new types.   Effectively until you upgrade BOTH a producer of OTLP and a consumer of OTLP it will continue to treat these types the way they were previously.

That should be documented in the protocol itself for consumers, instead of just the PR comment.

dashpole:
> can you add the details from your PR comment into the protocol as comments for interaction?

Done.  Added:
```proto
  // When receiving from older producers of OTLP, it is safe to assume the
  // zero-value for sub_type, GAUGE_SUBTYPE_NONE. When sending OTLP to older
  // consumers, it is safe for them to ignore the field. In both cases, the
  // previous behavior without sub_type is maintained until producer and
  // consumer are aware of the field.
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -331,6 +340,22 @@ enum DataPointFlags {
...
 }
 
+// GaugeSubType defines valid gauge "sub-types", which are used to describe
+// "special" kinds of gauges which have additional properties. These sub-types
+// generally apply to metrics which originate from protocols which support
+// these additional types, but are represented as gauges in OTLP.
+enum GaugeSubType {
+  // GAUGE_SUBTYPE_NONE is the default GaugeSubType, which represents a gauge
+  // that does not have a sub_type.
+  GAUGE_SUBTYPE_NONE = 0;
```
jack-berg, line 350: @tigrannajaryan just want to confirm that this doesn't collide with any of the work you recently did to normalize enum values. For other enums in OTLP, we have the following patterns:
- `{ENUM_NAME}_UNSPECIFIED`
- `{ENUM_NAME}_UNSET`
- `{ENUM_NAME}_DO_NOT_USE`
I think its fine but wanted to double check since we're stable now.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -331,6 +340,22 @@ enum DataPointFlags {
...
+
+  // GAUGE_SUBTYPE_UNKNOWN is a GaugeSubType which represents a metric for
+  // which the type is not known, such as a Prometheus metric without a TYPE
+  // comment. For example, the metric may be a monotonically-increasing counter,
+  // or histogram, but type information was not included at the source.
+  GAUGE_SUBTYPE_UNKNOWN = 1;
```
jack-berg, line 356: What other values do we anticipate adding to this enum?

Do we anticipate expanding this concept to other point types?

arminru, line 356:
```suggestion
  GAUGE_SUBTYPE_PROMETHEUS_UNKNOWN = 1;
```
Would it make sense to indicate that this is not part of the OTel spec/datamodel/API but is specific only to Prometheus?

dashpole: See https://github.com/open-telemetry/opentelemetry-specification/issues/3058#issuecomment-1616061089 for my analysis of other protocols with types that aren't fully represented in OTel.  For example, it might make sense (with proper justification) to add collectd's `absolute` type, although i'm not aware of anyone that has asked for it, and it seems to itself be a "legacy" type it inherited from a tool.

I think the concept could be expanded to sums as well, but needs its own justification.  See the linked comment for types that would potentially make sense as sum subtypes.

dashpole: I would prefer to keep prometheus as an example of something that _could_ use the enum, but not make the enum value prometheus-specific.  I think OTLP should avoid fields that are specific to another protocol, even if we don't currently have plans to use them for other purposes.  For example, I wouldn't have wanted the Summary type to be called PrometheusSummary, even if the initial usage is only for Prometheus.  I'm no expert, but there are other "typeless" protocols, like carbon, which might want to use this field.

I agree it makes sense to clarify that this cannot be produced by the OpenTelemetry API.  I'll make the update.

I'm not sure what it would mean to say it isn't part of the OTel datamodel, but is part of OTLP, since OTLP [is part of](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#opentelemetry-protocol-data-model) the OTel datamodel.

dashpole: GAUGE_SUBTYPE_NONE now includes: `Since the OpenTelemetry API enforces types, this subtype is used for all metrics from OpenTelemetry instrumentation.`

arminru: For data coming from the OpenTelemetry API, I would not expect any untyped/unknown data to be allowed there even in the future. If there is even just a slim chance that it might be used for other use cases, however, then keeping the enum values neutral but adding a comment should be fine, yes. The clarification in 24a76be LGTM.

jack-berg:
FWIW, I prefer the string approach proposed in this [comment](https://github.com/open-telemetry/opentelemetry-specification/issues/3058#issuecomment-162274010). I think the chances of collision are small, and I value the flexibility provided by string (i.e. no need to have lengthly discussions for each new proposal) more than I fear the collision potential of strings.

dashpole:
@jack-berg I'm happy to make that change if there is consensus.  I'll raise it tomorrow at the spec SIG meeting to see if there is consensus on the approach.

pirgeo:
I understand that OTel will go ahead with adding these types. Keeping in mind the potential addition of a couple more metric types that can/should not be produced by the SDKs (https://github.com/open-telemetry/opentelemetry-specification/issues/3058#issuecomment-1616061089), has the usage of a separate metric type instead of using a subtype been discussed somewhere?

jmacd (approved):
I read through all the feedback. I agree that just adding an unknown flag here is "just enough" and any other subtypes (including sum subtypes) could be debated later.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -195,6 +195,15 @@ message Metric {
...
   repeated NumberDataPoint data_points = 1;
+
+  // sub_type provides additional type information for gauges.
+  // 
+  // When receiving from older producers of OTLP, it is safe to assume the
+  // zero-value for sub_type, GAUGE_SUBTYPE_NONE. When sending OTLP to older
+  // consumers, it is safe for them to ignore the field. In both cases, the
+  // previous behavior without sub_type is maintained until producer and
+  // consumer are aware of the field.
+  GaugeSubType sub_type = 2;
```
bogdandrutu, line 206: Not sure if "Sub" is needed here, since we call it "GaugeType" it is a type of Gauge/sub-type of the main type.

dashpole: Removed Sub from GaugeSubType and GAUGE_SUBTYPE

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -337,6 +346,23 @@ enum DataPointFlags {
...
 
+// GaugeType defines valid gauge "sub-types", which are used to describe
+// gauges which have additional properties. These sub-types generally apply
+// to metrics which originate from protocols which support these additional
+// types, but are represented as gauges in OTLP.
+enum GaugeType {
+  // GAUGE_TYPE_NONE is the default GaugeType, which represents a gauge
+  // that does not have a sub_type. Since the OpenTelemetry API enforces types,
+  // this subtype is used for all metrics from OpenTelemetry instrumentation.
+  GAUGE_TYPE_NONE = 0;
```
bogdandrutu, line 357: GAUGE_TYPE_NONE -> GAUGE_TYPE_STANDARD or something around this idea?

dashpole: I'm OK with that naming.  Agree it doesn't make sense to have a NONE, if it isn't a subtype anymore

dashpole: Updated.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -337,6 +346,23 @@ enum DataPointFlags {
...
+  // GAUGE_TYPE_STANDARD is the default GaugeType, which represents a gauge
+  // that does not have a sub_type. Since the OpenTelemetry API enforces types,
+  // this subtype is used for all metrics from OpenTelemetry instrumentation.
+  GAUGE_TYPE_STANDARD = 0;
+
+  // GAUGE_TYPE_UNKNOWN is a GaugeType which represents a metric for
+  // which the type is not known, such as a Prometheus metric without a TYPE
+  // comment. For example, the metric may be a monotonically-increasing counter,
+  // or histogram, but type information was not included at the source.
+  GAUGE_TYPE_UNKNOWN = 1;
```
bogdandrutu, line 363: // GAUGE_TYPE_UNKNOWN is a GaugeType which represents a metric for
  // which the real type is unknown is treated as a Gauge, such as a Prometheus
  // metric without a TYPE comment. For example, the metric may be a
  // monotonically-increasing sum, or histogram bucket counter, but type information
  // was not included at the source.

bogdandrutu (approved):

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -201,6 +201,15 @@ message Metric {
...
   repeated NumberDataPoint data_points = 1;
+
+  // sub_type provides additional type information for gauges.
+  // 
+  // When receiving from older producers of OTLP, it is safe to assume the
+  // zero-value for sub_type, GAUGE_TYPE_STANDARD. When sending OTLP to older
+  // consumers, it is safe for them to ignore the field. In both cases, the
+  // previous behavior without sub_type is maintained until producer and
+  // consumer are aware of the field.
+  GaugeType sub_type = 2;
```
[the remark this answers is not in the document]
dashpole: done

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -337,6 +346,23 @@ enum DataPointFlags {
...
 
+// GaugeType defines valid gauge "sub-types", which are used to describe
+// gauges which have additional properties. These sub-types generally apply
+// to metrics which originate from protocols which support these additional
+// types, but are represented as gauges in OTLP.
+enum GaugeType {
+  // GAUGE_TYPE_STANDARD is the default GaugeType, which represents a gauge
+  // that does not have a sub_type. Since the OpenTelemetry API enforces types,
+  // this subtype is used for all metrics from OpenTelemetry instrumentation.
+  GAUGE_TYPE_STANDARD = 0;
```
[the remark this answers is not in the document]
dashpole: done

opentelemetry/proto/metrics/v1/metrics.proto
dashpole: done

dashpole:
closing in favor of https://github.com/open-telemetry/opentelemetry-proto/pull/514

PR #499, opened 2023-07-12
Fix EOF and invalid character issue in metrics payload

ZaradarBH:
The metrics example uses "I'm" which causes havoc with curl on WSL2 resulting in a EOF error. Changing to "I`m" resolves this annoying issue :)

### DIFF
--- a/examples/metrics.json
+++ b/examples/metrics.json
@@ -29,7 +29,7 @@
             {
               "name": "my.counter",
               "unit": "1",
-              "description": "I`m a Counter",
+              "description": "I am a Counter",
               "sum": {
                 "aggregationTemporality": 1,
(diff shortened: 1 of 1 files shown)


examples/metrics.json
```diff
@@ -29,7 +29,7 @@
             {
               "name": "my.counter",
               "unit": "1",
-              "description": "I'm a Counter",
```
tigrannajaryan, line 32: Can we change simply to "I am a Counter"?

ZaradarBH: Done :)

reyang (approved):

tigrannajaryan (approved):

PR #503, opened 2023-08-04
Add W3C-specified trace flags to v1 Span proto

jmacd:
As discussed in https://github.com/open-telemetry/opentelemetry-specification/issues/3411 and https://github.com/open-telemetry/opentelemetry-proto/issues/382, the OpenTelemetry Span object does not record the trace flags that were provided in its Trace Context. 

For a tail sampler to sample spans using mechanism consistent with head sampling, the flags must be available in the protocol. W3C reserves 8 bits for these flags, of which two are defined. I've added 32-bit fields in the protocol for the span and the span link here.

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -76,7 +76,7 @@ message ScopeSpans {
 
 // A Span represents a single operation performed by a single component of the system.
 //
-// The next available field id is 18.
+// The next available field id is 17.
 message Span {
   // A unique identifier for a trace. All spans from the same trace share
   // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes OR
@@ -107,7 +107,7 @@ message Span {
   // There are two flags defined in the W3C Trace Context Level 2 spec:
   // - sampled = 0x1
   // - random = 0x2
-  uint32 trace_flags = 17;
+  uint32 trace_flags = 16;
 
   // A description of the span's operation.
   //
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -302,6 +302,11 @@ message Status {
 //   (span.flags & SPAN_FLAGS_TRACE_FLAGS_MASK)
 //
 // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
+//
+// Note that Span flags were introduced in version 1.1 of the
+// OpenTelemetry protocol.  Older Span producers do not set this
+// field, consequently consumers should not rely on the absence of a
+// particular flag bit to indicate the presence of a particular feature.
 enum SpanFlags {
   // The zero value for the enum. Should not be used for comparisons.
   // Instead use bitwise "and" with the appropriate mask as shown above.
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -112,9 +112,15 @@ message Span {
   // Flags, a bit field. 8 least significant bits are the trace
   // flags as defined in W3C Trace Context specification. Readers
   // MUST not assume that 24 most significant bits will be zero.
-  // When creating new spans, the most-significant 24-bits MUST be
-  // zero.  To read the 8-bit W3C trace flag (use flags &
-  // SPAN_FLAGS_TRACE_FLAGS_MASK).  [Optional].
+  // To read the 8-bit W3C trace flag, use `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
+  //
+  // When creating span messages, if the message is logically forwarded from another source
+  // with an equivalent flags fields (i.e., usually another OTLP span message), the field SHOULD
+  // be copied as-is. If creating from a source that does not have an equivalent flags field
+  // (such as a runtime representation of an OpenTelemetry span), the high 24 bits MUST
+  // be set to zero.
+  //
+  // [Optional].
   //
   // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
   fixed32 flags = 16;
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -106,11 +106,12 @@ message Span {
   // Flags, a bit field. 8 least significant bits are the trace flags as
   // defined in W3C Trace Context specification. 24 most significant bits are reserved
   // and must be set to 0. Readers must not assume that 24 most significant bits
-  // will be zero and must correctly mask the bits when reading 8-bit trace flag (use
-  // flags & SPAN_FLAGS_TRACE_FLAGS_MASK). [Optional].
+  // will be zero.  When writing new spans, SDKs must set the most-significant 24-bits
+  // to zero.  To read the 8-bit trace flag (use flags & SPAN_FLAGS_TRACE_FLAGS_MASK).
+  // [Optional].
   //
   // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
-  fixed32 flags = 16;
+  fixed32 flags = 18;
 
   // A description of the span's operation.
   //
@@ -249,8 +250,9 @@ message Span {
     // Flags, a bit field. 8 least significant bits are the trace flags as
     // defined in W3C Trace Context specification. 24 most significant bits are reserved
     // and must be set to 0. Readers must not assume that 24 most significant bits
-    // will be zero and must correctly mask the bits when reading 8-bit trace flag (use
-    // flags & SPAN_FLAGS_TRACE_FLAGS_MASK). [Optional].
+    // will be zero.  When writing new spans, SDKs must set the most-significant 24-bits
+    // to zero.  To read the 8-bit trace flag (use flags & SPAN_FLAGS_TRACE_FLAGS_MASK).
+    // [Optional].
     //
     // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
     fixed32 flags = 6;
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -76,7 +76,7 @@ message ScopeSpans {
 
 // A Span represents a single operation performed by a single component of the system.
 //
-// The next available field id is 17.
+// The next available field id is 18.
 message Span {
   // A unique identifier for a trace. All spans from the same trace share
   // the same `trace_id`. The ID is a 16-byte array. An ID with all zeroes OR
@@ -102,6 +102,12 @@ message Span {
   // The `span_id` of this span's parent span. If this is a root span, then this
   // field must be empty. The ID is an 8-byte array.
   bytes parent_span_id = 4;
+
+  // Flags from the W3C traceparent, if provided.
+  // There are two flags defined in the W3C Trace Context Level 2 spec:
+  // - sampled = 0x1
+  // - random = 0x2
+  uint32 trace_flags = 17;
 
   // A description of the span's operation.
   //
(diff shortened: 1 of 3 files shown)


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -103,6 +103,12 @@ message Span {
   // field must be empty. The ID is an 8-byte array.
   bytes parent_span_id = 4;
 
+  // Flags from the W3C traceparent, if provided.
+  // There are two flags defined in the W3C Trace Context Level 2 spec:
+  // - sampled = 0x1
+  // - random = 0x2
+  uint32 trace_flags = 17;
```
Oberon00, line 110: This needs to be optional or have some other indicator of whether it was present or not, since zero is a valid flag combination

jmacd: I would like to know in what sense zero is a meaningful combination of flags? I see no reason to make an optional zero, since the present-zero and absent-zero should have identical meanings.

Oberon00: Zero would mean that the span is not sampled. Such a span will not be sent in default configurations when using the OTel SDK, but in principle is valid (e.g. you might want to debug some sampler that throws away too many spans).

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -76,7 +76,7 @@ message ScopeSpans {
 
 // A Span represents a single operation performed by a single component of the system.
 //
-// The next available field id is 17.
+// The next available field id is 18.
```
kalyanaj, line 79: In the existing Span structure, I can't seem to find a field with id 16. If field id 16 is not being used, should we use field id 16 for this newly added field?

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -236,6 +242,12 @@ message Span {
     // dropped_attributes_count is the number of dropped attributes. If the value is 0,
     // then no attributes were dropped.
     uint32 dropped_attributes_count = 5;
+
+    // Flags from the W3C traceparent, if provided.
```
kalyanaj, line 246: Nit: Since these are the flags from the "linking" span's context, I wonder if this comment should say something like "Flags from the linking span's trace context, if provided.".

opentelemetry/proto/trace/v1/trace.proto
jmacd: Yeah the comment said I should use 17 so I did, but I agree that 16 looks available and is not declared `reserved`.

opentelemetry/proto/trace/v1/trace.proto
jmacd: Fixed.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -103,6 +103,12 @@ message Span {
   // field must be empty. The ID is an 8-byte array.
   bytes parent_span_id = 4;
 
+  // Flags from the W3C traceparent, if provided.
+  // There are two flags defined in the W3C Trace Context Level 2 spec:
```
jsuereth, line 107: Maybe just have a link to the specification rather than call out the two values (in case of changes).

It may also be worth calling out this field may be empty / zero in older clients.

jsuereth:
Isn't this a duplciate of #384?

jmacd:
> Isn't this a duplciate of https://github.com/open-telemetry/opentelemetry-proto/pull/384?

Oh, yes!

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -103,6 +103,12 @@ message Span {
...
+  // - sampled = 0x1
+  // - random = 0x2
+  uint32 trace_flags = 16;
```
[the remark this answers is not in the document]
tigrannajaryan: And we have a candidate PR that could use 2 more bits from this bitfield: https://github.com/open-telemetry/opentelemetry-proto/pull/484

opentelemetry/proto/trace/v1/trace.proto
jmacd: Added 
```
// Note that Span flags were introduced in version 1.1 of the
// OpenTelemetry protocol.  Older Span producers do not set this
// field, consequently consumers should not rely on the absence of a
// particular flag bit to indicate the presence of a particular feature.
```

jsuereth (approved):

tigrannajaryan (approved):

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -103,6 +103,15 @@ message Span {
   // field must be empty. The ID is an 8-byte array.
   bytes parent_span_id = 4;
 
+  // Flags, a bit field. 8 least significant bits are the trace flags as
+  // defined in W3C Trace Context specification. 24 most significant bits are reserved
+  // and must be set to 0. Readers must not assume that 24 most significant bits
```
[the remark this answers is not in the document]
tigrannajaryan: This is a good point. Maybe rephrase to "24 most significant bits are reserved and when creating a Span must be set to 0" to indicate that forwarders don't need to zero them?

jmacd:
@atshaw

jmacd:
(@atshaw this is on the topic you asked about in Sampling SIG https://github.com/open-telemetry/opentelemetry-specification/issues/2918 just now)

opentelemetry/proto/trace/v1/trace.proto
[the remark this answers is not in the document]
jmacd: The comment(s) now read:

    // Flags, a bit field. 8 least significant bits are the trace flags as
    // defined in W3C Trace Context specification. 24 most significant bits are reserved
    // and must be set to 0. Readers must not assume that 24 most significant bits
    // will be zero and must correctly mask the bits when reading 8-bit trace flag (use
    // flags & SPAN_FLAGS_TRACE_FLAGS_MASK). [Optional].
    //
    // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.

opentelemetry/proto/trace/v1/trace.proto
kalyanaj: Thanks @jmacd. In the latest iteration of this PR, I see it is " fixed32 flags = 18;", should this be using field id 16?

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -103,6 +103,16 @@ message Span {
...
+  // will be zero.  When writing new spans, SDKs must set the most-significant 24-bits
```
tigrannajaryan, line 109: Nit: technically Collector can create spans too, not only SDKs. So perhaps simply drop the reference to SDKs.

opentelemetry/proto/trace/v1/trace.proto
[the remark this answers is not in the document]
Oberon00: Just to confirm: This still requires every writer to clear the bits?

opentelemetry/proto/trace/v1/trace.proto
Oberon00: I think what the bits should be set to in new objects should not be described in the protocol.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -109,6 +109,16 @@ message Span {
...
 
+  // Flags, a bit field. 8 least significant bits are the trace flags as
+  // defined in W3C Trace Context specification. 24 most significant bits are reserved
+  // and must be set to 0. Readers must not assume that 24 most significant bits
+  // will be zero.  When creating new spans, the most-significant 24-bits MUST be
+  // zero.  To read the 8-bit W3C trace flag (use flags & SPAN_FLAGS_TRACE_FLAGS_MASK).
+  // [Optional].
+  //
+  // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
+  fixed32 flags = 16;
```
tigrannajaryan, line 120: For the record, why this is using field number 16 and not 17 as it is suggested on line 85 see this comment https://github.com/open-telemetry/opentelemetry-proto/pull/484/files#r1301972011

TLDR: field 16 was never used in a stable version of OTLP is is available for use.

carlosalberto (approved):

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -280,3 +300,28 @@ message Status {
...
+// expression like:
+//
+//   (span.flags & SPAN_FLAGS_TRACE_FLAGS_MASK)
+//
+// See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
+//
+// Note that Span flags were introduced in version 1.1 of the
+// OpenTelemetry protocol.  Older Span producers do not set this
+// field, consequently consumers should not rely on the absence of a
+// particular flag bit to indicate the presence of a particular feature.
```
jack-berg, line 317: @tigrannajaryan does this mean that flags is not a good option to solve #484? Or is the plan to use two bits to encode whether parent is remote?

tigrannajaryan: Use 2 bits, e.g. 00 - Unknown, 01 - IsNotRemote, 10 - IsRemote.

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -109,6 +109,16 @@ message Span {
   // field must be empty. The ID is an 8-byte array.
   bytes parent_span_id = 4;
 
+  // Flags, a bit field. 8 least significant bits are the trace
+  // flags as defined in W3C Trace Context specification. Readers
+  // MUST not assume that 24 most significant bits will be zero.
+  // When creating new spans, the most-significant 24-bits MUST be
```
Oberon00, line 115: I think this is confusing. The protocol is about serializing spans, so "creating new spans" can be understood as "creating new span messages". I think this sentence simply does not belong into the protocol definition. It seems to be assuming that the in-memory object representation of a span in an SDK also has this flags field, which may or may not be the case.

carlosalberto:
@Oberon00 We are planning to merge this soon. Please take a last look, etc ;)

kalyanaj (approved):

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -109,6 +109,16 @@ message Span {
...
+  // zero.  To read the 8-bit W3C trace flag (use flags &
+  // SPAN_FLAGS_TRACE_FLAGS_MASK).  [Optional].
```
Oberon00, line 117: I think "when creating new spans" may still be a bit confusing in the protocol spec as it could be understood as "creating new span messages", especially since the remainder seems to assume that "new spans" have a 32 bit flags field. I could not find a shorter unambiguous way than this:

```suggestion
  // MUST not assume that 24 most significant bits will be zero.
  // To read the 8-bit W3C trace flag (use flags & SPAN_FLAGS_TRACE_FLAGS_MASK).
  // When creating span messages, if the message is logically forwarded from another source
  // with an equivalent flags fields (i.e., usually another OTLP span message), the field SHOULD
  // be copied as-is. If creating from a source that does not have an equivalent flags field
  // (such as a runtime representation of an OpenTelemetry span), the high 24 bits MUST
  // be set to zero.
  // [Optional].
```

Oberon00: Also, there seems to be a typo in:

> To read the 8-bit W3C trace flag (use flags & SPAN_FLAGS_TRACE_FLAGS_MASK).

I assumed it should mean:

> To read the 8-bit W3C trace flag, use `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.

Oberon00 (approved):

opentelemetry/proto/trace/v1/trace.proto
Oberon00: I think we forgot to apply an analogous change to the flags field in the link object.

PR #510, opened 2023-10-27
Fix the incorrect histogram example in metrics.json

songy23:
There are 2 buckets (-inf, 1), [1, +inf) but 3 bucket counts.

### DIFF
--- a/examples/metrics.json
+++ b/examples/metrics.json
@@ -81,11 +81,11 @@
                   {
                     "startTimeUnixNano": 1544712660300000000,
                     "timeUnixNano": 1544712660300000000,
-                    "count": 3,
-                    "sum": 3,
-                    "bucketCounts": [1,1,1],
+                    "count": 2,
+                    "sum": 2,
+                    "bucketCounts": [1,1],
                     "explicitBounds": [1],
-                    "min": 1,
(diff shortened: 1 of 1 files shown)


examples/metrics.json
```diff
@@ -81,11 +81,11 @@
...
-                    "count": 3,
-                    "sum": 3,
-                    "bucketCounts": [1,1,1],
+                    "count": 2,
+                    "sum": 2,
+                    "bucketCounts": [1,1],
                     "explicitBounds": [1],
-                    "min": 1,
+                    "min": 0,
                     "max": 1,
```
jack-berg, line 89: If the count is 2, and the min is zero, then one of the two meaurements is zero. Therefore for the sum to be 2, the other measurement must be 2.

songy23: Good catch, fixed.

jack-berg (approved):

tigrannajaryan:
@open-telemetry/specs-metrics-approvers please review.

MrAlias (approved):

cijothomas (approved):

songy23 (commented):
Thanks for the reviews, this should be good to go

PR #514, opened 2023-11-29
Add metric.metadata for supporting additional metadata on metrics

dashpole:
Part of https://github.com/open-telemetry/opentelemetry-specification/issues/3058

Alternative to #496.

Language is borrowed from scope attributes, with the addition of: `Attributes are non-identifying.`

The primary motivation is to support the prometheus unknown type without explicitly adding it to the data model.

This provides a place for additional information about metrics to be provided that isn't associated with individual data points.  It is non-identifying to ensure it can be safely ignored by exporters if they don't support them.

Examples of problems this could be used to solve (not proposed, just examples):

* **Support additional types**: prometheus.type = "unknown", or statsd.type = "set"
* **Support vendor-specific features**: gcp.value_type = "bool"
    * GCP currently solves this by using a "special" unit.
* **Support additional features on top-of OTel**: e.g. k8s.stability_level = "beta"
    * K8s currently encodes stability level in the description.

Alternatives:

* #496
    * Pros: An enum is a smaller change.  Additional attributes may cause confusion if users are looking for data point attributes.
    * Cons: Enum solves fewer use-cases, and would likely need follow-up changes for other additional types.  Adding attributes (proposed here) would likely mean fewer changes to the proto in the future to support new classes of uses-cases.
* `string` instead of `repeated opentelemetry.proto.common.v1.KeyValue.
    * Cons: A string would make it harder to use
* `attributes` on `Gauge` instead of on `Metric`.
    * Pros: Meets primary use-case (prometheus unknown)
    * Cons: Likely (?) to be extended to other data types eventually, which would result in a slightly more complex protocol.
* `attributes` on `DataPoint` instead of on `Metric`.
    * Cons: Duplicates functionality of existing `attributes` field of `DataPoint`, and would create confusion over which to use.

@bogdandrutu @jack-berg @jsuereth @jmacd @tigrannajaryan

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -196,7 +196,7 @@ message Metric {
   // for lossless roundtrip translation to / from another data model.
   // Attribute keys MUST be unique (it is not allowed to have more than one
   // attribute with the same key).
-  repeated opentelemetry.proto.common.v1.KeyValue metadata_attributes = 12;
+  repeated opentelemetry.proto.common.v1.KeyValue metadata = 12;
 }
 
 // Gauge represents the type of a scalar metric that always exports the
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -191,7 +191,7 @@ message Metric {
 
   // Additional metadata attributes that describe the metric. [Optional].
   // Attributes are non-identifying.
-  // Consumers SHOULD not need to be aware of these attributes.
+  // Consumers SHOULD NOT need to be aware of these attributes.
   // These attributes MAY be used to encode information allowing
   // for lossless roundtrip translation to / from another data model.
   // Attribute keys MUST be unique (it is not allowed to have more than one
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -189,14 +189,14 @@ message Metric {
     Summary summary = 11;
   }
 
-  // Additional attributes that describe the metric. [Optional].
+  // Additional metadata attributes that describe the metric. [Optional].
   // Attributes are non-identifying.
   // Consumers SHOULD not need to be aware of these attributes.
   // These attributes MAY be used to encode information allowing
   // for lossless roundtrip translation to / from another data model.
   // Attribute keys MUST be unique (it is not allowed to have more than one
   // attribute with the same key).
-  repeated opentelemetry.proto.common.v1.KeyValue attributes = 12;
+  repeated opentelemetry.proto.common.v1.KeyValue metadata_attributes = 12;
 }
 
 // Gauge represents the type of a scalar metric that always exports the
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -191,6 +191,9 @@ message Metric {
 
   // Additional attributes that describe the metric. [Optional].
   // Attributes are non-identifying.
+  // Consumers SHOULD not need to be aware of these attributes.
+  // These attributes MAY be used to encode information allowing
+  // for lossless roundtrip translation to / from another data model.
   // Attribute keys MUST be unique (it is not allowed to have more than one
   // attribute with the same key).
   repeated opentelemetry.proto.common.v1.KeyValue attributes = 12;
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -188,6 +188,12 @@ message Metric {
     ExponentialHistogram exponential_histogram = 10;
     Summary summary = 11;
   }
+
+  // Additional attributes that describe the metric. [Optional].
+  // Attributes are non-identifying.
+  // Attribute keys MUST be unique (it is not allowed to have more than one
+  // attribute with the same key).
+  repeated opentelemetry.proto.common.v1.KeyValue attributes = 12;
 }
 
 // Gauge represents the type of a scalar metric that always exports the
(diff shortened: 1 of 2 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -188,6 +188,12 @@ message Metric {
     ExponentialHistogram exponential_histogram = 10;
     Summary summary = 11;
   }
+
+  // Additional attributes that describe the metric. [Optional].
+  // Attributes are non-identifying.
+  // Attribute keys MUST be unique (it is not allowed to have more than one
+  // attribute with the same key).
+  repeated opentelemetry.proto.common.v1.KeyValue attributes = 12;
```
tigrannajaryan, line 196: What's the reason this field is here and not in the `*DataPoint` messages?

dashpole: `*DataPoint` messages already contain attributes, so a second set of attributes seems a bit odd.  More importantly, we are interested describing things about the metric, or aggregation (e.g. gauge, sum, etc).  My primary use case, representing the "prometheus type", is defined for an entire prometheus `MetricFamily`, which corresponds to the `Metric` in OTel.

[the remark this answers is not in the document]
dashpole: added both of these.

tigrannajaryan: Make sense.

jack-berg (approved):

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -188,6 +188,15 @@ message Metric {
...
   }
+
+  // Additional attributes that describe the metric. [Optional].
+  // Attributes are non-identifying.
+  // Consumers SHOULD not need to be aware of these attributes.
+  // These attributes MAY be used to encode information allowing
+  // for lossless roundtrip translation to / from another data model.
+  // Attribute keys MUST be unique (it is not allowed to have more than one
+  // attribute with the same key).
+  repeated opentelemetry.proto.common.v1.KeyValue attributes = 12;
```
tigrannajaryan, line 199: Do we expect that these will use attribute names from the same namespace as all other attributes? For product- or protocol-specific attributes we can just follow the regular attribute naming recommendation to use reverse FQDN, e.g. `io.prometheus.metric_type=unknown`.
This way we can safely use these attributes for any other orthogonal purposes knowing there won't be conflicts.

tigrannajaryan: >We might want to loosen the language to include "OSS Projects" if prometheus should be using a reverse fqdn. Right now, it seems to imply only individual companies need a reverse fqdn.

Yes, definitely. Any organization, whether a company or an OSS project.

>Do you think that should be documented in the proto?

I don't think we document this for any other attributes, so I am inclined to say no, unless we also document similarly other attributes fields. On the other hand all other attributes originate in the API/SDK, so it is clear that spec's attribute conventions apply to them, so perhaps for this one we should explicitly document it needs to follow the same rules as everything else. I am not sure what's the best way.

dashpole: I'm ok excluding it for now, unless others see a need.  It seems clear enough to me that attribute naming applies to all otel attributes.

tigrannajaryan (approved):

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
jmacd: How about just "metadata"?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -188,6 +188,15 @@ message Metric {
...
   }
+
+  // Additional metadata attributes that describe the metric. [Optional].
+  // Attributes are non-identifying.
+  // Consumers SHOULD not need to be aware of these attributes.
+  // These attributes MAY be used to encode information allowing
+  // for lossless roundtrip translation to / from another data model.
+  // Attribute keys MUST be unique (it is not allowed to have more than one
+  // attribute with the same key).
+  repeated opentelemetry.proto.common.v1.KeyValue metadata_attributes = 12;
```
jack-berg, line 199: To avoid potential confusion about misinterpreting these with attributes, how about:
```suggestion
  repeated opentelemetry.proto.common.v1.KeyValue metadata = 12;
```
or 
```suggestion
  repeated opentelemetry.proto.common.v1.KeyValue properties = 12;
```

carlosalberto (approved):

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
tigrannajaryan: Sorry for more bikeshedding, but what if we call this `non_identifying_attributes` (? I don't like just the abstract `metadata` without any qualifiers. It is not clear what data this metadata is about (and metadata by definition must be about some other data).

reyang (approved):

opentelemetry/proto/metrics/v1/metrics.proto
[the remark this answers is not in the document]
reyang: +1 on "metadata", it might be a good thing for traces/logs as well

jmacd (approved):

jsuereth (approved):

carlosalberto:
It seems we are good to go, specially after 1.1.0 was released. @bogdandrutu maybe a last look?

carlosalberto:
I think we are ready to merge (briefly mentioned this in the Spec call a week or two ago, and no reviews ever since). 

cc @bogdandrutu

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -188,6 +188,15 @@ message Metric {
     ExponentialHistogram exponential_histogram = 10;
     Summary summary = 11;
   }
+
+  // Additional metadata attributes that describe the metric. [Optional].
+  // Attributes are non-identifying.
+  // Consumers SHOULD not need to be aware of these attributes.
```
arminru, line 194:
```suggestion
  // Consumers SHOULD NOT need to be aware of these attributes.
```

jmacd (approved):

arminru (approved):

PR #516, opened 2023-12-15
fix: OTLP JSON examples

bripkens:
# Why
The example files do not follow the specification. 64 bit integers must be [string
encoded](https://protobuf.dev/programming-guides/proto3/#json).

# What

 - Fix the encoding in the examples
 - Add an example for log `severityNumber` to point at the enum encoding rules (to help folks with cases [like these](https://github.com/open-telemetry/opentelemetry-proto/issues/517)).

### DIFF
--- a/examples/logs.json
+++ b/examples/logs.json
@@ -27,8 +27,9 @@
           },
           "logRecords": [
             {
-              "timeUnixNano": 1544712660300000000,
-              "observedTimeUnixNano": 1544712660300000000,
+              "timeUnixNano": "1544712660300000000",
+              "observedTimeUnixNano": "1544712660300000000",
+              "severityNumber": 10,
               "severityText": "Information",
               "traceId": "5B8EFFF798038103D269B633813FC60C",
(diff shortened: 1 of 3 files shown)


examples/logs.json
```diff
@@ -27,8 +27,9 @@
           },
           "logRecords": [
             {
-              "timeUnixNano": 1544712660300000000,
```
tigrannajaryan, line 30: The numbers [are acceptable when decoding](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#json-protobuf-encoding) 64 bit int fields (so not completely wrong as an example), but I agree it is better to show examples that match our encoding requirements too.

tigrannajaryan (approved):
The numbers for in64 fields are acceptable when decoding (so not completely wrong as an example), but I agree it is better to show examples that match our encoding requirements too.

jsuereth (approved):

bogdandrutu (approved):

Issue #518, opened 2023-12-15
Improve metrics.proto by mentioning Resource Attributes and Scope


krisztianfekete:
I think [metrics.proto](https://github.com/open-telemetry/opentelemetry-proto/blob/ea449ae0e9b282f96ec12a09e796dbb3d390ed4f/opentelemetry/proto/metrics/v1/metrics.proto#L100-L106) would benefit from listing [Resource Attributes and Scope](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#opentelemetry-protocol-data-model).

I am writing a blog on a related topic and would like to cross-reference the illustration in the proto file with the `logging` exporter.

I also added `attribute` to the key-value pairs.

I am thinking of something like this:
```console
//     Metric
//  +--------------------+
//  |name                |
//  |resource attributes |
//  |description         |
//  |unit                |     +------------------------------------+
//  |data                |---> |Gauge, Sum, Histogram, Summary, ... |
//  +--------------------+     +------------------------------------+
//
//    Data [One of Gauge, Sum, Histogram, Summary, ...]
//  +-----------+
//  |...        |  // Metadata about the Data.
//  |points     |--+
//  +-----------+  |
//                 |      +---------------------------+
//                 |      |DataPoint 1                |
//                 |      |                           |
//                 |      |attribute                  |
//                 v      |+------+------+   +------+ |
//              +-----+   ||label |label |...|label | |
//              |  1  |-->||value1|value2|...|valueN| |
//              +-----+   |+------+------+   +------+ |
//              |  .  |   |+-----+                    |
//              |  .  |   ||value|                    |
//              |  .  |   |+-----+                    |
//              |  .  |   +---------------------------+
//              |  .  |                   .
//              |  .  |                   .
//              |  .  |                   .
//              |  .  |   +---------------------------+
//              |  .  |   |DataPoint M                |
//              |  .  |   |                           |
//              |  .  |   |attribute                  |
//              +-----+   |+------+------+   +------+ |
//              |  M  |-->||label |label |...|label | |
//              +-----+   ||value1|value2|...|valueN| |
//                        |+------+------+   +------+ |
//                        |+-----+                    |
//                        ||value|                    |
//                        |+-----+                    |
//                        +---------------------------+
```
Happy to open a PR as well. What do you think?

tigrannajaryan:
+1. I think it would be useful. Please use the same ASCII chart style for consistency.

PR #519, opened 2023-12-16
Add resource attributes and scope to metrics proto diagram

krisztianfekete:
Fixes https://github.com/open-telemetry/opentelemetry-proto/issues/518.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -116,14 +116,6 @@ message ScopeMetrics {
 // │          └── Summary
 // └── ...
 //
-//     Metric
-//  +------------+
-//  |name        |
-//  |description |
-//  |unit        |     +------------------------------------+
-//  |data        |---> |Gauge, Sum, Histogram, Summary, ... |
-//  +------------+     +------------------------------------+
-//
 //    Data [One of Gauge, Sum, Histogram, Summary, ...]
 //  +-----------+
 //  |...        |  // Metadata about the Data.
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -98,12 +98,14 @@ message ScopeMetrics {
 //   fields.
 //
 //     Metric
-//  +------------+
-//  |name        |
-//  |description |
-//  |unit        |     +------------------------------------+
-//  |data        |---> |Gauge, Sum, Histogram, Summary, ... |
-//  +------------+     +------------------------------------+
+//  +--------------------+
+//  |name                |
+//  |resource attributes |
+//  |scope               |
+//  |description         |
+//  |unit                |     +------------------------------------+
+//  |data                |---> |Gauge, Sum, Histogram, Summary, ... |
+//  +--------------------+     +------------------------------------+
 //
 //    Data [One of Gauge, Sum, Histogram, Summary, ...]
 //  +-----------+
@@ -112,6 +114,8 @@ message ScopeMetrics {
 //  +-----------+  |
 //                 |      +---------------------------+
 //                 |      |DataPoint 1                |
+//                 |      |                           |
+//                 |      |attribute                  |
 //                 v      |+------+------+   +------+ |
 //              +-----+   ||label |label |...|label | |
 //              |  1  |-->||value1|value2|...|valueN| |
@@ -125,6 +129,8 @@ message ScopeMetrics {
 //              |  .  |                   .
 //              |  .  |   +---------------------------+
 //              |  .  |   |DataPoint M                |
+//              |  .  |   |                           |
+//              |  .  |   |attribute                  |
 //              +-----+   |+------+------+   +------+ |
 //              |  M  |-->||label |label |...|label | |
 //              +-----+   ||value1|value2|...|valueN| |
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -97,15 +97,32 @@ message ScopeMetrics {
 // - DataPoint contains timestamps, attributes, and one of the possible value type
 //   fields.
 //
+// MetricsData
+// ├── ResourceMetrics
+// │ ├── Resource
+// │ ├── SchemaURL
+// │ └── ScopeMetrics
+// │    ├── Scope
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -98,12 +98,14 @@ message ScopeMetrics {
...
-//  +------------+
-//  |name        |
-//  |description |
-//  |unit        |     +------------------------------------+
-//  |data        |---> |Gauge, Sum, Histogram, Summary, ... |
-//  +------------+     +------------------------------------+
+//  +--------------------+
+//  |name                |
+//  |resource attributes |
+//  |scope               |
```
tigrannajaryan, line 104: I think his is misleading. Resource and Scope are in different messages, they are not part of the Metric message.

krisztianfekete: Hm, right, although then I am not sure what's the best way to visualise it as currently we only have a diagram for Metric, and other messages are missing as well. Maybe just visualising all messages at the top of the proto?

jsuereth: I'd suggest visualizing more or leaving things as-is.

Metric hierarchy is quite important to the DataModel, and it's a nuance *on top of* the:  `Resource{Signal}->Scope{Signal}->{Signal}s` hierarchy of the rest of the protocol.  Particularly because with metrics:

- We have two identities:  Metric Identity and *time series* identity.
- We are reporting aggregated values (accumulated within start/stop window) that have known aggregations for handling them.

krisztianfekete: Yes, now I am thinking maybe introducing a tree to describe these relations. Something like:
```console
MetricsData
├── ResourceMetrics
│ ├── Resource
│ ├── SchemaURL
│ └── ScopeMetrics
│    ├── InstrumentationScope
│    ├── SchemaURL
│    └── Metric
│       ├── Name
│       ├── Description
│       ├── Unit
│       └── data
│          ├── Gauge
│          ├── Sum
│          ├── Histogram
│          ├── ExponentialHistogram
│          └── Summary
└── ...
```

krisztianfekete: Thanks @bogdandrutu, I will fix it! Any thoughts on including something like that?

bogdandrutu: I think your proposal helps understand, will also be good to apply to the other signals to be consistent.

krisztianfekete: I pushed a commit, @bogdandrutu, @jsuereth, @tigrannajaryan would you mind checking it?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -97,6 +97,25 @@ message ScopeMetrics {
...
+// │       ├── Unit
+// │       └── data
+// │          ├── Gauge
+// │          ├── Sum
+// │          ├── Histogram
+// │          ├── ExponentialHistogram
+// │          └── Summary
+// └── ...
+//
 //     Metric
```
bogdandrutu, line 119: No longer need lines 119-126

krisztianfekete: Done.

tigrannajaryan (approved):

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -97,13 +97,24 @@ message ScopeMetrics {
...
 //   fields.
 //
-//     Metric
-//  +------------+
-//  |name        |
-//  |description |
-//  |unit        |     +------------------------------------+
-//  |data        |---> |Gauge, Sum, Histogram, Summary, ... |
-//  +------------+     +------------------------------------+
+// MetricsData
```
MrAlias, line 100: Why provide a definition of the `MetricsData` type for the `Metric`? This seems more appropriate in the `MetricsData` type documentation.

krisztianfekete: @jsuereth, @tigrannajaryan what do you think? Should I move this up?

tigrannajaryan: Works either for me, as long as we have this diagram somewhere.

krisztianfekete: @tigrannajaryan, I just pushed a commit. Can you please take a look?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -29,6 +29,25 @@ option go_package = "go.opentelemetry.io/proto/otlp/metrics/v1";
...
+// │       ├── Name
+// │       ├── Description
+// │       ├── Unit
+// │       └── data
+// │          ├── Gauge
+// │          ├── Sum
+// │          ├── Histogram
+// │          ├── ExponentialHistogram
+// │          └── Summary
+// └── ...
```
MrAlias, line 49: What does "..." refer to?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -97,14 +115,6 @@ message ScopeMetrics {
...
 //   fields.
 //
-//     Metric
-//  +------------+
-//  |name        |
-//  |description |
-//  |unit        |     +------------------------------------+
-//  |data        |---> |Gauge, Sum, Histogram, Summary, ... |
-//  +------------+     +------------------------------------+
-//
```
MrAlias, line 107: This doesn't need to be removed. It contextualizes the following diagram about metric data.

krisztianfekete: I removed that based on @bogdandrutu's advice: https://github.com/open-telemetry/opentelemetry-proto/pull/519#discussion_r1452724379

MrAlias: His comment was made when this PR was changing the documentation of this declaration. It doesn't seem relevant to me given this PR has moved the added documentation elsewhere.

krisztianfekete: Alright, I added it back.

tigrannajaryan:
Thank you @krisztianfekete

PR #520, opened 2023-12-28
fix typo in trace.proto that says end_time >= start_time

adriangb:


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -178,7 +178,7 @@ message Span {
   // is the time when the server's application handler starts running.
   // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
   //
-  // This field is semantically required and it is expected that end_time >= start_time.
+  // This field is semantically required and it is expected that end_time <= start_time.
```
reyang, line 181: I think the current wording is correct.

adriangb: Oh duh sorry I misread and got confused because the order is the same in the docstring below 🤦🏻

PR #522, opened 2024-01-09
Add TraceState to the LogRecord

bogdandrutu:


opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -208,4 +208,9 @@ message LogRecord {
   //   - the field is not present,
   //   - the field contains an invalid value.
   bytes span_id = 10;
+
+  // trace_state conveys information about request position in multiple distributed tracing graphs.
+  // It is a trace_state in w3c-trace-context format: https://www.w3.org/TR/trace-context/#tracestate-header
```
reyang, line 213:
```suggestion
  // It is a trace_state in w3c-trace-context format: https://www.w3.org/TR/trace-context/#tracestate-header.
```

reyang (approved):

sfc-gh-bdrutu:
> I am struggling to understand what is the meaning and purpose of the tracestate in a LogRecord. Can someone explain? What do the receivers of such LogRecord supposed to do with this data?

These are same reasons as why we propagate probability sampler [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md); TL;DR is that if you derive any metrics from these logs/events you can have a "weight" associated with that.

bogdandrutu:
> I am struggling to understand what is the meaning and purpose of the tracestate in a LogRecord. Can someone explain? What do the receivers of such LogRecord supposed to do with this data?

These are same reasons as why we propagate probability sampler [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md); TL;DR is that if you derive any metrics from these logs/events you can have a "weight" associated with that.

jmacd:
Please add a description to this PR and refer to issues where this was discussed. I believe we had decided not to do this. Sampling SIG has discussed creating semantic conventions for encoding information we might usefully extract from tracestate to apply to log records, e.g., `sampling.threshold`, `sampling.priority`, and `sampling.randomness`.

bogdandrutu:
> Please add a description to this PR and refer to issues where this was discussed. I believe we had decided not to do this. 

I don't have an issue, but if you have any prior discussion would be good to link as reference.

> Sampling SIG has discussed creating semantic conventions for encoding information we might usefully extract from tracestate to apply to log records, e.g., sampling.threshold, sampling.priority, and sampling.randomness.

Would this scale for everything we like to put in tracestate (since I believe we may add more things into the tracestate)? What about other's tracestate (e.g. snwoflake's entry)?

PR #529, opened 2024-01-29
Add SpanFlags enum prefix to comments about bit mask

breedx-splk:
Mostly a nitpick because I wasn't able to easily see where the bitmask could be found. This clarifies it to be part of the `SpanFlags` enum.

### DIFF
--- a/opentelemetry/proto/trace/v1/trace.proto
+++ b/opentelemetry/proto/trace/v1/trace.proto
@@ -113,7 +113,7 @@ message Span {
   //
   // Bits 0-7 (8 least significant bits) are the trace flags as defined in W3C Trace
   // Context specification. To read the 8-bit W3C trace flag, use
-  // `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
+  // `flags & SpanFlags.SPAN_FLAGS_TRACE_FLAGS_MASK`.
   //
   // See https://www.w3.org/TR/trace-context-2/#trace-flags for the flag definitions.
   //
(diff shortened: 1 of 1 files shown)


tigrannajaryan (approved):

arminru (approved):

Kielek (approved):

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -113,7 +113,7 @@ message Span {
   //
   // Bits 0-7 (8 least significant bits) are the trace flags as defined in W3C Trace
   // Context specification. To read the 8-bit W3C trace flag, use
-  // `flags & SPAN_FLAGS_TRACE_FLAGS_MASK`.
+  // `flags & SpanFlags.SPAN_FLAGS_TRACE_FLAGS_MASK`.
```
bogdandrutu, line 116: This is extremely language specific, a.k.a Java :))

breedx-splk: I mean ok sure, I definitely don't claim to know how every language applies scoping rules to weird enum/constants.....but I know enough to guess that it's flipping inconsistent. 😊 

I'm happy to change the `SpanFlags.` bit to something that's less language opinionated if there are other options. Even just a sentence below that says where the flag is defined would help, but it seemed more verbose.

Oberon00: I think the comment is sufficiently clear before this PR, seeing that SPAN_FLAGS_ is already at the start of the name. You should be able to find the definition with a simple text search in the proto.

breedx-splk: @Oberon00 I see where you're coming from, but I would expect many practitioners to be dealing with generated language bindings and not directly touching the protobuf definition files. In cases like that, the comments can carry over to the field in code, but someone reading the comments for said field can't readily find the mask field in code. At least that was my experience.

breedx-splk:
Unhelpful/unimportant. Whatever. Someone else can reopen if they like this.

PR #534, opened 2024-03-19
Add new profile signal

petethepig:
This is a follow up to [OTEP 239: Introduces Profiling Data Model v2](https://github.com/open-telemetry/oteps/pull/239)

The main motivation behind this PR is that this will allow us to start experimenting with the profiles proto in opentelemetry-collector. I marked the profiles part as `Experimental` to indicate that this is not a final version of the data model.

I copied the proto from the OTEP, and moved `pprofextended.proto` from `profiles/v1/alternatives/pprofextended.proto` to just `profiles/v1/pprofextended.proto`. I did this because I figured we no longer have alternative representations and this will reduce confusion for people outside of Profiling SIG. The rest of the proto stayed the same.

I tested this file with a collector fork and I it compiles properly.


Rperry2174 (approved):
Lgtm!

florianl (approved):

Kielek (approved):
Great to see that progress!

tigrannajaryan:
The [recommendation](https://github.com/open-telemetry/opentelemetry-proto?tab=readme-ov-file#experiments) is to put experimental signals in "experimental" subdirectory.

@open-telemetry/specs-approvers any thoughts on whether want to use something different to indicate the state of profiles in this repo?

opentelemetry/proto/collector/profiles/v1/profiles_service_http.yaml
```diff
@@ -0,0 +1,9 @@
+# This is an API configuration to generate an HTTP/JSON -> gRPC gateway for the
+# OpenTelemetry service using github.com/grpc-ecosystem/grpc-gateway.
+type: google.api.Service
+config_version: 3
+http:
+ rules:
+ - selector: opentelemetry.proto.collector.profiles.v1.ProfilesService.Export
+   post: /v1/profiles
```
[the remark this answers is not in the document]
tigrannajaryan: >The only way I think we should use /v1/profiles is if we guarantee we won't make any breaking changes to the protocol post-this PR being accepted.

I think it is too early to guarantee that. I would do what you suggest and explicitly signify that profiles are not yet stable.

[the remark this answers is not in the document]
petethepig: Updated `v1` to `v1experimental` everywhere. There was a little [conversation in CNCF Slack](https://cloud-native.slack.com/archives/C03J794L0BV/p1710874656981869) where we decided to go with that. Let me know what you think!

tigrannajaryan (approved):
LGTM. Use of `v1experimental` is clear.

jsuereth (approved):

tigrannajaryan:
This formally has enough approvals, but I would like to keep it open in case any other @open-telemetry/specs-approvers want to review.

README.md
```diff
@@ -43,6 +43,7 @@ components as indicated by the Maturity table below.
 | metrics/\*<br>collector/metrics/* | Stable | [Stable](docs/specification.md#json-protobuf-encoding) |
 | trace/\*<br>collector/trace/* | Stable | [Stable](docs/specification.md#json-protobuf-encoding) |
 | logs/\*<br>collector/logs/* | Stable | [Stable](docs/specification.md#json-protobuf-encoding) |
+| profiles/\*<br>collector/profiles/* | Experimental | [Experimental](docs/specification.md#json-protobuf-encoding) |
```
christophe-kamphaus-jemmic, line 46: The `maturity-matrix.yaml` does not list `experimental` for `apiMaturityLevels`. Would `alpha` be applicable?

jsuereth: We can update that YAML.  `experimental` matches the terminology we use elsewhere in the spec and we have not added anything to the Proto since going stable, where `experimental` would be needed.

felixge (approved):

christophe-kamphaus-jemmic (approved):

README.md
tigrannajaryan: Now that [OTEP0232](https://github.com/open-telemetry/oteps/blob/main/text/0232-maturity-of-otel.md#maturity-levels) is accepted we need to apply it to Otel including this repo. Not in this PR though, a separate one.

pantuza (approved):

christos68k (approved):

opentelemetry/proto/profiles/v1experimental/profiles.proto
```diff
@@ -0,0 +1,185 @@
...
+  // The resource for the profiles in this message.
+  // If this field is not set then no resource info is known.
+  opentelemetry.proto.resource.v1.Resource resource = 1;
+
+  // A list of ScopeProfiles that originate from a resource.
+  repeated ScopeProfiles scope_profiles = 2;
+
+  // This schema_url applies to the data in the "resource" field. It does not apply
+  // to the data in the "scope_profiles" field which have their own schema_url field.
+  string schema_url = 3;
```
jack-berg, line 117: In the other signals, the comment on this field provides more details about how to interpret schema_url. Should probably just copy / paste it for symmetry.

Same comment applies to `ScopeProfiles.schema_url`

opentelemetry/proto/profiles/v1experimental/profiles.proto
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// A collection of Profiles produced by an InstrumentationScope.
```
jack-berg, line 120:
```suggestion
// A collection of ProfileContainers produced by an InstrumentationScope.
```

opentelemetry/proto/profiles/v1experimental/profiles.proto
```diff
@@ -0,0 +1,185 @@
...
+  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
+  //
+  // This field is semantically required and it is expected that end_time >= start_time.
+  fixed64 end_time_unix_nano = 3;
+
+  // attributes is a collection of key/value pairs. Note, global attributes
+  // like server name can be set using the resource API. Examples of attributes:
+  //
+  //     "/http/user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36"
+  //     "/http/server_latency": 300
```
jack-berg, line 158: These are bad examples of attributes. They include characters we don't recommend, and the `/http/server_latency` is more like a metric than an attribute. I know they're copy / pasted from `trace.proto`, but we ought to fix them. Also, the phrasing above "global attributes like server name can be set using the resource API" needs work. 

Should probably key as is here, and open an issue to track fixing in all spots.

opentelemetry/proto/profiles/v1experimental/profiles.proto
```diff
@@ -0,0 +1,185 @@
...
+  // The OpenTelemetry API specification further restricts the allowed value types:
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute
+  // Attribute keys MUST be unique (it is not allowed to have more than one
+  // attribute with the same key).
+  repeated opentelemetry.proto.common.v1.KeyValue attributes = 4;
+
+  // dropped_attributes_count is the number of attributes that were discarded. Attributes
+  // can be discarded because their keys are too long or because there are too many
+  // attributes. If this value is 0, then no attributes were dropped.
+  uint32 dropped_attributes_count = 5;
```
jack-berg, line 171: Should probably punt on `dropped_attributes_count` until its clear that profiling will have SDK semantics that require attributes to conform to [attribute limits](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/common#attribute-limits).

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
```diff
@@ -0,0 +1,386 @@
...
+package opentelemetry.proto.profiles.v1experimental;
+
+import "opentelemetry/proto/common/v1/common.proto";
+
+option csharp_namespace = "OpenTelemetry.Proto.Profiles.V1Experimental";
+option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1experimental";
+
+// Represents a complete profile, including sample types, samples,
+// mappings to binaries, locations, functions, string table, and additional metadata.
+message Profile {
```
jack-berg, line 66: Does this effectively represent a fork of the upstream definition [here](https://github.com/google/pprof/blob/main/proto/profile.proto)? It seems like it must be we likely have different versioning rules than whatever governs the upstream.

It seems valuable to have comments explaining the relationship between this file and upstream pprof.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
```diff
@@ -0,0 +1,386 @@
...
+  int64 default_sample_type = 14;
+}
+
+// Represents a mapping between Attribute Keys and Units.
+message AttributeUnit {
+  // Index into string table.
+  int64 attribute_key = 1;
+  // Index into string table.
+  int64 unit = 2;
+}
```
jack-berg, line 132: This is new. Curious if there are plans / ideas to generalize this concept.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
```diff
@@ -0,0 +1,386 @@
...
+  CUMULATIVE is valid, it is not recommended. */
+  AGGREGATION_TEMPORALITY_CUMULATIVE = 2;
+}
+
+// ValueType describes the type and units of a value, with an optional aggregation temporality.
+message ValueType {
+  int64 type = 1; // Index into string table.
+  int64 unit = 2; // Index into string table.
+
+  AggregationTemporality aggregation_temporality = 3;
```
jack-berg, line 218: Why is AggregationTemporality redefined in this file instead of reusing the definition in `metrics.proto`?

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
```diff
@@ -0,0 +1,386 @@
...
+  //
+  // NOTE: While possible, having multiple values for the same label key is
+  // strongly discouraged and should never be used. Most tools (e.g. pprof) do
+  // not have good (or any) support for multi-value labels. And an even more
+  // discouraged case is having a string label and a numeric label of the same
+  // name on a sample.  Again, possible to express, but should not be used.
+  // [deprecated, superseded by attributes]
+  repeated Label label = 3;
+  // References to attributes in Profile.attribute_table. [optional]
+  repeated uint64 attributes = 10;
```
jack-berg, line 255: I don't disagree with this optimization, but its interesting that we choose to use it hear and not elsewhere in the protocol. Wouldn't other signals, especially metrics, benefit from having a shared pool of attributes with points which reference those by index?

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
```diff
@@ -0,0 +1,386 @@
...
+  // The leaf is at location_index[0]. [deprecated, superseded by locations_start_index / locations_length]
+  repeated uint64 location_index = 1;
+  // locations_start_index along with locations_length refers to to a slice of locations in Profile.location.
+  // Supersedes location_index.
+  uint64 locations_start_index = 7;
+  // locations_length along with locations_start_index refers to a slice of locations in Profile.location.
+  // Supersedes location_index.
+  uint64 locations_length = 8;
+  // A 128bit id that uniquely identifies this stacktrace, globally. Index into string table. [optional]
+  uint32 stacktrace_id_index = 9;
```
jack-berg, line 236: It reads funny that the fields are not ordered according to their field number, but I suppose it makes sense to keep related fields near each other. 

How do we know that pprof won't use these same field numbers for different purpose? (This question applies generally to any place in here where we've extended the upstream)

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
```diff
@@ -0,0 +1,386 @@
...
+  repeated Label label = 3;
+  // References to attributes in Profile.attribute_table. [optional]
+  repeated uint64 attributes = 10;
+
+  // Reference to link in Profile.link_table. [optional]
+  uint64 link = 12;
+
+  // Timestamps associated with Sample represented in ms. These timestamps are expected
+  // to fall within the Profile's time range. [optional]
+  repeated uint64 timestamps = 13;
```
jack-berg, line 262: This is new from the upstream pprof definition and diverges from the convention of using epoch ns elsewhere in the data model.

jack-berg (commented):
Handful of small comments for someone pretty familiar with OTLP, but not profiling. Don't want to slow things down, but might be a comment or two that prevents re-work later.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
florianl: I noticed this as well earlier and have created https://github.com/open-telemetry/oteps/issues/253. But it seems like, I have chosen the wrong place to report it.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
```diff
@@ -0,0 +1,386 @@
...
+  // Locations referenced by samples via location_indices.
+  repeated Location location = 4;
+  // Array of locations referenced by samples.
+  repeated int64 location_indices = 15;
+  // Functions referenced by locations.
+  repeated Function function = 5;
+  // Lookup table for attributes.
+  repeated opentelemetry.proto.common.v1.KeyValue attribute_table = 16;
+  // Represents a mapping between Attribute Keys and Units.
+  repeated AttributeUnit attribute_units = 17;
```
simonswine, line 90: In terms of inter-operability with the existing [pprof](https://github.com/google/pprof/blob/29b9bb013b0f45d835bcffd9a0995e0d77127453/proto/profile.proto) spec, I do think it would be a good idea to offset the additional field IDs by a an offset. I arbitrarily chose `+32`. If in the meantime the upstream pprof adds some fields (up to 15 🙂 ), we can stay interoperable. If we don't have that gap and for some reason additional fields are added, we are in trouble.


```suggestion
  // Array of locations referenced by samples.
  repeated int64 location_indices = 33;
  // Functions referenced by locations.
  repeated Function function = 5;
  // Lookup table for attributes.
  repeated opentelemetry.proto.common.v1.KeyValue attribute_table = 34;
  // Represents a mapping between Attribute Keys and Units.
  repeated AttributeUnit attribute_units = 35;
```

simonswine: Ideally we also make sure they are not getting accidentally used by defining them as reserved

```
  // Those field ids are reserved for future use by the upstream pprof definition
  reserved 15 to 32;
```

petethepig: Good point. Collisions like this would be really hard to handle in the future so I think it is worth trying to address this now.

One slight improvement I would suggest is that we could make it so that all new fields on all structs start from 50, this way it's a little easier to distinguish and gives us a little more room.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
petethepig: On the first point, we've received feedback on earlier drafts of the proto that having related fields be next to each other was preferable, so not going to change this at the moment.

The second point raises an important concern, I answered [here](https://github.com/open-telemetry/opentelemetry-proto/pull/534#discussion_r1556453478) and also brought it up in [#otel-profiles slack](https://cloud-native.slack.com/archives/C03J794L0BV/p1712612611120679) for broader discussion. I do agree we should address this now, because if we don't we might be setting ourselves up for a lot of pain down the line.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
petethepig: That is potentially true. As we were working on the spec within Profiles SIG we've identified multiple of these types of optimizations that could broadly be applied to the whole OTEL ecosystem. We considered introducing them to other groups within OTEL, but we quickly learned that if we try to push for every single one we end up in a bit of a "boiling the ocean" type of situation where we need to convince multiple groups of multiple changes and the problem quickly becomes almost combinatorial in nature.

So instead the approach we're taking is testing these optimizations out within the boundaries of our Profiles SIG. We intend to find out if these optimizations actually make things better and if they do we plan to introduce similar changes to other signals. That's also a big part of the reason profiles signal is marked as `experimental` at this point.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
petethepig: Fair point. I think if we go that route, the proper way of reusing this struct would be to move it to `common.proto` and then make it so that `metrics.proto` and `profiles.proto` link to it. 

The problem is that if we change that in this repo that would require changes in many other SDKs / collector and so pulling that off would require coordination between many groups within OTEL, so we prefer to keep it inside `profiles.proto` while the profiles signal is in `experimental` stage.

CC @jack-berg, @tigrannajaryan, @jsuereth — let me know if I'm wrong on some of these assumptions.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
felixge: I'm worried this would solve the wrong problem and will make our payloads bigger.

IMO we need to either make sure that we DO NOT become a fork of pprof OR give up on pprof compatibility completely. Otherwise we get all the legacy of pprof + all the complexity of trying to stay compatible with it, which means we're getting the worst of both worlds.

Protobuf field tags use 3 bits for the wire type, leaving us with 4 bits before we exceed the threshold for going from a 1-byte to a 2-byte varint for the tag. 4 bits are means that the field numbers `0-15` can be encoded efficiently, but `16+` will cause the tags to become twice as big. For fields inside of message types that are used infrequently this won't matter, but I could imagine it to be significant for message types like `Sample` that are used frequently. See [protobuf wire spec](https://protobuf.dev/programming-guides/encoding/) for more details.

tigrannajaryan: When was the last time we spoke with pprof team? I think this PR is a significant milestone and we should sync with them. Last info I have is they were willing to entertain the idea that pprof's canonical Protobuf definitions live at Otel. 
@jsuereth can we set up a call and see if we can make more progress on this?

petethepig: @felixge fair point on the size. I forgot that wire type goes into the same varint, and thought we had 7 bits to work with.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
```diff
@@ -0,0 +1,386 @@
...
+// context. The program context is typically a stack trace, perhaps
+// augmented with auxiliary information like the thread-id, some
+// indicator of a higher level request being handled etc.
+message Sample {
+  // The indices recorded here correspond to locations in Profile.location.
+  // The leaf is at location_index[0]. [deprecated, superseded by locations_start_index / locations_length]
+  repeated uint64 location_index = 1;
+  // locations_start_index along with locations_length refers to to a slice of locations in Profile.location.
+  // Supersedes location_index.
+  uint64 locations_start_index = 7;
```
florianl, line 231: With https://github.com/open-telemetry/oteps/issues/252 I have opened an issue to start a discusion around  `locations_start_index` and `locations_length`, as their use do have a negative impact on the network bandwidth.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
```diff
@@ -0,0 +1,386 @@
...
+  // For a heap profile, this might be:
+  //   [["allocations","count"], ["space","bytes"]],
+  // If one of the values represents the number of events represented
+  // by the sample, by convention it should be at index 0 and use
+  // sample_type.unit == "count".
+  repeated ValueType sample_type = 1;
+  // The set of samples recorded in this profile.
+  repeated Sample sample = 2;
+  // Mapping from address ranges to the image/binary/library mapped
+  // into that address range.  mapping[0] will be the main binary.
```
florianl, line 79: In case of full system profiling, a "main binary" can not be identified all the time. E.g. instruction are written directly into memory, the memory protection is set executable and one just executes this directly. Besides that, there can also not be a "main binary" identified if multiple sources are adding to a trace - e.g. libc, interpreted frames like Python and kernel frames.

To work around this, `profile.mapping[0]` should be an empty entry similar to `profile.string_table[0]`.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
jack-berg: > but we quickly learned that if we try to push for every single one we end up in a bit of a "boiling the ocean"

👍 

Let's just open an issue  just so we don't loose track of it and can discuss the pros / cons of bringing this optimization to the other signals.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
jack-berg: It would be ideal to if it were in common, but is it strictly necessary? What happens if `profiles.proto` depends on `metrics.proto`?

opentelemetry/proto/profiles/v1experimental/profiles.proto
jack-berg: Opened #543 to track this.

tigrannajaryan:
@petethepig can you please confirm that the format defined here is backward-compatible with [current pprof format](https://github.com/google/pprof/blob/main/proto/profile.proto) in the sense that:
- This format only adds new fields. No fields are deleted and there are no changes to existing fields.
- Any valid profile represented in the current pprof format is considered a valid Otel profile data.

Is this correct?

petethepig:
@tigrannajaryan yes, these two things are correct.

opentelemetry/proto/profiles/v1experimental/profiles.proto
petethepig: I imagine the way profiling SDKs use attributes will be very similar to tracing SDKs, so I'm relatively confident that we will need this.

Having said that, we'll definitely revisit this before going stable.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
petethepig: addressed in https://github.com/open-telemetry/opentelemetry-proto/pull/534/commits/f5b589717f79ddb5d143c9a78976f315e637cf04

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
petethepig: Let's discuss separately.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
petethepig: Let's discuss separately.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
petethepig: This part is still in the works. We will make sure to clarify the relationship before going stable.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
petethepig: Nothing concrete, but yes, we'll likely propose something about units to greater Otel once we gain some experience with the current version of the proto.

petethepig:
@tigrannajaryan I addressed the comments and resolved the outstanding conversations.

tigrannajaryan:
@jack-berg you had several comments, you OK with merging in the current state?

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
jack-berg: > I'd be inclined to leave it as for the v1experimental and merge it to common when we hit stable, that way we can potentially change it without having to change common during the beta

Change it how? Is there anything in mind? Its hard to imagine an additional enum value for AggregationTemporality.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
```diff
@@ -0,0 +1,386 @@
...
+  repeated Label label = 3;
+  // References to attributes in Profile.attribute_table. [optional]
+  repeated uint64 attributes = 10;
+
+  // Reference to link in Profile.link_table. [optional]
+  uint64 link = 12;
+
+  // Timestamps associated with Sample represented in nanoseconds. These timestamps are expected
+  // to fall within the Profile's time range. [optional]
+  repeated uint64 timestamps = 13;
```
jack-berg, line 262: To try to better match the naming convention elsewhere in the proto:
```suggestion
  repeated uint64 unix_nano_times = 13;
```

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
jack-berg: Opened #545 to track this.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
jack-berg: Opened #546 to track.

jack-berg (approved):
Left another comment with a naming suggestion. I think it would be best if the relationship between these proto definitions and pprof were hashed out already, since I think not defining that essentially guarantees breaking changes with the proto definitions which might otherwise be avoided.

Still, I'm fine merging this and iterating. Everyone should be on the same page that breaking changes are not only allowed, but very likely.

tigrannajaryan:
We have a rule of keeping spec and proto PRs open for 2 days since last change. I will merge this tomorrow.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
petethepig: @jack-berg I assume you meant `times_unix_nano`? I'm gonna change to `times_unix_nano` to better match other fields.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
```diff
@@ -0,0 +1,386 @@
...
+  repeated Label label = 3;
+  // References to attributes in Profile.attribute_table. [optional]
+  repeated uint64 attributes = 10;
+
+  // Reference to link in Profile.link_table. [optional]
+  uint64 link = 12;
+
+  // Timestamps associated with Sample represented in nanoseconds. These timestamps are expected
+  // to fall within the Profile's time range. [optional]
+  repeated uint64 unix_nano_times = 13;
```
petethepig, line 262:
```suggestion
  repeated uint64 timestamps_unix_nano = 13;
```

petethepig:
@tigrannajaryan could you please merge it if everything looks good, as it's been a few days since the last update? Thank you!

PR #564, opened 2024-05-31
Fix time typo and the Sample Link relation

yanglong1010:
Could I have a review of this simple typo fix ? I would appreciate it.

According to the definition of message Sample, the relation between Sample and Link should be n-1.


christos68k (approved):
Thanks, LGTM

felixge (approved):

yanglong1010:
Thank you all ！

PR #571, opened 2024-07-11
Remove HTTP 1.1 restriction from Protocol Details

svrnm:
In the `Protocol Details` section the spec states that it defines how OTLP is implemented over grpc and **http 1.1**, but later in the document in section `OTLP/HTTP` it states "Implementations MAY use HTTP/1.1 or HTTP/2 transports."

This may confuse end readers and I suggest that it is easier to drop the 1.1 in the Protocol Details section

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -72,7 +72,7 @@ OTLP defines the encoding of telemetry data and the protocol used to exchange
 data between the client and the server.
 
 This specification defines how OTLP is implemented over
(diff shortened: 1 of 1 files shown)


yurishkuro (approved):

tigrannajaryan (commented):
This is a Stable doc but I believe the change is allowed, since it is a bug (OTLP-over-HTTP indeed is not tied to any particular HTTP version).

tigrannajaryan (approved):

jsuereth (approved):

PR #576, opened 2024-08-23
Makefile: exclude Profiles protocol from breaking-changes

florianl:
The Profiles protocol is still experimental. Exclude it from the breaking-changes check as such changes are expected at this stage.

PR #575, opened 2024-08-22
profiles: drop Sample.stacktrace_id_index

florianl:
The value for Sample.stacktrace_id_index is hardly defined except for its length. It is also not defined how this value should be generated, interpreted and could be used.

Some kind of stacktrace_id can be most efficient in stateful protocols, where transmitting duplicate information should be avoided. As the OTel Profiling protocol is a stateless protocol, this element can be droped.

The proposed change is a breaking change. As the OTel Profiling signal is still marked as experimental this should be fine? Please let me know.

FYI: @open-telemetry/profiling-maintainers

PR #578, opened 2024-08-29
profiles: drop Location.type_index

florianl:
With https://github.com/open-telemetry/semantic-conventions/pull/1188 the semantic convention for Profiles that introduces frame types and well known values is about to be merged.

Therefore,  `Location.type_index` is no longer needed.

FYI: @open-telemetry/profiling-maintainers

Similar to https://github.com/open-telemetry/opentelemetry-proto/pull/575 this is a breaking change. And CI can only pass, once https://github.com/open-telemetry/opentelemetry-proto/pull/576 is merged.

### DIFF
--- a/opentelemetry/proto/profiles/v1experimental/pprofextended.proto
+++ b/opentelemetry/proto/profiles/v1experimental/pprofextended.proto
@@ -354,9 +354,6 @@ message Location {
   // profile changes.
   bool is_folded = 5;
 
-  // Type of frame (e.g. kernel, native, python, hotspot, php). Index into string table.
-  int64 type_index = 6;
-
   // References to attributes in Profile.attribute_table. [optional]
   repeated uint64 attributes = 7;
 }


christos68k (approved):
Thanks!

felixge (approved):
LGTM, thanks! 🙇

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
```diff
@@ -354,9 +354,6 @@ message Location {
   // profile changes.
   bool is_folded = 5;
 
-  // Type of frame (e.g. kernel, native, python, hotspot, php). Index into string table.
-  int64 type_index = 6;
-
   // References to attributes in Profile.attribute_table. [optional]
   repeated uint64 attributes = 7;
```
Kielek, line 358: What's about indexes for the following fields? Shouldn't it be changed to `-1`?

florianl: With this change, https://github.com/open-telemetry/opentelemetry-proto/pull/575 and more upcoming changes, it might be easier to update the indexing after all changes landed. That way it is less of a breaking change, until the index cleanup it done.

aalexand (approved):

tigrannajaryan:
Please resolve the conversations so that we can merge.

PR #587, opened 2024-09-11
Changes to the profiles protocol based on implementation experience.

jhalliday:
Context:

The Profiling SIG and TC representatives have agreed that the protocol will no longer aim for wire compatibility with pprof, but rather will be accompanied by a canonical method of mapping data from pprof to OpenTelemetry formats. This provides us greater flexibility in adapting pprof to OpenTelemetry use cases.

Early implementation experience has exposed areas in which the protocol could benefit from breaking changes. The SIG has agreed that some of these are worthwhile and better made sooner rather than later, such that the minimum number of existing implementations are impacted.

See slack #otel-profiles and the profiling SIG meeting minutes google doc for further background on the change motivation.

In light of these points, this change:

- Alters the message definitions without using a version field or package change, e.g. not 'v1experimental' -> 'v2experiemental' or 'v1experimental2'. This is a process design choice. It will cause client/server interop problems during the transition.  For the most part OpenTelemetry protos don't change once fully defined, so there exists little precedent here and the early-stage evolution process is open to debate. Early implementers on whom this will have an impact are invited to comment, as are those with an eye on how we may want to handle breaking changes further down the line when the blast zone is larger.

- Combines profiles.proto and pprofextended.proto to a single file, merging ProfileContainer and Profile into a single new entity, Profile. The fields now in Profile are the union of those in the two messages, with some largely cosmetic reordering to provide improved clarity on the relationship between pprof and OpenTelemetry. A design choice is made to reserve a small number of field ids to provide for potential future alignment with any additions that pprof may make in later versions.

- Fields which are array indexes into lookup tables are modified to consistently use int32 data type, so as to better support languages which don't feature 64bit array sizes. pprof was targeting golang only and was used as a storage file format, whereas OpenTelemetry has polyglot requirements and is intended as a wire message in which very large arrays are not expected. This change prioritizes implementation ease over ease of transforming pprof data to OpenTelemetry format, which as a result may require splitting any very large pprof file to multiple Otel messages.

- Fields which are array indexes into lookup tables are renamed to consistently use a suffix indicating their purpose, which reduces confusion in the resulting code-generated APIs, where e.g. int fields represent String data types. Design choice: naming is hard!  '_strindex' is used for indexes into the string table, whist indexes into all other lookup tables use '_index' and rely on the prefix for context indicating which table is referenced e.g. 'function_index' is clearly a reference into the function table without requiring the more verbose 'function_functionindex'.

There remain other breaking changes under consideration. These may follow separately. This PR limits the scope to just these few to keep the discussion focussed.


opentelemetry/proto/profiles/v1experimental/profiles.proto
```diff
@@ -169,23 +279,280 @@ message ProfileContainer {
...
+  int32 build_id_strindex = 6;  // Index into string table
+  // Specifies the kind of build id. See BuildIdKind enum for more details [optional]
+  BuildIdKind build_id_kind = 11;
+  // References to attributes in Profile.attribute_table. [optional]
+  repeated uint64 attributes = 12;
+  // The following fields indicate the resolution of symbolic info.
+  bool has_functions = 7;
+  bool has_filenames = 8;
+  bool has_line_numbers = 9;
+  bool has_inline_frames = 10;
```
christos68k, line 494: We can throw these away from the proto and turn them into attributes (as discussed in the SIG).

opentelemetry/proto/profiles/v1experimental/profiles.proto
```diff
@@ -127,35 +136,136 @@ message ScopeProfiles {
...
+  repeated Function function = 5;
+  // A common table for strings referenced by various messages.
+  // string_table[0] must always be "".
+  repeated string string_table = 6;
+  // frames with Function.function_name fully matching the following
+  // regexp will be dropped from the samples, along with their successors.
+  int32 drop_frames_strindex = 7;   // Index into string table.
+  // frames with Function.function_name fully matching the following
+  // regexp will be kept, even if it matches drop_frames.
+  int32 keep_frames_strindex = 8;  // Index into string table.
```
christos68k, line 209: Should we drop these as discussed?

opentelemetry/proto/profiles/v1experimental/profiles.proto
```diff
@@ -127,35 +136,136 @@ message ScopeProfiles {
...
 
   // end_time_unix_nano is the end time of the profile.
   // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
   //
   // This field is semantically required and it is expected that end_time >= start_time.
-  fixed64 end_time_unix_nano = 3;
+  fixed64 end_time_unix_nano = 22;
+
+  // Array of locations referenced by samples.
+  repeated int64 location_indices = 23;
```
christos68k, line 261:
```suggestion
  repeated int32 location_indices = 23;
```

opentelemetry/proto/profiles/v1experimental/profiles.proto
```diff
@@ -169,23 +279,280 @@ message ProfileContainer {
...
+  // abstractions like "[vdso]".
+  int32 filename_strindex = 5;  // Index into string table
+  // A string that uniquely identifies a particular program version
+  // with high probability. E.g., for binaries generated by GNU tools,
+  // it could be the contents of the .note.gnu.build-id field.
+  int32 build_id_strindex = 6;  // Index into string table
+  // Specifies the kind of build id. See BuildIdKind enum for more details [optional]
+  BuildIdKind build_id_kind = 11;
+  // References to attributes in Profile.attribute_table. [optional]
+  repeated uint64 attributes = 12;
```
christos68k, line 489:
```suggestion
  repeated int32 attributes = 12;
```

opentelemetry/proto/profiles/v1experimental/profiles.proto
```diff
@@ -169,23 +279,280 @@ message ProfileContainer {
...
+  // case the line information above represents one of the multiple
+  // symbols. This field must be recomputed when the symbolization state of the
+  // profile changes.
+  bool is_folded = 5;
+
+  // Type of frame (e.g. kernel, native, python, hotspot, php). Index into string table.
+  int32 type_strindex = 6;
+
+  // References to attributes in Profile.attribute_table. [optional]
+  repeated uint64 attributes = 7;
```
christos68k, line 531:
```suggestion
  repeated int32 attributes = 7;
```

opentelemetry/proto/profiles/v1experimental/profiles.proto
jhalliday: Yes, there are a number of additional changes including these and the 0-indexed array issue that I deliberately left out of this PR to keep the scope a bit more manageable and facilitate quicker review. Because git doesn't have good history tracking when combining files, we're already seeing potential merge problems e.g. the doc_url field change (#588) would have to be reapplied manually if it get approved. I'd prefer to get the switchover done and then add further steps after.

opentelemetry/proto/profiles/v1experimental/profiles.proto
jhalliday: Same answer as above - out of scope for this PR. Though, for the record: yes. Just maybe not right now.

opentelemetry/proto/profiles/v1experimental/profiles.proto
jhalliday: Oops, yes! Though I'm also going to add _indices whilst I'm on, I think.

jhalliday:
Thanks for the review, Christos. I've pushed changes for the things that are in the scope of this PR and kept a note to myself for the others. There will be at least one more separate PR before the dust settles I think, but I'm still working on the 0-index headache.

opentelemetry/proto/profiles/v1experimental/profiles.proto
```diff
@@ -120,42 +129,143 @@ message ResourceProfiles {
...
+  // Locations referenced by samples via location_indices.
+  repeated Location location_table = 4;
+  // Functions referenced by locations.
+  repeated Function function_table = 5;
+  // A common table for strings referenced by various messages.
+  // string_table[0] must always be "".
+  repeated string string_table = 6;
+  // frames with Function.function_name fully matching the following
+  // regexp will be dropped from the samples, along with their successors.
+  int32 drop_frames_strindex = 7;   // Index into string table.
```
aalexand, line 206: I left this comment in the SIG meeting notes doc, copying here for visibility. Not directly related to this PR.

> Drop / keep frames regex: These fields are informal, just telling users what has been dropped/kept. I’d be okay with dropping them. Don’t think we would have designed this into the protocol.

In the pprof symbolization process these fields are not informal. The collectors populate them when producing the profile - the frames can't be dropped at that point since function names for native language like C++ are not known yet. Then the offline symbolization uses the values to actually drop some frames during the symbolization.

Basically these fields decouple the mechanism from the policy.

opentelemetry/proto/profiles/v1experimental/profiles.proto
```diff
@@ -120,42 +129,143 @@ message ResourceProfiles {
...
   // This field is required.
-  bytes profile_id = 1;
+  bytes profile_id = 20;
 
   // start_time_unix_nano is the start time of the profile.
   // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
   //
   // This field is semantically required and it is expected that end_time >= start_time.
-  fixed64 start_time_unix_nano = 2;
+  fixed64 start_time_unix_nano = 21;
```
aalexand, line 252: There are already time_nanos and duration_nanos fields above. Can we just use those?

FWIW I have a slight preference to start+duration rather than start+end:
* During debugging it's useful to see the duration easily: "is this a 10 second profile or maybe zero duration profile like heap?"
* Duration is a smaller value and is encoded more effectively. This is a more minor thing since this value is just stored once.

opentelemetry/proto/profiles/v1experimental/profiles.proto
jhalliday: ok. They stay for now anyhow. We'll take that into account when deciding if they stay longer term or get folded into attributes, which would still leave them present for use in this role.
On similar note I'm tempted to remove the deprecated fields in various places, but that may cause headaches for pprof data conversion if they have no suitable replacements.

opentelemetry/proto/profiles/v1experimental/profiles.proto
jhalliday: Makes sense. Done.

opentelemetry/proto/profiles/v1experimental/profiles.proto
```diff
@@ -169,23 +279,280 @@ message ProfileContainer {
...
 
   // Specifies format of the original payload. Common values are defined in semantic conventions. [required if original_payload is present]
-  string original_payload_format = 6;
+  string original_payload_format = 29;
 
   // Original payload can be stored in this field. This can be useful for users who want to get the original payload.
   // Formats such as JFR are highly extensible and can contain more information than what is defined in this spec.
   // Inclusion of original payload should be configurable by the user. Default behavior should be to not include the original payload.
   // If the original payload is in pprof format, it SHOULD not be included in this field.
   // The field is optional, however if it is present `profile` MUST be present and contain the same profiling information.
```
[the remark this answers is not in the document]
jhalliday: Good catch, thanks. Done.

opentelemetry/proto/profiles/v1experimental/profiles.proto
```diff
@@ -120,42 +129,143 @@ message ResourceProfiles {
...
+  repeated int32 comment_strindices = 13; // Indices into string table.
+  // Index into the string table of the type of the preferred sample
+  // value. If unset, clients should default to the last sample value.
+  int32 default_sample_type_strindex = 14;
+
+  reserved 15; // doc_url field is not inherited from pprof.
+
+  // In the event that future pprof versions add additional fields, then
+  // future OpenTelementry versions MAY use these for corresponding purposes.
+  reserved 16,17,18,19;
```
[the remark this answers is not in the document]
jhalliday: Yes, it may well be futile, but OTOH it's essentially zero cost to try. Whilst there are several other messages where we could do this, only a couple are ones where we already add additional fields without leaving a gap. The biggest mess is Sample, which has gaps for no reason and without using 'reserved' for them.
There is going to be at least one more PR for changes soon, I'm inclined to leave this as is for now and change it in that later one if needed, after a policy discussion with the SIG to get some global consistency on our approach. Unless you feel really strongly about it I'd prefer to get this PR merged without waiting on that, as the loss of change history is going to make re-syncing other changes a pain if it goes on too long.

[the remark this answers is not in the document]
aalexand: We can discuss this in a later PR, yes. My understanding was that we are going to remove fields that act as "the old way of doing things" (such as labels) and so at the physical representation level the OTel proto simply completely diverges from the pprof's proto. Plus we'd re-number and re-order the OTel's proto fields and so the numbers even for fields that have the same semantical meaning as in pprof would diverge. With that in mind having field number gaps to allow pprof to add its new fields without breaking the representation compatibility seems like not achievable but maybe I'm missing something.

opentelemetry/proto/profiles/v1experimental/profiles.proto
```diff
@@ -120,42 +129,133 @@ message ResourceProfiles {
...
+
+
+  // second section: OpenTelemetry fields.
+
   // A globally unique identifier for a profile. The ID is a 16-byte array. An ID with
   // all zeroes is considered invalid.
   //
   // This field is required.
-  bytes profile_id = 1;
+  bytes profile_id = 20;
```
aalexand, line 248: I have mixed feelings about this field being required. I suspect this is modeled after other OTel signals, but I'm thinking of something like, say, a particular language runtime decides to generate its profiles in the OTel format. Say, Python. Requiring a low-level language runtime to take a dependency on a UUID generator seems excessive and unnecessary. Maybe it's a lesser deal than I imagine.

BTW this also means that pprof profiles cannot be converted into OTel profiles with just the information that is in the pprof profile - this ID will have to be "invented" at the conversion time. It's probably OK, but this is the first field of this kind, so I wanted to call this out.

aalexand:
+1 to submit this PR soon since it involves moving files which causes conflicts with other changes.

tigrannajaryan:
@open-telemetry/profiling-maintainers please use spec:profiles for profiling issues/PR. I filter on the label to see if there is anything pending merging/etc.

trask:
> @open-telemetry/profiling-maintainers please use spec:profiles for profiling issues/PR. I filter on the label to see if there is anything pending merging/etc.

we probably need to give them triage permission on this repo

tigrannajaryan:
> > @open-telemetry/profiling-maintainers please use spec:profiles for profiling issues/PR. I filter on the label to see if there is anything pending merging/etc.
> 
> we probably need to give them triage permission on this repo

@trask profiling-maintainers have "Write" permission to this repo. Isn't that a superset of permissions that "Triage" have? If it is not, I agree let's add triage permission.

trask:
> @trask profiling-maintainers have "Write" permission to this repo. Isn't that a superset of permissions that "Triage" have? If it is not, I agree let's add triage permission.

oh yes, I missed that, ignore my comment 👍

tigrannajaryan:
Hmm, @florianl seems to be unable to add labels. Florian, is the labels gear button disabled for you?
image

trask:
I don't see @florianl in @open-telemetry/profiling-maintainers

tigrannajaryan:
> I don't see @florianl in @open-telemetry/profiling-maintainers

Of course, sorry for the confusion.

PR #588, opened 2024-09-12
Add doc_url field to match its recent addition in pprof proto.

aalexand:
google/pprof/pull/888 added this field in pprof proto, this change adds it to the OTel profiling proto to match the change.


aalexand:
@florianl @jhalliday FYI

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
```diff
@@ -123,6 +123,14 @@ message Profile {
   // Index into the string table of the type of the preferred sample
   // value. If unset, clients should default to the last sample value.
   int64 default_sample_type = 14;
+  // Documentation link for this profile. The URL must be absolute,
+  // e.g., http://pprof.example.com/cpu-profile.html
+  //
+  // The URL may be missing if the profile was generated by older code or code
+  // that did not bother to supply a link.
+  int64 doc_url = 19;  // Index into string table.
```
florianl, line 131: With `ScopeProfiles.schema_url` the URL that describes the provided profile exists already. Where do you see the differences between the two, so that this additional field is required?

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
```diff
@@ -123,6 +123,14 @@ message Profile {
...
+
+  reserved 20 to max; // NEXT ID
```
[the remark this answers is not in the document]
aalexand: This is merely to make it clear what the next ID to use is. I.e. if a new field needs to be added it would be of ID = 20. I think this is orthogonal to renumbering the fields, but I'll remove it from this commit. After all I'm not sure if OTel has any recommendations on tracking / documenting the next ID in proto files. Sometimes people just use a comment for this, but the more recent way is to use this reserved trick.

opentelemetry/proto/profiles/v1experimental/pprofextended.proto
aalexand: schema_url points to https://opentelemetry.io/docs/specs/otel/schemas/#schema-url file which is not user docs. The doc_url here is for user-level documentation. I would also expect that the schema URL would be the same for many profiles as it corresponds to the version of the data schema or something like that rather than the nature of the data (CPU profile vs heap profile vs contention profile).

christos68k: Since this should be an optional field, what do you think about using an attribute? 

On a similar note, we had a discussion in the last SIG about the boolean [fields](https://github.com/open-telemetry/opentelemetry-proto/blob/6f589125b0b7d708c9b0f32916378182ac1e123d/opentelemetry/proto/profiles/v1experimental/pprofextended.proto#L319-L322) in `Mapping` and the consensus was to drop from the proto and turn them into attributes.

aalexand: I think the doc_url can be an attribute. I'll change this.

For the has_* fields I'm less sure. I added an agenda topic to the SIG notes doc with some thoughts on this.

aalexand:
We decided to make this an attribute instead.

PR #590, opened 2024-09-26
Fold the content of pprofextended.proto into profiles.proto

jhalliday:
This change removes the now unnecessary split of the profiling format between two files, combining them into one as with other protocols. In the process, the likewise unnecessary ProfileContainer message type is removed and its fields merged into Profile, saving one level of indirection.

Note this as discussed on slack otel-profiles, this is effectively a minimal subset of the now defunct #587, remaining parts of which will be resubmitted in subsequent PRs.  This PR effectively blocks any concurrent changes to pprofextended.proto e.g. #588 as git won't track such changes across the file 'merge'.


christos68k (approved):

felixge (approved):
LGTM, thanks! 🙇  (sorry for the delay, was on PTO followed by a team summit)

PR #591, opened 2024-09-27
Clarify aggregation temporality for Summary metric type

alanwest:
I recently came across a scenario where an OTLP consumer assumed the count and sum fields of a Summary metric represented delta values. Needless to say, this is a false assumption.

The specific use case that led me to this discovery was from a user using the collector's Prometheus receiver scraping a target that emitted summary metrics. Of course the start time of the metric emitted by the collector was constant signaling these were cumulative values.

If my assumption is correct that Summary metrics will always represent cumulative values (with the exception of the quantiles) I think it makes sense to clarify this in the description.


jmacd (approved):

reyang (approved):

reyang:
Editorial, merging now..

PR #592, opened 2024-10-21
Improve lookup table pattern use in profiles.

jhalliday:
Based on implementation experience with the draft version of the profiles spec, the use of lookup tables for storing values is space efficient but not particularly user friendly. This change makes improvements by changing field names to clarify their role in the pattern and changing their data types to better accommodate their use in languages limited to 32 bit array indexing e.g. Java.

### DIFF
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -377,7 +377,7 @@ message ValueType {
 // augmented with auxiliary information like the thread-id, some
 // indicator of a higher level request being handled etc.
 message Sample {
-  // The indices recorded here correspond to locations in Profile.location.
+  // References to locations in Profile.location_table.
   // The leaf is at location_index[0]. [deprecated, superseded by locations_start_index / locations_length]
   repeated int32 location_index = 1;
   // locations_start_index along with locations_length refers to to a slice of locations in Profile.location.
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -483,7 +483,7 @@ message Location {
 
 // Details a specific line in a source code, linked to a function.
 message Line {
-  // The index of the corresponding profile.Function for this line.
+  // Reference to function in Profile.function_table.
   int32 function_index = 1;
   // Line number in source code.
   int64 line = 2;
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -380,7 +380,7 @@ message Sample {
   // References to locations in Profile.location_table.
   // The leaf is at location_index[0]. [deprecated, superseded by locations_start_index / locations_length]
   repeated int32 location_index = 1;
-  // locations_start_index along with locations_length refers to to a slice of locations in Profile.location.
+  // locations_start_index along with locations_length refers to to a slice of locations in Profile.location_indices.
   // Supersedes location_index.
   int32 locations_start_index = 7;
   // locations_length along with locations_start_index refers to a slice of locations in Profile.location.
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -139,7 +139,7 @@ message ScopeProfiles {
   // A list of Profiles that originate from an instrumentation scope.
   repeated Profile profiles = 2;
 
-  // The Schema URL, if known. This is the identifier of the Schema that the metric data
+  // The Schema URL, if known. This is the identifier of the Schema that the profile data
   // is recorded in. To learn more about Schema URL see
   // https://opentelemetry.io/docs/specs/otel/schemas/#schema-url
   // This schema_url applies to all profiles in the "profiles" field.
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -478,13 +478,13 @@ message Location {
   bool is_folded = 5;
 
   // References to attributes in Profile.attribute_table. [optional]
-  repeated uint64 attributes = 7;
+  repeated int32 attribute_indices = 7;
 }
 
 // Details a specific line in a source code, linked to a function.
 message Line {
   // The index of the corresponding profile.Function for this line.
```
christos68k, line 486:
```suggestion
  // Reference to function in Profile.function_table.
```

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -455,7 +455,7 @@ message Location {
   // The index of the corresponding profile.Mapping for this location.
```
christos68k, line 455:
```suggestion
  // Reference to mapping in Profile.mapping_table.
```

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -394,10 +394,10 @@ message Sample {
   // lists of the originals.
   repeated int64 value = 2;
   // References to attributes in Profile.attribute_table. [optional]
-  repeated uint64 attributes = 10;
+  repeated int32 attribute_indices = 10;
 
   // Reference to link in Profile.link_table. [optional]
```
christos68k, line 399:
```suggestion
  // Reference to link in Profile.link_table. [optional]
```

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -379,13 +379,13 @@ message ValueType {
 message Sample {
   // The indices recorded here correspond to locations in Profile.location.
```
christos68k, line 380:
```suggestion
  // References to locations in Profile.location_table.
```

christos68k (commented):
Added some docstring nits for consistency

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -377,15 +377,15 @@ message ValueType {
 // augmented with auxiliary information like the thread-id, some
 // indicator of a higher level request being handled etc.
 message Sample {
-  // The indices recorded here correspond to locations in Profile.location.
+  // References to locations in Profile.location_table.
   // The leaf is at location_index[0]. [deprecated, superseded by locations_start_index / locations_length]
-  repeated uint64 location_index = 1;
+  repeated int32 location_index = 1;
   // locations_start_index along with locations_length refers to to a slice of locations in Profile.location.
```
christos68k, line 383: Should this be `location_indices` instead of [location](url)?


```suggestion
  // locations_start_index along with locations_length refers to to a slice of locations in Profile.location_indices.
```

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -377,15 +377,15 @@ message ValueType {
...
   // Supersedes location_index.
-  uint64 locations_start_index = 7;
+  int32 locations_start_index = 7;
   // locations_length along with locations_start_index refers to a slice of locations in Profile.location.
```
christos68k, line 386:
```suggestion
  // locations_length along with locations_start_index refers to a slice of locations in Profile.location_indices.
```

christos68k (approved):

trask (approved):
I like the clarity of the renames

felixge (approved):
LGTM. Nits (that could be addressed in follow-up PRs)

* Why `int32` instead of `uint32` for the indexes? I'd prefer the latter, but don't feel strongly.
* `key_strindex` and `link_index` feels inconsistent. My preference would be `key_string_index`.

jhalliday:
uint32 has the same problem as int64 - it can hold values that don't map cleanly to the signed 32 bit array range in Java. It would express intent more clearly, but potentially cause more headaches for implementers.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -194,13 +194,13 @@ message Profile {
...
   // If multiple binaries contribute to the Profile and no main
   // binary can be identified, mapping[0] has no special meaning.
-  repeated Mapping mapping = 3;
+  repeated Mapping mapping_table = 3;
   // Locations referenced by samples via location_indices.
-  repeated Location location = 4;
+  repeated Location location_table = 4;
   // Array of locations referenced by samples.
-  repeated int64 location_indices = 15;
+  repeated int32 location_indices = 15;
```
aalexand, line 201: The only thing that gives me pause about this loss of type width is that at least some pprof profiles I know (including ones generated by internal profilers) use the address value as the location ID and 32 bits is not enough for that. But on the other hand such sparse encoding already became unsupported as we switched to the dense array indexing, so it should be fine.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -407,10 +407,10 @@ message Sample {
 // Provides additional context for a sample,
 // such as thread ID or allocation size, with optional units. [deprecated]
 message Label {
```
aalexand, line 409: Should we get rid of the Label type? Keeping the use of both attributes and labels was part of trying to be wire compatible with pprof but we don't pursue that lower level goal anymore.

aalexand: Not in this PR, mostly want to discuss going forward.

aalexand (approved):

PR #594, opened 2024-10-31
WIP: Proposal: Support Timestamped Profiling Events

felixge:
**🚧 WORK IN PROGRESS 🚧:** This proposal is still in the early stages. I'm mostly looking for feedback from active participants in the profiling SIG before soliciting feedback from the wider community.

### DIFF
--- a/opentelemetry/proto/profiles/v1development/eventprofile.md
+++ b/opentelemetry/proto/profiles/v1development/eventprofile.md
@@ -1,0 +1,284 @@
+# Proposal: Support Timestamped Profiling Events
+
+**Author(s):** @felixge in collaboration with @thegreystone, @jbachorik et al.
+
+**🚧 WORK IN PROGRESS 🚧:** This proposal is still in the early stages. I'm mostly looking for feedback from active participants in the profiling SIG before soliciting feedback from the wider community.
+
+## Summary
+
+The OpenTelemetry profiling format is currently designed to record aggregated summaries of events. This proposal suggests to pivot the format toward recording individual events with timestamps while retaining the ability to record aggregated profiling data as well.
+
+This will enable the following use cases.
+
+1. Richer analysis of On-CPU as well as Off-CPU data.
+2. Powerful visualizations such as as Thread Timelines, Flame Charts, and FlameScopes.
+3. Better compatibility with JFR, Go Execution Traces and Linux perf.
+
+## Motivation
+
+Most of the current OpenTelemetry profiling format, including the concept of aggregating stack traces, is inherited from pprof.
+However, given that we have [decided](https://github.com/open-telemetry/opentelemetry-proto/issues/567#issuecomment-2286565449) that strict pprof compatibility is not a goal for the OpenTelemetry, we are now free to design a more flexible and extensible format that can be used for a wider range of profiling use cases than pprof.
+
+One use case that recently came up is the [collection of Off-CPU profiling data for the ebpf profiler](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/144). The attendees of the [Sep 5](https://docs.google.com/document/d/19UqPPPlGE83N37MhS93uRlxsP1_wGxQ33Qv6CDHaEp0/edit?tab=t.0#heading=h.lenvx4xd62c6) SIG meeting agreed that aggregated profiling data is not ideal for this use case, as it increases the difficulty to reason about the collected data and to correlate it with application behavior. This is especially true when it comes to the analysis of tail latency problems caused by outliers. So instead aggregation, it is much more useful to record this data as individual events with timestamps as well as additional context such as thread id, trace id and span id. This becomes even more powerful when On-CPU samples are also recorded with timestamps, as it allows users to identify spikes and stalls of CPU activity as well as Off-CPU examplars that explain the cause of the stalls and resulting latency (or lack of throughput).
+
+In addition to the manual analysis of such data, collecting timestamps also enables powerful visualizations such as [Thread Timelines](https://www.datadoghq.com/blog/continuous-profiler-timeline-view/), [Flame Charts](https://developer.chrome.com/docs/devtools/performance/reference#flame-chart), and [FlameScopes](https://www.brendangregg.com/blog/2018-11-08/flamescope-pattern-recognition.html). See [this document](https://docs.google.com/document/d/1285Rp1pSu6eOTInzPynCiJvDoG-r2cjqxWwTewaTncs/edit?tab=t.0#heading=h.jze26hdz58dy) or [this presentation](https://youtu.be/53UIPZfz-_U?si=x4uYpMtOCIJy8ihY) for more details.
+
+Last but not least, first-class support for timestamps allows greater compatibility with other profiling tools such as JFR, Go Execution Traces and Linux perf that also record timestamps for profiling events.
+
+## Design
+
+This proposal comes with a prelimary sketch for the new format to illustrate the feasibility of supporting timestamped as well as aggregated profiling data within the same format. However, the final design will be subject to further discussion and refinement and the changes can be merged via smaller PRs if needed.
+
+The goals for the design are to be a strict superset of pprof and a useful subset of JFR, Go Execution Traces and Linux perf. Efficiency is also a goal, but it is secondary to the goal of providing a flexible and extensible format that can be implemented by a wide range of profilers and analysis tools.
+
+TODO: Finish describing the design from a high level.
+
+## Examples
+
+TODO: Add example for Go mutex and goroutine profiles.
+
+### Aggregated CPU Profile
+
+Below is a simple example of an aggregated CPU profile that demonstrates how the new format can operate in the same way as pprof.
+
+```txtpb
+event_types: [
+    { # index 0
+        name_string_index: 1 # cpu_sample
+        unit_string_index: 2 # ms
+        default_value:     10 # each Event.count represents 10ms of CPU time - this avoid redundant 
+    }
+]
+strings: [
+    "",           # index 0
+    "cpu_sample", # index 1
+    "ms"          # index 2
+]
+
+events: [
+    {
+        event_type_index: 0 # cpu_sample (can be omitted on the wire)
+        stack_index:      1 # main;serve_request
+        count:            2 # samples, i.e. 20ms of CPU time
+    }
+    {
+        event_type_index: 0 # cpu_sample (can be omitted on the wire)
+        stack_index:      2 # main;foo
+        count:            1 # samples, i.e. 10ms of CPU time
+    },
+]
+
+# stacks, locations, functions, mappings, etc. omitted for brevity
+```
+
+### Timestamped CPU Profile
+
+```txtpb
+event_types: [
+    { # index 0
+        name_string_index: 1  # cpu_sample
+        unit_string_index: 2  # ms
+        default_value:     10 # each event represents 10ms of CPU time
+        clock_id:          0
+    }
+]
+clocks: [
+    { # index 0
+        frequency :      1_000_000_000          # 1 Ghz clock, i.e. 1 cycle = 1 ns
+        time_unix_nanos: 1_257_894_000_000_000_000
+    }
+]
+attributes: [
+    { # index 0
+        key_index: 4          # thread.id
+        value: {int_value: 1} # 1
+    },
+    { # index 1
+        key_index: 4,         # thread.id
+        value: {int_value: 2} # 2
+    }
+]
+strings: [
+    "",             # index 0
+    "cpu_time",     # index 1
+    "milliseconds", # index 2
+    "unix_epoch",   # index 3
+    "thread.id"     # index 4
+]
+
+events: [
+    {
+        event_type_index: 0     # cpu_sample (can be omitted on the wire)
+        stack_index: 1          # main;serve_request
+        time:        10_500_000 # 10.5ms since clock start
+        attributes:  [0]        # thread_id: 1
+    }
+    {
+        event_type_index: 0  # cpu_sample (can be omitted on the wire)
+        stack_index: 2       # main;foo
+        time:        300_000 # 0.3ms since previous event
+        attributes:  [1]     # thread_id: 2
+    },
+    {
+        event_type_index: 0    # cpu_sample (can be omitted on the wire)
+        stack_index: 1         # main;serve_request
+        time:        9_200_000 # 9.2ms since previous event
+        attributes:  [0]       # thread_id: 1
+    }
+]
+# stacks, locations, functions, mappings, etc. omitted for brevity
+```
+
+### Timestamped CPU Profile + Off-CPU Thread State Profile
+
+
+```txtpb
+event_types: [
+    { # index 0
+        name_string_index: 1  # "cpu_sample"
+        unit_string_index: 2  # "ms"
+        default_value:     10 # each event represents 10ms of CPU time
+        clock_id:          0
+    } 
+    { # index 1
+        name_string_index: 2  # "thread_state"
+        clock_id:          0
+    }
+]
+clocks: [
+    { # index 0
+        frequency :      1_000_000_000          # 1 Ghz clock, i.e. 1 cycle = 1 ns
+        time_unix_nanos: 1_257_894_000_000_000_000
+    }
+]
+attributes: [
+    { # index 0
+        key_index: 4          # thread.id
+        value: {int_value: 1} # 1
+    },
+    { # index 1
+        key_index: 4,         # thread.id
+        value: {int_value: 2} # 2
+    }
+    { # index 2
+        key_index: 5,         # thread.state
+        value: {key_index: 6} # mutex
+    }, # index 3
+    {
+        key_index: 5,         # thread.state
+        value: {key_index: 7} # sleep
+    }
+]
+strings: [
+    "",             # index 0
+    "cpu_sample",   # index 1
+    "thread_state", # index 2
+    "ms",           # index 3
+    "thread.id",    # index 4
+    "thread.state", # index 5
+    "mutex",        # index 6
+    "sleep",        # index 7
+]
+
+events: [
+    {
+        event_type_index: 0     # cpu_sample (can be omitted on the wire)
+        stack_index: 1          # main;serve_request
+        time:        10_500_000 # 10.5ms since clock start
+        attributes:  [0]        # thread_id: 1
+    }
+    {
+        event_type_index: 0     # cpu_sample (can be omitted on the wire)
+        stack_index: 2          # main;foo
+        time:        300_000    # 0.3ms since previous event
+        attributes: [1]         # thread_id: 2
+    },
+    {
+        event_type_index: 0     # cpu_sample (can be omitted on the wire)
+        stack_index: 1          # main;serve_request
+        time:        9_200_000  # 9.2ms since previous event
+        attributes:  [0]        # thread_id: 1
+    },
+    {
+        event_type_index: 1            # thread_state
+        stack_index:      1            # main;serve_request;mutex_lock
+        time:             300_000_000  # 300ms since previous event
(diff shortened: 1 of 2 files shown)


opentelemetry/proto/profiles/v1development/eventprofile.proto
```diff
@@ -0,0 +1,230 @@
...
+  uint32 event_type_index = 1;
+
+  // The index of the Stack associated with this event.
+  uint32 stack_index = 2;
+
+  // The index of the Link associated with this event if available.
+  optional uint32 link_index = 3;
+
+  // The indices of the attributes associated with this event.
+  repeated uint32 attribute_indices = 4;
```
aalexand, line 71: Attribute value units need to be specified somehow. In the current profiling format there is a sidecar field for that.

opentelemetry/proto/profiles/v1development/eventprofile.proto
```diff
@@ -0,0 +1,230 @@
...
+  // The index of the Clock associated with this event type. The absence of this
+  // field indiciates that events associated with this event type are expected
+  // to be aggregates containing a Event.count for the number of events that
+  // were aggregated into this event and an Event.value representing the sum of
+  // the values of the aggregated events.
+  optional uint32 clock_index = 3;
+  // The default value for the Event.value field. This is used when the value
+  // field is the same for all events of this type. For example CPU samples may
+  // typically represent the same value of 10ms.
+  uint64 default_value = 4;
```
aalexand, line 148: This feels more like weight than "default value" to me.

opentelemetry/proto/profiles/v1development/eventprofile.proto
```diff
@@ -0,0 +1,230 @@
...
+  // should derive it by multiplying the count with the default value.
+  uint64 sum = 9;
+
+  // TODO: Add more statistics for aggregated events (e.g. min, max,
+  // stddev, etc) or even histogram sketches?
+}
+
+// Clock describes the clock format used by the EventType. Clocks are expected
+// to be monotonic and include the time when a system is suspended.
+message Clock {
```
aalexand, line 113: I like the idea of having the clock type to allow configurable and different resolution of timestamps in the same or different profiles, but one downside is - does it make simple and typical case more verbose? E.g. perhaps many collectors might end up just using Unix epoch microseconds with no attributes, but they will have to configure at least one clock.

opentelemetry/proto/profiles/v1development/eventprofile.proto
```diff
@@ -0,0 +1,230 @@
...
+// for the fields time, duration, value, count and sum when considering that
+// they can be unset (0) or holding a value (>1). We need to be clear which
+// combinations have use cases and which should be considered invalid.
+message Event {
+  // The index of EventType associated with this event. The most frequent event
+  // type should use index 0 so it can be omitted on the wire.
+  uint32 event_type_index = 1;
+
+  // The index of the Stack associated with this event.
+  uint32 stack_index = 2;
```
beaubelgrave, line 65: It would be really nice in addition to the stack if we could also include raw bytes of the event, along with some way to define the format of the Event (maybe in EventType). This goes back to the issue you and I discussed about store JFR / DotNet like events in this format someway.

felixge: @beaubelgrave that's worth considering. protobuf's stdlib already has an [Any type](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/any.proto) that can be used to embed raw protobuf messages. Maybe we could do something similar here?

opentelemetry/proto/profiles/v1development/eventprofile.proto
felixge: You mean the [AttributeUnit messages](https://github.com/open-telemetry/opentelemetry-proto/blob/7312bdf63218acf27fe96430b7231de37fd091f2/opentelemetry/proto/profiles/v1development/profiles.proto#L280-L286)? Yeah, we can add that - I don't recall deleting it intentionally.

opentelemetry/proto/profiles/v1development/eventprofile.proto
```diff
@@ -0,0 +1,230 @@
...
+  // The indices of the attributes associated with this event.
+  repeated uint32 attribute_indices = 4;
+
+  // time represents the timestamp of the event. For events without a duration,
+  // this is the end time of the event. Time values for the same Clock must be
+  // delta encoded. This means that the time value of the first Event on the
+  // wire is the number of clock cycles since the start of the associated Clock.
+  // For subsequent Event messages for the same Clock, the time value is the
+  // number of clock cycles since the previous message on the wire.
+  int64 time = 5;
```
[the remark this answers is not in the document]
felixge: I'm open to alternative naming. My worry with `timestamp` is that people would miss the delta encoding requirement. But that could happen with the current name as well I guess. I don't feel strongly about it.

opentelemetry/proto/profiles/v1development/eventprofile.proto
felixge: I took the term `value` from pprof's `Sample.value` field. `weight` would also work for me. I don't feel strongly about it.

opentelemetry/proto/profiles/v1development/eventprofile.proto
felixge: Yeah, it's a tradeoff. I added the clocks for a few different reasons. If I had to put them in order, it would probably be:

1. **Leap Seconds**: IMO observability systems should avoid unix epoch timestamps until leap seconds are [abolished in 2035](https://www.theverge.com/2022/11/21/23470759/leap-second-scrapped-2035-time-computer-chaos). At least I would imagine that they make it significantly less fun to debug incidents that are caused by leap seconds. But unfortunately this is a problem for all other OpenTelemetry signals already.
2. **Data volumes**: Changing the resolution could save 1-2 bytes per event under varint encoding.
3.  **TSC Clocks:** Those can run at frequencies above 1 Ghz, so nanosecond resolution is not sufficient to store their values. OTOH, I can think of very few use cases that require sub-nanosecond resolution, especially for the stuff we target with OpenTelemetry.

That being said, the first problem is solvable without configurable clocks. If we decide 2 and 3 are not important enough, we should probably just require timestamps that originate from a non-leap unix epoch with nanosecond precision?

opentelemetry/proto/profiles/v1development/eventprofile.proto
beaubelgrave: Yeah, that Any type would work well, since it gives us a type url we can use to indicate format like we were kind of circling around in the issue discussion. I'd just want to ensure that we can put the payload in this Event object and also have an optional format specified for the payload in the EventType object.

The scenario is we dump raw bytes in the Event objects and label them as "perf_event" or "tracepoint" when we get our data from Linux tracepoints. Then we'd have the tracepoint format data within the EventType struct. That way we could have small payloads in the Event with a larger format description in the EventType that has a 1:N relationship with Event (if I read the patches right).

opentelemetry/proto/profiles/v1development/eventprofile.md
```diff
@@ -0,0 +1,284 @@
...
+
+In addition to the manual analysis of such data, collecting timestamps also enables powerful visualizations such as [Thread Timelines](https://www.datadoghq.com/blog/continuous-profiler-timeline-view/), [Flame Charts](https://developer.chrome.com/docs/devtools/performance/reference#flame-chart), and [FlameScopes](https://www.brendangregg.com/blog/2018-11-08/flamescope-pattern-recognition.html). See [this document](https://docs.google.com/document/d/1285Rp1pSu6eOTInzPynCiJvDoG-r2cjqxWwTewaTncs/edit?tab=t.0#heading=h.jze26hdz58dy) or [this presentation](https://youtu.be/53UIPZfz-_U?si=x4uYpMtOCIJy8ihY) for more details.
+
+Last but not least, first-class support for timestamps allows greater compatibility with other profiling tools such as JFR, Go Execution Traces and Linux perf that also record timestamps for profiling events.
+
+## Design
+
+This proposal comes with a prelimary sketch for the new format to illustrate the feasibility of supporting timestamped as well as aggregated profiling data within the same format. However, the final design will be subject to further discussion and refinement and the changes can be merged via smaller PRs if needed.
```
florianl, line 30: Should the focus be on timestamps for non aggregated profiling data?
Just a thought: With data-type profiling ,e.g. https://github.com/golang/go/issues/69699, it might not be possible to attach to every profiling information a timestamp. So maybe a generic approach might be best?
```suggestion
This proposal comes with a prelimary sketch for the new format to illustrate the feasibility of supporting event based as well as aggregated profiling data within the same format. However, the final design will be subject to further discussion and refinement and the changes can be merged via smaller PRs if needed.
```

opentelemetry/proto/profiles/v1development/eventprofile.proto
```diff
@@ -0,0 +1,230 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package opentelemetry.proto.profiles.v1development;
+
+import "opentelemetry/proto/common/v1/common.proto";
+
+// Profile holds a collection of events that represent a profile.
+message Profile {
```
florianl, line 22: Did you consider having a single `event_type` for a `message Profile`?
For `message Profile` with a single `event_type` processing could look like this:
```mermaid
flowchart TD
    A[Shared Envelope] --> B{" "}
    B --> D[Profile for on-CPU profiling]
    B --> E[Profile for off-CPU profiling]
    B --> F[Profile for memory allocation profiling]
```

The string table, attribute table and other shared data then should be part of the `Shared Envelope` (I'm bad in naming things :pray: ). This would allow faster handling of profiling data by processors based on `event_type`. With the suggested approach processors have to walk all `events` to handle a specific `event_type`.

opentelemetry/proto/profiles/v1development/eventprofile.md
```diff
@@ -0,0 +1,284 @@
...
+    }
+    { # index 1
+        name_string_index: 2 # liveheap
+        unit_string_index: 3 # By (bytes)
+    }
+]
+strings: [
+    "",              # index 0
+    "allocations",   # index 1
+    "liveheap",      # index 2
```
tsint, line 237: I think it's necessary to differentiate between virtual memory, physical memory, and situations like page faults. Typically, virtual memory consumed is more than physical memory because some memory regions might be zeroed and never referenced.  Some might be 'copy-on-write'.

opentelemetry/proto/profiles/v1development/eventprofile.proto
aalexand: Yes, that one.

opentelemetry/proto/profiles/v1development/eventprofile.proto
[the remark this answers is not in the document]
aalexand: Some ideas, with their pros and cons:

* An explicit name like `timestamp_or_delta`, `time_or_delta`.
* Oneof with `timestamp` and `time_delta` fields.

opentelemetry/proto/profiles/v1development/eventprofile.proto
aalexand: I see. No strong opinion, we can discuss later if anyone becomes more opinionated.

opentelemetry/proto/profiles/v1development/eventprofile.proto
aalexand: 2 and 3 could be solved by having some way of specifying the resolution of the timestamp per event type? E.g. nanos, micros, or millis.

aalexand: We should probably look if all other signals follow some specific approach with this. Being extra special is probably not worth it.

opentelemetry/proto/profiles/v1development/eventprofile.md
```diff
@@ -0,0 +1,284 @@
...
+        event_type_index: 1            # thread_state
+        stack_index:      1            # main;serve_request;mutex_lock
+        time:             300_000_000  # 300ms since previous event
+        duration:         50_000_500   # 50ms
+        attributes:       [0, 2]       # thread_id: 1, thread.state: mutex
+    },
+    {
+        event_type_index: 1             # thread_state
+        stack_index:      1             # main;foo;sleep
+        time:             5_000_000     # 5ms since previous event
```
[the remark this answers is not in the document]
beaubelgrave: IMHO, if we had a duration in addition to time, we could make off_cpu and stall events clearer. IE: Stall started at time X, lasted for Y. This gives a start/stop. This could be stored in an attribute, but feels somewhat wasteful to have to have an attribute per-duration value.

opentelemetry/proto/profiles/v1development/eventprofile.md
```diff
@@ -0,0 +1,284 @@
...
+    {
+        event_type_index: 0     # cpu_sample (can be omitted on the wire)
+        stack_index: 1          # main;serve_request
+        time:        10_500_000 # 10.5ms since clock start
+        attributes:  [0]        # thread_id: 1
+    }
+    {
+        event_type_index: 0  # cpu_sample (can be omitted on the wire)
+        stack_index: 2       # main;foo
+        time:        300_000 # 0.3ms since previous event
```
beaubelgrave, line 119: Why is time based on previous event? This implies we have events in perfect time order. Often during collection, especially off-CPU (cswitch) events, the start of the event is not in-order of when it completes. This would cause collectors to sort potentially large sets of data (cswitches can be massive).

Can we just have an absolute time option vs the delta time encoding scheme?

petethepig: I imagine delta encoding is used to reduce payload size

opentelemetry/proto/profiles/v1development/eventprofile.md
```diff
@@ -0,0 +1,284 @@
...
+
+**🚧 WORK IN PROGRESS 🚧:** This proposal is still in the early stages. I'm mostly looking for feedback from active participants in the profiling SIG before soliciting feedback from the wider community.
+
+## Summary
+
+The OpenTelemetry profiling format is currently designed to record aggregated summaries of events. This proposal suggests to pivot the format toward recording individual events with timestamps while retaining the ability to record aggregated profiling data as well.
+
+This will enable the following use cases.
+
+1. Richer analysis of On-CPU as well as Off-CPU data.
```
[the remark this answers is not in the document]
felixge: > Richer analysis of aggregated data, like On-CPU profiling

Not sure I follow. The use case enabled by this proposal is event based CPU profiling rather than aggregation. We already have aggregation?

opentelemetry/proto/profiles/v1development/eventprofile.proto
felixge: That's an interesting idea. I can see how this could simplify things. I'll think about this a bit more!

opentelemetry/proto/profiles/v1development/eventprofile.md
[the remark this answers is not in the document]
felixge: duration and value are separate fields in the current proposal, so I think the use cases brought up here should be covered? https://github.com/open-telemetry/opentelemetry-proto/pull/594/files/7d85bba21098f7950813d9717f06e86865da0726#diff-4a0777f946b5d445b17b64906a2ee5b121a070ee499189a656988ad0df350e11R81-R94

opentelemetry/proto/profiles/v1development/eventprofile.md
felixge: This needs to be clarified, probably via semantic conventions. But the example given here is allocations profiled by the Go runtime. Those are always measured in terms of virtual memory. I wrote up more details on this [here](https://www.datadoghq.com/blog/go-memory-metrics/).

opentelemetry/proto/profiles/v1development/eventprofile.md
felixge: The purpose of delta encoding is to reduce payload sizes. `time` is an `int64` right now, so negative values are possible. If we frequently need to deal with unordered events, we could define `time` as an `sint64` which allows to encode negative numbers more efficiently. However, it comes a the expense of encoding positive numbers.

Anyway, I don't feel very strongly about delta encoding. I'm planning to do some benchmarks later to provide some evidence (or lack of evidence) that it's going to be beneficial.

opentelemetry/proto/profiles/v1development/eventprofile.md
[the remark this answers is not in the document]
florianl: OTel ebpf profiler could aggregate on CPU profiling data. This is possible by setting [Sample.value](https://github.com/open-telemetry/opentelemetry-proto/blob/793e43e25f2eb8063c2693bc2e66e16672cf7ed8/opentelemetry/proto/profiles/v1development/profiles.proto#L379-L385) to a different value than 1 and provide a timestamp in [Sample.timestamps_unix_nano](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L394C19-L394C39) for each `Sample.value`. 
Currently, this aggregation is not done as it was rejected during the review process and so it is still a low hanging fruit for optimization.

opentelemetry/proto/profiles/v1development/eventprofile.md
```diff
@@ -0,0 +1,284 @@
...
+
+One use case that recently came up is the [collection of Off-CPU profiling data for the ebpf profiler](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/144). The attendees of the [Sep 5](https://docs.google.com/document/d/19UqPPPlGE83N37MhS93uRlxsP1_wGxQ33Qv6CDHaEp0/edit?tab=t.0#heading=h.lenvx4xd62c6) SIG meeting agreed that aggregated profiling data is not ideal for this use case, as it increases the difficulty to reason about the collected data and to correlate it with application behavior. This is especially true when it comes to the analysis of tail latency problems caused by outliers. So instead aggregation, it is much more useful to record this data as individual events with timestamps as well as additional context such as thread id, trace id and span id. This becomes even more powerful when On-CPU samples are also recorded with timestamps, as it allows users to identify spikes and stalls of CPU activity as well as Off-CPU examplars that explain the cause of the stalls and resulting latency (or lack of throughput).
```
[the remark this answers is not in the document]
felixge: Ack. I'm not trying to suggest that aggregated Off-CPU profiles are useless. But they tend to be dominated by idle threads, so it can be tricky to navigate them. Additionally they can't really be used to analyze tail latencies since all they show are averages.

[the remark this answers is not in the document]
beaubelgrave: FWIW, we use a mix of both. Overlaying active CPU over aggregate wait CPU shows where the CPU is trying to execute and getting delayed. From there we allow zooming into individual time sections involved in the aggregate to debug thread interactions, etc. We need both within the same profile to achieve the same thing we have in our private formats.

opentelemetry/proto/profiles/v1development/eventprofile.md
beaubelgrave: 64-bits are already variable encoded, and if they were deltas off of a stable epoch (like profile start time). They would be quite small without deltas between each other. I get why we would want delta encoding, but it comes at a sorting cost that is likely really bad for off-cpu data sources (since they complete at random times compared to the start time).

opentelemetry/proto/profiles/v1development/eventprofile.md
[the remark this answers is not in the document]
beaubelgrave: Yeah, sorry I missed that! Definitely covers what I was asking for, thanks!

felixge:
The problem of efficiently encoding timestamped profiling data has been addressed by https://github.com/open-telemetry/opentelemetry-proto/pull/649 . Closing this PR.

PR #595, opened 2024-10-31
Replace has_* fields with an enum.

aalexand:


aalexand:
@jhalliday @felixge @florianl FYI - sending this for feedback.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -423,6 +423,54 @@ message Label {
   int64 num_unit = 4;  // Index into string table
 }
 
+// Specifies the availability of the function and file names, line numbers and
+// inline frames for a mapping.
+enum SymbolizationLevel {
```
christos68k, line 428: @aalexand Why not have this as an attribute? (I admit I already forgot the relevant part in today's discussion).

One reason that springs to mind is implementors not having to deal with OTel [KeyValue](https://github.com/open-telemetry/opentelemetry-proto/blob/7312bdf63218acf27fe96430b7231de37fd091f2/opentelemetry/proto/common/v1/common.proto#L64-L67) which IIRC is a concern that you had in the past. Just wondering if there is something else, feel fee to correct my assumptions 😅

aalexand: The most elaborate "life of a profile" in terms of symbolization that I can think of is something like:

1) A collector gathers a C++ profile. It sets the symbolization level to `SYMBOLIZATION_LEVEL_NONE` to indicate that only IP addresses are available and nothing else.
2) The profile is transferred to the backend and stored there.
3) The offline symbolizer processes the profile. At this point, initially, only stripped binary is available to the symbolizer so it symbolizes the profile at `SYMBOLIZATION_LEVEL_SYMBOLS` level.
4) The offine symbolizer at some point gets full DWARF for the binary and re-symbolizes the profile, storing it with `SYMBOLIZATION_LEVEL_LINES_INLINE` level now.

The re-symbolization part is not something that all profilers may want to support, but I wanted to indicate that the level upgrade is possible.

Of course, replacing an attribute in a proto in memory is not a huge deal - iterate over the `repeated opentelemetry.proto.common.v1.KeyValue attributes` list and replace or append the attribute with the desired key and new value, but it's still microchurn and for something that is so common as symbolization level it feels that having a simple explicit field is nicer. But it's definitely not a dealbreaker, so I'd love to hear what others think.

As a side note, I wonder if it would be better if `repeated opentelemetry.proto.common.v1.KeyValue attributes` would be a [`Map<Key, Value>`](https://protobuf.dev/programming-guides/proto3/#maps) instead. I thought maybe it's to support multi-value attributes, but there is an explicit support for lists in this type, so it's not that.

One other thing with commony used attributes also is that the definition of allowed values in their semantics is [so far away from the proto itself](https://github.com/open-telemetry/semantic-conventions/tree/main/model/profile). With an enum one gets compile-time checks, code completion etc. With attributes I don't even know what kind of presubmit or code generation support exists. This aspect probably shouldn't solely guide our decisions on what should be an attribute vs a field, but I think it's still a factor.

florianl: With https://github.com/open-telemetry/opentelemetry-proto/pull/595 I started to transform these levels into OTel SemConv attributes.

Will you remove this enum from the PR and also remove the _has* fields? Or should we remove the _has* fields in a dedicated PR?

aalexand: Done - updated the PR and title/description.

P.S. In your comment I think you meant https://github.com/open-telemetry/semantic-conventions/pull/2522, not #595 which is this PR.

florianl (approved):

christos68k (approved):

felixge (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
aalexand: Resolving - we did decide to make these attributes after all, see the description, it was updated.

aalexand:
@tigrannajaryan This is ready to be merged.

PR #606, opened 2024-12-06
profile: drop duplicate field in message Profile

florianl:
The message Profile contains two fields of type `opentelemetry.proto.common.v1.KeyValue`, `attribute_table` and `attributes`. Both fields are intended for the same use. Therefore drop the later one.

ping @open-telemetry/profiling-maintainers @jhalliday


opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -240,20 +244,6 @@ message Profile {
...
-  //     "/http/user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36"
-  //     "/http/server_latency": 300
-  //     "abc.com/myattribute": true
-  //     "abc.com/score": 10.239
-  //
-  // The OpenTelemetry API specification further restricts the allowed value types:
-  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute
-  // Attribute keys MUST be unique (it is not allowed to have more than one
-  // attribute with the same key).
-  repeated opentelemetry.proto.common.v1.KeyValue attributes = 18;
```
aalexand, line 255: I think `Profile.attributes` and `Profile.attribute_table` serve different purpose.

`Profile.attributes` is the list of attributes for the profile itself.

`Profile.attribute_table` is a table referenced by Sample.attribute_indices.

I'm not sure we can remove either of them. At least I don't think this is a typo or an accidental duplication.

jhalliday:
yes, not actually duplicates. The attributes_table is there to be referenced by fields of attributes in other places as a de-duplication / size reduction measure, whereas the attributes field is for things that belong directly to the profile itself. It would be correct, though not especially useful, to make the attributes field a repeated int indexing into the attributes table, but removing it entirely is not appropriate.

florianl:
Thanks for the feedback. The differences between `attribute_table` and `attributes` were not clear, so that gave me the impression of a duplicate field. To address your feedback, I have added `attribute_indices` for the global attributes in message Profile, that can reference into `attribute_table`.
This should help differentiating the attributes while having only a single `attribute_table` which also helps to deduplicate elements.

opentelemetry/proto/profiles/v1development/profiles.proto
florianl: https://github.com/open-telemetry/opentelemetry-proto/pull/606/commits/bcfb44e09110272a3016c951b8b5c88ff15b3191 adds a list of indices into `attributes_table`  - does this address your concern?

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -269,6 +259,17 @@ message Profile {
...
+  // References to attributes in attribute_table. [optional]
+  // It is a collection of key/value pairs. Note, global attributes
+  // like server name can be set using the resource API. Examples of attributes:
+  //
+  //     "/http/user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36"
+  //     "/http/server_latency": 300
+  //     "abc.com/myattribute": true
+  //     "abc.com/score": 10.239
+  //
+  repeated int32 attribute_indices = 22;
```
aalexand, line 272: Do we want to simply reuse the field ID of the removed field maybe? It's not a backward compatible change, but we recently did a field renumbering and I assume before releasing version 1 we'd want to make sure we don't have ID gaps, so we might as well do this now. @jhalliday WDYT

opentelemetry/proto/profiles/v1development/profiles.proto
aalexand: Yes, this looks good. AFAICT this brings all uses of attributes into consistency as they are not all referenced via the indices. One note is that in this case this indirection optimization has little value since, unlike other messages, there is no repetition / nesting for Profile, plus the indirection potentially makes debugging a little harder (such as dumping the profile as a proto and reading the profile attributes in the dump), but it's hard to weigh whether hypothetically easier debugging is worth inconsistency, so the change LGTM.

opentelemetry/proto/profiles/v1development/profiles.proto
florianl: For the moment I would vote for just using a new field ID and not reuse the existing one, change its field name and field type. Before declaring this proto as stable, closing these gaps make sense, but I think this will still take a while.

christos68k (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
aalexand: SGTM

aalexand (approved):

jhalliday:
@dmathieu does this have the potential to cause headaches in the collector? Adding attributes to the 'top level' entity in a given signal type seems like one of the most common mutations in the collector pipeline use cases and I think all the others use common.v1.KeyValue and hence probably have shared pdata golang types/code for doing so. Are we adding undue complexity here if profiles works a different way?

dmathieu:
Processors should be able to get a command to "add an attribute to a profile/sample", and have that turned into the proper append/find in the attributes table, and indices.
So I don't think it's so much of an added complexity.

It's true that the entirety of these mapping tables are pretty unique to profiles, and none of the other signals use it though, making profiles rather a snowflake.

jhalliday (approved):

lmolkova:
We'd like to cut a release before holidays, draft PR - https://github.com/open-telemetry/opentelemetry-proto/pull/611

Should we try to get this PR in since it seems to be pretty close? I can wait a day or two.

felixge:
> We'd like to cut a release before holidays, draft PR - #611
> 
> Should we try to get this PR in since it seems to be pretty close, @open-telemetry/profiling-approvers ? I can wait a day or two.

I think this is read to go from our perspective. We just need @open-telemetry/spec-sponsors approval 🙇

jsuereth (approved):

PR #609, opened 2024-12-09
profiles: move attribute_table to message ScopeProfiles

florianl:
Move `attribute_table` from `message Profile` to `message ScopeProfiles`. This reduces the amount of duplicate attributes significantly if more than one kind of profile is reported - e.g. on CPU and memory heap profiling data.

ping @open-telemetry/profiling-maintainers

### DIFF
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -144,6 +144,9 @@ message ScopeProfiles {
   // https://opentelemetry.io/docs/specs/otel/schemas/#schema-url
   // This schema_url applies to all profiles in the "profiles" field.
   string schema_url = 3;
+
+  // Lookup table for attributes.
+  repeated opentelemetry.proto.common.v1.KeyValue attribute_table = 4;
 }
 
 // Profile is a common stacktrace profile format.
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -144,6 +144,9 @@ message ScopeProfiles {
   // https://opentelemetry.io/docs/specs/otel/schemas/#schema-url
   // This schema_url applies to all profiles in the "profiles" field.
   string schema_url = 3;
+
+  // Lookup table for attributes.
+  repeated opentelemetry.proto.common.v1.KeyValue attribute_table = 4;
```
florianl, line 149: The argument can be made to even move `attribute_table` to `ResourceProfiles` or `ProfilesData` instead of `ScopeProfiles`. If you prefer this, please let me know.

florianl: Not moving `attribute_table` for the named reason to `ProfilesData` or `ResourceProfiles` sounds good to me. When moving "up" `attributes_table` to `ScopeProfiles` I just looked at it from a protocol perspective and not from a potential synchronization issues of processors.

> Doesn't the same reasoning (reduce amount of duplicate data) also apply to moving the string table?

yes - the same applies to `string_table`. I can add it here as well.

> Also, should we merge https://github.com/open-telemetry/opentelemetry-proto/pull/606 into this PR, otherwise it can be a little confusing (e.g. [doesn't update](https://github.com/florianl/opentelemetry-proto/blob/9a0dee7addf3d815ac996e93806542dce36169c9/opentelemetry/proto/profiles/v1development/profiles.proto#L256) Profile attributes to be references)

To me, these are independent proposals that should be handled and decided on separately. So I didn't merge everything into a big PR. Depending on which PR gets merged first, the other still can be updated.

florianl:
Putting this PR on hold for the moment.

florianl:
Closing in favor of https://github.com/open-telemetry/opentelemetry-proto/issues/628#issuecomment-2772566325

PR #611, opened 2024-12-13
Prepare release 1.5.0

lmolkova:
- profiles: Remove unused `Label` definition. [#602](https://github.com/open-telemetry/opentelemetry-proto/pull/602)
- all: Add note about `schema_url` format (including version). [#605](https://github.com/open-telemetry/opentelemetry-proto/pull/605)
- logs: Add top-level `event_name` field to log records instead of `event.name` attribute. [#600](https://github.com/open-telemetry/opentelemetry-proto/pull/600)
- profiles: drop duplicate `attributes` field in message Profile. [#606](https://github.com/open-telemetry/opentelemetry-proto/pull/606)

PR #613, opened 2024-12-17
Add extra link check to release.md

lmolkova:
See https://github.com/open-telemetry/semantic-conventions/issues/1009 and https://github.com/open-telemetry/opentelemetry-proto/pull/611#issuecomment-2542654361 for the context.

TL;DR: [markdown-link-check](https://github.com/tcort/markdown-link-check/) does not check cross-file anchor links.
Today broken links are discovered only when a new proto release is published on otel.io, where these links needs to be manually overridden and fixed.

The general tooling for this is tracked in https://github.com/open-telemetry/semantic-conventions/issues/1009, in the meantime, we should leverage otel.io workflow that can run against proto `main` branch and detect broken anchors. 

It finds a few broken anchors in almost every semconv and spec release, so 1) it's useful 2) it adds just a few minutes to the release process and not too bad.

This PR adds this as a manual step until we have a full automation.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -2,7 +2,7 @@
 
 _Instruction for Maintainers only._
 
-- Run [opentelemetry.io workflow](https://github.com/open-telemetry/opentelemetry.io/actions/workflows/build-dev.yml)
+- Run the [opentelemetry.io dev-build workflow](https://github.com/open-telemetry/opentelemetry.io/actions/workflows/build-dev.yml)
   against `opentelemetry-proto` submodule as a smoke-test for docs. Fix broken links, if any.
 
 - Prepare the release by updating [CHANGELOG.md](CHANGELOG.md), see for example
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -2,6 +2,9 @@
 
 _Instruction for Maintainers only._
 
+- Run [opentelemetry.io workflow](https://github.com/open-telemetry/opentelemetry.io/actions/workflows/build-dev.yml)
+  against `opentelemetry-proto` submodule as a smoke-test for docs. Fix broken links, if any.
+
 - Prepare the release by updating [CHANGELOG.md](CHANGELOG.md), see for example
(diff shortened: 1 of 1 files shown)


trask (approved):

RELEASING.md
```diff
@@ -2,6 +2,9 @@
 
 _Instruction for Maintainers only._
 
+- Run [opentelemetry.io workflow](https://github.com/open-telemetry/opentelemetry.io/actions/workflows/build-dev.yml)
```
tigrannajaryan, line 5: How do you run this against a submodule? Can you please provide more details. I don't know how to do this.

lmolkova: 1. open that link
2. type opentelemetry-proto into the submodule field 
image
3. Run it
4. wait for it to complete
6. Both `Build and check links` step along with `warnings in build log?` should pass.
7. If any of them fails, check  `Build and check links`/`Run npm run log:test-and-fix` logs - broken links should be there
8. Ignore "REFCACHE updates?" result. 

Did I capture it right @chalin ?

run with broken links

image

run without issues:

image

I don't think we normally go into this level of details in release doc, but it's easy to figure out the moment you try it.

chalin, line 5:
```suggestion
- Run the [opentelemetry.io dev-build workflow](https://github.com/open-telemetry/opentelemetry.io/actions/workflows/build-dev.yml)
```

chalin: The workflow accepts a regex as a argument:

> image

You can leave the argument as is with the default value, or change it to `opentelemetry-proto` as suggested in the next line (line 6) below.

chalin: Yes, thanks about it. 👍🏻

tigrannajaryan: Great, thanks! Would it be possible to add the distilled version of this to RELEASING.md?

lmolkova:
I'm going to close it in favor of https://github.com/open-telemetry/opentelemetry-proto/pull/614

PR #615, opened 2025-01-09
Simplify profile stack trace representation

felixge:
- Introduce a first-class Stack message type and lookup table.
- Replace location index range based stack trace encoding on Sample with a single stack_index reference.
- Remove the location_indices lookup table.

The primary motivation is laying the ground work for [timestamp based profiling][timestamp proposal] where the same stack trace needs to be referenced much more frequently compared to aggregation based on low
cardinality attributes.

Timestamp based profiling is also expected to be used with the upcoming [Off-CPU profiling][off-cpu pr] feature in the eBPF profiler. Off-CPU stack traces have a different distribution compared to CPU stack traces. In particular Off-CPU stack traces are much more repetitive because they typically occur at special leaf functions such as leaf (async preemption being a notable exception). For the same reason it is also uncommon to see a stack trace that are a root-prefix of a previously observed stack trace.

We might need to revisit the previous [previous benchmarks][benchmarks] to confirm these claims, as the previous analysis seems to have shown that the location range based encoding is always either comparable or better than the mechanism proposed here.

The secondary motivation is simplicitly. Arguably the proposed change here will make it easier to write exporters, processors as well as receivers.

It seems like we had rough consensus around this change in previous SIG meetings, and it seems like a good incremental step to make progress on the timestamp proposal.

[timestamp proposal]: https://github.com/open-telemetry/opentelemetry-proto/pull/594
[off-cpu pr]: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196
[benchmarks]: https://docs.google.com/spreadsheets/d/1Q-6MlegV8xLYdz5WD5iPxQU2tsfodX1-CDV1WeGzyQ0/edit?gid=2069300294#gid=2069300294

### DIFF
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -163,24 +163,24 @@ message ScopeProfiles {
 // - On-disk, the serialized proto must be gzip-compressed.
 //
 // - The profile is represented as a set of samples, where each sample
-//   references a sequence of locations, and where each location belongs
+//   references a stack trace which is a list of locations, each belonging
 //   to a mapping.
-// - There is a N->1 relationship from sample.location_id entries to
-//   locations. For every sample.location_id entry there must be a
+// - There is a N->1 relationship from Stack.location_indices entries to
+//   locations. For every Stack.location_indices entry there must be a
 //   unique Location with that index.
 // - There is an optional N->1 relationship from locations to
 //   mappings. For every nonzero Location.mapping_id there must be a
 //   unique Mapping with that index.
 
-// Represents a complete profile, including sample types, samples,
-// mappings to binaries, locations, functions, string table, and additional metadata.
-// It modifies and annotates pprof Profile with OpenTelemetry specific fields.
+// Represents a complete profile, including sample types, samples, mappings to
+// binaries, stacks, locations, functions, string table, and additional
+// metadata.  It modifies and annotates pprof Profile with OpenTelemetry
+// specific fields.
 //
 // Note that whilst fields in this message retain the name and field id from pprof in most cases
 // for ease of understanding data migration, it is not intended that pprof:Profile and
 // OpenTelemetry:Profile encoding be wire compatible.
 message Profile {
-
   // A description of the samples associated with each Sample.value.
   // For a cpu profile this might be:
   //   [["cpu","nanoseconds"]] or [["wall","seconds"]] or [["syscall","count"]]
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -374,11 +374,8 @@ message ValueType {
 // augmented with auxiliary information like the thread-id, some
 // indicator of a higher level request being handled etc.
 message Sample {
-  // locations_start_index along with locations_length refers to to a slice of locations in Profile.location_indices.
-  int32 locations_start_index = 1;
-  // locations_length along with locations_start_index refers to a slice of locations in Profile.location_indices.
-  // Supersedes location_index.
-  int32 locations_length = 2;
+  // Reference to stack in Profile.stack_table.
+  int32 stack_index = 1;
```
beaubelgrave, line 378: It would be nice to have an optional int64 instruction_ip property here that would take the common stack at stack_index, but also add the instruction_ip to it. This would allow for common on-CPU callstacks to be respresented cheaply that are in the same function but hit different parts of it.

aalexand: Shouldn't be just int64 since that won't play well with symbolization and inlines, would have to be a location ID.

Also, this optimization would only be most useful to profile types like CPU sampling where the leaf frame is indeed very random. For profiles like heap profiling it would not be as valuable since there even the leaf is a callsite, not a sampled IP.

And if we are looking at such "encode the tree more efficiently" optimizations then we should probably look at using the actual tree encoding as discussed a bit [here](https://cloud-native.slack.com/archives/C03J794L0BV/p1736442340357979).

beaubelgrave: Agree, it's only useful for on-CPU sampling with skids, etc. So I don't think it should be required, but definitely useful when needed as an option.

rockdaboot: Just an idea not very well thought of:
Did we already think about aggregating and sending the stacktraces as flamegraph to achieve prefix-compression on the protocol level? At least within a binary, stacktraces often have a large large common part (at Elastic we saw ~60% when introducing prefix compression for stacktraces in the backend).

It boils down to
- client aggregates stack traces into flamegraph
- add an index for the leaf location in `sample`
- add an index for the parent location in `location` (index == 0 means end-of-stacktrace)

Pros: reduced network payload, reduced memory consumption on the receiver side
Cons: increases CPU usage on client (hashing and comparing each frame, additional memory allocations)

beaubelgrave: @rockdaboot We've (Microsoft) found that approach of callstack representation very useful, especially when crossing billions (now in trillions) of callstacks. We have our details covered in a patent on our approach to this: https://patents.justia.com/patent/12112151

rockdaboot: Thanks for the link, very insightful!
Since creating graphs from stacks is much older than the patent, do you see any legal blocker in case we consider doing this on the protocol level? Also using delta encoding for addresses would be interesting to further reduce the network payload for profiles.

beaubelgrave: Callstack representation methods have been around for some time, but the particular methods and systems in the patent were created to deal with shortcomings of those known methods when dealing with large numbers of stacks.

felixge:
I made a mistake in the branch management for this PR which prevents me from updating it now. Will close and reopen a new PR and link them together.

PR #619, opened 2025-02-05
Clarify when bucket_counts array may be empty

tigrannajaryan:
The current definition says:

"The number of elements in bucket_counts array must be by one
 greater than the number of elements in explicit_bounds array."

However, it also says "bucket_counts is an optional field", which 
means its length can be 0, which then in turn means the length of 
explicit_bounds must be -1, which is impossible.

We clarify that it is a valid situation to have both bucket_counts and explicit_bounds arrays empty.

### DIFF
--- a/opentelemetry/proto/metrics/v1/metrics.proto
+++ b/opentelemetry/proto/metrics/v1/metrics.proto
@@ -461,7 +461,9 @@ message HistogramDataPoint {
   // The sum of the bucket_counts must equal the value in the count field.
   //
   // The number of elements in bucket_counts array must be by one greater than
-  // the number of elements in explicit_bounds array.
+  // the number of elements in explicit_bounds array. The exception to this rule
+  // is when the length of bucket_counts is 0, then the length of explicit_bounds
+  // must also be 0.
   repeated fixed64 bucket_counts = 6;
 
   // explicit_bounds specifies buckets with explicitly defined bounds for values.
@@ -477,6 +479,9 @@ message HistogramDataPoint {
   // Histogram buckets are inclusive of their upper boundary, except the last
   // bucket where the boundary is at infinity. This format is intentionally
   // compatible with the OpenMetrics histogram definition.
+  //
+  // If bucket_counts length is 0 then explicit_bounds length must also be 0,
+  // otherwise the data point is invalid.
   repeated double explicit_bounds = 7;
 
   // (Optional) List of exemplars collected from


cijothomas (approved):

jmacd (approved):

reyang (approved):

jsuereth (approved):

mx-psi:
Does this resolve open-telemetry/opentelemetry-specification/issues/3848? Should it be solved first? I am always unsure about the exact relationship between the proto and the data model

tigrannajaryan:
> Does this resolve [open-telemetry/opentelemetry-specification/issues/3848](https://github.com/open-telemetry/opentelemetry-specification/issues/3848)? Should it be solved first? I am always unsure about the exact relationship between the proto and the data model

@mx-psi this PR is only about the regular Histogram. https://github.com/open-telemetry/opentelemetry-specification/issues/3848 is about Exponential Histogram.

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -461,7 +461,9 @@ message HistogramDataPoint {
   // The sum of the bucket_counts must equal the value in the count field.
   //
   // The number of elements in bucket_counts array must be by one greater than
-  // the number of elements in explicit_bounds array.
+  // the number of elements in explicit_bounds array. The exception to this rule
+  // is when the length of bucket_counts is 0, then the length of explicit_bounds
+  // must also be 0.
   repeated fixed64 bucket_counts = 6;
```
jack-berg, line 467: If the `bucket_counts` is zero, then all the other fields also have to be zero / omitted, else the data would be contradictory.

I suppose this isn't necessarily invalid, but its definitely an edge case.

mx-psi: The data model [says](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#histogram)

> Bucket counts are optional. A Histogram without buckets conveys a population in terms of only the sum and count, and may be interpreted as a histogram with single bucket covering (-Inf, +Inf).

So the rest of the fields need not be zero in this case

tigrannajaryan: Is it worth copying that phrase about bucket counts being optional from the data model spec to the comment in the proto file?

pellared:
CC @open-telemetry/specs-metrics-approvers

opentelemetry/proto/metrics/v1/metrics.proto
jack-berg: Ah I forgot that there is a separate top level `count` field. Ok so there's still useful data that can be transmitted without bucket counts.

jack-berg (approved):

bogdandrutu (approved):

tigrannajaryan:
This was open for several days and appears to match the data model spec. Merging.

PR #621, opened 2025-02-07
profiles: clarify use of optional fields in message Line and message Function

florianl:
Add clarifying comments for the use of optional fields in message Line and message Function

For visibility: @open-telemetry/profiling-maintainers


opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -463,13 +463,17 @@ message Line {
...
+  int32 name_strindex = 1;
+  // Index into Profile.string_table to the name of the function, as identified
+  // by the system. For instance, it can be a C++ mangled name.
+  // Use index 0 if value is not available.
+  int32 system_name_strindex = 2;
+  // Index into Profile.string_table to the source file containing the function.
+  // Use index 0 if value is not available.
+  int32 filename_strindex = 3;
   // Line number in source file.
   int64 start_line = 4;
```
Kielek, line 478: Can we make a note here? 0 if not available/making it optional?

florianl: My understanding was, that we use `optional` for `function_index`, similar to `optional int32 mapping_index = 1;` to have consistency. And improve on the documentation of the fields within `message Function`, pointing to the special meaning of `Profile.string_table[0]`.

christos68k: I don't see it written down in the agenda but I do remember the point of "specialness" for tracking a missing value for strings was raised and the final consensus was to use `optional`. Maybe @felixge can chime in.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -463,13 +463,17 @@ message Line {
...
-  // Name of the function, in human-readable form if available.
-  int32 name_strindex = 1; // Index into string table
-  // Name of the function, as identified by the system.
-  // For instance, it can be a C++ mangled name.
-  int32 system_name_strindex = 2; // Index into string table
-  // Source file containing the function.
-  int32 filename_strindex = 3; // Index into string table
+  // Index into Profile.string_table to the name of the function, in
+  // human-readable form.
+  // Use index 0 if value is not available.
```
aalexand, line 468: Describing the semantics of how we index strings at every field is becoming repetitive IMO.

For *_strindex fields I would suggest we focus the documentation on the semantics of this string field. The strindex itself points to the fact that this is a string index, with the same meaning universally (including index 0 for empty value).

So, maybe put the documentation simply as

```
  // Function name. Empty string if not available.
  int32 name_strindex = 1;
  // Function name, as identified by the system. For instance,
  // it can be a C++ mangled name. Empty string if not available.
  int32 system_name_strindex = 2;
  // Source file containing the function. Empty string if not available.
  int32 filename_strindex = 3;
  ...
```

?

opentelemetry/proto/profiles/v1development/profiles.proto
aalexand: Unfortunately I don't recall if we decided to change how we handle strindex fields (we should capture notes better), but thinking more about this I would propose we leave them without optional keyword and with special zero index value meaning empty for now. We can discuss later.

Motivation: optional field at least in Go are encoded at runtime as a field by pointer rather than by value, so if we make them all optional we end up with a lot more "*int32" fields instead of "int32". These dynamic allocations can add up and have memory usage and GC CPU costs.

florianl:
The Profiling SIG (2025-02-20) wants to explore the [protobuf Opaque API](https://go.dev/blog/protobuf-opaque) first - before making further steps.

opentelemetry/proto/profiles/v1development/profiles.proto
florianl: Applied this suggestion to use `Empty string if not available`.

opentelemetry/proto/profiles/v1development/profiles.proto
florianl: Based on the feedback around the `optional` field in https://cloud-native.slack.com/archives/C01N7PP1THC/p1741664907074709 (There is a historical reason why OTel sparsely make use of `optional`) I have applied the suggestions from https://github.com/open-telemetry/opentelemetry-proto/pull/621#discussion_r1947132154 and set `optional` only for int64 values.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -454,21 +454,21 @@ message Line {
   // Reference to function in Profile.function_table.
   int32 function_index = 1;
   // Line number in source code.
-  int64 line = 2;
+  optional int64 line = 2;
```
aalexand, line 457: Line and column are 1-based, so we could easily say here "0 means unset" I think? Not sure if optional buys us much here. I do like the purity and the cleanness of optional, but all the potential quirks and potential performance penalty of using it does give me a pause.

florianl: "0 means unset" sounds also good to me. Will update it.

christos68k: They're 1-indexed in DWARF but I'm not sure if that's the case for every interpreted language we support. I'd expect line numbers to be 1-indexed across the board but a quick search for column numbers in Python shows 0-indexing:

https://docs.python.org/3/library/traceback.html#traceback.FrameSummary.colno

christos68k (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
aalexand: That example is for columns, maybe we could say that the column number is only valid if the line number is non-zero and assume line numbers are 1-indexed? I would hope that the line indexation is always 1-based and seems true for that example above as well, it says

> end_lineno[¶](https://docs.python.org/3/library/traceback.html#traceback.FrameSummary.end_lineno)
The last line number of the source code for this frame. By default, it is set to lineno and indexation starts from 1.

(interestingly there is no this clarifying comment there for the `lineno` field, but I assume it's the same, 1-based)

Kielek (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -453,22 +453,22 @@ message Location {
...
 message Function {
-  // Name of the function, in human-readable form if available.
-  int32 name_strindex = 1; // Index into string table
-  // Name of the function, as identified by the system.
-  // For instance, it can be a C++ mangled name.
-  int32 system_name_strindex = 2; // Index into string table
-  // Source file containing the function.
-  int32 filename_strindex = 3; // Index into string table
-  // Line number in source file.
+  // Function name. Empty string if not available.
```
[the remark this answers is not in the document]
aalexand: There is a global convention in this proto that fields with _strindex suffix in the name index into the string table and are effectively string-typed. Documenting this for every field is repetitive, so it's omitted.

tigrannajaryan:
Please resolve all comments to get this merged.

florianl:
@tigrannajaryan I have resolved open comments.

PR #623, opened 2025-02-12
Clarify client behavior for 502 and 504 response codes

tigrannajaryan:
The spec previously said that the client should retry if 502 or 504 response is received. However, it was not specified what retry strategy should be used if Retry-After header is missing.

The spec now says that the behavior when Retry-After is missing is the same for all 4 expected codes: 429,502,503,504.


docs/specification.md
```diff
@@ -591,8 +591,9 @@ overloaded, the server SHOULD respond with `HTTP 429 Too Many Requests` or
 recommended time interval in seconds to wait before retrying.
 
 The client SHOULD honour the waiting interval specified in the "Retry-After"
-header if it is present. If the client receives an `HTTP 429` or an `HTTP 503`
-response and the "Retry-After" header is not present in the response, then the
+header if it is present. If the client receives a retryable error code
+`HTTP 429`, `HTTP 502`, `HTTP 503` or `HTTP 504`
+and the "Retry-After" header is not present in the response, then the
```
bogdandrutu, line 596:
```
When client receives any retyable errors (maybe create a link to the previous table):
1. If the "Retry-After" header is present in the response, the client SHOULD honour the waiting interval specified in the "Retry-After" header.
2. If the "Retry-After" header is not present in the response, the client SHOULD implement an exponential backoff strategy between retries.
```

I would clarify this way, that for any retryable error these are the steps.

bogdandrutu (approved):

docs/specification.md
tigrannajaryan: That changes the meaning of spec though. Currently the spec says that Retry-After should be honoured regardless of response code, not just for retryable errors.

bogdandrutu: >Currently the spec says that Retry-After should be honoured regardless of response code, not just for retryable errors.

But also the spec says:
```
All other 4xx or 5xx response status codes MUST NOT be retried.
```

tigrannajaryan: According to the spec it does not have to be a retry. It can be the next regular HTTP request. The spec says:
>The client SHOULD honour the waiting interval specified in the "Retry-After" header if it is present.

This applies to all responses, not just retryable ones.

I am not sure this was ever the intent of the spec, but technically this means if you send a POST, get 200 in return with Retry-After, then the next POST should honour the Retry-After interval. Pointless? Most likely never intended to be like this, we can treat it as a bug.

tigrannajaryan: No, I think the behavior of "all other 4xx and 5xx" remains unchanged.

bogdandrutu: I will let you decide, but I feel that you are asking me to implement a retry on 200 the way it is written, and I don't think that is right.

tigrannajaryan: >I feel that you are asking me to implement a retry on 200 the way it is written, and I don't think that is right.

Agree. I don't want to ask that. The spec implies that and I think it is wrong.

tigrannajaryan: >Also, I am not sure I understand if I need to retry if I get a Retry-After for a 400 code.

No. 400 is not a retryable response code. Only 4 codes listed in the table are retryable. The spec is clear on that. It says "All other 4xx or 5xx response status codes MUST NOT be retried."

The spec also explicitly calls out 400 and says "The client MUST NOT retry the request when it receives HTTP 400 Bad Request response.".

If client receives a Retry-After for a 400 code, it does not imply in any way that the request must be retried. Also server should never send Retry-After with 400, the spec does not ask for it.

tigrannajaryan: I suggest that in a separate PR we change this

>The client SHOULD honour the waiting interval specified in the "Retry-After" header if it is present.

to this

>_For retryable response codes_ the client SHOULD honour the waiting interval specified in the "Retry-After" header if it is present.

Retry-After does not make sense for non-retryable codes. I think we should consider this a bug. It was never the intent to have a waiting interval for non-retryable response codes. That's pointless. Whoever wrote the phrase (me?) just used poor wording.

I also think the spec does not need to say anything about what happens with Retry-After for non-retryable codes. The client must simply ignore Retry-After in those cases, just like it ignores all other irrelevant headers that may be present in the response.

bogdandrutu: I would not repeat though the  retryable codes here as you do right now because that leads to inconsistency.

reyang: > I would not repeat though the retryable codes here as you do right now because that leads to inconsistency.

+1

tigrannajaryan: Agree. Let me see if I can add a link to the table.

tigrannajaryan: Done.

tigrannajaryan: Added issue https://github.com/open-telemetry/opentelemetry-proto/issues/624

docs/specification.md
```diff
@@ -563,6 +563,8 @@ The requests that receive a response status code listed in following table SHOUL
 be retried.
 All other `4xx` or `5xx` response status codes MUST NOT be retried.
 
+<a name="retryable-response-codes"></a>
```
bogdandrutu, line 566: I would make it a header L5?

arminru (approved):

arminru (approved):

Issue #628, opened 2025-02-19
OTLP resources for the opentelemetry-ebpf-profiler


Gandem:
# Recap

Currently, by default, the profiler generates OTLP payloads containing a single Profile object with stack traces for all processes on a host. This creates issues with resource attributes, particularly container attributes, which were initially planned to be attached to Sample attributes, not top-level resource attributes. This incompatibility makes the profiler's approach inconsistent with other signals and processors like the `k8sattributesprocessor`.

This leads to the following challenges:
* Defining resources for the profiler, considering various granularities (per process, per container, or another concept). 
* Impact on lookup tables and payload size when splitting processes into multiple ResourceProfiles.

The goal of this issue is to start a discussion on these challenges with impacted SIGs and determine next steps.

# Context
The intent of this issue is to discuss potential incompatibilities between the current OpenTelemetry protocol specification for profiles and the [opentelemetry-ebpf-profiler](https://github.com/open-telemetry/opentelemetry-ebpf-profiler), with regards to usage of OTLP resources.

The [opentelemetry-ebpf-profiler](https://github.com/open-telemetry/opentelemetry-ebpf-profiler) is a profiler which allows CPU profiling for all processes running on a single host. 

The profiler will be built in a standalone collector distribution, where it would be configured as a collector receiver ([RFC](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/main/design-docs/159-collector-api)). This distribution would be deployed on each host for which the user wants to collect profiling data.

# Problem Statement

Currently, by default, OTLP profiles generated by the opentelemetry-ebpf-profiler **always contain a single [Profile](https://github.com/open-telemetry/opentelemetry-proto/blob/298a60fa953bbdcfa8be54a9dccb6f90cb0711d9/opentelemetry/proto/profiles/v1development/profiles.proto#L151-L182) object (in a single ScopeProfile, in a single ResourceProfile)**, which contains stack traces for all the processes running on the host (when off CPU profiling is enabled, this adds another ScopeProfile object, under the same ResourceProfile).

As a consequence:
* **The [top level resource](https://github.com/open-telemetry/opentelemetry-proto/blob/298a60fa953bbdcfa8be54a9dccb6f90cb0711d9/opentelemetry/proto/profiles/v1development/profiles.proto#L117-L119) attributes only hold general information about the host** on which the profiler is running.
[...]

PR #634, opened 2025-03-05
profiles: allow optional higher granularity resources

Gandem:
Exploring a potential solution to https://github.com/open-telemetry/opentelemetry-proto/issues/628 (please refer to the problem statement in that issue!)

This adds a new `sub_resources` field in `ResourceProfiles`, with the goal of using it in the opentelemetry-ebpf-profiler:
* The `resource` field will refer to the host, as a resource, and hold host-level attributes.
* The `sub_resources` field will contain higher granularity resources (for example, container, pod, ...), and hold relevant attributes. Sub-resources are then referred to by their index in the list in each sample.

For example:
```
ResourceProfiles.sub_resources = [{"container_id": "deadbeef"}]

Sample.sub_resource_index = 0
```

Means that the sample has been taken by the profiler for a process running in a container with the `container_id`: `deadbeef`.

This approach has the following upsides:
* **No performance impact:** for lookup tables, as opposed to other alternatives (such as splitting the host profiles to multiple resource profiles, one per sub-resource), this approach allows keeping all samples in a single `Profiles` object, and have them share the underlying lookup tables. It also doesn't impact merging multiple profiling payload (for example in a collector gateway).
* **Minimal changes to the spec and implementation:** Implementing this in the profiler would not require an important overhaul. Additionally, most components handling the profiles signal payload (processors, runtime profilers, ...) will be able to ignore the sub_resources field, when not relevant. For the `k8sattributeprocessor`, the implementation can just iterate on the `sub_resources` field and apply the same processing that is applied on each individual `resource`'s attributes.
* **Storage efficiency:** sub-resource level attributes will be de-duplicated between samples, as each sample will refer to the sub-resource (instead of containing all the container-level attributes, for example).

This will introduce a slight discrepancy with other signals, however, such a change _could_ generally make sense for host-level collected observability (e.g. traces with Beyla?).


christos68k:
I think it's helpful to examine why we'd want to expose what's currently per-`Sample` resource information in a more convenient form than having it buried with each `Sample`:

1. `k8sattributeprocessor` can enrich `ResourceProfiles` with container/K8s metainformation without having to parse `eeper (`Sample`-level) into the message.
2. Other processors can filter based on that information, without also having to parse `Sample`-level information.

[Maybe other reasons]

Does this proposal solve these issues? I think we still can't easily do 1. as `k8sattributeprocessor` will still need to iterate and enrich `Samples`, even though it now writes that information to `ResourceProfiles` instead of each `Sample`. Other processors still can't easily filter based on `ResourceProfiles.sub_resources` as the data to filter is buried into the message and they'll need to parse and restructure it.

PR #597, opened 2024-11-12
Clarify experimental additions to Stable components

tigrannajaryan:
The Entities SIG is planning a backward-compatible addition to the Resource, see the prototype here: https://github.com/tigrannajaryan/opentelemetry-proto/pull/2

The change is compliant with existing stability requirements in this repository.

However, our stability definition does not specify how such experimental additions to stable signals are expected to be done and what happens during and after the experiment.

This PR adds more clarity in the process and further refines the stability definition to reinforce our strong commitment ot OTLP's stability.

PR #635, opened 2025-03-06
Add EntityRef and use it in Resource.

jsuereth:
As per [OTEP 264](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/entities/0264-resource-and-entities.md) - This is the proposal for representing "Entity" in OTLP.

- Adds a new `EntityRef` message where an Entity is an object of interest in observability.
  - Entity has a type (e.g. `host`, `k8s.pod`, `service`).
  - Entity has a set of identifying attributes (these contribute to signal identity)
  - Entity has a set of descriptive attirbutes (these are used to aide signal retreival/querying).
  - Entity can be defined and versioned (e.g. semantic conventions via schema_url).
- Adds a new `entity_refs` field to `Resource` that allows distinguishing which key-value pairs in `Resource` come from which reported `Entity`.

Please read the OTEP for details and rationale behind this protocol design.  High level this meant to be:

- Sufficient to represent entities
- Non-breaking for all OpenTelemetry users.

Prototypes using this protocol

- java: https://github.com/open-telemetry/opentelemetry-java/pull/6855
- collector: https://github.com/open-telemetry/opentelemetry-collector/pull/11958
- go: https://github.com/open-telemetry/opentelemetry-go/pull/5918/

### DIFF
--- a/opentelemetry/proto/common/v1/common.proto
+++ b/opentelemetry/proto/common/v1/common.proto
@@ -90,7 +90,7 @@ message EntityRef {
   // https://opentelemetry.io/docs/specs/otel/schemas/#schema-url
   //
   // This schema_url applies to the data in this message and to the Resource attributes
-  // referenced by id_attr_keys and descr_attr_keys.
+  // referenced by id_keys and description_keys.
   // TODO: discuss if we are happy with this somewhat complicated definition of what
   // the schema_url applies to.
   //
--- a/opentelemetry/proto/common/v1/common.proto
+++ b/opentelemetry/proto/common/v1/common.proto
@@ -81,7 +81,7 @@ message InstrumentationScope {
 }
 
 // A reference to an Entity.
-// Entity represents an object of interest associated with produced telemetry: e.g spans, metrics or logs.
+// Entity represents an object of interest associated with produced telemetry: e.g spans, metrics, profiles, or logs.
 //
 // Status: [Development]
 message EntityRef {
--- a/opentelemetry/proto/common/v1/common.proto
+++ b/opentelemetry/proto/common/v1/common.proto
@@ -79,3 +79,37 @@ message InstrumentationScope {
   repeated KeyValue attributes = 3;
   uint32 dropped_attributes_count = 4;
 }
+
+// A reference to an Entity.
+// Entity represents an object of interest associated with produced telemetry: e.g spans, metrics or logs.
+//
+// Status: [Development]
+message EntityRef {
+  // The Schema URL, if known. This is the identifier of the Schema that the entity data
+  // is recorded in. To learn more about Schema URL see
+  // https://opentelemetry.io/docs/specs/otel/schemas/#schema-url
+  //
+  // This schema_url applies to the data in this message and to the Resource attributes
+  // referenced by id_attr_keys and descr_attr_keys.
+  // TODO: discuss if we are happy with this somewhat complicated definition of what
+  // the schema_url applies to.
+  //
+  // This field obsoletes the schema_url field in ResourceMetrics/ResourceSpans/ResourceLogs.
+  string schema_url = 1;
+
+  // Defines the type of the entity. MUST not change during the lifetime of the entity.
+  // For example: "service" or "host". This field is required and MUST not be empty
+  // for valid entities.
+  string type = 2;
+
+  // Attribute Keys that identify the entity.
+  // MUST not change during the lifetime of the entity. The Id must contain at least one attribute.
+  // These keys MUST exist in the containing {message}.attributes.
+  repeated string id_keys = 3;
+
+  // Descriptive (non-identifying) attribute keys of the entity.
+  // MAY change over the lifetime of the entity. MAY be empty.
+  // These attribute keys are not part of entity's identity.
+  // These keys MUST exist in the containing {message}.attributes.
+  repeated string description_keys = 4;
+}
--- a/opentelemetry/proto/resource/v1/resource.proto
+++ b/opentelemetry/proto/resource/v1/resource.proto
@@ -37,7 +37,7 @@ message Resource {
 
   // Set of entities that participate in this Resource.
   //
-  // Note: keys in the references MUST existing in attributes of this message.
+  // Note: keys in the references MUST exist in attributes of this message.
   //
   // Status: [Development]
   repeated opentelemetry.proto.common.v1.EntityRef entity_refs = 3;
--- a/opentelemetry/proto/resource/v1/resource.proto
+++ b/opentelemetry/proto/resource/v1/resource.proto
@@ -34,4 +34,11 @@ message Resource {
   // dropped_attributes_count is the number of dropped attributes. If the value is 0, then
   // no attributes were dropped.
   uint32 dropped_attributes_count = 2;
+
+  // Set of entities that participate in this Resource.
+  //
+  // Note: keys in the references MUST existing in attributes of this message.
+  //
+  // Status: [Development]
+  repeated opentelemetry.proto.common.v1.EntityRef entity_refs = 3;
 }


opentelemetry/proto/common/v1/common.proto
```diff
@@ -79,3 +79,37 @@ message InstrumentationScope {
   repeated KeyValue attributes = 3;
   uint32 dropped_attributes_count = 4;
 }
+
+// A reference to an Entity.
+// Entity represents an object of interest associated with produced telemetry: e.g spans, metrics or logs.
+//
+// Status: [Development]
+message EntityRef {
```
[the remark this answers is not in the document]
jsuereth: Yes - that's exactly the reason

dyladan (approved):

opentelemetry/proto/common/v1/common.proto
Kielek, line 87: Both, class and field are marked as `Development` and put into the stable proto files. 

What is the strategy to release/version it?

opentelemetry/proto/common/v1/common.proto
```diff
@@ -79,3 +79,37 @@ message InstrumentationScope {
   repeated KeyValue attributes = 3;
   uint32 dropped_attributes_count = 4;
 }
+
+// A reference to an Entity.
+// Entity represents an object of interest associated with produced telemetry: e.g spans, metrics or logs.
```
smith, line 84:
```suggestion
// Entity represents an object of interest associated with produced telemetry: e.g spans, metrics, profiles, or logs.
```

opentelemetry/proto/common/v1/common.proto
```diff
@@ -79,3 +79,37 @@ message InstrumentationScope {
...
+//
+// Status: [Development]
+message EntityRef {
+  // The Schema URL, if known. This is the identifier of the Schema that the entity data
+  // is recorded in. To learn more about Schema URL see
+  // https://opentelemetry.io/docs/specs/otel/schemas/#schema-url
+  //
+  // This schema_url applies to the data in this message and to the Resource attributes
+  // referenced by id_attr_keys and descr_attr_keys.
```
smith, line 93:
```suggestion
  // referenced by id_keys and description_keys.
```

opentelemetry/proto/resource/v1/resource.proto
```diff
@@ -34,4 +34,11 @@ message Resource {
   // dropped_attributes_count is the number of dropped attributes. If the value is 0, then
   // no attributes were dropped.
   uint32 dropped_attributes_count = 2;
+
+  // Set of entities that participate in this Resource.
+  //
+  // Note: keys in the references MUST existing in attributes of this message.
```
smith, line 40:
```suggestion
  // Note: keys in the references MUST exist in attributes of this message.
```

smith (approved):
Looks good; just a few suggestions for comment changes.

opentelemetry/proto/resource/v1/resource.proto
```diff
@@ -34,4 +34,11 @@ message Resource {
...
+  //
+  // Status: [Development]
+  repeated opentelemetry.proto.common.v1.EntityRef entity_refs = 3;
```
jack-berg, line 43: Imagine I'm consuming this data. I see data coming in from the opentelemetry java agent. It has a resource with `service.name` other `service.*` attributes, since it represents data from a service, and `host.*` attributes which the otel java agent automatically detects and includes. It has two entities in `entity_refs`, representing a `service` and `host`:
- `{"type":"service","id_keys":[...],"descriptive_keys":[...]}`
- `{"type":"host","id_keys":[...],"descriptive_keys":[...]}`

The corresponding [entities data model PR](https://github.com/open-telemetry/opentelemetry-specification/pull/4442/files) says:

> For example, telemetry produced using OpenTelemetry SDK is normally associated with
a `service` entity. Similarly, OpenTelemetry defines system metrics for a `host`. The `host` is the
entity we want to associate metrics with in this case.

> Entities may be also associated with produced telemetry indirectly.
For example a service that produces
telemetry is also related with a process in which the service runs, so we say that
the `service` entity is related to the `process` entity. The process normally also runs
on a host, so we say that the `process` entity is related to the `host` entity.

Based on the data contained in these proto messages, how do I determine which entity produced the data (i.e. `service` in this case) vs which entities are related (i.e. `host` in this case)?

jsuereth: We are not choosing a producing entity.  We actually aren't sure we can and still support an open ecosystem.  

A more long winded narrative on the problem:

Specifically, think about `container` and `k8s.container`.   When generating docker metrics, you could consider both of these the producing entity.  Similarly when thinking about `k8s.node`, `host` and `aws.ec2.instance` (or you favorite cloud provider here), *all* of the above would be a producing entity when it comes to host metrics.

One of our principles in Entities was that we don't want to force the kubernetes resource detectors (and ecosystem) to fully understand all possible cloud providers to generate valid "ids" and "entities" on a resource.  This means they need a way to interact with each other such that if just `k8s.node` exists, things are fine, but if `k8s.node`, `host` and `aws.ec2.instance` all exist, things are also fine.  See the ["telescoping" principle](https://github.com/open-telemetry/opentelemetry-specification/pull/4442/files#diff-9cdfc29107083d5809260d026ac978393fdac80f00dbbd16187fb10fccce9f55R53) in the corresponding spec PR.

In the past year of working, we never had a proposal that would allow us to identify a single producing entity without ALSO creating hard coupling between entity types and ecosystems, thereby locking all entity definitions into OpenTelemetry itself and semantic conventions.  This is why we're going with the flexible approach.

In practice, we think the producing entity will be derived from the relationships.  For example, if I get a resource that contains four entities:  `host`, `process`, `service.instance` and `service` from a Java SDK.  I would have a set of relationships:

- `process` "runs on" `host
- `service.instance` "run on" `process`
- `service.instance` "is part of" `service`

We can derive that `service.instance` is the producing entity as it's the "bottom" of the hierarchy.

In practice, the things folks want to *use* resource for are all applicable (i.e. making queries by process or service name or host name faster), and the identity is no less strong for having the extra information beyond `service.instance.id` (a UUID).

We'd written a bunch of docs discussing this in the SIG, Perhaps @dmitryax can add some color context here.

opentelemetry/proto/common/v1/common.proto
jsuereth: Great question - We documented the answer in https://github.com/open-telemetry/opentelemetry-proto/pull/597

If you need further details, happy to provide them.  

I will clarify, this proposal had a high-bar before opening the PR. We have three prototypes in different language SDKs (@dyladan just added javascript) and a Collector prototype.  So we're confident this is the right model going forward, but want to land the rest of the specification and SDK/Collector work before marking this as stable.

opentelemetry/proto/resource/v1/resource.proto
jack-berg: > For example, if I get a resource that contains four entities: host, process, service.instance and service from a Java SDK. I would have a set of relationships: [...] We can derive that service.instance is the producing entity as it's the "bottom" of the hierarchy.

Would you model this in semantic conventions?

How would you extend this logic to this: 

I get a resource that contains two entities from the collector's `hostmetrics` receiver: `host` and `container`. `host` since its the the host metrics receiver, `container` since the collector is running in a container. 

* `container` "runs on" `host` 

Suggesting that `container` is the bottom of the hierarchy, and thus that `container` is the producing entity. Yet in this case the `host` ought to be the producing entity.

jsuereth: > Would you model this in semantic conventions?

Yes.  See [our proposed guide](https://docs.google.com/document/d/1GmB0kDx9opHxjj2oy6A4h0fqE8UVW-3ebWgp6zhtbVE/edit?tab=t.0#heading=h.jcpv5hm93mzq) we discussed with semconv SIG.

> I get a resource that contains two entities from the collector's hostmetrics receiver: host and container. host since its the the host metrics receiver, container since the collector is running in a container.

I'd actually consider this an invalid configuration or bug.  While a user could configure the collector to attach a container entity to the resource for host metrics, we would recommend against this in docs.

The collector, specifically, has to be more careful with resource as it can be a *remote* observer, so just adding entities discovered about where the collector is running is often not the right thing.  E.g. see `k8sattribute` vs. `resourcedetection` processor - One can handle remotely observing kubernetes systems, the other should only be used when observing locally.

opentelemetry/proto/common/v1/common.proto
Kielek: @jsuereth, thank you. I was not aware about this PR.

Based on what I see, after release, the field(repeated opentelemetry.proto.common.v1.EntityRef entity_refs = 3;)  in resources will stay with definition forever (in the worst case as Obsolete).

I would like to just double check if it is the goal and confirm that it is mature enough decision.

jsuereth: We've been prototyping this change for the past year, so yes - we consider this a mature enough decision to propose now.

We've been working with a fork of open-telemetry proto for quite a while as we explored this and have decided this is our path forward.  Otherwise I would not have raised this PR

carlosalberto (approved):

opentelemetry/proto/resource/v1/resource.proto
dmitryax: Potentially, we could add another field to this message that would identify if it's the emitting entity vs a parent. That wouldn't be a breaking change. The most important decision is that we want start with one resource with multiple entities. Looks there are no objections to that.

Kielek (approved):

PR #636, opened 2025-03-17
Generate Mermaid diagrams for proto files

smith:
These were generated by [proto-gen-md-diagrams](https://github.com/GoogleCloudPlatform/proto-gen-md-diagrams).

This is not meant to be merged as-is, since I just generated these in the root directory, but if people find them useful maybe the diagrams could be added to the docs.


dmathieu:
If we add this, how about a CI check that ensures the doc is always up to date? (generates the doc, and checks the GIT repository is still clean).
I'd also move this under a subfolder in `docs`.

PR #644, opened 2025-04-16
profiles: Move lookup tables to ProfilesData

florianl:
Implements https://github.com/open-telemetry/opentelemetry-proto/issues/628#issuecomment-2772566325.

PR #645, opened 2025-04-17
Profiles: simplify profile stack trace representation

felixge:
## Changes

- Introduce a first-class Stack message type and lookup table.
- Replace location index range based stack trace encoding on Sample with a single stack_index reference.
- Remove the location_indices lookup table.

## Motivation

**tl;dr:** Let's KISS. The stack trace encoding proposed in this PR is competitive with more advanced stack trace encoding formats, but much simpler to implement. See table below or @christos68k's [benchmark for details](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/524).

|              | Uncompressed | vs Dedup | Compressed | vs Dedup |
| ------------ | ------------ | -------- | ---------- | -------- |
| ExportDedup (current)  | 688709       | 0.0%    | 277955     | 0.0%    |
| ExportStacks (this PR) | 703704       | 2.8%    | 277713     | \-0.1%  |
| ExportArrays ([PR #651][array-encoding proposal]) | 602920       | \-12.5% | 305235     | 9.8%    |

The current stack trace encoding is somewhat complex as it tries to optimize for CPU profiling stack traces where it's relatively common to see prefixes of a stack trace.

For example, a program that looks like this ...

```a.go
func a() { b() }
func b() { c() }

func c() {
	for {
		d()
		e()
	}
}
```

... will produce a CPU profile with stack traces like the ones shown below. The number indicate samples taken at different locations (program counters) within the same function:

```
a;b;c;d1
a;b;c;d2
a;b;c;d3
...
a;b;c;e1
a;b;c;e2
a;b;c;e3
...
a;b;c
```

This makes it attractive to optimize away the prefix repetition (`a;b;c`) in the stack trace dictionary table. But the current format doesn't quite succeed in this, as it would only allow us to get rid of the last stack trace in the table. This probably explains why it doesn't end up performing very differently from the encoding proposed in this PR.

The alternative [double array proposal][array-encoding proposal] from @aalexand achieves over 12% reduction in profile size before compression, because it can exploit this pattern effectively. However, for some reason the resulting data patterns end up being much harder to compress, and the encoding loses by almost 10% after compression has been applied.

Last but not least, not all profiling data will benefit from prefix-encoding as much as CPU profiles. For example memory allocations and especially Off-CPU stack traces have significantly less prefix repetition. We don't have a good benchmark for this, but I hope this argument is easy to follow regardless.

Given the above, I'm proposing to chose the most trivial stack trace encoding format as outlined in this PR as it seems to hit a sweet spot between complexity and efficiency. The main tasks for reviewers is to validate benchmarks from @christos68k's and confirm if they're okay with this direction.

It seems like we had rough consensus around this change in previous SIG meetings.

[array-encoding proposal]: https://github.com/open-telemetry/opentelemetry-proto/pull/651
[timestamp proposal]: https://github.com/open-telemetry/opentelemetry-proto/pull/594
[off-cpu pr]: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/196
[benchmarks]: https://docs.google.com/spreadsheets/d/1Q-6MlegV8xLYdz5WD5iPxQU2tsfodX1-CDV1WeGzyQ0/edit?gid=2069300294#gid=2069300294

### DIFF
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -99,7 +99,7 @@ message ProfilesDictionary {
   // so that _index fields can use 0 to indicate null/unset.
   repeated Mapping mapping_table = 1;
 
-  // Locations referenced by samples via Profile.location_indices.
+  // Locations referenced by samples via Stack.location_indices.
   repeated Location location_table = 2;
 
   // Functions referenced by locations via Line.function_index.
@@ -119,6 +119,9 @@ message ProfilesDictionary {
 
   // Represents a mapping between Attribute Keys and Units.
   repeated AttributeUnit attribute_units = 7;
+
+  // Stacks referenced by samples via Sample.stack_index.
+  repeated Stack stack_table = 8;
 }
 
 // ProfilesData represents the profiles data that can be stored in persistent storage,
@@ -201,18 +204,19 @@ message ScopeProfiles {
 //   information present to determine the original sampled values.
 //
 // - The profile is represented as a set of samples, where each sample
-//   references a sequence of locations, and where each location belongs
+//   references a stack trace which is a list of locations, each belonging
 //   to a mapping.
-// - There is a N->1 relationship from sample.location_id entries to
-//   locations. For every sample.location_id entry there must be a
+// - There is a N->1 relationship from Stack.location_indices entries to
+//   locations. For every Stack.location_indices entry there must be a
 //   unique Location with that index.
 // - There is an optional N->1 relationship from locations to
 //   mappings. For every nonzero Location.mapping_id there must be a
 //   unique Mapping with that index.
 
-// Represents a complete profile, including sample types, samples,
-// mappings to binaries, locations, functions, string table, and additional metadata.
-// It modifies and annotates pprof Profile with OpenTelemetry specific fields.
+// Represents a complete profile, including sample types, samples, mappings to
+// binaries, stacks, locations, functions, string table, and additional
+// metadata. It modifies and annotates pprof Profile with OpenTelemetry
+// specific fields.
 //
 // Note that whilst fields in this message retain the name and field id from pprof in most cases
 // for ease of understanding data migration, it is not intended that pprof:Profile and
@@ -227,27 +231,24 @@ message Profile {
   // The set of samples recorded in this profile.
   repeated Sample sample = 2;
 
-  // References to locations in ProfilesDictionary.location_table.
-  repeated int32 location_indices = 3;
-
-  // The following fields 4-14 are informational, do not affect
+  // The following fields 3-13 are informational, do not affect
   // interpretation of results.
 
   // Time of collection (UTC) represented as nanoseconds past the epoch.
-  int64 time_nanos = 4;
+  int64 time_nanos = 3;
   // Duration of the profile, if a duration makes sense.
-  int64 duration_nanos = 5;
+  int64 duration_nanos = 4;
   // The kind of events between sampled occurrences.
   // e.g [ "cpu","cycles" ] or [ "heap","bytes" ]
-  ValueType period_type = 6;
+  ValueType period_type = 5;
   // The number of events between sampled occurrences.
-  int64 period = 7;
+  int64 period = 6;
   // Free-form text associated with the profile. The text is displayed as is
   // to the user by the tools that read profiles (e.g. by pprof). This field
   // should not be used to store any machine-readable information, it is only
   // for human-friendly content. The profile must stay functional if this field
   // is cleaned.
-  repeated int32 comment_strindices = 8; // Indices into ProfilesDictionary.string_table.
+  repeated int32 comment_strindices = 7; // Indices into ProfilesDictionary.string_table.
 
   // A globally unique identifier for a profile. The ID is a 16-byte array. An ID with
   // all zeroes is considered invalid. It may be used for deduplication and signal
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -373,11 +373,8 @@ message ValueType {
 // augmented with auxiliary information like the thread-id, some
 // indicator of a higher level request being handled etc.
 message Sample {
-  // locations_start_index along with locations_length refers to to a slice of locations in Profile.location_indices.
-  int32 locations_start_index = 1;
-  // locations_length along with locations_start_index refers to a slice of locations in Profile.location_indices.
-  // Supersedes location_index.
-  int32 locations_length = 2;
+  // Reference to stack in Profile.stack_table.
+  int32 stack_index = 1;
```
beaubelgrave, line 377: It would still be nice to have an optional location_index to indicate a location in addition to the common Stack locations. For on-CPU profiling (or memory profiling with virtual top-frame insertions for the memory type) it would useful to have common locations mixed with per-sample location to aid in reducing the size of Stack entries.\

In other words, there are often times that many locations are the same and only differ by the leaf location due to skidding or intentional splits/label of a stack.

rockdaboot: Maybe we can come up with numbers for certain kinds of scenarios!?

Iirc, when we did this some years ago (very mixed language scenario), the leaf frame was responsible for ~25% of additional cardinality of stacktraces. Using a graph (same as prefix compression for reversed stacktraces) showed a ~60% reduction of cardinality. 
So removing the leaf frame from the `Stack` seems like a good compromise between this PR and a graph-based prefix compression, which may be too CPU intensive for an agent that strives for < 1% CPU usage.

If there is enough interest and time allows, I could figure out the numbers for a Java heavy workload (80% Elasticsearch stacktraces from 200+ clusters).

florianl (approved):
Especially mixed reports of profiles, e.g. on- and off-CPU profiles, will benefit from this approach.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -197,10 +197,10 @@ message Profile {
   // If multiple binaries contribute to the Profile and no main
   // binary can be identified, mapping[0] has no special meaning.
   repeated Mapping mapping_table = 3;
-  // Locations referenced by samples via location_indices.
+  // Locations referenced by stacks via location_indices.
   repeated Location location_table = 4;
-  // Array of locations referenced by samples.
-  repeated int32 location_indices = 5;
+  // Stacks referenced by samples via stack_index.
+  repeated Stack stack_table = 5;
```
aalexand, line 203: The table should move to ProfilesData given that #644 was merged.

christos68k: Let's wait for #650 to get merged as it moves all lookup tables outside of `ProfilesData`.

opentelemetry/proto/profiles/v1development/profiles.proto
aalexand: We can consider something like #651 if we want to optimize for this.

christos68k:
See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/524 for benchmarks that also cover this proposal.

opentelemetry/proto/profiles/v1development/profiles.proto
felixge: Done.

felixge:
This PR has been rebased and the description has been updated. It's ready for another round of review.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -448,6 +446,13 @@ message Mapping {
   bool has_inline_frames = 9;
 }
 
+// A Stack represents a stack trace as a list of locations. The first location
```
christos68k, line 449: We should also add this to and update the helper ascii graph at the beginning of the file.

christos68k (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
aalexand: Resolving this conversation, we are not going to do #651, we are going with the approach in this PR.

opentelemetry/proto/profiles/v1development/profiles.proto
aalexand, line 449: nit: "The first location is the leaf frame" - would it make sense to put this in the field comment rather than message?

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -119,6 +119,9 @@ message ProfilesDictionary {
 
   // Represents a mapping between Attribute Keys and Units.
   repeated AttributeUnit attribute_units = 7;
+
+  // Stacks referenced by samples via Sample.stack_index.
+  repeated Stack stack_table = 8;
```
aalexand, line 124: Maybe group stack_table with location_table and friends and keep attribute_units last since it's somewhat special in semantics?

aalexand: I guess with #672 it's less special .

opentelemetry/proto/profiles/v1development/profiles.proto
christos68k: Opened https://github.com/felixge/opentelemetry-proto/pull/2

christos68k: I added this [here](https://github.com/felixge/opentelemetry-proto/pull/2/files).

christos68k:
> We are planning to release 1.8.0. Do you need this merged before that?

Before if possible, otherwise we'll need to trigger another release very soon.

tigrannajaryan:
> > We are planning to release 1.8.0. Do you need this merged before that?
> 
> Before if possible, otherwise we'll need to trigger another release very soon.

OK, in that case please address / close open comments and resolve the merge conflicts so that this PR can be merged. We can wait a day or so but if it is going to take much longer then it will have to wait until the release after that.

pellared:
@christos68k, should this PR be closed given https://github.com/open-telemetry/opentelemetry-proto/pull/708 is merged?

PR #648, opened 2025-04-30
Switch the otlp profiles root element to be ProfilesData

dmathieu:
The OTLP profiles endpoint takes a `ResourceProfiles` as its root slice. However, the data model wants a `ProfilesData`.
See https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L46-L66

This was unseen until now because ProfilesData only held ResourceProfiles.
But now that we have the various index maps in there, we need to handle this properly.

### DIFF
--- a/opentelemetry/proto/collector/profiles/v1development/profiles_service.proto
+++ b/opentelemetry/proto/collector/profiles/v1development/profiles_service.proto
@@ -31,12 +31,13 @@ service ProfilesService {
 }
 
 message ExportProfilesServiceRequest {
-  // An array of ResourceProfiles.
-  // For data coming from a single resource this array will typically contain one
-  // element. Intermediary nodes (such as OpenTelemetry Collector) that receive
-  // data from multiple origins typically batch the data before forwarding further and
-  // in that case this array will contain multiple elements.
-  repeated opentelemetry.proto.profiles.v1development.ResourceProfiles resource_profiles = 1;
+  // The ProfilesData.
+  // For data coming from a single resource this object will typically contain
+  // one element resource profiles. Intermediary nodes (such as OpenTelemetry
+  // Collector) that receive data from multiple origins typically batch the
+  // data before forwarding further and in that case there will be multiple
+  // resources.
+  opentelemetry.proto.profiles.v1development.ProfilesData profiles_data = 1;
 }
 
 message ExportProfilesServiceResponse {


florianl (approved):
Thanks for the quick fix!

rockdaboot (approved):

tigrannajaryan:
This deviates from how the other signals work. To make it consistent we could duplicate all tables into ExportProfilesServiceRequest but that seems too much duplication. I am not sure if there is a better way, so probably we should accept that ExportProfilesServiceRequest is not going to be consistent with other Request messages.

dmathieu:
Would it make more sense to keep the `ResourceProfiles` as they are, drop `ProfilesData` and move all the lookup tables into `ResourceProfiles`?

tigrannajaryan:
We made `Data` and `Request` distinct for the following reason: the `Request` is intended to be used for RPC calls, the `Data` is intended for local serialization. They are the same today but may evolve differently, e.g. we may add RPC-specific fields to the `Request` (see for example [this PR](https://github.com/open-telemetry/opentelemetry-proto/pull/219/files) that was not accepted, but gives an idea of what an RPC-specific field can be).

dmathieu:
We can move the lookup tables into `ResourceProfiles` and keep `ProfilesData` for future use then.

opentelemetry/proto/collector/profiles/v1development/profiles_service.proto
```diff
@@ -31,12 +31,13 @@ service ProfilesService {
...
-  // element. Intermediary nodes (such as OpenTelemetry Collector) that receive
-  // data from multiple origins typically batch the data before forwarding further and
-  // in that case this array will contain multiple elements.
-  repeated opentelemetry.proto.profiles.v1development.ResourceProfiles resource_profiles = 1;
+  // The ProfilesData.
+  // For data coming from a single resource this object will typically contain
+  // one element resource profiles. Intermediary nodes (such as OpenTelemetry
+  // Collector) that receive data from multiple origins typically batch the
+  // data before forwarding further and in that case there will be multiple
+  // resources.
```
christos68k, line 39: Can we add some of text from the proto here (in light of recent changes):

```suggestion
  // The ProfilesData which contains an array of ResourceProfiles.
  // For data coming from an SDK profiler, this array will typically contain one
  // element. Host-level profilers will usually create one ResourceProfile per
  // container, as well as one additional ResourceProfile grouping all samples
  // from non-containerized processes. Intermediary nodes (such as OpenTelemetry
  // Collector) that receive data from multiple origins typically batch the
  // data before forwarding further and in that case there will also be multiple
  // ResourceProfiles.
```

christos68k (approved):

jhalliday:
I'm inclined towards Alexey's #3 option...

Whilst I think we have good reasons for wanting profiles to deviate from the way other signal types work, I also think there are reusable patterns in our work that may flow back into future refinement of those older signal types or development of new ones. I'd like to find a clean way to express a pattern for more portability. For example

```
message ${Signal}ReferenceTables {
   repeated string string_table = 1;
   repeated Foo foo_table = 2;
   repeated Bar bar_table = 3;
   ...
}

message ${Signal}Data {
  repeated Resource${Signal} resource_${signal} = 1;
  ${Signal}ReferenceTables lookup_tables = 2;
}

message Export${Signal}ServiceRequest {
  repeated opentelemetry.proto.${signal}.Resource${Signal} resource_${signal} = 1;
  opentelemetry.proto.${signal}.${Signal}ReferenceTables lookup_tables = 2;
}
```

Any signal which may benefit from the space savings of shared reference data may define a signal-specific bundle of lookup tables as a message and if desired retrofit it in a backwards compatible manner to its existing `_Data` or `Export_ServiceRequest` messages.

Treating the reference tables as a separate building block may also allow for more abstract handling in tools or pipelines that wish to e.g. reduce a stream of (tables,datapoints),(tables,datapoints),... tuples by merging the tables and recomputing-in-place the datapoints to point to the unified tables, which seems likely to be a common and desirable space saving task, particularly for back-end ingest.
Likewise, and at risk of reopening the stateful vs stateless protocol debate, adding a unique identifier to a `${Signal}ReferenceTables` structure would allow for a set of samples (e.g. `message Profile`) to explicitly say which tables message they are referring to rather than implicitly referring to the one in the same message/file heirarchy, opening the way for cases where a `_Data` or `Export_ServiceRequest` doesn't populate the tables field, but carries via the resources field, samples which refer to tables stored or transmitted in a separate `Data` or `ServiceRequest` instance.

Scope creep is a thing, but framing the change thus as an optional, flexible and compatible space saving addition to the existing Data/ExportRequest structures rather than a redefinition/repurposing/(polluting?) of them is perhaps a more interesting way of considering the proposal, particularly at TC level.

tigrannajaryan:
> PUlling out all lookup tables into a separate message makes sense.

+1.

dmathieu:
This is superseded by #650.

Issue #633, opened 2025-03-05
[pdata/pprofile] proto docs may need to update


YifanYang6:
### Component(s)

pdata/pprofile

### Describe the issue you're reporting

Hello,

I've encountered a potential inconsistency between the proto definition for profile data and the actual data observed during debugging. Specifically, I'm concerned with the behavior of the Value and TimestampsUnixNano fields within the Sample structure.

According to the proto definition https://github.com/open-telemetry/opentelemetry-collector/blob/7ee8a2b86015d380f9217dbe3dfe12f620cfe800/pdata/internal/data/protogen/profiles/v1development/profiles.pb.go#L745-L767 it states:

"All samples must have the same number of values, the same as the length of Profile.sample_type"

However, my debugging output shows the following:

Each Value entry appears to correspond to a specific Timestamp, rather than being determined by the Profile.sample_type length.
Samples within the same profile do not consistently have the same number of Value or Timestamp entries.
Here's an example of my debugging output:
```
Profile open-telemetry/opentelemetry-collector#1
    Profile ID     : c2d01967aa7b578f1b70e3c35fd6e379
    SampleType #%d : %!s(int=0)
    TypeStringIndex: events
    UnitStringIndex: nanoseconds
    Start time     : 2025-03-05 04:56:59.894095226 +0000 UTC
    Duration       : 1970-01-01 00:00:01.756312228 +0000 UTC

    Sample #0
        Location start: 0
        Location length: 6
        Value: [903976986 839861575 883929035 463890732 668006176]
        Attributes:
             -> container.id: /
             -> thread.name: kworker/0:0
             -> process.pid: %!s(int64=2548176)
        Timestamp: 1741150617770086999
        Timestamp: 1741150618610126241
        Timestamp: 1741150619494089058
        Timestamp: 1741150619958085800
        Timestamp: 1741150620626113008
    Sample open-telemetry/opentelemetry-collector#1
        Location start: 6
        Location length: 6
        Value: [20025858 1007966728 7908106]
        Attributes:
             -> container.id: /
             -> thread.name: kworker/12:2
             -> process.pid: %!s(int64=4041331)
        Timestamp: 1741150618074111301
        Timestamp: 1741150619082111579
        Timestamp: 1741150619090073827
```
Based on these observations, I have the following questions:

1. What is the intended relationship between Value, TimestampsUnixNano, and Profile.sample_type?
2. Why do individual samples contain multiple Value and Timestamp entries? What does this represent?
[...]

PR #649, opened 2025-05-02
profiles: use single Profile.sample_type and clarify use of timestamps

florianl:
Implements https://github.com/open-telemetry/opentelemetry-proto/issues/633#issuecomment-2772648117.


opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -207,13 +207,10 @@ message ScopeProfiles {
 message Profile {
   // A description of the samples associated with each Sample.value.
```
christos68k, line 208:
```suggestion
  // The type and unit of all Sample.values in this profile.
```

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -207,13 +207,10 @@ message ScopeProfiles {
...
   // For a cpu profile this might be:
-  //   [["cpu","nanoseconds"]] or [["wall","seconds"]] or [["syscall","count"]]
+  //   ["cpu","nanoseconds"] or ["wall","seconds"] or ["syscall","count"]
```
christos68k, line 210:
```suggestion
  //   ["cpu","count"] or ["off_cpu","nanoseconds"]
```

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -207,13 +207,10 @@ message ScopeProfiles {
...
   // For a heap profile, this might be:
-  //   [["allocations","count"], ["space","bytes"]],
-  // If one of the values represents the number of events represented
-  // by the sample, by convention it should be at index 0 and use
-  // sample_type.unit == "count".
-  repeated ValueType sample_type = 1;
+  //   ["allocations","count"] or ["space","bytes"],
```
christos68k, line 212:
```suggestion
  //   ["allocated_objects","count"] or ["allocated_space","bytes"],
```

christos68k:
> Should default sample type go to the scope message?

That would be less flexible as it'd force us to have separate instrumentation scopes (`ScopeProfiles`) for different sample types. With the current proposal that's a possibility but not a requirement. The OTel documentation for [instrumentation scope](https://opentelemetry.io/docs/concepts/instrumentation-scope/) is sufficiently vague to allow both possibilities. For example, with the OTel eBPF profiling agent, we could have a single instrumentation scope (the eBPF profiler receiver/library) produce profiles with different sample types (cpu and memory).

Do you have a use-case in mind where associating different sample types with separate instrumentation scopes would be a benefit?

aalexand:
> > Should default sample type go to the scope message?
> 
> That would be less flexible as it'd force us to have separate instrumentation scopes (`ScopeProfiles`) for different sample types. With the current proposal that's a possibility but not a requirement. The OTel documentation for [instrumentation scope](https://opentelemetry.io/docs/concepts/instrumentation-scope/) is sufficiently vague to allow both possibilities. For example, with the OTel eBPF profiling agent, we could have a single instrumentation scope (the eBPF profiler receiver/library) produce profiles with different sample types (cpu and memory).
> 
> Do you have a use-case in mind where associating different sample types with separate instrumentation scopes would be a benefit?
> 
> (I don't have a strong opinion on this, just that if we don't have a good reason to force a separate instrumentation scope per sample type, I'd go for more flexibility).

I mostly want to make sure there is some way to capture the default sample type. The current PR removes the field IIUC.

florianl:
> Should default sample type go to the scope message?
> I mostly want to make sure there is some way to capture the default sample type. The current PR removes the field IIUC.

The main purpose of the proposal is to align values with timestamps as laid out in https://github.com/open-telemetry/opentelemetry-proto/issues/633#issuecomment-2772648117.

In general [google/pprof](https://github.com/google/pprof/blob/main/proto/profile.proto) it is possible to have multiple `sample_types` per `Profile` with special handling of [sample_type.unit=count](https://github.com/google/pprof/blob/c0086092b71a70d39d35d5a1130a8938b2f0e31a/proto/profile.proto#L52-L54) but does not have detailed timestamps but [just a durations](https://github.com/google/pprof/blob/c0086092b71a70d39d35d5a1130a8938b2f0e31a/proto/profile.proto#L78-L81). With a slightly more focus on timestamps in OTel Profiling, I think, the concept of having multiple `sample_types`, like in [google/pprof](https://github.com/google/pprof/blob/c0086092b71a70d39d35d5a1130a8938b2f0e31a/proto/profile.proto#L47-L55), conflicts.  IIRC @felixge and I discussed the situation, where one wants to report profiles with different sample types and we concluded that it should be fine, to have multiple `message Profile` with their dedicated `sample_type`. I think, it also reduces complexity by not having a concept of a default sample_type and simplifies also the handling of `message Profiles` in the backend, if they just want to focus on a certain `sample_type`.

christos68k:
> > Should default sample type go to the scope message?
> > I mostly want to make sure there is some way to capture the default sample type. The current PR removes the field IIUC.
> 
> The main purpose of the proposal is to align values with timestamps as laid out in [#633 (comment)](https://github.com/open-telemetry/opentelemetry-proto/issues/633#issuecomment-2772648117).
> 
> In general [google/pprof](https://github.com/google/pprof/blob/main/proto/profile.proto) it is possible to have multiple `sample_types` per `Profile` with special handling of [sample_type.unit=count](https://github.com/google/pprof/blob/c0086092b71a70d39d35d5a1130a8938b2f0e31a/proto/profile.proto#L52-L54) but does not have detailed timestamps but [just a durations](https://github.com/google/pprof/blob/c0086092b71a70d39d35d5a1130a8938b2f0e31a/proto/profile.proto#L78-L81). With a slightly more focus on timestamps in OTel Profiling, I think, the concept of having multiple `sample_types`, like in [google/pprof](https://github.com/google/pprof/blob/c0086092b71a70d39d35d5a1130a8938b2f0e31a/proto/profile.proto#L47-L55), conflicts. IIRC @felixge and I discussed the situation, where one wants to report profiles with different sample types and we concluded that it should be fine, to have multiple `message Profile` with their dedicated `sample_type`. I think, it also reduces complexity by not having a concept of a default sample_type and simplifies also the handling of `message Profiles` in the backend, if they just want to focus on a certain `sample_type`.

@aalexand (I misread your previous comment and thought you wanted to lift `sample_type` to the scope message). Regarding `default_sample_type`, are you referring to introducing it in the scope message in order to make `sample_type` in each profile optional or do you have something else in mind? Lile @florianl mentioned the current proposal splits different `sample_types` across different profiles.

aalexand:
> > > Should default sample type go to the scope message?
> > > I mostly want to make sure there is some way to capture the default sample type. The current PR removes the field IIUC.
> 
> @aalexand (I misread your previous comment and thought you wanted to lift `sample_type` to the scope message). Regarding `default_sample_type`, are you referring to introducing it in the scope message in order to make `sample_type` in each profile optional or do you have something else in mind? Lile @florianl mentioned the current proposal splits different `sample_types` across different profiles.

As one example, Go heap profiles [have four sample types](https://github.com/golang/go/blob/68ca584ffaf3f647eda2706500fba5280f2ad19c/src/runtime/pprof/protomem.go#L20) today: `alloc_objects`, `alloc_space`, `inuse_objects`, `inuse_space`. There are two profile types with this sample type set:
* The heap profile [sets the default sample type to empty string](https://github.com/golang/go/blob/30b2b767d6d902787b90476fd00eee4c9b3a3f15/src/runtime/pprof/pprof.go#L593) (that is, unsets it) which means that tools visualizing the profile should use the last sample type as the default one (the "use the last one if the default is not set" is a non-intuitive convention but this is what the pprof format has historically).
* The alloc profile on the other hand [sets the default sample type to `alloc_space`](https://github.com/golang/go/blob/30b2b767d6d902787b90476fd00eee4c9b3a3f15/src/runtime/pprof/pprof.go#L599), so that tools like pprof focus the user on that metric by default.

The question is how sucha  profile with four sample types and one of them being the preferred default as chosen by the profile producer should be encoded in OpenTelemetry profiles. Before this change it would be the array of sample types plus the default sample type. What would it be after this change and how would the default be captured?

christos68k:
> The question is how such a profile with four sample types and one of them being the preferred default as chosen by the profile producer should be encoded in OpenTelemetry profiles. Before this change it would be the array of sample types plus the default sample type. What would it be after this change and how would the default be captured?

The current proposal doesn't have an explicit default sample type (and we probably should steer away from introducing implicit conventions) so assuming four different sample types in a single OTel messsage, samples will be grouped according to sample type to separate profiles. One possible encoding would have four different profiles, each with a different sample type (of course one could further split the data according to other criteria).

Some options to accommodate the notion of a default sample type (e.g. for a consumer visualization tool) include:

1. Use an `InstrumentationScope` attribute
2. Introduce `default_sample_type` in `ScopeProfiles`

1 is more restricted than 2 as it could be seen as purely advisory metadata that a visualization tool can use. Adding `default_sample_type` to `ScopeProfiles` could also act as a fallback `sample_type` for profiles that do not declare it.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -384,21 +379,16 @@ message Sample {
...
   // References to attributes in ProfilesData.attribute_table. [optional]
   repeated int32 attribute_indices = 4;
 
   // Reference to link in ProfilesData.link_table. [optional]
   optional int32 link_index = 5;
 
-  // Timestamps associated with Sample represented in nanoseconds. These timestamps are expected
-  // to fall within the Profile's time range. [optional]
+  // Timestamps associated with Sample represented in nanoseconds. These
+  // timestamps are expected to fall within the Profile's time range. [optional]
```
felixge, line 391: I think we need to be more explicit about the interactions between `values` and `timestamps_unix_nano`. Maybe like this?

```diff
diff --git a/opentelemetry/proto/profiles/v1development/profiles.proto b/opentelemetry/proto/profiles/v1development/profiles.proto
index bc6e5c2b94..7b16e3e8af 100644
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -369,10 +369,14 @@
   AggregationTemporality aggregation_temporality = 3;
 }
 
-// Each Sample records values encountered in some program
-// context. The program context is typically a stack trace, perhaps
-// augmented with auxiliary information like the thread-id, some
-// indicator of a higher level request being handled etc.
+// Each Sample records values encountered in some program context. The program
+// context is typically a stack trace, perhaps augmented with auxiliary
+// information like the thread-id, some indicator of a higher level request
+// being handled etc.
+//
+// A Sample MUST have have at least one values or timestamps_unix_nano entry. If
+// both fields are populated, they MUST contain the same number of elements, and
+// the elements at the same index MUST refer to the same event.
 message Sample {
   // locations_start_index along with locations_length refers to to a slice of locations in Profile.location_indices.
   int32 locations_start_index = 1;
@@ -388,7 +392,7 @@
   optional int32 link_index = 5;
 
   // Timestamps associated with Sample represented in nanoseconds. These
-  // timestamps are expected to fall within the Profile's time range. [optional]
+  // timestamps are expected to fall within the Profile's time range.
   repeated uint64 timestamps_unix_nano = 6;
 }
 ```

https://github.com/florianl/opentelemetry-proto/pull/2

LMK what you think.

**Review by felixge — COMMENTED** [review]
LGTM in general. Sorry for the delay. PTAL at my comments. (I still need to find time to reply to some of the comments above, will do soon)

File: opentelemetry/proto/profiles/v1development/profiles.proto (hunk 4 above)
Comment (christos68k) [reply opentelemetry/proto/profiles/v1development/profiles.proto:391]: I'd rather we use 'should' to be consistent with OTel terminology:

```suggestion
  // timestamps should fall within the Profile's time range. [optional]
```

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -205,15 +205,12 @@ message ScopeProfiles {
 // for ease of understanding data migration, it is not intended that pprof:Profile and
 // OpenTelemetry:Profile encoding be wire compatible.
 message Profile {
-  // A description of the samples associated with each Sample.value.
+  // The type and unit of all Sample.values in this profile.
   // For a cpu profile this might be:
-  //   [["cpu","nanoseconds"]] or [["wall","seconds"]] or [["syscall","count"]]
+  //   ["cpu","count"] or ["off_cpu","nanoseconds"]
```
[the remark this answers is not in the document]
florianl: applyed https://github.com/florianl/opentelemetry-proto/pull/1

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -178,6 +178,13 @@ message ScopeProfiles {
   // https://opentelemetry.io/docs/specs/otel/schemas/#schema-url
   // This schema_url applies to all profiles in the "profiles" field.
   string schema_url = 3;
+
+  // The preferred type and unit of Samples in at least one Profile.
+  // For a cpu or off-cpu profile this might be:
```
christos68k:
```suggestion
  // Also used as fallback for Profiles with missing sample_type.
  // For a cpu or off-cpu profile this might be:
```

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -210,15 +217,12 @@ message ScopeProfiles {
 // for ease of understanding data migration, it is not intended that pprof:Profile and
 // OpenTelemetry:Profile encoding be wire compatible.
 message Profile {
-  // A description of the samples associated with each Sample.value.
-  // For a cpu profile this might be:
-  //   [["cpu","nanoseconds"]] or [["wall","seconds"]] or [["syscall","count"]]
+  // The type and unit of all Sample.values in this profile.
+  // For a cpu or off-cpu profile this might be:
+  //   ["cpu","nanoseconds"] or ["off_cpu","nanoseconds"]
   // For a heap profile, this might be:
```
christos68k, line 224:
```suggestion
  // See ScopeProfiles.default_sample_type for some possible values.
```

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -210,15 +217,12 @@ message ScopeProfiles {
...
-  //   [["allocations","count"], ["space","bytes"]],
-  // If one of the values represents the number of events represented
-  // by the sample, by convention it should be at index 0 and use
-  // sample_type.unit == "count".
-  repeated ValueType sample_type = 1;
+  //   ["allocated_objects","count"] or ["allocated_space","bytes"],
```
christos68k, line 225:
```suggestion
```

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -379,31 +381,30 @@ message ValueType {
...
   // Supersedes location_index.
   int32 locations_length = 2;
-  // The type and unit of each value is defined by the corresponding
-  // entry in Profile.sample_type. All samples must have the same
-  // number of values, the same as the length of Profile.sample_type.
-  // When aggregating multiple samples into a single sample, the
-  // result has a list of values that is the element-wise sum of the
-  // lists of the originals.
-  repeated int64 value = 3;
+  // The type and unit of each value is defined by Profile.sample_type.
```
christos68k, line 415:
```suggestion
  // The type and unit of each value is defined by Profile.sample_type.
```

christos68k (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
florianl: As Profile.sample_type is not optional, naming this field as fallback sounds confusing and conflicts in some way.

christos68k: Every field is optional in proto3, we don't have to explicitly mark it as such. Additionally, since `Profile.sample_type` is a message field we will also have an auto-generated `has` accessor that can be used to check for presence.

What conflict are you referring to? (We decided not to use 'optional' keyword, but we still have fields that can be missing).

christos68k: Looking at the new proto where we now group profiles by container, this fallback behavior probably isn't worth the complexity so let's leave it out.

opentelemetry/proto/profiles/v1development/profiles.proto
christos68k: Alternatively, have `ScopeProfiles.default_sample_type` refer to `Profile.sample_type` for possible values, we shouldn't duplicate the values in two places.

florianl: Applied with https://github.com/open-telemetry/opentelemetry-proto/pull/649/commits/5f0767dd04a875015b994ef9ac11800adf9a357d

opentelemetry/proto/profiles/v1development/profiles.proto
florianl: With https://github.com/open-telemetry/opentelemetry-proto/pull/649/commits/5f0767dd04a875015b994ef9ac11800adf9a357d only `Profile.sample_type` holds these examples now.

opentelemetry/proto/profiles/v1development/profiles.proto
florianl: The suggested change seems to be a NOP.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -379,31 +378,30 @@ message ValueType {
...
   // References to attributes in ProfilesDictionary.attribute_table. [optional]
   repeated int32 attribute_indices = 4;
 
   // Reference to link in ProfilesDictionary.link_table. [optional]
   optional int32 link_index = 5;
 
-  // Timestamps associated with Sample represented in nanoseconds. These timestamps are expected
-  // to fall within the Profile's time range. [optional]
+  // Timestamps associated with Sample represented in nanoseconds. These
+  // timestamps should fall within the Profile's time range.
```
atoulme, line 425: are those changes related to the change to the proto file?

atoulme: maybe that's just a rewording given the change to having timestamps organized in this manner. Disregard.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -178,6 +178,10 @@ message ScopeProfiles {
   // https://opentelemetry.io/docs/specs/otel/schemas/#schema-url
   // This schema_url applies to all profiles in the "profiles" field.
   string schema_url = 3;
+
+  // The preferred type and unit of Samples in at least one Profile.
+  // See Profile.sample_type for possible values.
+  ValueType default_sample_type = 4;
```
atoulme, line 188: Does it need to be a dedicated field? Could it be a scope attribute?

christos68k: We could use an attribute (it was the other option [here](https://github.com/open-telemetry/opentelemetry-proto/pull/649#issuecomment-2855491964)) but then we'd need to ensure key uniqueness and possibly introduce a new semantic convention. 

Given that all fields are optional in proto3, and that this specific attribute seems only relevant to profiling, it's simpler I think to have this field here? We could also update the field documentation to mention that it's optional.

opentelemetry/proto/profiles/v1development/profiles.proto
florianl: The change here is mostly about rewording.

Resolving this comment. Please feel free to reopen it, if needed.

opentelemetry/proto/profiles/v1development/profiles.proto
florianl: The Profiling SIG discussed this topic and agreed to prefer a dedicated field over some other mechanism.

felixge (approved):
This LGTM, including the proposed `default_sample_type` field. Sorry for the delay 😞.

jhalliday (approved):

aalexand (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -383,32 +382,31 @@ message ValueType {
...
-// Each Sample records values encountered in some program
-// context. The program context is typically a stack trace, perhaps
-// augmented with auxiliary information like the thread-id, some
-// indicator of a higher level request being handled etc.
+// Each Sample records values encountered in some program context. The program
+// context is typically a stack trace, perhaps augmented with auxiliary
+// information like the thread-id, some indicator of a higher level request
+// being handled etc.
+//
+// A Sample MUST have have at least one values or timestamps_unix_nano entry. If
```
aalexand, line 392: It might be good to describe a bit more the patterns of using these fields that we expect. The original proposal [comment](https://github.com/open-telemetry/opentelemetry-proto/issues/633#issuecomment-2772648117) has them, copying everything would probably be too verbose, but maybe we can do that in some more compact way here?

/cc @felixge

florianl: How about adding this to the comment?
```
// Examples:
// Report of a stacktrace at five different timestamps:
//    values: []
//    timestamps_unix_nano: [1, 2, 3, 4, 5]
//
// Report of a stacktrace that got reported five times without timestamps:
//   values: [5]
//    timestamps_unix_nano: []
//
// Report of a stacktrace that is reported five times with timestamps and values:
//    values: [10, 30, 20, 90, 110]
//    timestamps_unix_nano: [1, 2, 3, 4, 5]
```

/cc @felixge

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -182,6 +182,10 @@ message ScopeProfiles {
   // https://opentelemetry.io/docs/specs/otel/schemas/#schema-url
   // This schema_url applies to all profiles in the "profiles" field.
   string schema_url = 3;
+
+  // The preferred type and unit of Samples in at least one Profile.
+  // See Profile.sample_type for possible values.
+  ValueType default_sample_type = 4;
```
aalexand, line 188: Should this be just the name or the index of the profile to be considered as the default metric? Otherwise we are duplicating the unit information which doesn't feel nice...

opentelemetry/proto/profiles/v1development/profiles.proto
florianl: @aalexand is the suggested comment ok for you?

aalexand: > How about adding this to the comment?
> 
> ```
> // Examples:
> // Report of a stacktrace at five different timestamps:
> //    values: []
> //    timestamps_unix_nano: [1, 2, 3, 4, 5]
> //
> // Report of a stacktrace that got reported five times without timestamps:
> //   values: [5]
> //    timestamps_unix_nano: []
> //
> // Report of a stacktrace that is reported five times with timestamps and values:
> //    values: [10, 30, 20, 90, 110]
> //    timestamps_unix_nano: [1, 2, 3, 4, 5]
> ```
> 
> /cc @felixge

For the first example, should we clarify that consumers must treat this as value of 1?

Maybe write all examples to represent the same overall weight of the sample? I think it would be easier to follow.

Something like

```
// Examples of different ways of representing a sample with the total value of 10:
//
// Report of a stacktrace at 10 timestamps (consumers must assume the value is 1 for each point):
//    values: []
//    timestamps_unix_nano: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
//
// Report of a stacktrace with an aggregated value without timestamps:
//   values: [10]
//    timestamps_unix_nano: []
//
// Report of a stacktrace at 4 timestamps where each point records a specific value:
//    values: [2, 2, 3, 3]
//    timestamps_unix_nano: [1, 2, 3, 4]
```

florianl: Applied with https://github.com/open-telemetry/opentelemetry-proto/pull/649/commits/7210fda09ada75c19581078df9d7a64ab60c8234

opentelemetry/proto/profiles/v1development/profiles.proto
florianl: Being explicit and have `ValueType` here reduces complexity I think. Any tooling has to check `Profile.sample_type` anyway, as it can be different to the default one.

PR #650, opened 2025-05-05
Move the profiles lookup tables into a new Dictionary message

dmathieu:
As discussed in https://github.com/open-telemetry/opentelemetry-proto/pull/648

### DIFF
--- a/opentelemetry/proto/collector/profiles/v1development/profiles_service.proto
+++ b/opentelemetry/proto/collector/profiles/v1development/profiles_service.proto
@@ -39,7 +39,7 @@ message ExportProfilesServiceRequest {
   repeated opentelemetry.proto.profiles.v1development.ResourceProfiles resource_profiles = 1;
 
   // The reference table containing all data shared by profiles across the message being sent.
-  opentelemetry.proto.profiles.v1development.ProfilesReferenceTables reference_tables = 2;
+  opentelemetry.proto.profiles.v1development.ProfilesDictionary dictionary = 2;
 }
 
 message ExportProfilesServiceResponse {
--- a/opentelemetry/proto/collector/profiles/v1development/profiles_service.proto
+++ b/opentelemetry/proto/collector/profiles/v1development/profiles_service.proto
@@ -38,8 +38,8 @@ message ExportProfilesServiceRequest {
   // in that case this array will contain multiple elements.
   repeated opentelemetry.proto.profiles.v1development.ResourceProfiles resource_profiles = 1;
 
-  // The lookup table containing all data shared by profiles across the message being sent.
-  opentelemetry.proto.profiles.v1development.ProfilesReferenceTables lookup_tables = 2;
+  // The reference table containing all data shared by profiles across the message being sent.
+  opentelemetry.proto.profiles.v1development.ProfilesReferenceTables reference_tables = 2;
 }
 
 message ExportProfilesServiceResponse {
--- a/opentelemetry/proto/collector/profiles/v1development/profiles_service.proto
+++ b/opentelemetry/proto/collector/profiles/v1development/profiles_service.proto
@@ -38,8 +38,8 @@ message ExportProfilesServiceRequest {
   // in that case this array will contain multiple elements.
   repeated opentelemetry.proto.profiles.v1development.ResourceProfiles resource_profiles = 1;
 
-  // The lookup tables containing all data shared by profiles across the message being sent.
-  repeated opentelemetry.proto.profiles.v1development.ProfilesReferenceTables lookup_tables = 2;
+  // The lookup table containing all data shared by profiles across the message being sent.
+  opentelemetry.proto.profiles.v1development.ProfilesReferenceTables lookup_tables = 2;
 }
 
 message ExportProfilesServiceResponse {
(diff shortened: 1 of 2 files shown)


opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -134,6 +117,30 @@ message ProfilesData {
...
+  // For data coming from an SDK profiler, this array will typically contain one
+  // element. Host-level profilers will usually create one ResourceProfile per
+  // container, as well as one additional ResourceProfile grouping all samples
+  // from non-containerized processes.
+  // Other resource groupings are possible as well and clarified via
+  // Resource.attributes and semantic conventions.
+  repeated ResourceProfiles resource_profiles = 1;
+
+  // An array of ProfilesReferenceTables
+  repeated ProfilesReferenceTables lookup_tables = 2;
```
tigrannajaryan, line 141: Should this be repeated or we want a single instance of ProfilesReferenceTables? I thought we use one set of reference tables for all ResourceProfiles. Did I misunderstand that?

opentelemetry/proto/collector/profiles/v1development/profiles_service.proto
```diff
@@ -37,6 +37,9 @@ message ExportProfilesServiceRequest {
   // data from multiple origins typically batch the data before forwarding further and
   // in that case this array will contain multiple elements.
   repeated opentelemetry.proto.profiles.v1development.ResourceProfiles resource_profiles = 1;
+
+  // The lookup tables containing all data shared by profiles across the message being sent.
+  repeated opentelemetry.proto.profiles.v1development.ProfilesReferenceTables lookup_tables = 2;
```
aalexand, line 42: Given that this is proposed to be a pattern for all signals, can we unify the terminology?

We use _index in the field names to reference the items in the tables.

This field uses both "lookup" and "reference" terms.

This seems too many.

aalexand: How about "dictionary" here (both type and field) and _index in fields?

dmathieu: It's unlikely any other signal will start doing this without breaking changes (and thefore v2 and all that entitles).
But sure, I'm open to any name proposal.

christos68k (commented):
We should also update the references in documentation strings from `ProfilesData.XY_table` to refer to `ProfilesDictionary.XY_table`.

opentelemetry/proto/collector/profiles/v1development/profiles_service.proto
christos68k: `ProfilesDictionary dictionary` SGTM. I think we're already using `_index` everywhere we should?

aalexand: Yes, I think the `_index` / `_strindex` naming style is uniform already.

jhalliday:
Thanks, Damien. I like it, this feels like a more refined direction than #648 was.

I agree with points on streamlining terminology. Mixing 'reference' and 'lookup' isn't great. 'lookup_tables' field should probably become 'reference_tables' for consistency with the message name? Likewise if going with 'Dictionary'. I think the use of _index/_indices for pointers into the tables is already consistent, though if we're going with 'reference' everywhere I'm inclined to _ref/_reference/_references suffixes as an alternative and that may be a point in favour of using 'Reference' over 'Dictionary' for the message type. Whilst I'm the reason _strindex exists as a special distinction, I'm no longer sold on its advantages, it could be simplified to _index/_ref without great loss.

string_table is the only field likely to be common across ${signal}ReferenceTables messages in future, there may be some small clarity/consistency advantage to establishing a convention of ordering it as the first field in the message.

The idea of adding a 'bytes tables_id' field in ProfilesReferenceTables to allow for separately transmitted messages referencing into a particular set of tables rather than one default/bundled one still has some attraction, but at present I'm leaning towards leaving it out as YAGNI.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -90,26 +90,9 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
...
-  // Other resource groupings are possible as well and clarified via
-  // Resource.attributes and semantic conventions.
-  repeated ResourceProfiles resource_profiles = 1;
-
+// ProfilesReferenceTables represents the profiles data shared across the
+// entire message being sent.
+message ProfilesReferenceTables {
   // Mappings from address ranges to the image/binary/library mapped
   // into that address range referenced by locations via Location.mapping_index.
   repeated Mapping mapping_table = 2;
```
felixge, line 98: NIT: We should update the field numbers. You can apply this from: https://github.com/dmathieu/opentelemetry-proto/pull/1

opentelemetry/proto/collector/profiles/v1development/profiles_service.proto
felixge: I like the term dictionary as well.

felixge (commented):
LGTM. I like this direction! Just got some nits.

opentelemetry/proto/collector/profiles/v1development/profiles_service.proto
dmathieu: I have switched the term to dictionary.

aalexand (approved):

florianl (approved):

felixge (approved):

christos68k (approved):

reyang (approved):

Gandem (approved):

rockdaboot (approved):

aalexand (approved):

jsuereth (approved):

PaulinaParangerHr:
Really

PR #659, opened 2025-05-19
profiles: avoid 'optional' keyword usage

jhalliday:
Prefer reference tables with an explicit null/unused entry at index=0, rather than 'optional' pointers into the tables.

Following discussion in the profiling SIG, we note that other OpenTelemetry signal types don't use 'optional' protobuf fields and, whilst the Dictionary lookup tables pattern in the profile signal type require some form of 'null pointer' value for indicating unset indexes into the tables, the string_table convention inherited from pprof prefers table[0] = "" i.e. treats a 0 value pointer as null.

In the interest of improving consistency with other signal types and within the profiling signal itself, we therefore make mapping and link tables/references follow the same convention of having an explicit 'not used' default value as the first element in the dictionary table.

Note that I've chosen to not update the docs for the other dictionary tables to require the same use of the [0] slot, as at present no _index fields optionally reference into them, so I'm treating it as unnecessary at this time. However, it can be argued that greater clarity and consistency would be achieved by making the convention universal in anticipation before adoption becomes widespread and it gets harder to change.

### DIFF
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -196,8 +196,6 @@ message ScopeProfiles {
 //   that is most useful to humans.  There should be enough
 //   information present to determine the original sampled values.
 //
-// - On-disk, the serialized proto must be gzip-compressed.
-//
 // - The profile is represented as a set of samples, where each sample
 //   references a sequence of locations, and where each location belongs
 //   to a mapping.
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - uses: fossas/fossa-action@93a52ecf7c3ac7eb40f5de77fd69b1a19524de94 # v1.5.0
+      - uses: fossas/fossa-action@3ebcea1862c6ffbd5cf1b4d0bd6b3fe7bd6f2cac # v1.7.0
         with:
           api-key: ${{secrets.FOSSA_API_KEY}}
           team: OpenTelemetry
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -23,7 +23,7 @@ jobs:
         with:
           persist-credentials: false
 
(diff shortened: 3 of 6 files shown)


jhalliday:
@open-telemetry/profiling-maintainers

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -95,6 +95,8 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
 message ProfilesDictionary {
   // Mappings from address ranges to the image/binary/library mapped
   // into that address range referenced by locations via Location.mapping_index.
+  // mapping_table[0] must always be set to an invalid default mapping,
```
christos68k, line 98: Nit: Use same wording for consistency

```suggestion
  // mapping_table[0] must always be set to a 'null' default mapping,
```

christos68k (approved):

felixge (approved):

felixge:
Thanks for the PR 🙇 

> Note that I've chosen to not update the docs for the other dictionary tables to require the same use of the [0] slot, as at present no _index fields optionally reference into them, so I'm treating it as unnecessary at this time. However, it can be argued that greater clarity and consistency would be achieved by making the convention universal in anticipation before adoption becomes widespread and it gets harder to change.

I'm in favor of considering this.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -104,6 +106,8 @@ message ProfilesDictionary {
   repeated Function function_table = 3;
 
   // Links referenced by samples via Sample.link_index.
+  // link_table[0] must always be set to an invalid default link,
```
aalexand, line 109: Similar wording nuance to above.

opentelemetry/proto/profiles/v1development/profiles.proto
christos68k: Either "zero value" or "all-empty" SGTM.

florianl (approved):
Thanks for working on this :pray:

jsuereth (approved):

jsuereth:
> Note that I've chosen to not update the docs for the other dictionary tables to require the same use of the [0] slot, as at present no _index fields optionally reference into them, so I'm treating it as unnecessary at this time. However, it can be argued that greater clarity and consistency would be achieved by making the convention universal in anticipation before adoption becomes widespread and it gets harder to change.

@jhalliday can you open an issue to doing this work so it's not lost?

tigrannajaryan (approved):

PR #664, opened 2025-05-28
Support grpc-retry-pushback-ms

### LINKED ISSUE #4513: Question about gRPC error handling specification
For gRPC OTLP exporters, I'm wondering if we should just use gRPC's retry config: https://grpc.io/docs/guides/retry/#retry-configuration, rather than having custom back off/retry code...  I'm refactoring the python gRPC exporter a bit now and came across this.

gRPC retry config does not respect the `RetryInfo` proto described in https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures.

There's nothing that mentions `RetryInfo` on the gRPC repo / website. `RetryInfo` is mentioned in some Google AIP's, it's a Google published proto, it is used by some Google APIs (mainly for when quota based errors should be retried). 

gRPC has a supported header (`grpc-retry-pushback-ms`) for server pushback: https://github.com/grpc/proposal/blob/master/A6-client-retries.md#pushback. I tested this header and it is supported by gRPC's retry config.

I'm wondering if it's acceptable to use gRPC retry config given it diverges from the spec a little, or if we should update the spec to recommend this other header..

I feel it's nice to let the gRPC library handle this stuff instead of hand rolling our own code. If gRPC changes something around retries I'm assuming they'd update their own retry handling code and we wouldn't have to worry about it.

DylanRussell:
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/4513

I've tried to change the wording such that it's acceptable for SDKs to use gRPC retry config, which only respects the `grpc-retry-pushback-ms` header to handle all backoff / retry logic. 

The advantage of this is that SDKs don't need to implement / maintain any backoff/retry logic and can hand it off to the official gRPC library.


docs/specification.md
```diff
@@ -339,29 +341,41 @@ Here is a snippet of sample Go code to illustrate:
...
-// the delay between retries based on `retry_delay` until either a maximum
-// number of retries has been reached, or a maximum retry delay cap has been
-// reached.
+  # On the client side use gRPC retry config: https://grpc.io/docs/guides/retry/, which will automatically parse this header
+  # and handle all backoff and retry logic.
+  # Or manually parse the header:
+  
+  pushback = dict(error.trailing_metadata()).get("grpc-retry-pushback-ms")
+  if pushback is not None:
+    wait_period = int(pushback)
```
pellared, line 366: I think that the gRPC clients are supposed to handle this out-of-the-box. From https://github.com/grpc/proposal/blob/master/A6-client-retries.md#pushback

> When a client receives an explicit pushback response from a server, and it is appropriate to retry the RPC, it will retry after exactly that delay.

docs/specification.md
```diff
@@ -339,29 +341,41 @@ Here is a snippet of sample Go code to illustrate:
...
 ```
 
-The value of `retry_delay` is determined by the server and is implementation
-dependant. The server SHOULD choose a `retry_delay` value that is big enough to
+When the client receives this signal, it SHOULD follow the recommendations
+outline in the documentation for [`grpc-retry-pushback-ms`](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#pushback)
```
pellared, line 370: This is already a specification (design) of gRPC clients.

pellared (commented):
In general, LGTM

docs/specification.md
```diff
@@ -339,29 +341,41 @@ Here is a snippet of sample Go code to illustrate:
...
+  # Do this on the server side.
+  context.set_trailing_metadata(
+        (
+            (
+                "grpc-retry-pushback-ms",
+                "5000", # 5 seconds
+            ),
+        )
+    )
+  context.abort(StatusCode.UNAVAILABLE, "")
```
pellared: I think it would be nice to provide an example in Go for consistency:


```suggestion
```go
# Do this on the server side.
trailer := metadata.Pairs("grpc-retry-pushback-ms", "5000")
grpc.SetTrailer(ctx, trailer)
```

Based on: https://github.com/grpc/grpc-go/blob/master/examples/features/metadata/server/main.go

docs/specification.md
```diff
@@ -313,8 +310,13 @@ error with code [Unavailable](https://godoc.org/google.golang.org/grpc/codes)
 and MAY supply additional
 [details via status](https://godoc.org/google.golang.org/grpc/status#Status.WithDetails)
 using
-[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40).
-Here is a snippet of sample Go code to illustrate:
+[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40)
+or [through the gRPC metadata key `grpc-retry-pushback-ms`](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#pushback).
```
pellared: AFAIK it `grpc-retry-pushback-ms` must be be sent as [trailer](https://grpc.io/docs/guides/metadata/#trailers).

See https://github.com/search?q=repo%3Agrpc%2Fgrpc-go+grpc-retry-pushback-ms&type=code

```suggestion
or via Trailer [with the gRPC metadata key `grpc-retry-pushback-ms`](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#pushback).
```

docs/specification.md
pellared: I think if they do NOT use  gRPC retry config then they should not parse `grpc-retry-pushback-ms` as it is part of the the gRPC retry mechanism.

In short the OTLP exporters can have two retry mechanisms:
- provided by gRPC client via gRPC retry config
- custom that can also interpret `RetryInfo`

docs/specification.md
```diff
@@ -227,15 +227,12 @@ retryable and not-retryable:
...
-[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40)
-containing 0 value of RetryDelay. Here is a sample Go code to illustrate:
+[details via status](https://godoc.org/google.golang.org/grpc/status#Status.WithDetails).
+Here is a sample Go code to illustrate:
 
 ```go
   // Do this on server side.
-  st, err := status.New(codes.Unavailable, "Server is unavailable").
-    WithDetails(&errdetails.RetryInfo{RetryDelay: &duration.Duration{Seconds: 0}})
+  st, err := status.New(codes.Unavailable, "Server is unavailable")
```
pellared: :+1: 

Saying that RetryDelay should be set to 0 does not make any sense and conflicts with other part of the specification.

I find this a bugfix and should be better done in a separate PR.

docs/specification.md
DylanRussell: >> I think if they do NOT use gRPC retry config then they should not parse grpc-retry-pushback-ms as it is part of the gRPC retry mechanism.

Why not though ?

docs/specification.md
DylanRussell: Ack ! Sent out https://github.com/open-telemetry/opentelemetry-proto/pull/669

docs/specification.md
pellared: If the gRPC client in given language (e.g. new version) would start doing it then you could do the retry (exponentially?) more times: first by gRPC client and then second time by custom `grpc-retry-pushback-ms ` handling in the exporter implementation. It is safer to leave the `grpc-retry-pushback-ms` implementation to the gRPC clients.

pellared:
@DylanRussell, given https://github.com/open-telemetry/opentelemetry-proto/pull/669 is merged I think you can update this PR :wink:

docs/specification.md
DylanRussell: Alright we can do that it way, I've updated this PR so it's like that, let me know what you think

docs/specification.md
DylanRussell: Ack, I've removed this from this PR

docs/specification.md
```diff
@@ -340,29 +344,23 @@ Here is a snippet of sample Go code to illustrate:
     }
   }
 ```
+Here is a snippet of sample Python code to illustrate using `grpc-retry-pushback-ms`:
```
```suggestion
Here is a snippet of sample Go code to illustrate using `grpc-retry-pushback-ms`:
```

docs/specification.md
```diff
@@ -340,29 +344,23 @@ Here is a snippet of sample Go code to illustrate:
...
-// retrying.
-//
-// Clients should wait until `retry_delay` amount of time has passed since
-// receiving the error response before retrying.  If retrying requests also
-// fail, clients should use an exponential backoff scheme to increase gradually
-// the delay between retries based on `retry_delay` until either a maximum
-// number of retries has been reached, or a maximum retry delay cap has been
-// reached.
+```go
+# Do this on the server side.
```
```suggestion
// Do this on the server side.
```

docs/specification.md
```diff
@@ -314,8 +313,13 @@ error with code [Unavailable](https://godoc.org/google.golang.org/grpc/codes)
 and MAY supply additional
 [details via status](https://godoc.org/google.golang.org/grpc/status#Status.WithDetails)
 using
-[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40).
-Here is a snippet of sample Go code to illustrate:
+[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40)
+or via Trailer [with the gRPC metadata key `grpc-retry-pushback-ms`](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#pushback).
+
+Some clients may only respect one of these pushback mechanisms, so it's recommended for the server to implement both if it's
+critical that it be respected.
```
pellared: It would be good to use normative language

```suggestion
The client SHOULD honor at least one of these backpressure mechanisms.
In order to have the backpressure respected, the server SHOULD implement both mechanisms.
```

docs/specification.md
```diff
@@ -340,29 +344,23 @@ Here is a snippet of sample Go code to illustrate:
...
 ```
 
-The value of `retry_delay` is determined by the server and is implementation
-dependant. The server SHOULD choose a `retry_delay` value that is big enough to
+On the client side using gRPC retry config (https://grpc.io/docs/guides/retry/) 
+will cause gRPC to automatically parse this header and handle all backoff and retry logic.
+
+When the client receives this signal, if it isn't using gRPC retry config, it SHOULD 
+follow the recommendations or outlined in documentation for
+[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40).
```
pellared: > if it isn't using gRPC retry config

This is not clear for me.
Does it suggest that the client should support only one retry pushback mechanism?
Would it be bad to support both of them?

How about

```suggestion
The description of how the client should respect `RetryInfo` is provided
in its [documentation](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L34-L39).
```

docs/specification.md
```diff
@@ -340,29 +344,23 @@ Here is a snippet of sample Go code to illustrate:
...
+```go
+# Do this on the server side.
+trailer := metadata.Pairs("grpc-retry-pushback-ms", "5000")
+grpc.SetTrailer(ctx, trailer)
 ```
 
-The value of `retry_delay` is determined by the server and is implementation
-dependant. The server SHOULD choose a `retry_delay` value that is big enough to
+On the client side using gRPC retry config (https://grpc.io/docs/guides/retry/) 
+will cause gRPC to automatically parse this header and handle all backoff and retry logic.
```
```suggestion
will cause the gRPC client to automatically parse the `grpc-retry-pushback-ms` trailer metadata
and handle all backoff and retry logic.
```

docs/specification.md
DylanRussell: Looks good ! Updated

docs/specification.md
DylanRussell: Done

docs/specification.md
DylanRussell: Done

docs/specification.md
DylanRussell: Done

docs/specification.md
DylanRussell: Yeah I think that's better. Thanks for the suggestion.

docs/specification.md
```diff
@@ -340,29 +345,23 @@ Here is a snippet of sample Go code to illustrate:
     }
   }
 ```
+Here is a snippet of sample Go code to illustrate using `grpc-retry-pushback-ms`:
```
pellared, line 349: nit: Add blank line. Otherwise some linters will complain

```suggestion

Here is a snippet of sample Go code to illustrate using `grpc-retry-pushback-ms`:
```

pellared (dismissed):

pellared:
@open-telemetry/spec-sponsors, @open-telemetry/technical-committee, PTAL

docs/specification.md
```diff
@@ -314,8 +313,14 @@ error with code [Unavailable](https://godoc.org/google.golang.org/grpc/codes)
 and MAY supply additional
 [details via status](https://godoc.org/google.golang.org/grpc/status#Status.WithDetails)
 using
-[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40).
-Here is a snippet of sample Go code to illustrate:
+[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40)
+or via Trailer [with the gRPC metadata key `grpc-retry-pushback-ms`](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#pushback).
```
MrAlias, line 317: Please link to a permanent link instead of one that can break.

docs/specification.md
```diff
@@ -314,8 +313,14 @@ error with code [Unavailable](https://godoc.org/google.golang.org/grpc/codes)
...
+
+The client SHOULD honor at least one of these backpressure mechanisms.
+
+In order to have the backpressure respected, the server SHOULD implement both mechanisms.
```
MrAlias, line 321: Does this mean the server should be responding with both mechanisms at the same time? The example below only has one mechanism being used at a time.

This seems odd to me. It is usually the client that is recommended to handle multiple versions of a response and the server is expected to support at least one.

Is there a preference for which mechanism the server should support?

DylanRussell: I don't think there is a preference. About half the languages are using gRPC retry config which supports only the `pushback-ms` header, and the other half is rolling their own retry logic and only support the `RetryInfo`..

If a client uses gRPC retry config, it is offloading all the retry logic to gRPC internal library, so it can't also support `RetryInfo`.

docs/specification.md
```diff
@@ -341,28 +346,23 @@ Here is a snippet of sample Go code to illustrate:
...
 ```
 
-When the client receives this signal, it SHOULD follow the recommendations
-outlined in documentation for
-[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40):
+Here is a snippet of sample Go code to illustrate using `grpc-retry-pushback-ms`:
 
+```go
+// Do this on the server side.
+trailer := metadata.Pairs("grpc-retry-pushback-ms", "5000")
```
tigrannajaryan, line 353: Is this expected to be done _only_ if the server wants to pushback? It is not clear from the sample if this is some sort of permanent option that the server must always set or only set when the server is overloaded and needs to pushback the clients.

tigrannajaryan (commented):
@DylanRussell can you please also attach a prototype implementation that shows how this works? You can take existing implementation in one of the languages or in the Collector and fork/modify it to demonstrate the change. Collectors otlp receiver/exporter fork would be ideal since it shows both the client and server side.

docs/specification.md
```diff
@@ -314,8 +313,14 @@ error with code [Unavailable](https://godoc.org/google.golang.org/grpc/codes)
 and MAY supply additional
 [details via status](https://godoc.org/google.golang.org/grpc/status#Status.WithDetails)
 using
-[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40).
-Here is a snippet of sample Go code to illustrate:
+[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40)
+or via Trailer [with the gRPC metadata key `grpc-retry-pushback-ms`](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#pushback).
+
+The client SHOULD honor at least one of these backpressure mechanisms.
```
tigrannajaryan, line 319: What happens with the clients that honor the new mechanism that attempt to work with a server that honors only the old mechanism (existing implementations)? They would be incompatible, right?

tigrannajaryan:
I think we need a description of how the new and old implementations should interoperate. Explain the operation in the following 3 scenarios:
- Old client implementing the spec before this change connects to a new server implementing the spec after this change.
- New client connects to old server.
- New client connects to new server.

docs/specification.md
DylanRussell: Below it's recommended below for the server to implement both mechanisms, then the client is guaranteed to respect it..

Right now clients are split between the 2 mechanisms, that's the current state, the server needs to do both if it wants a guarantee.

Alternatively lets just recommend the retry - config mechanism in all cases for everyone ? Since the clients are already split, it's likely that server's timeouts aren't being respected in all cases. If a retry-timeout isn't respected it's not that bad anyway, as all the clients have their own back off timeout logic already.

docs/specification.md
bogdandrutu: Isn’t the problem that permanent link may include outdated information?

PR #665, opened 2025-05-29
profiles: make `profile_id` optional

### LINKED ISSUE #652: Profile: `profile_id` is required
I'm trying to understand why the spec says `Profile::profile_id` [is required](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L247). I see why one would to assign a unique ID to a profile; yet, I don't see why this behavior should be enforced. I think the spec should leave this choice to the user, and simply assume two profiles not to be related if `profile_id` is not set.

The main reason to remove this from the spec is that generating good UUIDs is not trivial, especially if utility functions are not available in the standard library (e.g. C++11). Forcing people to use fallback to bad UUID does more harm than good in this case.

fandreuz:
Fixes #652


florianl (approved):
I think it is a good idea to make this field optional. Speaking for OTel eBPF profiler it is just a [randomly generated value](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/afbda3604166b4446c31e310b56e81292d05ab96/reporter/internal/pdata/generate.go#L51-L58). So based on the profile_id reported by OTel eBPF profiler, no conclusions can be drawn to the actual data it reports.

christos68k (approved):

aalexand (approved):

petethepig (approved):

jhalliday (approved):

tigrannajaryan:
Markdown lint failed (unrelated to changes in this PR). Please file an issue to fix it.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -251,9 +251,11 @@ message Profile {
   int32 default_sample_type_index = 9;
 
   // A globally unique identifier for a profile. The ID is a 16-byte array. An ID with
-  // all zeroes is considered invalid.
-  //
-  // This field is required.
+  // all zeroes is considered invalid. It may be used for deduplication and signal
```
atoulme, line 254: if the ID is a 16-byte array, and is unset, then all bytes are set to zero - is a profile ID unset therefore invalid per this comment?

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -251,9 +251,11 @@ message Profile {
...
+  // correlation purposes. It is acceptable to treat two profiles with different values
+  // in this field as not equal, even if they represented the same object at an earlier
+  // time.
+  // This field is optional; an ID may be assigned to an ID-less profile in a later step.
```
atoulme, line 258: What value is it set to if unset and optional?

opentelemetry/proto/profiles/v1development/profiles.proto
fandreuz: Unset will look like "field not set" (e.g. for Python `message.HasField("profile_id") == false`) or an empty byte array depending on the Protobuf implementation. But in both cases, unset won't look like a 16-byte array of zeros.

fandreuz: I think the idea behind stating that `An ID with all zeroes is considered invalid` is just to provide a null value that the community can use in implementations when a well-formed value must be set for practical reasons.

opentelemetry/proto/profiles/v1development/profiles.proto
fandreuz: There was some discussion in #652 about this, the idea is that if a profile does not have an ID it may receive one for deduplication purposes later, e.g. in the collector. The value would be a real valid identifier, including 16 zeros if that makes sense.

atoulme: the proto file comment says 16 zeros is invalid. I cannot reconcile it with your statement above.

fandreuz: Right, that's a bit unclear, let me rephrase.

> The value would be a real valid identifier, including 16 zeros if that makes sense.

The receiver of an id-less profile (i.e. `profile_id` absent) is allowed to assign an arbitrary value to `profile_id`, including the null value (16 zeros).

I'm not sure whether the semantic difference between unset and 16 zeros is part of the spec though. I'm just guessing.

PR #669, opened 2025-06-03
Minor spec change around UNAVAILABLE / RetryInfo

DylanRussell:
It doesn't make sense that we recommend a RetryInfo with time 0, none of the implementations i've seen special case this value, and it goes against the rest of the spec which says to set it to a real integer.

It also doesn't make since to say the server "MUST" use UNAVAILABLE for retryable errors, and then go on to mention other retryable error status codes..


dashpole:
> rest of the spec which says to set it to a real integer

Where does it say that?

DylanRussell:
https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlpgrpc-throttling - this is the section that talks about how to use `RetryInfo`.

The only other mention of `RetryInfo` in the spec is:

>> The client SHOULD interpret RESOURCE_EXHAUSTED code as retryable only if the server signals that the recovery from resource exhaustion is possible. This is signaled by the server by returning [a status](https://godoc.org/google.golang.org/grpc/status#Status.WithDetails) containing [RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40). In this case the behavior of the server and the client is exactly as described in [OTLP/gRPC Throttling](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlpgrpc-throttling) section. If no such status is returned, then the RESOURCE_EXHAUSTED code SHOULD be treated as non-retryable.

This part of the spec I'm removing doesn't make sense.. Why must retryable errors return a RetryInfo of 0 and UNAVAILABLE  ?

tigrannajaryan (changes requested):
Blocking temporarily, since I am not sure this is not a breaking change.

We have removed a MUST requirement from the servers. Removing it means clients that strictly rely on this requirement may not work correctly anymore.

We must preserve the interoperability of (old,new) and (new,old) pairs of clients and servers. Please show the analysis that demonstrates this interoperability.

pellared:
@DylanRussell, shouldn't we also change 

https://github.com/open-telemetry/opentelemetry-proto/blob/68f9c6329ca91f96333ee5b240e305945b122b70/docs/specification.md?plain=1#L311

to

docs/specification.md
```diff
@@ -224,18 +224,12 @@ retryable and not-retryable:
...
 
-The server MUST indicate retryable errors using code
-[Unavailable](https://godoc.org/google.golang.org/grpc/codes) and MAY supply
-additional
-[details via status](https://godoc.org/google.golang.org/grpc/status#Status.WithDetails)
-using
-[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40)
-containing 0 value of RetryDelay. Here is a sample Go code to illustrate:
+For retryable errors the server is recommended to use code
+[Unavailable](https://godoc.org/google.golang.org/grpc/codes).
```
pellared: Maybe to less style changes? I think it is also good to keep the MAY with `RetryInfo`, but just without the `0` value which only brings confusion.

```suggestion
The server SHOULD indicate retryable errors using code
[Unavailable](https://godoc.org/google.golang.org/grpc/codes) and MAY supply
additional
[details via status](https://godoc.org/google.golang.org/grpc/status#Status.WithDetails)
using
[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40).
Here is a sample Go code to illustrate:
```

DylanRussell:
Yeah those suggestions SG. I've updated the PR

pellared (approved):

tigrannajaryan:
This fixes a bug in the spec and should be allowed.

tigrannajaryan:
Can't merge, the checks are stuck.

tigrannajaryan:
@open-telemetry/technical-committee can we have a few more eyes on this? I want to make sure we didn't miss anything.

docs/specification.md
```diff
@@ -224,18 +224,18 @@ retryable and not-retryable:
...
-[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40)
-containing 0 value of RetryDelay. Here is a sample Go code to illustrate:
+[RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40).
+Here is a sample Go code to illustrate:
 
 ```go
   // Do this on server side.
-  st, err := status.New(codes.Unavailable, "Server is unavailable").
-    WithDetails(&errdetails.RetryInfo{RetryDelay: &duration.Duration{Seconds: 0}})
+  st, err := status.New(codes.Unavailable, "Server is unavailable")
```
```suggestion
  st, err := status.New(codes.Unavailable, "Server is unavailable").
```

jack-berg (approved):

PR #672, opened 2025-06-13
profiles: improve attribute encoding in ProfilesDictionary

jhalliday:
Subsequent to https://github.com/open-telemetry/opentelemetry-proto/pull/659 (profiles: avoid 'optional' keyword usage) and in accordance with discussions in the profiling SIG, this PR updates the dictionary message docs to impose a more consistent handling of 'null pointer' semantics in the encoding pattern.

### DIFF
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -115,10 +115,7 @@ message ProfilesDictionary {
   repeated string string_table = 5;
 
   // A common table for attributes referenced by various messages.
-  repeated opentelemetry.proto.common.v1.KeyValue attribute_table = 6;
-
-  // Represents a mapping between Attribute Keys and Units.
-  repeated AttributeUnit attribute_units = 7;
+  repeated KeyValueAndUnit attribute_table = 6;
 }
 
 // ProfilesData represents the profiles data that can be stored in persistent storage,
@@ -501,4 +498,13 @@ message Function {
   int32 filename_strindex = 3;
   // Line number in source file. 0 means unset.
   int64 start_line = 4;
+}
+
+// A custom 'dictionary native' style of encoding attributes which is more specialized
+// for our use than opentelemetry.proto.common.v1.KeyValue
+// Specifically, uses the string table for keys and allows optional unit information.
+message KeyValueAndUnit {
+  int32 key_strindex  = 1;
+  opentelemetry.proto.common.v1.AnyValue value = 2;
+  int32 unit_strindex = 3; // non-zero where the unit cannot be derived from the name by semconv.
 }
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -93,15 +93,10 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
 // ProfilesDictionary represents the profiles data shared across the
 // entire message being sent.
 message ProfilesDictionary {
-
-  // Note all fields in this message MUST have a zero value encoded as the first element.
-  // This allows for _index fields pointing into the dictionary to use a 0 pointer value to indicate 'null' / 'not set'
-  // without requiring 'optional' encoding, which would be less compact.
-  // Unless otherwise defined, a 'zero value' message value is one with all default field values,
-  // so as to minimize wire encoded size.
-
   // Mappings from address ranges to the image/binary/library mapped
   // into that address range referenced by locations via Location.mapping_index.
+  // mapping_table[0] must always be set to a zero value default mapping,
+  // so that _index fields can use 0 to indicate null/unset.
   repeated Mapping mapping_table = 1;
 
   // Locations referenced by samples via Profile.location_indices.
@@ -111,6 +106,8 @@ message ProfilesDictionary {
   repeated Function function_table = 3;
 
   // Links referenced by samples via Sample.link_index.
+  // link_table[0] must always be set to a zero value default link,
+  // so that _index fields can use 0 to indicate null/unset.
   repeated Link link_table = 4;
 
   // A common table for strings referenced by various messages.
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -93,10 +93,15 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
 // ProfilesDictionary represents the profiles data shared across the
 // entire message being sent.
 message ProfilesDictionary {
+
+  // Note all fields in this message MUST have a zero value encoded as the first element.
+  // This allows for _index fields pointing into the dictionary to use a 0 pointer value to indicate 'null' / 'not set'
+  // without requiring 'optional' encoding, which would be less compact.
+  // Unless otherwise defined, a 'zero value' message value is one with all default field values,
+  // so as to minimize wire encoded size.
+
   // Mappings from address ranges to the image/binary/library mapped
   // into that address range referenced by locations via Location.mapping_index.
-  // mapping_table[0] must always be set to a zero value default mapping,
-  // so that _index fields can use 0 to indicate null/unset.
   repeated Mapping mapping_table = 1;
 
   // Locations referenced by samples via Profile.location_indices.
@@ -106,8 +111,6 @@ message ProfilesDictionary {
   repeated Function function_table = 3;
 
   // Links referenced by samples via Sample.link_index.
-  // link_table[0] must always be set to a zero value default link,
-  // so that _index fields can use 0 to indicate null/unset.
   repeated Link link_table = 4;
 
   // A common table for strings referenced by various messages.
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -67,9 +67,9 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
 //   │                                n-1
 //   │ 1-n         ┌───────────────────────────────────────┐
 //   ▼             │                                       ▽
-// ┌──────────────────┐   1-n   ┌──────────────┐      ┌──────────┐
-// │      Sample      │ ──────▷ │   KeyValue   │      │   Link   │
-// └──────────────────┘         └──────────────┘      └──────────┘
+// ┌──────────────────┐   1-n   ┌─────────────────┐   ┌──────────┐
+// │      Sample      │ ──────▷ │ KeyValueAndUnit │   │   Link   │
+// └──────────────────┘         └─────────────────┘   └──────────┘
 //   │                    1-n       △      △
 //   │ 1-n        ┌─────────────────┘      │ 1-n
 //   ▽            │                        │
@@ -286,14 +286,6 @@ message Profile {
   repeated int32 attribute_indices = 13;
 }
 
-// Represents a mapping between Attribute Keys and Units.
-message AttributeUnit {
-  // Index into string table.
-  int32 attribute_key_strindex = 1;
-  // Index into string table.
-  int32 unit_strindex = 2;
-}
-
 // A pointer from a profile Sample to a trace Span.
 // Connects a profile sample to a trace span, identified by unique trace and span IDs.
 message Link {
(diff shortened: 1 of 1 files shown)


jhalliday:
ref https://github.com/open-telemetry/opentelemetry-proto/pull/659#issuecomment-2893175871
@open-telemetry/profiling-maintainers @jsuereth

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -93,10 +93,15 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
 // ProfilesDictionary represents the profiles data shared across the
 // entire message being sent.
 message ProfilesDictionary {
+
+  // Note all fields in this message MUST have a zero value encoded as the first element.
```
christos68k: We don't have index lookups for `attribute_units` and a zero value encoded as the first element isn't needed for that field, so this is a little confusing as-is. Maybe we can say "all fields except attribute_units".

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -93,10 +93,15 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
...
+  // This allows for _index fields pointing into the dictionary to use a 0 pointer value to indicate 'null' / 'not set'
+  // without requiring 'optional' encoding, which would be less compact.
+  // Unless otherwise defined, a 'zero value' message value is one with all default field values,
+  // so as to minimize wire encoded size.
```
christos68k: I think we can trim this text down a bit:

```suggestion
  // This allows for _index fields pointing into the dictionary to use a 0 pointer value
  // to indicate 'null' / 'not set'. Unless otherwise defined, a 'zero value' message value is one with all default field
  // values, so as to minimize wire encoded size.
```

opentelemetry/proto/profiles/v1development/profiles.proto
felixge: Oh, interesting. That implies that `attribute_units` isn't really a dictionary table. Maybe we should move it to `ProfilesData`?

I'm okay with adding a comment for now and dealing with `attribute_units` as a separate issue.

dmathieu: If we want attribute_units shared across multiple profiles, they should be good in this dictionary.
How about an explicit comment for that attribute specifying it doesn't need zero padding ?

aalexand: Conceptually `attribute_units` is `map<int32, int32>`, mapping attribute names to the unit. I agree with Felix it's not really a dictionary. Also, the number of attribute names is usually small (5-10), so sharing this table between profiles is not terribly important. We could probably move it to the profile message.

jhalliday: I don't think it's always a map keyed on name, because it's possible for attributes to exist such that they have the same name but different type, or even the same name but different usage. That is, attribute_table is potentially the result of serializing and then concatenating multiple maps, each of which may contain the same key, resulting in the need to express 'attribute "foo" at attribute_table[3] has unit x' at the same time as 'attribute "foo" at attribute_table[5] has unit y".
Arguably the unit should, like the type, be part of the array element tuple in attribute_table, which allows for each foo to have distinct type whilst having a shared name, but whereas the type is always required for deserialization reasons, the unit is not, so that could be inefficient.

jhalliday:
```
message AttributeUnit {
  // Index into attribute_table. i.e. points at a {key,type} tuple (was previously string table)
  int32 attribute_index = 1;
  // Index into string table.
  int32 unit_strindex = 2;
}
```

may allow attribute_units to work as needed, at the cost of requiring additional entries where the same name appears multiple times in the attribute_table, regardless of if they actually have different units or not. A possible optimization here is to require AttributeUnit to be provided only where the Attribute name is NOT one defined by OTel semconv as having a well known unit.

jhalliday: One additional problem: When the SDK API is in terms of Key,Value tuples as it is for Attributes in OTel, the API can't express the unit name string, so there exists no easy way for the user to provide the required information for serialization into the AttributeUnit. That's ok for standard ones where we can code a name to unit lookup in the serializer, but for those we can just use the reverse table on the deserialization side and don't need the unit info on the wire anyhow. We only need AttributeUnit instances for 'extensions' that use names that are not known.
At the API level, a {Key,Value,Unit?} tuple is perhaps cleaner, though that's a design decision distinct from how the elements of the tuple are encoded on the wire

jhalliday: On the same note, because we pulled in KeyValue from common, it's defined as
```
message KeyValue {
  string key = 1;
  AnyValue value = 2;
}
```
rather than using string_table for the name and thus attribute_table potentially stores the name multiple times, which is nice class reuse at the API/usability level, but inefficient on the wire. If we're revisiting the design here anyhow to support unit at as part of the tuple at the API level, it may make sense to ditch KeyValue and define something else for the wire proto.
```
message KeyValueAndUnit {
  int32 key_strindex  = 1;
  AnyValue value = 2;
  int32 unit_strindex = 3; // non-zero if the unit name cannot be derived from the name by semconv.
}
...
// in ProfilesDictionary
repeated KeyValueAndUnit attribute_table = 6; // was previously opentelemetry.proto.common.v1.KeyValue
```

florianl: This might be a controversial approach:
Assuming the attributes in the OTel Profiling signal following the OTel SemConv, then we could drop `attribute_units`.
We could drop `attribute_units` as OTel SemConv defines units for each attribute - e.g. https://opentelemetry.io/docs/specs/semconv/system/hardware-metrics/#hwmemory---memory-module-metrics. But this approach works only for OTel SemConv compliant attributes.

If `attribute_units` gets dropped, there will be some complexity to translate pprof <-> OTel Profiling, as the conversion needs to be aware of OTel SemConv.

For the moment, I think, receivers of the OTel Profiling signal need to check for every element in `ProfilesDictionary.attribute_table` if there is a corresponding entry in `ProfilesDictionary.attribute_units`, which feels odd from an API/SDK perspective.

florianl: Extending a bit more on my last controversal approach in https://github.com/open-telemetry/opentelemetry-proto/pull/672#discussion_r2200179829:

If I follow the conversion from google/pprof to the OTel format, I still think, dropping `attribute_unit` should be fine.

From google/pprof:
```
  // Use arbitrary string (for example, "requests") as a custom count unit.
  // If no unit is specified, consumer may apply heuristic to deduce the unit.
  // Consumers may also  interpret units like "bytes" and "kilobytes" as memory
  // units and units like "seconds" and "nanoseconds" as time units,
  // and apply appropriate unit conversions to these.
```
https://github.com/google/pprof/blob/6e76a2b096b5fa52e4bb3f7f7a357bd6e6b3b7b1/proto/profile.proto#L148-L152

As there is no list of accepted or preferred units, units can be arbitrary. So using a heuristic approach to get the unit from the attribute name/key should be fine to start with for custom attributes, that do not follow OTel SemConv.

Otherwise, the `message KeyValueAndUnit` approach as suggested in https://github.com/open-telemetry/opentelemetry-proto/pull/672#discussion_r2198148492 by @jhalliday also sounds good to me.

jhalliday: Discussed in SIG meeting, KeyValueAndUnit way forward has been decided.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -502,3 +499,12 @@ message Function {
   // Line number in source file. 0 means unset.
   int64 start_line = 4;
 }
+
+// A custom 'dictionary native' style of encoding attributes which is more specialized
+// for our use than opentelemetry.proto.common.v1.KeyValue
```
rockdaboot:
```suggestion
// A custom 'dictionary native' style of encoding attributes which is more convenient
// for profiles than opentelemetry.proto.common.v1.KeyValue.
```

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -502,3 +499,12 @@ message Function {
...
+// Specifically, uses the string table for keys and allows optional unit information.
+message KeyValueAndUnit {
+  int32 key_strindex  = 1;
+  opentelemetry.proto.common.v1.AnyValue value = 2;
+  int32 unit_strindex = 3; // non-zero where the unit cannot be derived from the name by semconv.
```
rockdaboot:
```suggestion
  int32 unit_strindex = 3; // non-zero if the unit is not defined by semconv.
```

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -502,3 +499,12 @@ message Function {
...
+}
```
rockdaboot: Add a linefeed at the end of the file.

rockdaboot (approved):
lgtm, just a few nits

florianl (approved):
From a Profiling protocol perspective I appreciate this change. It drops the complexity of needing to handle two tables (`attribute_table` and `attribute_units`), while at the same time it also makes benefit of the string table (using `key_strindex` and `unit_strindex`).

As the OTel ecosystem uses `opentelemetry.proto.common.v1.KeyValue` in various signal, we might need to adapt some parts of this ecosystem. But I think, this should not be a blocker for this great step forward!

opentelemetry/proto/profiles/v1development/profiles.proto
christos68k: We should make it clear here that not all keys will have units as otherwise this reads a bit strange.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -115,10 +115,7 @@ message ProfilesDictionary {
   repeated string string_table = 5;
 
   // A common table for attributes referenced by various messages.
-  repeated opentelemetry.proto.common.v1.KeyValue attribute_table = 6;
-
-  // Represents a mapping between Attribute Keys and Units.
-  repeated AttributeUnit attribute_units = 7;
+  repeated KeyValueAndUnit attribute_table = 6;
```
christos68k, line 136: We should also add this to and update the ascii graph at the beginning of the file (We can just rename [KeyValue](https://github.com/open-telemetry/opentelemetry-proto/blob/b553517a730dc72097beb60292815ca221766598/opentelemetry/proto/profiles/v1development/profiles.proto#L71) to `KeyValueAndUnit`)

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -115,10 +115,7 @@ message ProfilesDictionary {
   repeated string string_table = 5;
 
   // A common table for attributes referenced by various messages.
-  repeated opentelemetry.proto.common.v1.KeyValue attribute_table = 6;
-
-  // Represents a mapping between Attribute Keys and Units.
-  repeated AttributeUnit attribute_units = 7;
```
christos68k, line 142: [AttributeUnit](https://github.com/open-telemetry/opentelemetry-proto/blob/b553517a730dc72097beb60292815ca221766598/opentelemetry/proto/profiles/v1development/profiles.proto#L293-L298) is still present in the proto though now unused and should be removed.

opentelemetry/proto/profiles/v1development/profiles.proto
tigrannajaryan, line 136: I was under the impression the Profiling SIG settled on the common `KeyValue` to be consistent with other signals. Has this decision been changed?

florianl: Yes this decision changed. The KeyValueAndUnit elements referenced in this lookup table are only used by submessages and so can not benefit anyway from commonly used functionality in other signals. Additionally, the Profiling SIG missed a `unit` indicator for these attributes. So there were two lookup tables, one for the attribute and its value and one for the attribute and its units  (see `attribute_units`) - which made attribute handling error-prone.
Besides that, the new `KeyValueAndUnit` also allows to reference commonly used keys in the string lookup table, which reduces the data pushed over the network.

tigrannajaryan:
Please resolve the conflicts.

PR #679, opened 2025-07-10
profiles: remove default_sample_type

felixge:
This resolves the ambiguity when a `ScopesProfiles` message contains
more than one `profiles` entry matching the `default_sample_type`.

Also add a comment to clarify which profile viewers should display by
default. This was previously also ambiguous.

This change creates a minor problem when it comes to ensuring that pprof
profiles can round-trip through otel conversion as it may require the
`ScopeProfiles.profiles` entries to be reordered in order to honor the
`default_sample_type` of the original pprof message. Naively converting
the resulting otel profile back to pprof would cause the order of
`Profile.sample_types` to be changed.

Merging this change indiciates consensus within the profiling group that
this issue should be handled by adding a `pprof.profile_order` attribute
(name TBD) to the `ScopeProfiles.attributes` during the initial
pprof->otel conversion. This label will allow converting the resulting
otel profile back to pprof without any information loss.

See [gh-633][] for additional details.

[gh-633]: https://github.com/open-telemetry/opentelemetry-proto/issues/633#issuecomment-3056622636

### DIFF
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -42,6 +42,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@181d5eefc20863364f96762470ba6f862bdef56b # v3.29.2
+        uses: github/codeql-action/upload-sarif@51f77329afa6477de8c49fc9c7046c15b9a4e79d # v3.29.5
         with:
           sarif_file: results.sarif


christos68k (approved):

jhalliday (approved):

aalexand (approved):

reyang (approved):

reyang:
@felixge might be good to have a changelog entry?

felixge:
In the last SIG meeting I promised that I would provide an example of how a `pprof.profile_order` (name TBD) label could be used to ensure that we can round trip pprof files without information loss.

Example pprof input (inspired by the Go memory profiles, but simplified):

```
# Profile
sample_type: [
  {type: "allocations", unit: "bytes"},
  {type: "allocations", unit: "count"},
],
default_sample_type: "allocations/count",
samples: [
  {value: [320, 10]},
  {value: [640, 2]},
],
```

Converted to OpenTelemetry:

```
# ProfilesData
resource_profiles: [
  {
    scope_profiles: [
      scope: {
        attributes: [
          {key: "pprof.profile_order": [1, 0]},
        ],
      },
      profiles: [
        {
          sample_type: {type: "allocations", unit: "count"},
          sample: [
            {values: [10]},
            {values: [2]},
          ]
        },
        {
          sample_type: {type: "allocations", unit: "bytes"},
          sample: [
            {values: [320]},
            {values: [640]},
          ]
        }
      ],
    ],
  },
]
```

This should round trip back to the original pprof file.

However, there is a small ambiguity here when `default_sample_type` is omitted.  pprof says that in this case, clients should default to the last sample (type?) value. We will probably need an additional or more complex label if we want to round trip pprof files that omit the `default_sample_type` field precisely.

Anyway, I don't think this is a blocker for this PR. It's certainly a solvable problem.

Issue #680, opened 2025-07-18
Improve documentation so that more can be published


thompson-tomo:
I want to have all the OTLP spec docs published on the website rather than having links to github and by adopting the new structure we can get more docs logically included. 

Allowing for the sub folder solution means down the track contributing docs etc could be published.

chalin:
Hi @thompson-tomo, can you provide some context and motivation for this request?

On the website under Docs > Specs we have, **up until now, only published the specs** themselves. That is only formally binding parts. For a change that was done in support of this see:

- #471 

/cc @svrnm @lmolkova @tigrannajaryan

chalin:
And, as I mention in https://github.com/open-telemetry/opentelemetry-proto/pull/681#issuecomment-3230169242, we don't need to reorganize the `docs` folder to publish more content.

thompson-tomo:
Hi @chalin as mentioned in other comment, there is 2 additional pages currently published via links to github on the website. This is mentioned in the issue description.

As mentioned above split is so that contributing docs could be published which makes no sense to be under specs.

chalin:
> there is 2 additional pages currently published via links to github on the website.

Can you show me (via a screenshot) what you mean by this?

Unless there is a change in policy, https://opentelemetry.io/docs/specs/ **only hosts spec pages that are _normative_**. For everything else (requirements, design docs, examples), we direct the user to the repo.

AFAIK, the only page that is normative on the proto repo is the page https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md.

PR #681, opened 2025-07-18
Make the spec readme the entry point for website #680

thompson-tomo:
These changes should enable the readme to the entry point of the public docs and the other docs in the specs to also be published

Closes: #680


docs/README.md
```diff
@@ -1,8 +1,7 @@
-# OpenTelemetry Protocol (OTLP)
+# OTLP Docs
 
-This is the specification of the OpenTelemetry Protocol (OTLP).
+The below table defines how the OpAMP docs get mounted to the Open Telemetry website.
```
arminru: This seems to be a copy-paste error.

thompson-tomo: Thanks for spoting that.

docs/README.md
```diff
@@ -1,8 +1,7 @@
-# OpenTelemetry Protocol (OTLP)
+# OTLP Docs
```
florianl: Abbreviations should not be used before they are defined.

This helps people not familiar with internals and details to understand the document:

```suggestion
# OpenTelemetry Protocol (OTLP) Docs
```

svrnm (approved):
lgtm overall, but I assume we might need to do some rewiring on the oteldotio repo as well, @chalin: WDYT?

chalin:
@svrnm @lmolkova - can we hold on this update until we get feedback from @tigrannajaryan?

thompson-tomo:
So as it currently stands both the design goals & requirements docs are also published in the navigation panel of the website but rather than being integrated seamlessly into the site it is provided as a link to the github repo.

Reason for introducing the specs folder is that provides a clear mapping of what docs gets mapped into the specs folder under docs on the website, allows for docs to be excluded from the website easily & it would enable other folders to be mounted at different locations. 

An example of what these other folders could be used for could be:
- A folder for describing the project for the community section
- A folder for describing how to contribute to the project for another location.

chalin:
> So as it currently stands both the design goals & requirements docs are also published in the navigation panel

Which nav panel?

thompson-tomo:
The main nav panel.
image

chalin:
Oh, those are legacy entries (from the time that those pages were a part of the spec), which it **might be time to remove**. WDYT @lmolkova @tigrannajaryan et all? /cc @svrnm

PR #683, opened 2025-07-21
all: add notes about the attribute values restrictions

pellared:
Per https://github.com/open-telemetry/opentelemetry-specification/pull/4595#discussion_r2214862515

Towards https://github.com/open-telemetry/opentelemetry-specification/blob/8aa1d27d9cf8c224299dcf0b58bfd5fcaacf88eb/oteps/4485-extending-attributes-to-support-complex-values.md?plain=1#L223-L227

The intent of this PR is to describe the current state of art.
You can consider this PR as a "bugfix". Thanks to it would be easier for us to communicate future changes in the OTLP protocol (new supported attribute types).

Related OTEP: https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4485-extending-attributes-to-support-complex-values.md

Some of the language is based on JSON RFC https://datatracker.ietf.org/doc/html/rfc8259#section-4:

> the behavior of software that receives such an object is unpredictable

Notice that this kind of description should be present in the comments even without the intent to go forward with https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4485-extending-attributes-to-support-complex-values.md

Related spec PR:
- https://github.com/open-telemetry/opentelemetry-specification/pull/4614


aalexand:
The OTEP says

> Instead of introducing a second set of "extended" attributes that can be used on spans and events, we propose to extend the standard attributes.

but IMO we effectively still create a new convention of "plain old standard attributes", spelling it in plain text like comments in this PR. I think it would be better to have an explicit name for it.

lmolkova (approved):

MrAlias (approved):

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -206,10 +206,18 @@ message Span {
...
+  //
+  // The attribute values SHOULD NOT contain empty values.
+  // The attribute values SHOULD NOT contain bytes values.
+  // The attribute values SHOULD NOT contain array values different than array of string values, bool values, int values,
+  // double values.
+  // The attribute values SHOULD NOT contain kvlist values.
+  // The behavior of software that receives attributes containing such values can be unpredictable.
+  // These restritions can change in a minor release.
+  // The restritions take origin from the OpenTelemetry specification:
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute.
```
lmolkova: nit: let's put a permalink so it won't change as we change the doc. And eventually we'll remove the comment?

pellared: Addressed abd7ae9

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -282,10 +282,18 @@ message Profile {
   //     "abc.com/myattribute": true
   //     "abc.com/score": 10.239
   //
-  // The OpenTelemetry API specification further restricts the allowed value types:
-  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute
   // Attribute keys MUST be unique (it is not allowed to have more than one
   // attribute with the same key).
+  //
+  // The attribute values SHOULD NOT contain empty values.
```
christos68k: The profiles protocol has multiple references to attributes (multiple `attribute_indices` fields across various messages: `Profile`, `Sample`, `Mapping`, `Location`) so we can either move this explanatory text to the [attribute table](https://github.com/open-telemetry/opentelemetry-proto/blob/b553517a730dc72097beb60292815ca221766598/opentelemetry/proto/profiles/v1development/profiles.proto#L118) and only have it be present in the file once, or replicate it across the various messages.

My preference would be to leverage the fact that we have a common attribute table and move the text there to avoid redundancy.

pellared: I will move the text to the attribute table.

pellared: Done 70335e6

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -377,6 +377,16 @@ message NumberDataPoint {
...
   // attribute with the same key).
+  //
+  // The attribute values SHOULD NOT contain empty values.
+  // The attribute values SHOULD NOT contain bytes values.
+  // The attribute values SHOULD NOT contain array values different than array of string values, bool values, int values,
+  // double values.
+  // The attribute values SHOULD NOT contain kvlist values.
+  // The behavior of software that receives attributes containing such values can be unpredictable.
+  // These restritions can change in a minor release.
+  // The restritions take origin from the OpenTelemetry specification:
```
christos68k:
```suggestion
  // The restrictions take origin from the OpenTelemetry specification:
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -377,6 +377,16 @@ message NumberDataPoint {
...
   // Attribute keys MUST be unique (it is not allowed to have more than one
   // attribute with the same key).
+  //
+  // The attribute values SHOULD NOT contain empty values.
+  // The attribute values SHOULD NOT contain bytes values.
+  // The attribute values SHOULD NOT contain array values different than array of string values, bool values, int values,
+  // double values.
+  // The attribute values SHOULD NOT contain kvlist values.
+  // The behavior of software that receives attributes containing such values can be unpredictable.
+  // These restritions can change in a minor release.
```
christos68k:
```suggestion
  // These restrictions can change in a minor release.
```

pellared:
> Hit approve too quickly 😅
> 
> (The same typo is repeated in the other signals)

No worries. Fixes applied :wink: 
7ac55b2
Thanks a lot for your review.

opentelemetry/proto/metrics/v1/metrics.proto
pellared: https://github.com/open-telemetry/opentelemetry-proto/pull/683/commits/7ac55b24830460863bce4bfbb7f6a62e7fee435c

opentelemetry/proto/metrics/v1/metrics.proto
pellared: https://github.com/open-telemetry/opentelemetry-proto/pull/683/commits/7ac55b24830460863bce4bfbb7f6a62e7fee435c

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -377,6 +377,16 @@ message NumberDataPoint {
...
+  //
+  // The attribute values SHOULD NOT contain empty values.
+  // The attribute values SHOULD NOT contain bytes values.
+  // The attribute values SHOULD NOT contain array values different than array of string values, bool values, int values,
+  // double values.
+  // The attribute values SHOULD NOT contain kvlist values.
+  // The behavior of software that receives attributes containing such values can be unpredictable.
+  // These restrictions can change in a minor release.
+  // The restrictions take origin from the OpenTelemetry specification:
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/v1.47.0/specification/common/README.md#attribute.
```
tigrannajaryan, line 389: Any way we can avoid repeating this several times? Maybe have it in common.proto once?

pellared: SIG meeting notes:
This restriction does NOT apply to attributes on log records. This is why I decided to repeat it everywhere.
I think repeating is also safer. Eventually there would be (hopefully) no type restrictions.
During the SIG meeting we agreed to resolve the comment.

tigrannajaryan (approved):

PR #688, opened 2025-07-25
profiles: Clarify ProfilesDictionary table[0] element use docs.

jhalliday:
Makes clear that all dictionary table fields should encode a 'null' element at index 0, such that optional fields pointing into the tables can use 0 value to indicate null.

### DIFF
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -93,10 +93,14 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
 // ProfilesDictionary represents the profiles data shared across the
 // entire message being sent.
 message ProfilesDictionary {
+
+  // Note all fields in this message MUST have a zero value encoded as the first element.
+  // This allows for _index fields pointing into the dictionary to use a 0 pointer value
+  // to indicate 'null' / 'not set'. Unless otherwise defined, a 'zero value' message value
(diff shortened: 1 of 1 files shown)


jhalliday:
@open-telemetry/profiling-maintainers as discussed in the SIG meeting, this PR factors out the straightforward doc change, leaving #672 for the attribute encoding change.

dmathieu (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -93,10 +93,14 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
 // ProfilesDictionary represents the profiles data shared across the
 // entire message being sent.
 message ProfilesDictionary {
+
+  // Note all fields in this message MUST have a zero value encoded as the first element.
```
christos68k, line 97: Should we also add this to make the text completely unambiguous?

```suggestion
  // All fields in this message MUST be present and MUST have a zero value encoded as the first element.
```

christos68k (approved):

aalexand (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
felixge: I don't think that's needed. Requiring the first element in a repeated list to be present implies the field to be present. Resolving this to make the PR mergable. Feel free to raise a new PR if you want to push this forward.

Issue #691, opened 2025-08-07
[profile] Use uint64 (fixed64) for TimeNanos


bogdandrutu:
To be consistent with the other signals, we should use unsigned types. It is very unlikely that you need negative timestamps.

cc @open-telemetry/profiling-maintainers

christos68k:
We are already using unsigned type for [timestamps_unix_nano](https://github.com/open-telemetry/opentelemetry-proto/blob/b14635e74b7909c9f55db39666c667d4ea3092b7/opentelemetry/proto/profiles/v1development/profiles.proto#L422-L424). 

However, [Profile.time_nanos](https://github.com/open-telemetry/opentelemetry-proto/blob/b14635e74b7909c9f55db39666c667d4ea3092b7/opentelemetry/proto/profiles/v1development/profiles.proto#L235) is still signed and should be changed.

PR #692, opened 2025-08-07
Unsigned Profile time_nanos and duration_nanos

christos68k:
### Summary

We are already using unsigned type for [Sample.timestamps_unix_nano](https://github.com/open-telemetry/opentelemetry-proto/blob/b14635e74b7909c9f55db39666c667d4ea3092b7/opentelemetry/proto/profiles/v1development/profiles.proto#L422-L424) but not for the time-related fields in [Profile](https://github.com/open-telemetry/opentelemetry-proto/blob/b14635e74b7909c9f55db39666c667d4ea3092b7/opentelemetry/proto/profiles/v1development/profiles.proto#L235-L237).

Also see:
- #691

### DIFF
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -232,9 +232,9 @@ message Profile {
   // interpretation of results.
 
   // Time of collection (UTC) represented as nanoseconds past the epoch.
-  uint64 time_nanos = 4;
+  fixed64 time_unix_nano = 4;
   // Duration of the profile, if a duration makes sense.
-  uint64 duration_nanos = 5;
+  uint64 duration_nano = 5;
   // The kind of events between sampled occurrences.
   // e.g [ "cpu","cycles" ] or [ "heap","bytes" ]
   ValueType period_type = 6;
@@ -421,7 +421,7 @@ message Sample {
 
   // Timestamps associated with Sample represented in nanoseconds. These
   // timestamps should fall within the Profile's time range.
-  repeated uint64 timestamps_unix_nano = 6;
+  repeated fixed64 timestamps_unix_nano = 6;
 }
 
 // Describes the mapping of a binary in memory, including its address range,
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -232,9 +232,9 @@ message Profile {
   // interpretation of results.
 
   // Time of collection (UTC) represented as nanoseconds past the epoch.
-  int64 time_nanos = 4;
+  uint64 time_nanos = 4;
   // Duration of the profile, if a duration makes sense.
-  int64 duration_nanos = 5;
+  uint64 duration_nanos = 5;
```
bogdandrutu: Should you use fixed64? Since we are using anyway a large part of the bytes (Time.now()), I think is more efficient.

christos68k: I also renamed to `_unix_nano` to be consistent with the other signals.

florianl (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -421,7 +421,7 @@ message Sample {
 
   // Timestamps associated with Sample represented in nanoseconds. These
   // timestamps should fall within the Profile's time range.
-  repeated uint64 timestamps_unix_nano = 6;
+  repeated fixed64 timestamps_unix_nano = 6;
```
felixge, line 434: I like this. It should decrease bytes per timestamp from 9 ([see for yourself](https://gist.github.com/felixge/9b24df80d134ff6a86f56212edec2682)) to 8.

Note: This comment can be resolved for merging, just not resolving it right away to give people a chance to see it.

tigrannajaryan (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
tigrannajaryan: Yes. Historical context: we initially had `google.protobuf.Timestamp` then moved to [sfixed64](https://github.com/open-telemetry/opentelemetry-proto/pull/33) then [to fixed64](https://github.com/open-telemetry/opentelemetry-proto/pull/46).

PR #698, opened 2025-08-14
Move zero-element comment about dictionary tables

brancz:
This at least gives language generators to put this comment into the generated code as well.

Even though I frequent the meeting around this format a lot myself, I almost implemented handling of this incorrectly because the comment did not end up in the generated Go code.

@florianl @felixge

### DIFF
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -92,13 +92,11 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
 
 // ProfilesDictionary represents the profiles data shared across the
 // entire message being sent.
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -92,13 +92,11 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
 
 // ProfilesDictionary represents the profiles data shared across the
 // entire message being sent.
+// Note all fields in this message MUST have a zero value encoded as the first element.
```
pellared:
```suggestion
//
// Note that all fields in this message MUST have a zero value encoded as the first element.
```

pellared (approved):

florianl (approved):

tigrannajaryan (approved):

PR #707, opened 2025-08-26
all: drop attribute values restrictions

pellared:
This MUST NOT be merged before `v1.8.0` release of `opentelemetry-proto`.

Towards https://github.com/open-telemetry/opentelemetry-specification/issues/4602

### DIFF
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -123,16 +123,6 @@ message ProfilesDictionary {
   //     "/http/server_latency": 300
   //     "abc.com/myattribute": true
   //     "abc.com/score": 10.239
-  //
-  // The attribute values SHOULD NOT contain empty values.
-  // The attribute values SHOULD NOT contain bytes values.
-  // The attribute values SHOULD NOT contain array values different than array of string values, bool values, int values,
-  // double values.
-  // The attribute values SHOULD NOT contain kvlist values.
-  // The behavior of software that receives attributes containing such values can be unpredictable.
-  // These restrictions can change in a minor release.
-  // The restrictions take origin from the OpenTelemetry specification:
-  // https://github.com/open-telemetry/opentelemetry-specification/blob/v1.47.0/specification/common/README.md#attribute.
   repeated KeyValueAndUnit attribute_table = 6;
 }
 
--- a/opentelemetry/proto/common/v1/common.proto
+++ b/opentelemetry/proto/common/v1/common.proto
@@ -54,8 +54,10 @@ message ArrayValue {
 message KeyValueList {
   // A collection of key/value pairs of key-value pairs. The list may be empty (may
   // contain 0 elements).
+  //
   // The keys MUST be unique (it is not allowed to have more than one
-  // value with the same key).
+  // attribute with the same key).
+  // The behavior of software that receives duplicated keys can be unpredictable.
   repeated KeyValue values = 1;
 }
 
@@ -76,6 +78,7 @@ message InstrumentationScope {
   // Additional attributes that describe the scope. [Optional].
   // Attribute keys MUST be unique (it is not allowed to have more than one
   // attribute with the same key).
+  // The behavior of software that receives duplicated keys can be unpredictable.
   repeated KeyValue attributes = 3;
   uint32 dropped_attributes_count = 4;
 }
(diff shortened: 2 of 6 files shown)


lmolkova (approved):

tigrannajaryan (changes requested):
I approved, but blocking to prevent accidental merging before 1.8.0

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -129,16 +129,6 @@ message ProfilesDictionary {
   //     "/http/server_latency": 300
   //     "abc.com/myattribute": true
   //     "abc.com/score": 10.239
-  //
-  // The attribute values SHOULD NOT contain empty values.
-  // The attribute values SHOULD NOT contain bytes values.
-  // The attribute values SHOULD NOT contain array values different than array of string values, bool values, int values,
```
aalexand, line 154: I think given that this is a dictionary table, the values are unique and we currently do require that attributes are of primitive types only. Otherwise it's more difficult to support the uniqueness check of arbitrarily complex JSON-like attributes and the ROI seems low.

@open-telemetry/profiling-approvers FYI

pellared: Here are some threads  from the OTEP that contains some answers for these concerns: 
- https://github.com/open-telemetry/opentelemetry-specification/pull/4485#discussion_r2058488908
-  https://github.com/open-telemetry/opentelemetry-specification/pull/4485#discussion_r2059387040

TL;DR;  (quick summary)
> it's more difficult to support the uniqueness check of arbitrarily complex JSON-like attributes

We need to support it for other signals anyway so we can reuse the same logic. The performance for primitive values should remain intact as the performance overhead would come in play when dealing with complex values.

> the ROI seems low

I think making it possible to reuse the same types/abstractions and similarity of the design as well as possible code reuse is not a small ROI.

florianl: > I think making it possible to reuse the same types/abstractions and similarity of the design as well as possible code reuse is not a small ROI.

As the profiles protocol uses `KeyValueAndUnit` and not `opentelemetry.proto.common.v1.KeyValue`, like other protocols, profiling does not benefit from code reuse.

jhalliday: Hmm. The profiles spec is unusual in another way: it specifies the wire proto but not the API, unlike all other signal types. At the language API level, it's possible to reuse e.g. Java's io.opentelemetry.api.common.Attributes and have them encoded into the proto's KeyValueAndUnit, at least in the common case where you don't need the optional unit part. Even where you do, extending the interface to add the additional field may be the best design. I'd certainly be concerned if it became impossible to serialize a valid io.opentelemetry.api.common.Attributes or similar to KeyValueAndUnit due to restrictions in the profiles proto. Unless confronted with a benchmark showing it's unduly costly, I'm in favor of removing the restrictions.

pellared: > The profiles spec is unusual in another way: it specifies the wire proto but not the API, unlike all other signal types. 

While I do not think Profiling API would be ever needed, yet I do imagine we could have an Profiling SDK at some point (so that profiling could be added to "manual instrumentation", set the same resources as for other signals, and maybe allow adding some customer profiling processors (for instance to filter something).

> As the profiles protocol uses `KeyValueAndUnit` and not `opentelemetry.proto.common.v1.KeyValue`, like other protocols, profiling does not benefit from code reuse.

 `KeyValueAndUnit` has a `KeyValue` field.

https://github.com/open-telemetry/opentelemetry-proto/blob/57a5849f7887b7fc149aa651de313dabd3e4704d/opentelemetry/proto/profiles/v1development/profiles.proto#L498-L505

>  Unless confronted with a benchmark showing it's unduly costly, I'm in favor of removing the restrictions.

Not sure what kind of benchmarks do you have in mind. But when working on https://github.com/open-telemetry/opentelemetry-go/pull/6809 I benchmarked that for "primitive values" the performance of the equality checking remains the same. As I have written before:

> The performance for primitive values should remain intact as the performance overhead would come in play when dealing with complex values.

carlosalberto (approved):

pellared:
> I approved, but blocking to prevent accidental merging before 1.8.0

@tigrannajaryan , it can be unblocked per https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v1.8.0. Thanks a lot for your help :+1:

bogdandrutu (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
pellared: @aalexand, can this be resolved?

reyang (approved):

PR #711, opened 2025-09-04
Document more explicitly the dictionary zero element requirement.

aalexand:
Right now we have a comment to the dictionary proto and a comment to the string_table field. This requirement is important enough I think that it's worth documenting it more explicitly, especially the part that even if there are no references to the particular dictionary, the zero index element must still be present.

### DIFF
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -106,19 +106,28 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
 message ProfilesDictionary {
   // Mappings from address ranges to the image/binary/library mapped
   // into that address range referenced by locations via Location.mapping_index.
+  //
+  // mapping_table[0] must always be zero value (Mapping{}) and present.
   repeated Mapping mapping_table = 1;
 
   // Locations referenced by samples via Stack.location_indices.
+  //
+  // location_table[0] must always be zero value (Location{}) and present.
   repeated Location location_table = 2;
 
   // Functions referenced by locations via Line.function_index.
+  //
+  // function_table[0] must always be zero value (Function{}) and present.
   repeated Function function_table = 3;
 
   // Links referenced by samples via Sample.link_index.
+  //
+  // link_table[0] must always be zero value (Link{}) and present.
   repeated Link link_table = 4;
 
   // A common table for strings referenced by various messages.
-  // string_table[0] must always be "".
+  //
+  // string_table[0] must always be "" and present.
   repeated string string_table = 5;
 
   // A common table for attributes referenced by various messages.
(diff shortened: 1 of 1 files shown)


opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -106,19 +106,28 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
 message ProfilesDictionary {
   // Mappings from address ranges to the image/binary/library mapped
   // into that address range referenced by locations via Location.mapping_index.
+  //
+  // mapping_table[0] must always be zero value (Mapping{}) and present.
```
bogdandrutu, line 110: @aalexand probably you have some logic behind this, do you mind documenting here so others who are using this understand why?

aalexand: There is a note in the message level comment:

> // Note that all fields in this message MUST have a zero value encoded as the first element.
> // This allows for _index fields pointing into the dictionary to use a 0 pointer value
> // to indicate 'null' / 'not set'. Unless otherwise defined, a 'zero value' message value
> // is one with all default field values, so as to minimize wire encoded size.

Replicating it is probably too verbose, but I can add a note like "For the motivation, see the message-level comment above" if that works.

aalexand: I mean I can add a note to every field's comment.

pellared: We would like to understand why a zero value is needed even if there are no references to the particular dictionary.

christos68k: I've previously [raised](https://docs.google.com/document/d/19UqPPPlGE83N37MhS93uRlxsP1_wGxQ33Qv6CDHaEp0/edit?tab=t.0#bookmark=id.kc0gwgn1otw2) this as a discussion item in the Profiling SIG. I was absent from the meeting when the actual discussion took place, but according to the agenda we went for simplicity/consistency.

pellared: It looks like a simplicity from the receiver (consumer) side and not necessarily from a producer side. It also increases the payload size.

Without having the reasoning documented anywhere this can easily get removed in future.
Notice that, for instance, the attributes restrictions (that keys have to be unique) are referencing the OTel Specification.

It is important for long-term maintainability of the project.

aalexand: We decided to follow this pattern (require that the dictionary fields are never empty and have at least the zeroth zero value element) to ensure that the `*_index` fields are never logically out-of-bounds. If the dictionary array would be allowed to be empty (e.g. link_table) then the link_index field that is empty, points to the zeroth element and that would be OOB. Consumers could special case this check, but it's better if they don't have to do that for a well-formed profile.

pellared: > Consumers could special case this check, but it's better if they don't have to do that for a well-formed profile.

I think they should do it anyways following the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle).

aalexand: It depends on what we call robust perhaps. If this requirement is a part of the standard (OTLP), the consumer could choose to check the input but reject the input that is invalid (empty dictionaries in this case). I.e. "robust" == "don't crash" in this case. But at least the rest of the code wouldn't need to be doing checks for "index zero and empty dictionary".

But if allowing empty dictionaries appears to be a strong preference, I can bring this up in Profiling SIG for discussion. One suspicion I have is that documenting that will look clumsy: we'll need to special case this single OOB access case basically along the lines of "indices point to the dictionary and they must be valid except the case where the index is zero from a default field and the dictionary is empty which is allowed". Maybe it's not too bad, but I'm not in love at first sight with this.

tigrannajaryan (approved):

PR #712, opened 2025-09-04
Rename line -> lines and sample -> samples since these are repeated f…

aalexand:
…ields.

Protobuf
[docs](https://protobuf.dev/programming-guides/style/#field-names) say:

> Use pluralized names for repeated fields.

Other fields in profiles.proto and other OTLP types follow that recommended style AFAICT. The only fields that do not are `line` and `sample`, and I think those are remnants of the now-gone exact pprof format compatibility we explored early on.

As we approach the Alpha release of the profiles signal, let's be consistent in how we name repeated fields and use plural here.


bogdandrutu (approved):

jhalliday (approved):

reyang:
@aalexand would you update the changelog? Thanks!

aalexand:
> @aalexand would you update the changelog? Thanks!

Done.

CHANGELOG.md
```diff
@@ -1,8 +1,12 @@
 # Changelog
 
+The full list of changes can be found in the compare view for the respective release at <https://github.com/open-telemetry/opentelemetry-proto/releases>.
+
 ## Unreleased
 
-The full list of changes can be found in the compare view for the respective release at <https://github.com/open-telemetry/opentelemetry-proto/releases>.
```
reyang, line 5: Thanks for fixing this!

PR #713, opened 2025-09-04
Add a note about cardinality implications for attribute values.

aalexand:


opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -121,14 +121,21 @@ message ProfilesDictionary {
...
+  // A common table for attributes referenced by the Profile, Sample, Mapping
+  // and Location messages below through attribute_indices field.
+  //
+  // It is a collection of key/value pairs with an optional unit. Examples of
+  // attributes:
   //
   //     "/http/user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36"
-  //     "/http/server_latency": 300
   //     "abc.com/myattribute": true
   //     "abc.com/score": 10.239
```
christos68k: Should we also remove this as it goes against recommendation?

```suggestion
```

jhalliday:
It's hopefully clear from context, particularly for anyone who's worked with dictionary encoding before, but in the spirit of clarity and beginner friendliness, perhaps it's worth adding a note on collection semantics whilst we're at it?
In every other OTel proto, a 'repeated KeyValue' field, despite being an array, actually operates with map semantics - the key must be unique. That is somewhat confusingly not true of attribute_table!
I [removed the incorrect restriction from the comments](https://github.com/open-telemetry/opentelemetry-proto/commit/53b4bbdc609a8144de5ef64b3d43702efdf37adb#diff-ddd0cf081a2fe76f424c617a5e0a85d00b440ddda2f9407fa4f8f670e9767f38R73-L127) but now I'm wondering if we could benefit from going further and explicitly noting it's required to be supported, though any individual 'repeated attribute_indices' indexing into it still needs to preserve the unique key semantics.

opentelemetry/proto/profiles/v1development/profiles.proto
aalexand: Good point, done. Thank you!

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -121,14 +121,22 @@ message ProfilesDictionary {
...
+  // cardinality. Using an attribute of double type in a sample can quickly make
+  // every attribute value unique, defeating the purpose of the dictionary and
+  // impractically increasing the profile size.
+  //
+  // Examples of attributes:
   //     "/http/user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36"
-  //     "/http/server_latency": 300
   //     "abc.com/myattribute": true
-  //     "abc.com/score": 10.239
+  //     "allocation_size": 128 bytes
```
pellared, line 150: Then `allocation_size` is of `string` type instead of `int`?

pellared (changes requested):
The notes about the optional unit in attributes do not seem aligned with other signals and also are not clear how the unit should be set.

opentelemetry/proto/profiles/v1development/profiles.proto
aalexand: The allocation_size attribute is of int type in this example, yes. Not sure I understand the question.

pellared: Then how is the optional unit set?

pellared: I have missed that it is of type `KeyValueAndUnit`

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -121,14 +121,22 @@ message ProfilesDictionary {
...
-  // like server name can be set using the resource API. Examples of attributes:
+  // A common table for attributes referenced by the Profile, Sample, Mapping
+  // and Location messages below through attribute_indices field. Each entry is
+  // a key/value pair with an optional unit. Since this is a dictionary table,
+  // multiple entries with the same key may be present, unlike direct attribute
+  // tables like Resource.attributes. The referencing attribute_indices fields,
+  // though, do maintain the key uniqueness requirement.
   //
+  // It's recommended to use attributes for categorical values of limited
+  // cardinality. Using an attribute of double type in a sample can quickly make
```
tigrannajaryan: >categorical values of limited cardinality

I am not sure I understand what this means. Can you rephrase please?

aalexand: Changed to

>  // It's recommended to use attributes for variables with bounded set of
>  // values, such as categorical variables
>  // (https://en.wikipedia.org/wiki/Categorical_variable). Using an attribute
>  // of a floating point type (e.g., CPU time) in a sample can quickly make
>  // every attribute value unique, defeating the purpose of the dictionary and
>  // impractically increasing the profile size.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -130,14 +130,24 @@ message ProfilesDictionary {
...
+  // A common table for attributes referenced by the Profile, Sample, Mapping
+  // and Location messages below through attribute_indices field. Each entry is
+  // a key/value pair with an optional unit. Since this is a dictionary table,
+  // multiple entries with the same key may be present, unlike direct attribute
+  // tables like Resource.attributes. The referencing attribute_indices fields,
+  // though, do maintain the key uniqueness requirement.
   //
+  // It's recommended to use attributes for variables with bounded set of
+  // values, such as categorical variables
+  // (https://en.wikipedia.org/wiki/Categorical_variable). Using an attribute
```
tigrannajaryan: Nit: Otel typically uses the term "cardinality" when referring to the number of unique values (and you also use it in the PR title), so I'd use that more commonly used term. That is, unless there is some nuance that I am missing.

I haven't heard of "Categorical variables" before (and thanks for the Wikipedia link). Can be just me :-)

tigrannajaryan (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
aalexand: Ah, let me add the cardinality term back

aalexand: Done

tigrannajaryan (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
tigrannajaryan: Thanks!

aalexand:
I think this can be merged now.

PR #714, opened 2025-09-05
profiles: improve Sample message clarity and usage documentation.

jhalliday:


jhalliday:
@open-telemetry/profiling-approvers

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -393,31 +393,42 @@ message ValueType {
 // both fields are populated, they MUST contain the same number of elements, and
 // the elements at the same index MUST refer to the same event.
 //
+// For the purposes of efficiently representing aggregated data observations, a Sample is regarded
+// as having a shared identity and an associated collection of per-observation data points.
+// Samples having the same identity may be combined by inserting timestamps and values to the data arrays.
```
christos68k: Can we switch this to `SHOULD`? We're not introducing any hard requirements but instead can signify best practices by using this term. Also since we're mentioning timestamps here, I don't think this wording conflicts with our other recommendation below (avoiding summing by encoding distinct values in separate Samples)

```suggestion
// Samples having the same identity SHOULD be combined by inserting timestamps and values to the data arrays.
```

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -393,31 +393,42 @@ message ValueType {
...
 // Report of a stacktrace at 10 timestamps (consumers must assume the value is 1 for each point):
 //    values: []
 //    timestamps_unix_nano: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
 //
 // Report of a stacktrace with an aggregated value without timestamps:
+// Note that in such case, it is not possible to determine how many aggregated observations summed to form the value.
+// If lossless encoding is preferred to space efficiency, it is recommended to instead encode the value as an attribute,
+// i.e. add more unique Sample messages. This may be required only where timestamps are not present.
 //   values: [10]
 //    timestamps_unix_nano: []
```
christos68k, line 411: Can we move this after the example, it currently reads awkward as the previous sentence ends with ':'. Also in the suggestion below, I've removed the 'value as an attribute' part to avoid confusion with the other attributes we have.

```suggestion
//   values: [10]
//    timestamps_unix_nano: []
// Note that in this case, it is not possible to determine how many aggregated observations summed to form the value.
// If lossless encoding is preferred to space efficiency, it is recommended to instead encode the value without summing,
// i.e. add more unique Sample messages, each encoding a separate value. This may be required only where timestamps are not present.
```

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -393,31 +393,42 @@ message ValueType {
...
   // References to attributes in ProfilesDictionary.attribute_table. [optional]
-  repeated int32 attribute_indices = 3;
-
+  repeated int32 attribute_indices = 2;
   // Reference to link in ProfilesDictionary.link_table. [optional]
   // It can be unset / set to 0 if no link exists, as link_table[0] is always a 'null' default value.
-  int32 link_index = 4;
+  int32 link_index = 3;
+
+  // The following fields may contain per-observation data and do not form part of the Sample's key.
```
christos68k:
```suggestion
  // The following fields may contain per-observation data and do not form part of the Sample's identity.
```

aalexand (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
jhalliday: The attributes bit is critical to understanding the model - without it we're apparently contradicting our own advise that Sample messages with the same key should be merged.

christos68k: `encode the value as an attribute` is confusing to me as someone could read it literally and try to use a `Sample` attribute when instead we're recommending that one use separate `Samples`. I'm simply rewording this text to remove the ambiguity around the word `attribute`.

Our advice is that Samples with the same identity should be combined by `inserting timestamps and values ..`. But for this specific example we don't have timestamps,  thus we're documenting it as a possible counterexample to the `SHOULD`. Essentially, we're saying that multiple `Samples` with the same identity (dupe keys) are fine in this case (lossless encoding), albeit inefficient. 

If that's not the case, are we implying that encoding separate values (that should not be summed) via distinct `Samples` is not enough and one needs to also use distinct attributes? Doesn't this go against the low-cardinality recommendation for attributes that we [recently](https://github.com/open-telemetry/opentelemetry-proto/pull/713) introduced?

jhalliday: Encoding multiple Samples with the same key is not effective - nothing prevents a processor further downstream, such as an intermediate collector, from combining them. That's perfectly legitimate architecture in e.g. cases where the originating node lacks the capability/resource to efficiently dedup, so it's left to a more capable collector node to repack the data. There is no way to say 'I'm deliberately not merging these, and you shouldn't either'. The only way to do that is to change the key, since Samples with different keys cannot be merged.  Yes, that drives up cardinality. That's the price you pay if you don't have timestamps and can't tolerate lossy encoding.

christos68k: My understanding from the last SIG meeting is that the absence of timestamps should be seen as a 'do not merge these Samples with the same identity but containing distinct values if you don't want to lose information' hint. It's true that there is no way to enforce this, but I think we're OK with documenting that this should not (or must not) take place. I don't think we mandated encoding values in Sample attributes in order to make the Sample identity unique (although it's certainly allowed).

Maybe others can chime in here, there's not a lot of clarifying information in the agenda.

Essentially, if the producer chooses to encode Samples with the same identity and no timestamps as distinct Samples then the producer strongly signals that it's not safe to merge these Samples. Any downstream processor needs to respect this choice or risk losing information.

CC: @felixge @aalexand

aalexand: Encoding the value as an attribute is unclear to me. What exactly is the value? What guarantees that that attribute is unique? Values could repeat.

Trying to step back a little, I think in general the fact that we are trying to have a way to represent both aggregated (pprof style) and disaggregated (tracing style) data adds the complexity and ambiguity. pprof doesn't have this problem because individual "rows" / Samples do not matter in terms of identity as they can be summed as long as their keys match (stack + labels). But here we are trying to come with a way to reliably carry disaggregated data and this does not seem to be trivial.

Trying to pick an example we can walk through:

Say, we want to represent a live heap profile. Let's say there are these individual allocations:

```
main -> foo; 128 bytes
main -> foo; 128 bytes
main -> foo; 250 bytes
main -> bar; 500 bytes
```

Normally in pprof this would be recorded in an aggregated form, like

```
main -> foo; space|count=256b|2; tag[size]=128b
main -> foo; space|count=250b|1; tag[size]=250b
main -> bar; space|count=500b|1; tag[size]=500b
```

In OpenTelemetry format in the aggregated form this would be recorded as two profiles:

```
space:
  main -> foo; values=[256b], timestamps=[]; attr[size]=128b
  main -> foo; values=[250b]; timestamps=[]; attr[size]=250b
  main -> bar; values=[500b]; timestamps=[]; attr[size]=500b
 
count:
  main -> foo; values=[2], timestamps=[]; attr[size]=128b
  main -> foo; values=[1]; timestamps=[]; attr[size]=250b
  main -> bar; values=[1]; timestamps=[]; attr[size]=500b
```

Now, IIUC we want to say that someone wants to disaggregate the first item and carry the individual recordings separately. Something like:

```
space:
  main -> foo; values=[256b, 256b], timestamps=[]; attr[size]=128b
  main -> foo; values=[250b]; timestamps=[]; attr[size]=250b
  main -> bar; values=[500b]; timestamps=[]; attr[size]=500b
 
count:
  main -> foo; values=[1, 1], timestamps=[]; attr[size]=128b
  main -> foo; values=[1]; timestamps=[]; attr[size]=250b
  main -> bar; values=[1]; timestamps=[]; attr[size]=500b
```

Is that what we want? I think the first problem I am seeing with this overall is that for an array of size one I don't know whether it's one event or it is the aggregated value.

Maybe I'm missing something. But I think we need to use an example like the above to capture discussion points, this will help nail down the issues and then we can transfer that to our docs.

christos68k: How about:

```
space:
  main -> foo; values=[128b, 128b], timestamps=[]
  main -> foo; values=[250b]; timestamps=[]
  main -> bar; values=[500b]; timestamps=[]
```

Meaning, we don't need to also separately encode `count` as it can be inferred from the values we send. 

A processor may decide to merge the first two Samples without losing information as such:

```
space:
  main -> foo; values=[128b, 128b, 250b], timestamps=[]
  main -> bar; values=[500b]; timestamps=[]
```

But if the values are aggregated, some information is lost:

```
space:
  main -> foo; values=[256b], timestamps=[]
  main -> foo; values=[250b]; timestamps=[]
  main -> bar; values=[500b]; timestamps=[]
```

or

```
space:
  main -> foo; values=[506b], timestamps=[]
  main -> bar; values=[500b]; timestamps=[]
```

aalexand: I think there is ambiguity in case of array of size 1.

Let's say the profile is

```
space:
  main -> foo; values=[256b], timestamps=[]
```

and I want to run something like `pprof_otel --some-flag-to-ask-to-show-counts ... profile.pb.gz`. How will the consumer know whether the count is here is 1 or `unknown because this is an aggregated value`?

christos68k: We can tag with the allocation size (as in your example, assumed to be low cardinality) then the count can be derived. But that's only needed if aggregation has taken place, so we could specify that absent any such tagging, the values are discrete events (and count implicitly 1 for each of them).

One reason I don't like splitting related information (space and count) across two different Profiles, is that we'd be introducing ordering requirements that link the two Profiles together, but these are implicit, there's nothing in the protocol that makes them concrete.

jhalliday: > We can tag with the allocation size (as in your example, assumed to be low cardinality) then the count can be derived.

Yes, that. You trade cardinality increase vs data loss. If you can't tolerate data loss, you prevent it by increasing cardinality. Expressing that technique is why I want to keep the bit about attributes in the doc in some form. The attributes are how you increase the cardinality.

> But that's only needed if aggregation has taken place, so we could specify that absent any such tagging, the values are discrete events (and count implicitly 1 for each of them).

You can't easily tell if aggregation has taken place or not. The only way to do that is to delve into the attributes, locate and interpret the special one (e.g. 'size'), which involves semantics that are unspecified. Further, the merge gets harder, because (again), you can't tell the difference between 'I didn't aggregate this, but you can' and 'I didn't aggregate this, and you should not either'. It's also why I toyed briefly with adding an 'occurrence_count' field to the proto message - it avoids the need to use a special attribute and instead uses a special field.

An alternative is to require the values and timestamps arrays to always have the same length and use a '0' timestamp to indicate 'not known'. Thus `main -> foo; values=[128], timestamps=[0]` is always unambiguously a single unaggregated observation, albeit one that occurred at an unspecified time, and `main -> foo; values=[128], timestamps=[0]` + `main -> foo; values=[128], timestamps=[0]` always merges as `main -> foo; values=[128,128], timestamps=[0,0]`  but never as `main -> foo; values=[256], timestamps=[0]`. Here we trade wire message size to prevent cardinality increases. It would be even cheaper (at least without compression) if we went back to uint64 timestamps. My concern with this is not so much the wire bytes cost, but the need for renderers to poke about in the timestamps to figure out what they can reasonably do with the data, rather than simply looking at the 'shape' of the arrays to distinguish 'this can/can't be meaningfully visualized on a timeline'. It's also less flexible, in that you can't choose data loss even if you want to.

PR #717, opened 2025-09-15
Fixes #207 - Add Style guide for proto comments.

### LINKED ISSUE #207: Consistent style for documentation comments
I notice there are three patterns for comments

    Go-style, repeating name in javadoc e.g. Sum represents the type
    Starting with It, otherwise similar to go-style
    Java-style, documenting behavior directly e.g. Represents the type of a metric

Should this be made consistent? I have a personal preference to Java-style for protos since I've seen that more in the past, but it looks like we use Go-style more here so consolidating on that would make sense too.

jsuereth:
Fixes #207

- Pick a naming convention for proto comments
- document it in CONTRIBUTING
- apply it to all .proto files

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -25,11 +25,11 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@d3678e237b9c32a6c9bffb3315c335f976f3549f # v3.30.2
+        uses: github/codeql-action/init@192325c86100d080feab897ff886c34abd4c83a3 # v3.30.3
         with:
           languages: actions
 
       - name: Perform CodeQL analysis
-        uses: github/codeql-action/analyze@d3678e237b9c32a6c9bffb3315c335f976f3549f # v3.30.2
+        uses: github/codeql-action/analyze@192325c86100d080feab897ff886c34abd4c83a3 # v3.30.3
         with:
           category: "/language:actions"
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -42,6 +42,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@d3678e237b9c32a6c9bffb3315c335f976f3549f # v3.30.2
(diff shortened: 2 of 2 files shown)


tigrannajaryan (approved):

opentelemetry/proto/common/v1/common.proto
```diff
@@ -59,24 +59,35 @@ message KeyValueList {
...
   string key = 1;
+
+  // The value of the pair.
   AnyValue value = 2;
 }
 
 // InstrumentationScope is a message representing the instrumentation scope information
 // such as the fully qualified name and version. 
 message InstrumentationScope {
+  // A fully qualified name.
```
tigrannajaryan: I am not sure if we define that this is "fully qualified" name anywhere in the spec (and what that means). Do we?

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -58,7 +58,7 @@ message MetricsData {
   // For data coming from a single resource this array will typically contain
   // one element. Intermediary nodes that receive data from multiple origins
   // typically batch the data before forwarding further and in that case this
-  // array will contain multiple elements.
+// array will contain multiple elements.
```
tigrannajaryan:
```suggestion
  // array will contain multiple elements.
```

opentelemetry/proto/metrics/v1/metrics.proto
```diff
@@ -228,25 +228,28 @@ message Metric {
 // AggregationTemporality is not included. Consequently, this also means
 // "StartTimeUnixNano" is ignored for all data points.
 message Gauge {
+  // The time series point values for this gauge.
```
tigrannajaryan: It may be data points all with the same timestamp and different attributes and calling that "time series" I think may be confusing.

Perhaps just say
> // The data points for this gauge.

Although I am not sure there is much value in such comments. Are comments mandated? Should we skip low-value comments like this?

opentelemetry/proto/common/v1/common.proto
jsuereth: Ah, above just uses "such as", I'll update this.

opentelemetry/proto/metrics/v1/metrics.proto
jsuereth: Comments are mandated now, which is why i added.

This structure actually HAS caused some confusion in the past.  The wording I use is to match our datamodel.  I think what's confusing you is that "time series" is both singular and plural, and I'm using the plural here.  I actually WANTED to have this comment denote that you'll get multiple time series in here (where time series = unique label sets) for a given metric, even if they all have the same timestamp.

SO I think we both agree there's something important to call out here.  Not sure how to phrase it.

jsuereth: Ok - I reworded and focused on what I think is the important bit to call out, PTAL

lmolkova (approved):

PR #673, opened 2025-06-13
Fix Markdown checks

fandreuz:
In this PR I:
- Pinpoint `ycheeverse/lychee` to `sha-3a09227-alpine` (check lycheeverse/lychee#1729)
- Change the `Makefile` recipe `markdownlint` to run a Docker container
- Disable [MD024](https://github.com/DavidAnson/markdownlint/blob/main/doc/md024.md, since it seems to be buggy(?)
- Disable [MD059](https://github.com/DavidAnson/markdownlint/blob/main/doc/md059.md), the transgressions I see don't look that bad to me:
```
CONTRIBUTING.md:27:176 MD059/descriptive-link-text Link text should be descriptive [Context: "[here]"]
```
```md
View meeting notes of previous SIG calls in this [google doc](https://docs.google.com/spreadsheets/d/1SYKfjYhZdm2Wh2Cl6KVQalKg_m4NhTPZqq-8SzEVO6s/edit?gid=0#gid=0) as stated [here](https://github.com/open-telemetry/community/?tab=readme-ov-file#governing-bodies) to stay up to date.
```

PR #721, opened 2025-10-01
Exclude buf.yaml from release archives via .gitattributes (#604)

### LINKED ISSUE #604: opentelemetry-proto tar archive contains spurious files
The file opentelemetry-proto-1.4.0.tar.gz contains the following:

* `buf.yaml`
* `package.json`
* `package-lock.json`

which are also present in git in the opentelemetry-proto repository.

These do not look like items that should be maintained in git, and should be removed.

RajaVarmaGVSSR:
Adds an export-ignore rule for buf.yaml to prevent it from appearing in GitHub release
tarballs and zip files.

Background:
- Release archives currently include all tracked files, which previously caused
  buf.yaml to appear in the tarball.
- package.json and package-lock.json are no longer in the repo, so no action needed.

Impact:
- Low-risk, does not affect .proto files or SDK consumers.
- Existing release workflow remains unchanged.

Closes #604

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
 * text=auto eol=lf
+
+buf.yaml export-ignore


.gitattributes
```diff
@@ -1 +1,3 @@
 * text=auto eol=lf
+
+buf.yaml export-ignore
```
tigrannajaryan, line 3: Do we also want to exlclude package.json and package-lock.json as was reported on the [issue](https://github.com/open-telemetry/opentelemetry-proto/issues/604)? @marcalff you opened the issue, any thoughts?

RajaVarmaGVSSR: I see those files have been deleted as part of https://github.com/open-telemetry/opentelemetry-proto/pull/673

lmolkova (changes requested):
Is there actually a need to remove this file? Does it create any problems?

It's part of the validation and code-generation and we ship proto files as they are in the archive. We expect consumers of the release to generate code. In a generic case, without this file, consumers might not be able to generate code we provide.

PR #724, opened 2025-10-03
profiles: improve Sample message clarity and usage documentation.

jhalliday:


opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -363,17 +367,21 @@ message ValueType {
 //    values: [2, 2, 3, 3]
 //    timestamps_unix_nano: [1, 2, 3, 4]
 message Sample {
+
+  // A Sample's identity (i.e. 'primary key') is the tuple of {stack_index, sorted(attribute_indices), link_index}
```
felixge: NIT: Sorting seems like an implementation detail, would be easier to just talk about treating the attribute_indices as a set for comparison purposes here.

jhalliday: changed to set_of(attribute_indices)

aalexand:
We also discussed in one of the SIG meetings that we should document that the "shape" of the samples in terms of values vs timestamps should be the same for all samples in the given profile. For example, we disallow having one sample with timestamps and another a timestamp-less aggregated value sample.

Do we want to add that note here or in a separate PR?

christos68k (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -349,6 +349,10 @@ message ValueType {
 // both fields are populated, they MUST contain the same number of elements, and
 // the elements at the same index MUST refer to the same event.
 //
+// For the purposes of efficiently representing aggregated data observations, a Sample is regarded
+// as having a shared identity and an associated collection of per-observation data points.
```
jsuereth, line 368: I think this is *against* the goal of sharing Resource  and its notion of identity with OpenTelemetry protos.

jsuereth (changes requested):
As discussed offline, I don't think we should move this direction.

- This does not allow sharing of resource/transformation code in the collector because we just move the data from resource elsewhere, so no sharing
- We should figure out efficiency issues for profiling separately.

opentelemetry/proto/profiles/v1development/profiles.proto
christos68k: I don't quite understand this but I've been away for a few weeks and missed the last two SIG meetings where this may have been discussed. 

Based on the recent Slack discussion, my impression is that the point of contention regarding Profiles is grouping Profiles (and thus Samples) by container (which forces us to encode process metadata as Sample attributes) rather than grouping Profiles by process (which would let us encode process metadata as Resource attributes).

This PR mostly clarifies how Sample datapoints (e.g. value, timestamp) are to be encoded in order to avoid Sample duplication, meaning that the clarification would apply regardless of which grouping we choose.

Is there something else here that is related to the Resource attribute discussion that I missed?

aalexand: +1 to @christos68k, this change merely improves the documentation around the general rules of what constitutes the identity of a sample in the profiling signal OTLP. It is orthogonal to specific labels, like resource reference, should we introduce that.

aalexand: I'm resolving this. @jsuereth reopen if needed.

felixge (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -364,6 +364,10 @@ message ValueType {
...
```
jsuereth, line 368: Shouldn't the identity (implicitly) have the Resource->Scope in it?

Just want to confirm understanding, the wording here is fine for now.

aalexand: Yes, the wording about the identity here is for the samples within a single `Profile`. Different profiles (i.e. profiles on the different "branches" of the Resource -> Scope tree) can each have their own sample with the same {stack_index, set_of(attribute_indices), link_index} identity.

jsuereth (approved):
Thanks for updating this folks!  Approving

PR #727, opened 2025-10-17
Fix schema_url comments to include scope field coverage

Gandem:
We were reviewing the profiling spec, and noticed that for all signals, there's a schema_url field on:
* The ResourceProfiles/Logs/... level, that applies to resource attributes ([here](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L62-L63))
* The ScopeProfiles/Logs/... level, that applies to all attributes under profiles/... ([here](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L81))

The spec is currently not explicit on whether the scope `schema_url` applies to scope attributes. After discussing in [this slack thread](https://cloud-native.slack.com/archives/C01N7PP1THC/p1760622388508689), it sounds like the original intent is for this to be the case.

This change updates comments in all four signal proto files (logs, trace, metrics, profiles) to reflect that the schema_url field applies to both the instrumentation scope and the telemetry data fields, not just the telemetry data.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -25,11 +25,11 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@f443b600d91635bebf5b0d9ebc620189c0d6fba5 # v4.30.8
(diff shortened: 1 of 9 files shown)


felixge (approved):
LGTM, but I suspect other reviewers would benefit from a bit more explanation on why this change is being proposed.

lmolkova (approved):

PR #729, opened 2025-10-22
profiles: drop profile.comment_strindices

florianl:
Follow up as https://github.com/open-telemetry/semantic-conventions/pull/2861 got merged.

### DIFF
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -149,16 +149,6 @@ message ProfilesDictionary {
   //     "abc.com/myattribute": true
   //     "allocation_size": 128 bytes
   //
-  // The attribute values SHOULD NOT contain empty values.
-  // The attribute values SHOULD NOT contain bytes values.
-  // The attribute values SHOULD NOT contain array values different than array of string values, bool values, int values,
-  // double values.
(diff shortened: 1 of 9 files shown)


christos68k (approved):

aalexand (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -285,25 +285,19 @@ message Profile {
...
-  // is cleaned.
-  repeated int32 comment_strindices = 7; // Indices into ProfilesDictionary.string_table.
 
   // A globally unique identifier for a profile. The ID is a 16-byte array. An ID with
   // all zeroes is considered invalid. It may be used for deduplication and signal
   // correlation purposes. It is acceptable to treat two profiles with different values
   // in this field as not equal, even if they represented the same object at an earlier
   // time.
   // This field is optional; an ID may be assigned to an ID-less profile in a later step.
-  bytes profile_id = 8;
```
lmolkova, line 292: we're still in development status, so it's not that important, but I believe we should not update the indices of existing fields - https://protobuf.dev/programming-guides/proto3/#deleting

aalexand: We've been doing this a few times as we agreed we'll do this before we go to alpha.

PR #732, opened 2025-10-27
Clarify dictionary guidelines: duplicates and orphaned items

aalexand:


christos68k (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -97,12 +97,25 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
...
+//    minimize wire encoded size.
+// 3) There SHOULD NOT be dupes in a dictionary. The identity of dictionary
+//    items is based on their value, recursively as needed. If a particular
+//    implementation does emit duplicated items, it MUST NOT attempt to give
+//    them meaning based on the index or order. A profile processor may remove
+//    duplicate items and this MUST NOT have any observable effects for
+//    consumers.
+// 4) There SHOULD NOT be orphaned (unreferenced) items in a dictionary. A
+//    profile processor may remove ("garbage-collect") orphaned items and this
+//    MUST NOT have any observable effects for consumers.
```
pellared: nit: Can we use consistent formatting, e.g. following https://github.com/open-telemetry/opentelemetry-proto/blob/8222b4281d6c00cde99ce7eac4418ce3a196061d/opentelemetry/proto/profiles/v1development/profiles.proto#L239-L259

```suggestion
//
// - A dictionary is an array of dictionary items. Users of the dictionary
//   compactly reference the items using the index within the array.
//
// - A dictionary MUST have a zero value encoded as the first element. This
//   allows for _index fields pointing into the dictionary to use a 0 pointer
//   value to indicate 'null' / 'not set'. Unless otherwise defined, a 'zero
//   value' message value is one with all default field values, so as to
//   minimize wire encoded size.
//
// - There SHOULD NOT be dupes in a dictionary. The identity of dictionary
//   items is based on their value, recursively as needed. If a particular
//   implementation does emit duplicated items, it MUST NOT attempt to give
//   them meaning based on the index or order. A profile processor may remove
//   duplicate items and this MUST NOT have any observable effects for
//   consumers.
//
// - There SHOULD NOT be orphaned (unreferenced) items in a dictionary. A
//   profile processor may remove ("garbage-collect") orphaned items and this
//   MUST NOT have any observable effects for consumers.
```

pellared:
Can you please an entry in `CHANGELOG.md`?

lmolkova:
@aalexand could you please address comments? I'm planning to release proto asap and would like to include this PR there.

thanks!

opentelemetry/proto/profiles/v1development/profiles.proto
aalexand: Done.

PR #733, opened 2025-10-30
[resource] introduce reference based attributes

florianl:
Reduce duplicate entries by using a reference based approach in combination with a lookup table.

It should help drive the discussion around the use or Resources.

We will include a int32 string_value_ref field within the AnyValue message. This allows data points to reference strings via an integer index rather than carrying the full string value.

- Collector Efficiency: This is the most pragmatic path to simplify the OpenTelemetry Collector's internal data model (pdata) and streamline code generation.

- Consolidation: While creating a separate "Ref" version of AnyValue was considered, it was determined that the confusion caused by a new type would be equal to or greater than simply adding a field to the existing message.

- Minimized Impact: In signals where no dictionary is provided (like standard Traces or Metrics), this field has no context and should be ignored, minimizing the risk of accidental misuse.

### DIFF
--- a/.github/workflows/build-check.yaml
+++ b/.github/workflows/build-check.yaml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Run docker pull
         run: make docker-pull
 
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Generate C++
         run: make gen-cpp
 
@@ -32,7 +32,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Generate C#
         run: make gen-csharp
 
@@ -41,7 +41,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Generate GoLang
         run: make gen-go
 
@@ -50,7 +50,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Generate Java
         run: make gen-java
 
@@ -59,7 +59,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Generate ObjC
         run: make gen-objc
 
@@ -68,7 +68,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Generate OpenAPI
         run: make gen-openapi
 
@@ -77,7 +77,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Generate PhP
         run: make gen-php
 
@@ -86,7 +86,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Generate Python
         run: make gen-python
 
@@ -95,7 +95,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Generate Ruby
         run: make gen-ruby
     
@@ -104,7 +104,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Generate Kotlin
         run: make gen-kotlin
 
@@ -113,7 +113,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       # breaking-change checks against last published release which is determined
       # using the last published tag
       - name: Get tags
@@ -130,7 +130,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: run markdown-link-check
         run: make markdown-link-check
@@ -139,7 +139,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: run markdownlint
         run: make markdownlint
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -22,14 +22,14 @@ jobs:
       security-events: write  # for github/codeql-action/analyze to upload SARIF results
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@0499de31b99561a6d14a36a5f662c2a54f91beee # v4.31.2
+        uses: github/codeql-action/init@6bc82e05fd0ea64601dd4b465378bbcf57de0314 # v4.32.1
         with:
           languages: actions
 
       - name: Perform CodeQL analysis
-        uses: github/codeql-action/analyze@0499de31b99561a6d14a36a5f662c2a54f91beee # v4.31.2
+        uses: github/codeql-action/analyze@6bc82e05fd0ea64601dd4b465378bbcf57de0314 # v4.32.1
         with:
           category: "/language:actions"
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -12,7 +12,7 @@ jobs:
   fossa:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - uses: fossas/fossa-action@3ebcea1862c6ffbd5cf1b4d0bd6b3fe7bd6f2cac # v1.7.0
         with:
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -19,7 +19,7 @@ jobs:
       # Needed for GitHub OIDC token if publish_results is true
       id-token: write
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
 
@@ -33,7 +33,7 @@ jobs:
       # uploads of run results in SARIF format to the repository Actions tab.
       # https://docs.github.com/en/actions/advanced-guides/storing-workflow-data-as-artifacts
       - name: "Upload artifact"
-        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         with:
           name: SARIF file
           path: results.sarif
@@ -42,6 +42,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@0499de31b99561a6d14a36a5f662c2a54f91beee # v4.31.2
+        uses: github/codeql-action/upload-sarif@6bc82e05fd0ea64601dd4b465378bbcf57de0314 # v4.32.1
         with:
           sarif_file: results.sarif
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,13 @@ The full list of changes can be found in the compare view for the respective rel
 
 ### Changed
 
+- profiles: clarify Sample message usage guidance docs. [#742](https://github.com/open-telemetry/opentelemetry-proto/pull/742)
+- profiles: improve Sample message clarity and usage documentation. [#724](https://github.com/open-telemetry/opentelemetry-proto/pull/724)
+- profiles: clarify the profile and sample timestamps relationship. [#744](https://github.com/open-telemetry/opentelemetry-proto/pull/744)
+
 ### Fixed
+
+- logs: `SEVERITY_NUMBER_UNSPECIFIED` can be used as the `severity_number` field is optional. [#736](https://github.com/open-telemetry/opentelemetry-proto/pull/736)
 
 ### Removed
 
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ To generate the raw gRPC client libraries, use `make gen-${LANGUAGE}`. Currently
 components as indicated by the Maturity table below.
 
 | Component | Binary Protobuf Maturity | JSON Maturity |
-| --------- |--------------- | ------------- |
+| --------- | --------------- | ------------- |
 | common/* | Stable | [Stable](docs/specification.md#json-protobuf-encoding) |
 | resource/* | Stable | [Stable](docs/specification.md#json-protobuf-encoding) |
 | metrics/\*<br>collector/metrics/* | Stable | [Stable](docs/specification.md#json-protobuf-encoding) |
--- a/dependencies.Dockerfile
+++ b/dependencies.Dockerfile
@@ -1,3 +1,3 @@
 # This is a renovate-friendly source of Docker images.
-FROM davidanson/markdownlint-cli2:v0.18.1@sha256:173cb697a255a8a985f2c6a83b4f7a8b3c98f4fb382c71c45f1c52e4d4fed63a AS markdownlint
+FROM davidanson/markdownlint-cli2:v0.20.0@sha256:8f7487027e6786a583da1c8be8b0541bbd17582a4b8fbe3e27a29d9f29fa77d0 AS markdownlint
 FROM lycheeverse/lychee:sha-3a09227-alpine@sha256:5853bd7c283663a1200dbb15924a5047f8d4c50adfa7a4c212a94f04bbac831c AS lychee
--- a/opentelemetry/proto/logs/v1/logs.proto
+++ b/opentelemetry/proto/logs/v1/logs.proto
@@ -85,7 +85,6 @@ message ScopeLogs {
 
 // Possible values for LogRecord.SeverityNumber.
 enum SeverityNumber {
-  // UNSPECIFIED is the default SeverityNumber, it MUST NOT be used.
   SEVERITY_NUMBER_UNSPECIFIED = 0;
   SEVERITY_NUMBER_TRACE  = 1;
   SEVERITY_NUMBER_TRACE2 = 2;
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -284,9 +284,14 @@ message Profile {
   // The following fields 3-12 are informational, do not affect
   // interpretation of results.
 
-  // Time of collection (UTC) represented as nanoseconds past the epoch.
+  // Time of collection. Value is UNIX Epoch time in nanoseconds since 00:00:00
+  // UTC on 1 January 1970.
   fixed64 time_unix_nano = 3;
-  // Duration of the profile, if a duration makes sense.
+  // Duration of the profile. For instant profiles like live heap snapshot, the
+  // duration can be zero but it may be preferable to set time_unix_nano to the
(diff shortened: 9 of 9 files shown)


tigrannajaryan:
I would like to see some benchmarks showing the impact of the change on profiling signals. I think we want to see the following with and without ref-based attributes on a few production-like payloads:
- Wire size uncompressed.
- Wire size compressed with zstd and/or gzip.
- In-memory size in one popular language (preferably Go for Collector impact).
- Serialization/deserialization speed (CPU time), including dictionary construction if string interning is not already somehow happening at the source.

This is a pretty significant change so I want benchmarks to support the change.

opentelemetry/proto/resource/v1/resource.proto
```diff
@@ -42,4 +42,21 @@ message Resource {
...
+  repeated StringKeyValue = 4;
 }
+
+// Represents a reference based key-value pair that is used to store attributes.
+message StringKeyValue {
+  // Reference to the key name of the pair in ProfilesDictionary(?).string_table.
+  int32 key_index = 1;
+
+  // Reference to the value of the pair in ProfilesDictionary(?).string_table.
+  int32 value_index = 2;
```
felixge: Do we want the value type to be restricted to strings? If not, we should probably do it [like we do for the attributes in profiling](https://github.com/open-telemetry/opentelemetry-proto/blob/024778c324d10ccaefa8b5bf7ff50da126f6fdd7/opentelemetry/proto/profiles/v1development/profiles.proto#L459C3-L459C52)?

Of course this means no dictionary for the values, but they're supposed to be [low cardinality](https://github.com/open-telemetry/opentelemetry-proto/blob/024778c324d10ccaefa8b5bf7ff50da126f6fdd7/opentelemetry/proto/profiles/v1development/profiles.proto#L140-L141).

```suggestion
  opentelemetry.proto.common.v1.AnyValue value = 2;
```

Edit: I see @tigrannajaryan asked similar questions above.

felixge:
> This is a pretty significant change so I want benchmarks to support the change.

Agreed. @Gandem and I started working on this yesterday. Our rough plan is the following:

- **Capture OTLP data from a suitable workload**. We're thinking a python app that forks one process per CPU core on a 16+ core machine. We can add more workload recordings later.
- **Implement a otlp-bench tool in [sig-profiling](https://github.com/open-telemetry/sig-profiling):** The tool takes OTLP profiling payloads and converts them into the new proposed dictionary encoding to compare the resulting size. This tool should also be suitable for other experiments in the future.

We hope to have it done by the Nov 13 profiling SIG meeting.

opentelemetry/proto/resource/v1/resource.proto
florianl: Just wanted to get the discussion started.

The recommendation in https://github.com/open-telemetry/opentelemetry-proto/pull/733#issuecomment-3468275186 sounds promissing:
```
message AnyValue {
  oneof value {
    ...
    uint32 string_ref = 8;
  }
}
```

I think, we definitely should also reference string values.

tigrannajaryan:
> * **Implement a otlp-bench tool in [sig-profiling](https://github.com/open-telemetry/sig-profiling):** The tool takes OTLP profiling payloads and converts them into the new proposed dictionary encoding to compare the resulting size. This tool should also be suitable for other experiments in the future.

@felixge My old bench from early OTLP designs is here: https://github.com/tigrannajaryan/exp-otelproto

Feel free to borrow anything if there is any useful/relevant stuff there. Might be outdated/bitrotten since it's been a while.

opentelemetry/proto/resource/v1/resource.proto
felixge: Yeah, giving AnyValue the ability to reference into the string table would be useful. But it seems like that would be an additional step we could take that doesn't need to be in the scope of this PR? Especially if we apply the change I suggested.

opentelemetry/proto/resource/v1/resource.proto
```diff
@@ -42,4 +42,21 @@ message Resource {
...
   // Status: [Development]
   repeated opentelemetry.proto.common.v1.EntityRef entity_refs = 3;
+
+  // Set of attributes that describe the resource.
+  // Key and Value combination MUST be unique.
+  //
+  // Note: This is currently used exclusively in the Profiling signal.
+  //
+  // Status: [Development]
+  repeated StringKeyValue = 4;
```
felixge: This is missing a field name. You can merge https://github.com/florianl/opentelemetry-proto/pull/3 to fix it if you like.

florianl: I must have missed that. Thanks for the update - approved and merged :white_check_mark:

opentelemetry/proto/resource/v1/resource.proto
```diff
@@ -42,4 +42,21 @@ message Resource {
...
+  //
+  // Note: This is currently used exclusively in the Profiling signal.
+  //
+  // Status: [Development]
+  repeated StringKeyValue dict_attributes = 4;
 }
+
+// Represents a reference based key-value pair that is used to store attributes.
+message StringKeyValue {
+  // Reference to the key name of the pair in ProfilesDictionary(?).string_table.
```
florianl:
```suggestion
  // Reference to the key name of the pair in ProfilesDictionary.string_table.
```

opentelemetry/proto/resource/v1/resource.proto
```diff
@@ -42,4 +42,21 @@ message Resource {
...
+  // Status: [Development]
+  repeated StringKeyValue dict_attributes = 4;
 }
+
+// Represents a reference based key-value pair that is used to store attributes.
+message StringKeyValue {
+  // Reference to the key name of the pair in ProfilesDictionary.string_table.
+  int32 key_index = 1;
+
+  // Reference to the value of the pair in ProfilesDictionary(?).string_table.
```
florianl:
```suggestion
  // Reference to the value of the pair in ProfilesDictionary.string_table.
```

opentelemetry/proto/resource/v1/resource.proto
florianl: Updating AnyValue with the ability to use a reference as a next step sound good to me. Applying the changes to `message Resource` in a first step limits also the number of follow up changes in other components, like OTel collector or SDKs.

tigrannajaryan: >Applying the changes to message Resource in a first step limits also the number of follow up changes in other components, like OTel collector or SDKs.

I disagree that this is a good approach. It creates an intermediary state that is going to be short-lived, which we have to get rid of once AnyValue gains ref capability. So the anticipated changes to the protocol look like this:

- Add ref attributes to Resource.
- Add ref to AnyValue.
- Deprecate ref attributes in Resource, but keep supporting both refs.
- Wait for a few years for the next major version of OTLP where we can make breaking changes.
- Delete ref attributes from Resource.

Consider instead this:
- Add refs to AnyValue. Done.

opentelemetry/proto/resource/v1/resource.proto
```diff
@@ -42,4 +42,21 @@ message Resource {
...
   // Status: [Development]
   repeated opentelemetry.proto.common.v1.EntityRef entity_refs = 3;
+
+  // Set of attributes that describe the resource.
+  // Key and Value combination MUST be unique.
+  //
+  // Note: This is currently used exclusively in the Profiling signal.
+  //
+  // Status: [Development]
+  repeated StringKeyValue dict_attributes = 4;
```
tigrannajaryan: What happens if the same key is present here and in `attributes`? Do we need precedence rules? Is it invalid state?

tigrannajaryan (changes requested):
Blocking until we have clarity on the way forward.

opentelemetry/proto/resource/v1/resource.proto
florianl: My suggestion is that if there are elements in `dict_attributes`, then `attributes` must not hold any element.

opentelemetry/proto/resource/v1/resource.proto
florianl: @tigrannajaryan applied the requested changes with https://github.com/open-telemetry/opentelemetry-proto/pull/733/commits/c8b22c66cb8777212fd03cfb35b1a37cb2490282.

tigrannajaryan:
> The rough plan is that the new `dict_attributes` will be opaque to the collector. The idea is to adjust the codegen for pdata so that [Resource.Attributes()](https://github.com/open-telemetry/opentelemetry-collector/blob/a5343f9ec6713b7a9274ea72e7704f83a76f9671/pdata/pcommon/generated_resource.go#L47-L50) returns the regular attributes as well as any attributes from `dict_attributes`. What do you think?

Let's set aside what happens inside the Collector for a moment. 

I am looking at a scenario where a pair of OTLP sender and receiver of different versions try to communicate (can be SDK, Collector, a backend, etc).

What happens if the OTLP sender uses the new version of protocol and encodes some attributes via dictionary, however the receiver is an old version and does not know anything about dictionaries? The receiver will see `AnyValue` with a oneof set to an unknown proto field number. (I am assuming we go with the current state of this PR that add `string_ref` to `AnyValue`). It will be most likely treated as an empty `AnyValue` by most receivers and will result in data loss or at best detected as unknown/invalid `AnyValue` (and still result in data loss or impossibility to communicate).

I do not think this is acceptable.

>Can you sync with @jsuereth on this? We discussed this with him in the last [SIG meeting](https://docs.google.com/document/d/19UqPPPlGE83N37MhS93uRlxsP1_wGxQ33Qv6CDHaEp0/edit?tab=t.0) and he didn't see large obstacles, e.g. the need for capability negotiation.

@jsuereth let's discuss in spec meeting today if you are around. @felixge and other @open-telemetry/profiling-maintainers it would be good if you can join too.

opentelemetry/proto/resource/v1/resource.proto
```diff
@@ -42,4 +42,22 @@ message Resource {
...
   repeated opentelemetry.proto.common.v1.EntityRef entity_refs = 3;
+
+  // Set of attributes that describe the resource.
+  // Key and Value combination MUST be unique.
+  // attributes MUST NOT be used if dict_attributes is used.
+  //
+  // Note: This is currently used exclusively in the Profiling signal.
+  //
+  // Status: [Development]
+  repeated RefKeyValue dict_attributes = 4;
```
bogdandrutu: I would put the dictionary in the "ResourceLogs/ResourceSpans/ResourceMetrics" and be shared between all attributes coming from a resource?

happy to hear others what they think

opentelemetry/proto/common/v1/common.proto
```diff
@@ -36,6 +36,13 @@ message AnyValue {
     ArrayValue array_value = 5;
     KeyValueList kvlist_value = 6;
     bytes bytes_value = 7;
+    // Reference to the string value in ProfilesDictionary.string_table.
+    // string_value MUST NOT be set if string_ref is used.
+    //
+    // Note: This is currently used exclusively in the Profiling signal.
+    //
+    // Status: [Development]
+    int32 string_ref = 8;
```
bogdandrutu: Should we use fixed32 instead? It is faster but a bit larger on the wire.

opentelemetry/proto/resource/v1/resource.proto
Gandem: Thanks for the suggestion. Quick context: the technical committee (cc @jsuereth) recommended we follow the OpenTelemetry resource model in the [opentelemetry-ebpf-profiler](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/) and align with the SDKs’ resource definition.

Today the profiler emits one resource per container (container.id) plus one for everything else. Aligning with the SDKs would mean one resource per process id.

In forking runtimes like Python, a single container often runs many processes. Those processes share almost identical resource attributes (container.id, container.name, etc.). Replicating them across 10 to 20 processes increases the profiling payload, especially the collector’s in-memory footprint.

The benchmarks I’m running with @felixge aim to quantify that overhead and the savings from a dictionary defined at the top level.

If we scope the dictionary to each Resource or inside ResourceLogs/ResourceSpans/ResourceMetrics, it can’t de-duplicate across separate Resource instances once we move to per-process resources. That defeats the goal and wouldn’t prevent the payload growth we’re trying to avoid.

opentelemetry/proto/common/v1/common.proto
aalexand: I think I would keep varint because then we could use lower values for more commonly referenced values and save wire size.

aalexand: In profiles.proto we use `*_strindex` / `*_strindices` as the suffix for string reference fields. Should we use it here as well for consistency? E.g as `string_strindex` or simply as `strindex`.

opentelemetry/proto/resource/v1/resource.proto
```diff
@@ -42,4 +42,22 @@ message Resource {
...
+}
+
+// Represents a reference based key-value pair that is used to store attributes.
+message RefKeyValue {
+  // Reference to the key name of the pair in ProfilesDictionary.string_table.
+  int32 key_index = 1;
```
aalexand: key_strindex?

opentelemetry/proto/common/v1/common.proto
florianl: resolving with https://github.com/open-telemetry/opentelemetry-proto/pull/733/commits/52a8e7bacc2457c7e32e8ed5dbcb7216db18c2ae

opentelemetry/proto/resource/v1/resource.proto
florianl: resolving with https://github.com/open-telemetry/opentelemetry-proto/pull/733/commits/52a8e7bacc2457c7e32e8ed5dbcb7216db18c2ae

opentelemetry/proto/resource/v1/resource.proto
florianl, line 33: As an alternative and to prevent other signals to adapt this pattern, one way could also be something like this:

```suggestion
  oneof attributes {
    repeated opentelemetry.proto.common.v1.KeyValue common_attributes = 1;
    repeated RefKeyValue dict_attributes = 4;
  }
```

opentelemetry/proto/resource/v1/resource.proto
tigrannajaryan: >If we scope the dictionary to each Resource or inside ResourceLogs/ResourceSpans/ResourceMetrics, it can’t de-duplicate across separate Resource instances once we move to per-process resources. That defeats the goal and wouldn’t prevent the payload growth we’re trying to avoid.

Do you think you can run the benchmarks with both approaches (what you suggest and what Bogdan suggests) so that we can see the difference?

opentelemetry/proto/resource/v1/resource.proto
florianl: Unfortunatly this approach does not work, as `repeated` can not be used inside of `oneof`.

tigrannajaryan:
> I'm curious about this, how will it impact Profiling generally? I.e. I don't think we're going to allow non-dictionary profiles...

Any Profiling implementation going forward (after we accept this change) will have support for dictionaries. There is no need to allow non-dictionary profiles. Although, we could make it possible as a config option if we wanted to.

opentelemetry/proto/resource/v1/resource.proto
jsuereth: Let's add that as an invariant in the comments then

florianl:
> There is no need to allow non-dictionary profiles. 

@tigrannajaryan there is no option to have Profiles without dictionary support. In Profiles dictionaries are established for some time now.

The proposed change just reflects, that resources are making use of this dictionary and benefit from it. This was not the case so far. And as Profiles is still in development, this is not a breaking change.

florianl:
I have updated this PR to reflect the changes suggested in the [OTEP](https://docs.google.com/document/d/1v-XlgxnyyIsT-azCsfHJdRzVx-Ei2bIKB4hrYZQ6rGE/edit?tab=t.0).

felixge (approved):

bogdandrutu:
@felixge these days zstd is more of a standard than the gzip, can you try that as well?

jack-berg:
> However, we want to leave the door open for other (including currently stable) signals to support attribute dictionaries in the future. We should prefer to design attribute dictionaries for profiling in a way that allows reusing the same mechanism consistently for other signals in the future.

This is nice because it avoids us having to solve the thorny problems of evolving the stable signals to use dictionaries, but its means that OTLP receivers have to operate in a split brain world, with some signals encoding using dictionaries and others not. This has consequences for things like pdata, which now need to accommodate both worlds.

I wonder if the move to reference based attributes in protos, which are occurring here and with entities, are a big enough shift that we should consider moving them to an OTLP v2.

florianl:
As the tendency is towards `AnyRefValue` rather than just adding a reference field in the oneof field in AnyValue the PR got updated.

opentelemetry/proto/resource/v1/resource.proto
```diff
@@ -42,4 +43,15 @@ message Resource {
...
+  // Set of reference based attributes that describe the resource.
+  // Attribute keys MUST be unique (it is not allowed to have more than one
+  // attribute with the same key).
+  // The behavior of software that receives duplicated keys can be unpredictable.
+  // attributes MUST NOT be used if dict_attributes is used.
+  //
+  // Note: This is currently used exclusively in the Profiling signal.
+  //
+  // Status: [Development]
+  repeated opentelemetry.proto.common.v1.KeyRefValue dict_attributes = 4;
```
jsuereth: What would the performance look like if you used `KeyValueAndUnit` refs here (where unit is always empty) to see if this saved more data?

florianl: I don't have data to back this, but I assume no difference to the benchmark results we see with https://github.com/open-telemetry/opentelemetry-proto/pull/733#issuecomment-3586564002.
As most of the time, the additional `Unit` field is not set it will not cause an overhead. And for the time it is actually set, it will be neglectable.

The Profiling SIG introduced `KeyValueAndUnit` as we see the need to assign units to values. As an additional field, like `Unit`, is just an additional and not a breaking change, shouldn't the discussion around `Unit` separated from the reference based attributes?

opentelemetry/proto/common/v1/common.proto
```diff
@@ -39,13 +39,47 @@ message AnyValue {
...
+//
+// Note: This is currently used exclusively in the Profiling signal.
+//
+// Status: [Development]
+message AnyRefValue {
+  // The value is one of the listed fields. It is valid for all values to be unspecified
+  // in which case this AnyRefValue is considered to be "empty".
+  oneof value {
+    // Reference to the string value in ProfilesDictionary.string_table.
+    int32 string_value_ref = 8;
```
felixge, line 49: This should have index 1, right?

```
int32 string_value_ref = 1;
```

florianl: Fixed with https://github.com/open-telemetry/opentelemetry-proto/pull/733/commits/82ae8931c3ac19ac00889f81bc27ceaa8a136166

opentelemetry/proto/common/v1/common.proto
```diff
@@ -39,13 +39,47 @@ message AnyValue {
...
+//
+// Note: This is currently used exclusively in the Profiling signal.
+//
+// Status: [Development]
+message AnyRefValue {
+  // The value is one of the listed fields. It is valid for all values to be unspecified
+  // in which case this AnyRefValue is considered to be "empty".
+  oneof value {
+    // Reference to the string value in ProfilesDictionary.string_table.
+    int32 string_value_ref = 1;
```
christos68k: @bogdandrutu can you please clarify your views on:

1. references folded into `AnyValue` or introduced with a new message `AnyRefValue` as done here? IIRC you expressed a preference for the former in the last Zoom chat we had. @jsuereth and @tigrannajaryan expressed preference for the latter.
2. references existing in the proto vs pdata? If I'm not mistaken, you've also expressed a preference for the former.

bogdandrutu: @tigrannajaryan @jsuereth my understanding of the decision is that we will add `string_value_ref` in the current `AnyValue` and will comment that is not available for other signals.

florianl: > add string_value_ref in the current AnyValue

I will update this PR, once @tigrannajaryan @jsuereth also agree on that.

opentelemetry/proto/common/v1/common.proto
```diff
@@ -36,6 +36,12 @@ message AnyValue {
     ArrayValue array_value = 5;
     KeyValueList kvlist_value = 6;
     bytes bytes_value = 7;
+    // Reference to the string value in ProfilesDictionary.string_table.
+    //
+    // Note: This is currently used exclusively in the Profiling signal.
+    //
+    // Status: [Development]
+    int32 string_value_ref = 8;
```
bogdandrutu, line 49: Just for me to understand, in comment you suggest uint and here you are using the signed version. What is the expectation?

florianl: Sorry for the typo - should be `int32`. Updated the comment accordingly.

christos68k (approved):

opentelemetry/proto/common/v1/common.proto
jsuereth, line 49: As discussed in the SIG call - I'm fine with this as a first step.

I *do* think you need an update to `KeyValue` as well where the `key` is a `key_ref`, otherwise we aren't solving the bloat problem that motivated this.

Additionaly, just re-expressing my preference that we have a `ProfileResource` that mirrors (but is different than) `Resource` and uses  a `KeyValueRef`.  We then have the Collector pdata abstract between `KeyValueRef`/`KeyValue` and `ProfileResource`/`Resource` with the *same* datastructure to represent both, and serialize apporpriately on the other end.  

However, if that's not an option for the collector (due to performance or other concerns) we should move forward with this.   cc @bogdandrutu / @tigrannajaryan

opentelemetry/proto/common/v1/common.proto
```diff
@@ -36,6 +36,12 @@ message AnyValue {
     ArrayValue array_value = 5;
     KeyValueList kvlist_value = 6;
     bytes bytes_value = 7;
+    // Reference to the string value in ProfilesDictionary.string_table.
+    //
+    // Note: This is currently used exclusively in the Profiling signal.
```
jsuereth, line 41: Nit: Can you add a comment about how other signals should handle seeing this value?

E.g. if I'm implementing OTLP receiver for trace/metrics/logs - should I drop data that has this? Should I error? Should I ignore?

My suggestion - we suggest implementers consider it a non-fatal issue.  Log an error, and pretend the value is empty.

opentelemetry/proto/common/v1/common.proto
bogdandrutu: The problem to have "two things" that act the same in golang is that you need to use any (interfaces). Interfaces in golang most of the time require an extra allocation which is unnecessary.

opentelemetry/proto/common/v1/common.proto
bogdandrutu: Same as the value is really empty, in proto is valid (for the presence) to send an empty AnyValue.

opentelemetry/proto/common/v1/common.proto
bogdandrutu: So the option would be to actually behind the scene in the collector generate only one message Resource (and do exactly what I suggest here but hide in the back that there are 2 messages).

My problem with 2 messages is that profiles wants to add Unit for attributes for example and by being different they feel that is ok and added it without making a global decision.

opentelemetry/proto/common/v1/common.proto
florianl: @jsuereth I have updated the comment with https://github.com/open-telemetry/opentelemetry-proto/pull/733/commits/b67fa919a4b2d62cec72d7ca5cd92095635d71ad - please let me know, if this works for you.

opentelemetry/proto/common/v1/common.proto
florianl: @jsuereth I have added the changes for KeyValue with https://github.com/open-telemetry/opentelemetry-proto/pull/733/commits/2542b13d53e858f03c35dbe2a3bac99df8c4a60a - hope this works for you?

@bogdandrutu do you have concerns about this?

opentelemetry/proto/common/v1/common.proto
jsuereth: LGTM

opentelemetry/proto/common/v1/common.proto
jsuereth: > My problem with 2 messages is that profiles wants to add Unit for attributes for example and by being different they feel that is ok and added it without making a global decision.

I think there's some allowable inconsistency between signals.  The decision to have units - I think would be inline w/ a multivariate-metric signal, if we were to add one.  Additionally, the decision to add unit *was* part of the -proto repo, approved by proto maintainers and merged after accepting it as a global decision.  Can you clarify how it wasn't?  We decided to accept this inconsistency, given trade offs.

If we don't like the inconsitency, then let's propose something that *works* (or at least, also propose how to deal with all the downsides) of an alternative.  Right now, when the two options are "be consistent, but break the ecosystem" or "be inconsistent, but profiling is successful and existing integrations do not break" it's hard to understand the consistency argument.

opentelemetry/proto/common/v1/common.proto
```diff
@@ -65,10 +76,24 @@ message KeyValueList {
...
 message KeyValue {
   // The key name of the pair.
+  // key_ref MUST NOT be set if key is used.
   string key = 1;
 
   // The value of the pair.
   AnyValue value = 2;
+
+  // Reference to the string key in ProfilesDictionary.string_table.
+  // key MUST NOT be set if key_ref is used.
```
jsuereth, line 86: NIT: In proto3, wihtout the use of optional, there's not really a way to check "existence" or non-existence.  I think we could just say `key_ref` is not set or the proto3 default value of empty.

tigrannajaryan: I assume 0 value is invalid for a reference, which makes it pretty easy to check for "existence" without relying on Protobuf's native existence APIs. 0 is also not going to be present on the wire, so it is safe to set it to 0 on the sender's side to indicate that the key is not refcounted.

florianl: Right, by [convention](https://github.com/open-telemetry/opentelemetry-proto/blob/8ee341638ae14fda2c61fc8d35be3f7d1659debd/opentelemetry/proto/profiles/v1development/profiles.proto#L106-L110) all dictionary tables use index 0  to indicate 'null' / 'not set'.

Would prefer this sentence rephrased to something like this:
```
key MUST NOT be set if key_ref is set to a value other than 0.
```

jhalliday: A zero value ref is not invalid. string_table[0] must always be "" thus a zero ref is a valid ref to an empty string.
https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto#L147
For all non-string tables (i.e. pointers to message types only in profiles.proto) x_table[0] is an 'empty value', defined as 'whatever the default for each field in the message is'. String behaviour is a carryover from pprof. Anyhow, upshot is that short of declaring the ref 'optional', there exists no way to distinguish 'null' from 'used for empty string'. You'd have to use e.g. -1, assuming it's even needed.

tigrannajaryan: >there exists no way to distinguish 'null' from 'used for empty string'.

I think that's unnecessary. If the key is an empty string the `key` field can be set to "" and the `key_ref` field can be set to 0 and recipients have a clear signal the key is an empty string. The recipients logic is straightforward:

```go
if kv.key_ref == 0 {
  keyString = kv.key
} else {
  keyString = string_table[kv.key_ref]
}
```

This works without the need to check for the presence of `key_ref` and distinguishing 0 from non-present state.

So, I think we are good, nothing else is needed.

tigrannajaryan:
I would love to have the results of 2 experiments that we were planning on pdata (@aalexand's idea and @bogdandrutu's idea), before we merge this. It will give us confidence that we have a good solution for pdata.

PR #744, opened 2025-11-27
Clarify the profile and sample timestamps relationship.

aalexand:


opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -284,9 +284,13 @@ message Profile {
   // The following fields 3-12 are informational, do not affect
   // interpretation of results.
 
-  // Time of collection (UTC) represented as nanoseconds past the epoch.
+  // Time of collection represented as nanoseconds past the epoch.
```
tigrannajaryan: Nit: the wording we use [elsewhere](https://github.com/open-telemetry/opentelemetry-proto/blob/be43d11062a8686f940e66a9f0faebda58b02913/opentelemetry/proto/trace/v1/trace.proto#L189) is "Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970."

aalexand: Done. Though, as a nit, strictly speaking "Unix Epoch time" [implies](https://en.wikipedia.org/wiki/Unix_epoch_time) the "since 00:00:00 UTC on 1 January 1970" part.

PR #774, opened 2026-03-09
[chore] Makefile: Enable breaking change check for profiles

### LINKED ISSUE #577: Follow up: breaking changes in the Profiling protocol
With https://github.com/open-telemetry/opentelemetry-proto/pull/576 it is proposed to exclude the Profiling protocol from the breaking-changes test.

Currently the Profiling protocol is marked as experimental and breaking changes are accepted. But once, the Profiling protocol is declared stable, https://github.com/open-telemetry/opentelemetry-proto/pull/576 needs to be reverted.

florianl:
#576 exluded profiles from breaking changes check. Moving beyond the experimental phase, profiles must now avoid making breaking changes, which were previously acceptable.

Fixes https://github.com/open-telemetry/opentelemetry-proto/issues/577

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -162,10 +162,9 @@ gen-ruby:
 	$(PROTOC) --ruby_out=./$(PROTO_GEN_RUBY_DIR) --grpc-ruby_out=./$(PROTO_GEN_RUBY_DIR) opentelemetry/proto/collector/logs/v1/logs_service.proto
 	$(PROTOC) --ruby_out=./$(PROTO_GEN_RUBY_DIR) --grpc-ruby_out=./$(PROTO_GEN_RUBY_DIR) opentelemetry/proto/collector/profiles/v1development/profiles_service.proto
 
-# The Profiling protocol is still development. So it is excluded from the breaking-change check.
 .PHONY: breaking-change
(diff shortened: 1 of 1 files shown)


Makefile
florianl, line 1: I didn't add a changelog entry, as this is more a tooling/CI change and no core component (proto) is affected. Please let me know, if you want to have a changelog entry.

felixge (approved):
SGTM

christos68k (approved):

florianl:
> Is this too early? Do we want to wait until "Beta"?

Looking back, the profiling signal did have regular breaking changes in the past OTLP versions. I think, we can agree that with moving on to an Alpha state, we want a breaking change to be the exception rather than regular business. Having this check in place pushes us being more careful and coordinated going forward.

bogdandrutu (approved):

tigrannajaryan:
@florianl I can merge once you resolve all conversations.

florianl:
Converting back to draft at this point. There is at least one breaking change for profiles, with https://github.com/open-telemetry/opentelemetry-proto/pull/786, that should land first.

PR #780, opened 2026-03-25
Minor documentation updates to profiles.proto

christos68k:
For consistency with the data format documentation in the [spec](https://github.com/open-telemetry/opentelemetry-specification/pull/4965).

CC: @open-telemetry/profiling-maintainers @open-telemetry/profiling-approvers

### DIFF
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -67,17 +67,17 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
 //   │                                n-1
 //   │ 1-n         ┌───────────────────────────────────────┐
 //   ▼             │                                       ▽
-// ┌──────────────────┐   1-n   ┌─────────────────┐   ┌──────────┐
+// ┌──────────────────┐   n-n   ┌─────────────────┐   ┌──────────┐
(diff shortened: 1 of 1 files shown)


aalexand (approved):

christos68k:
Did a few more updates.

jsuereth (approved):

Kielek (approved):

tigrannajaryan:
@christos68k please resolve the conflicts.

pellared (approved):

jhalliday (commented):
Urg. Went on vacation and missed the review window. Just one nit, for which I'll issue a new PR to patch if I ever get caught up far enough...

// A Stack represents a stack trace as a list of locations (leaf first).
// For example, the stack trace resulting from the call stack
// main -> foo -> bar would be encoded into the location_indices list
// [2, 1, 0] which references the locations in location_table as:
// [Location{"main"}, Location{"foo"}, Location{"bar"}].

location_table[0] is explicitly a null/unused value. That's for consistency with the semantics of all the other dictionary tables, though since the only thing pointing into it is a 'repeated' field, it's not strictly necessary - there is no need to have a way to represent 'unset' as there is with non-repeated, non-optional fields.
However, it does mean the example here is not quite right. The encoding should really be

// [3, 2, 1] which references the locations in location_table as:
// [Location(}, Location{"main"}, Location{"foo"}, Location{"bar"}].

PR #781, opened 2026-03-30
docs: add response size limitation

pellared:
Add response body size limitation to mitigate memory usage risks

Reference: https://cwe.mitre.org/data/definitions/789.html

OTel implementations: 
- https://github.com/open-telemetry/opentelemetry-go/pull/8108
- https://github.com/open-telemetry/opentelemetry-dotnet/pull/7017
- https://github.com/open-telemetry/opentelemetry-java/pull/8224
- https://github.com/open-telemetry/opentelemetry-kotlin/pull/361
- https://github.com/open-telemetry/opentelemetry-js/pull/6552

### Follow-ups 

- [x] Add limit for request on the server side (and client side?) https://github.com/open-telemetry/opentelemetry-proto/pull/781#issuecomment-4156611205
  - https://github.com/open-telemetry/opentelemetry-proto/pull/782
- [x] Create issues for all SDKs https://github.com/open-telemetry/opentelemetry-proto/pull/781#issuecomment-4156632535
   - All SIGs have been already been notified (even a few times).
- [x] Add limit for response on server side https://github.com/open-telemetry/opentelemetry-proto/pull/782
- [x] Change wording (not-retryable -> non-retryable) https://github.com/open-telemetry/opentelemetry-proto/pull/781#discussion_r3016594120

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -164,10 +164,10 @@ the specific message to use in the [Full Success](#full-success),
 The client MUST enforce a message size limit when receiving the response to
 mitigate possible excessive memory usage caused by a misconfigured or malicious
 server. gRPC client implementations typically enforce a default incoming message
-size limit of 4 MiB, which is acceptable to use. If the limit is exceeded, the
-client MUST treat the response as a not-retryable error. Note that in such
-scenario, the gRPC client implementations are reporting a `RESOURCE_EXHAUSTED`
-code to the caller.
+size limit of 4 MiB, which is acceptable to use. Implementations MAY allow this
+limit to be configured. If the limit is exceeded, the client MUST treat the
+response as a not-retryable error. Note that in such scenario, the gRPC client
+implementations are reporting a `RESOURCE_EXHAUSTED` code to the caller.
 
 ##### Full Success
 
@@ -489,9 +489,10 @@ below for the specific message to use in the [Full Success](#full-success-1),
 
 The client MUST limit the size of the response body when parsing it, including
 after decompression, to mitigate possible excessive memory usage caused by a
-misconfigured or malicious server. It is RECOMMENDED to limit the response body
-to 4 MiB. If the limit is exceeded, the client MUST treat the response as a
-not-retryable error and SHOULD record the fact that the response was discarded.
+misconfigured or malicious server. It is RECOMMENDED to use 4 MiB
+as the default limit. Implementations MAY allow this limit to be configured. If
+the limit is exceeded, the client MUST treat the response as a not-retryable
+error and SHOULD record the fact that the response was discarded.
 
 The server MUST set "Content-Type: application/x-protobuf" header if the
 response body is binary-encoded Protobuf payload. The server MUST set
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -161,11 +161,13 @@ The response MUST be the appropriate message (see below for
 the specific message to use in the [Full Success](#full-success),
 [Partial Success](#partial-success) and [Failure](#failures) cases).
 
-The client MUST limit the size of the response body when parsing it, including
-after decompression, to mitigate possible excessive memory usage caused by a
-misconfigured or malicious server. It is RECOMMENDED to limit the response body
-to 32 KiB. If the limit is exceeded, the client MUST treat the response as a
-not-retryable error and SHOULD record the fact that the response was discarded.
+The client MUST enforce a message size limit when receiving the response to
+mitigate possible excessive memory usage caused by a misconfigured or malicious
+server. gRPC client implementations typically enforce a default incoming message
+size limit of 4 MiB, which is acceptable to use. Implementations MAY apply a
+tighter limit (e.g. 32 KiB) since OTLP response messages are expected to be
+small. If the limit is exceeded, the client MUST treat the response as a
+not-retryable error (typically by returning `RESOURCE_EXHAUSTED` code).
 
 ##### Full Success
 
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -164,7 +164,8 @@ the specific message to use in the [Full Success](#full-success),
 The client MUST limit the size of the response body when parsing it, including
 after decompression, to mitigate possible excessive memory usage caused by a
 misconfigured or malicious server. It is RECOMMENDED to limit the response body
-to 32 KiB.
+to 32 KiB. If the limit is exceeded, the client MUST treat the response as a
+non-retryable error and SHOULD record the fact that the response was discarded.
 
 ##### Full Success
 
@@ -487,7 +488,8 @@ below for the specific message to use in the [Full Success](#full-success-1),
 The client MUST limit the size of the response body when parsing it, including
 after decompression, to mitigate possible excessive memory usage caused by a
 misconfigured or malicious server. It is RECOMMENDED to limit the response body
-to 32 KiB.
+to 32 KiB. If the limit is exceeded, the client MUST treat the response as a
+non-retryable error and SHOULD record the fact that the response was discarded.
 
 The server MUST set "Content-Type: application/x-protobuf" header if the
 response body is binary-encoded Protobuf payload. The server MUST set
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -165,7 +165,7 @@ The client MUST limit the size of the response body when parsing it, including
 after decompression, to mitigate possible excessive memory usage caused by a
 misconfigured or malicious server. It is RECOMMENDED to limit the response body
 to 32 KiB. If the limit is exceeded, the client MUST treat the response as a
-non-retryable error and SHOULD record the fact that the response was discarded.
+not-retryable error and SHOULD record the fact that the response was discarded.
 
 ##### Full Success
 
@@ -305,7 +305,7 @@ This is signaled by the server by returning
 [RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40).
 In this case the behavior of the server and the client is exactly as described in
 [OTLP/gRPC Throttling](#otlpgrpc-throttling) section. If no such status is returned,
-then the `RESOURCE_EXHAUSTED` code SHOULD be treated as non-retryable.
+then the `RESOURCE_EXHAUSTED` code SHOULD be treated as not-retryable.
 
 #### OTLP/gRPC Throttling
 
@@ -489,7 +489,7 @@ The client MUST limit the size of the response body when parsing it, including
 after decompression, to mitigate possible excessive memory usage caused by a
 misconfigured or malicious server. It is RECOMMENDED to limit the response body
 to 32 KiB. If the limit is exceeded, the client MUST treat the response as a
-non-retryable error and SHOULD record the fact that the response was discarded.
+not-retryable error and SHOULD record the fact that the response was discarded.
 
 The server MUST set "Content-Type: application/x-protobuf" header if the
 response body is binary-encoded Protobuf payload. The server MUST set
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -164,9 +164,9 @@ the specific message to use in the [Full Success](#full-success),
 The client MUST enforce a message size limit when receiving the response to
 mitigate possible excessive memory usage caused by a misconfigured or malicious
 server. gRPC client implementations typically enforce a default incoming message
-size limit of 4 MiB, which is acceptable to use.
-If the limit is exceeded, the client MUST treat the response as a
-not-retryable error (typically by returning `RESOURCE_EXHAUSTED` code).
+size limit of 4 MiB, which is acceptable to use. If the limit is exceeded, the
+client MUST treat the response as a not-retryable error (typically by returning
+`RESOURCE_EXHAUSTED` code).
 
 ##### Full Success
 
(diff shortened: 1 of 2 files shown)


docs/specification.md
```diff
@@ -161,6 +161,12 @@ The response MUST be the appropriate message (see below for
 the specific message to use in the [Full Success](#full-success),
 [Partial Success](#partial-success) and [Failure](#failures) cases).
 
+The client MUST limit the size of the response body when parsing it, including
+after decompression, to mitigate possible excessive memory usage caused by a
+misconfigured or malicious server. It is RECOMMENDED to limit the response body
```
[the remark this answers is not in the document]
pellared: https://github.com/open-telemetry/opentelemetry-proto/pull/781/commits/dbdf2d7515289a6088f90c997ab84d8863b81405

Kielek (approved):

martincostello (approved):

dmathieu (approved):

docs/specification.md
```diff
@@ -161,6 +161,14 @@ The response MUST be the appropriate message (see below for
 the specific message to use in the [Full Success](#full-success),
 [Partial Success](#partial-success) and [Failure](#failures) cases).
 
+The client MUST enforce a message size limit when receiving the response to
+mitigate possible excessive memory usage caused by a misconfigured or malicious
+server. gRPC client implementations typically enforce a default incoming message
+size limit of 4 MiB, which is acceptable to use. Implementations MAY apply a
+tighter limit (e.g. 32 KiB) since OTLP response messages are expected to be
+small. If the limit is exceeded, the client MUST treat the response as a
```
arminru: > since OTLP response messages are expected to be small

Do we have any data to back that claim?

docs/specification.md
```diff
@@ -479,6 +487,12 @@ The response body MUST be the appropriate serialized Protobuf message (see
 below for the specific message to use in the [Full Success](#full-success-1),
 [Partial Success](#partial-success-1) and [Failure](#failures-1) cases).
 
+The client MUST limit the size of the response body when parsing it, including
+after decompression, to mitigate possible excessive memory usage caused by a
+misconfigured or malicious server. It is RECOMMENDED to limit the response body
+to 32 KiB. If the limit is exceeded, the client MUST treat the response as a
```
arminru: An exporter can send an OTLP/HTTP message with a default size of 4 MiB. If there's are issues with most/all span, data points, or logs in there, an OTLP receiver could return a fairly verbose partial success message. A maximally verbose message, perhaps when the exporter is in a debug mode to aid troubleshooting, could look like the below. For just two rejected data points it would already amount to ~1.5 KiB, but there could be hundreds more being rejected from the incoming export message. Same for spans and logs.

If we go with the same 4 MiB default sanity limit as we have for the export (and for gRPC response) message, we would already address the currently unbound excessive memory usage risks, right?


<details>

<summary>verbose response body example (~1.5 KiB for just 2 data points)</summary>

```
Rejected data point - rejection reason: invalid character '' (U+0000) in value attribute 'k8s.storageclass.name'
- metric: k8s.resourcequota.persistentvolumeclaim_count.used
- value: 12
- resource
	- k8s.resourcequota.uid: a1b2c3d4-0002-0002-0002-000000000002
	- k8s.resourcequota.name: storage-limits
	- k8s.namespace.name: data-platform
	- k8s.cluster.name: eu-west-staging
	- k8s.cluster.uid: f1e2d3c4-aaaa-bbbb-cccc-000000000002
	- host.id: i-0a1b2c3d4e5f60002
	- host.name: ip-172-16-2-44.eu-west-1.compute.internal
	- host.type: t3.large
	- host.arch: amd64
	- os.type: linux
	- os.name: Ubuntu
	- os.version: 22.04
	- os.description: Ubuntu 22.04.4 LTS
	- service.name: otelcol
	- service.version: 0.122.0
	- service.namespace: monitoring
	- service.instance.id: 4d63009a-8d0f-11ee-aad7-000000000002
	- telemetry.sdk.name: opentelemetry
	- telemetry.sdk.language: go
	- telemetry.sdk.version: 1.35.0
- attributes
	- k8s.storageclass.name: gold.storageclass.storage.k8s.io

Rejected data point - rejection reason: invalid character '' (U+0000) in value attribute 'k8s.storageclass.name'
- metric: k8s.resourcequota.persistentvolumeclaim_count.used
- value: 12
- resource
	- k8s.resourcequota.uid: a1b2c3d4-0002-0002-0002-000000000002
	- k8s.resourcequota.name: storage-limits
	- k8s.namespace.name: data-platform
	- k8s.cluster.name: eu-west-staging
	- k8s.cluster.uid: f1e2d3c4-aaaa-bbbb-cccc-000000000002
	- host.id: i-0a1b2c3d4e5f60003
	- host.name: ip-172-16-2-44.eu-west-1.compute.internal
	- host.type: t3.large
	- host.arch: amd64
	- os.type: linux
	- os.name: Ubuntu
	- os.version: 22.04
	- os.description: Ubuntu 22.04.4 LTS
	- service.name: otelcol
	- service.version: 0.122.0
	- service.namespace: monitoring
	- service.instance.id: 4d63009a-8d0f-11ee-aad7-000000000003
	- telemetry.sdk.name: opentelemetry
	- telemetry.sdk.language: go
	- telemetry.sdk.version: 1.35.0
- attributes
	- k8s.storageclass.name: gold.storageclass.storage.k8s.io
```

</details>

docs/specification.md
pellared: Do I understand correctly that you suggest to use 4MiB per: https://github.com/open-telemetry/opentelemetry-proto/pull/781#discussion_r3010008338 ?

pellared:
```suggestion
size limit of 4 MiB, which is acceptable to use.
If the limit is exceeded, the client MUST treat the response as a
```

docs/specification.md
pellared:
```suggestion
misconfigured or malicious server. It is RECOMMENDED to limit the response body
to 4 MiB. If the limit is exceeded, the client MUST treat the response as a
```

docs/specification.md
pellared: Addressed https://github.com/open-telemetry/opentelemetry-proto/pull/781/commits/a232252a366862b66849e559f41dbda143783a1f

docs/specification.md
pellared: Addressed https://github.com/open-telemetry/opentelemetry-proto/pull/781/commits/a232252a366862b66849e559f41dbda143783a1f

docs/specification.md
```diff
@@ -161,6 +161,13 @@ The response MUST be the appropriate message (see below for
 the specific message to use in the [Full Success](#full-success),
 [Partial Success](#partial-success) and [Failure](#failures) cases).
 
+The client MUST enforce a message size limit when receiving the response to
+mitigate possible excessive memory usage caused by a misconfigured or malicious
+server. gRPC client implementations typically enforce a default incoming message
+size limit of 4 MiB, which is acceptable to use.
+If the limit is exceeded, the client MUST treat the response as a
```
yurishkuro: Confusing wording - client is typically an exporter in the SDK, it doesn't "return" anything. I think this was meant to say "act as if the server returned this response" but it doesn't read like that.

pellared: Which part is confusing? Is it the following one:

> If the limit is exceeded, the client MUST treat the response as a not-retryable error (typically by returning `RESOURCE_EXHAUSTED` code).

If so then what do you think of

> If the limit is exceeded, the client MUST treat the response as a not-retryable error. Note that in such scenario, the gRPC clients are reporting a `RESOURCE_EXHAUSTED` code to the caller.

docs/specification.md
```diff
@@ -161,6 +161,13 @@ The response MUST be the appropriate message (see below for
 the specific message to use in the [Full Success](#full-success),
 [Partial Success](#partial-success) and [Failure](#failures) cases).
 
+The client MUST enforce a message size limit when receiving the response to
+mitigate possible excessive memory usage caused by a misconfigured or malicious
+server. gRPC client implementations typically enforce a default incoming message
+size limit of 4 MiB, which is acceptable to use. If the limit is exceeded, the
+client MUST treat the response as a not-retryable error (typically by returning
+`RESOURCE_EXHAUSTED` code).
```
yurishkuro: > typically by returning `RESOURCE_EXHAUSTED` code

this is the confusing part - a client/exporter does not "return" codes.

pellared: Addressed https://github.com/open-telemetry/opentelemetry-proto/pull/781/commits/ec2c63e370eb46dbc086c5891e342af6960cf8de

docs/specification.md
pellared: Addressed https://github.com/open-telemetry/opentelemetry-proto/pull/781/commits/ec2c63e370eb46dbc086c5891e342af6960cf8de

yurishkuro (approved):

docs/specification.md
```diff
@@ -479,6 +487,12 @@ The response body MUST be the appropriate serialized Protobuf message (see
 below for the specific message to use in the [Full Success](#full-success-1),
 [Partial Success](#partial-success-1) and [Failure](#failures-1) cases).
 
+The client MUST limit the size of the response body when parsing it, including
+after decompression, to mitigate possible excessive memory usage caused by a
+misconfigured or malicious server. It is RECOMMENDED to limit the response body
+to 4 MiB. If the limit is exceeded, the client MUST treat the response as a
```
jack-berg: 4 mib is bigger than I expected, and still seems pretty exposed to a malicious server.

I'm fine with it as a default, but seems like it should be configurable. I suggest adding an option to the spec [exporter.md](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md) doc, but to keep it simple / uncontroversial, omitting adding new env vars. I.e. this would only be configurable via programmatic interface, and eventually declarative config.

jack-berg (approved):

docs/specification.md
pellared: I thought the same :wink: Discussion: https://github.com/open-telemetry/opentelemetry-proto/pull/781#discussion_r3010008338

We can add the option once users would need in a separate PR.

XSAM (approved):

arminru (approved):

docs/specification.md
pellared: I decided to include a little statement (that this MAY be configurable) in this PR here: https://github.com/open-telemetry/opentelemetry-proto/pull/781/commits/d52f73aa235856e5b9fcb9fef4e29dfec21d718e as I also wanted to include it this in https://github.com/open-telemetry/opentelemetry-proto/pull/782

docs/specification.md
```diff
@@ -299,7 +307,7 @@ This is signaled by the server by returning
 [RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40).
 In this case the behavior of the server and the client is exactly as described in
 [OTLP/gRPC Throttling](#otlpgrpc-throttling) section. If no such status is returned,
-then the `RESOURCE_EXHAUSTED` code SHOULD be treated as non-retryable.
+then the `RESOURCE_EXHAUSTED` code SHOULD be treated as not-retryable.
```
reyang, line 310: First time I've seen "not-retryable", @pellared are you sure this is the right use? I don't know, I'm not a good English speaker.

martincostello: Unless this is a specific technical term for protobuf, "non-retryable" would be more usual in English.

reyang (approved):

docs/specification.md
pellared: I started with "non-retryable", but there were a lot (all but one!) existing occurrences of "not-retryable" :laughing: 

I will make a separate PR to change `not-retryable` -> `non-retryable`.

Added as follow-up to the PR description.

docs/specification.md
```diff
@@ -479,6 +487,13 @@ The response body MUST be the appropriate serialized Protobuf message (see
 below for the specific message to use in the [Full Success](#full-success-1),
 [Partial Success](#partial-success-1) and [Failure](#failures-1) cases).
 
+The client MUST limit the size of the response body when parsing it, including
+after decompression, to mitigate possible excessive memory usage caused by a
+misconfigured or malicious server. It is RECOMMENDED to use 4 MiB
```
tigrannajaryan, line 492: Is 4 MiB compressed or uncompressed size limit?

pellared: uncompressed

> including after decompression

tigrannajaryan: Can we reword it to make it clearer? Does it means that the limit is applied twice, to the compressed size, and to the uncompressed size?

pellared: Decompression (e.g., gzip) should be performed in a streaming manner, so that size limits are enforced during both reading and decompression. In such an implementation, the limit is effectively applied once, as the data is processed incrementally.

If, however, an implementation first reads the entire compressed body and then decompresses it separately, the limit would need to be enforced twice: once on the compressed size and once on the decompressed size. _I would not recommend such approach as it is not efficient._

Do you have a proposal how this can be reworded to make it clearer?

pellared: > to mitigate possible excessive memory usage caused by a misconfigured or malicious server

This should also hopefully make the readers (developers) think when implementing the fixes :wink:

pellared: Maybe we could add something like if you and others would find it helpful

```suggestion
misconfigured or malicious server.
Implementations SHOULD process compressed messages incrementally
and enforce the size limit on the decompressed data as it is read.
It is RECOMMENDED to use 4 MiB
```

For me it feels unnecessary and I prefer keeping the wording as it is. 

Maybe you can try addressing your comment (improving the clarity) in a separate PR?
Is this comment a blocker?

dashpole (approved):

tigrannajaryan (approved):

florianl (approved):

PR #782, opened 2026-03-31
docs: add request size limitation

pellared:
Per https://github.com/open-telemetry/opentelemetry-proto/pull/781#issuecomment-4156611205

Add request size limitation for HTTP body and gRPC messages to mitigate memory usage risks.

Reference: https://cwe.mitre.org/data/definitions/789.html

The values are taken from ~~[otlpreceiver](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/config.md) defaults~~ https://github.com/open-telemetry/opentelemetry-proto/pull/782#discussion_r3032163377
- gRPC - ~~4 MiB~~ 64 MiB
- HTTP - ~~20 MiB~~ 64 MiB

OTel implementations:
- https://github.com/open-telemetry/opentelemetry-go/pull/8157

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -120,12 +120,14 @@ The server MUST enforce a message size limit when receiving the request,
 including after decompression, to mitigate possible excessive memory usage
 caused by a misconfigured or malicious client sending an oversized request.
 The server implementations typically enforce a default incoming message size
-limit of 4 MiB, which is acceptable to use. If the limit is exceeded, the client
-MUST treat the error as non-retryable. Note that in such a scenario, the gRPC
-server implementations return a `RESOURCE_EXHAUSTED` status code to the caller.
+limit of 4 MiB. However, it is RECOMMENDED to use 32 MiB as the default limit.
+Implementations MAY allow this limit to be configured. If the limit is
+exceeded, the client MUST treat the error as non-retryable. Note that in such a
+scenario, the gRPC server implementations return a `RESOURCE_EXHAUSTED` status
+code to the caller.
 
 The client SHOULD limit the size of the request message, including before
-compression, to avoid overwhelming the server. It is RECOMMENDED to use 4 MiB
+compression, to avoid overwhelming the server. It is RECOMMENDED to use 32 MiB
 as the default limit. Implementations MAY allow this limit to be configured.
 
 #### OTLP/gRPC Concurrent Requests
@@ -498,14 +500,14 @@ sides.
 
 The server MUST limit the size of the request body when parsing it, including
 after decompression, to mitigate possible excessive memory usage caused by a
-misconfigured or malicious client client sending an oversized request. It is
-RECOMMENDED to use 20 MiB as the default limit. Implementations MAY allow this
+misconfigured or malicious client sending an oversized request. It is
+RECOMMENDED to use 32 MiB as the default limit. Implementations MAY allow this
 limit to be configured. If the limit is exceeded, the server SHOULD respond with
 `HTTP 413 Content Too Large`. The client MUST NOT retry the request when it
 receives `HTTP 413 Content Too Large` response.
 
 The client SHOULD limit the size of the request body, including before
-compression, to avoid overwhelming the server. It is RECOMMENDED to use 20 MiB
+compression, to avoid overwhelming the server. It is RECOMMENDED to use 32 MiB
 as the default limit. Implementations MAY allow this limit to be configured.
 
 #### OTLP/HTTP Response
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -116,13 +116,13 @@ paths._
 
 #### OTLP/gRPC Request
 
-The server MUST limit the size of the request message when parsing it, including
-after decompression, to mitigate possible excessive memory usage caused by a
-misconfigured or malicious client. It is RECOMMENDED to limit the request
-message to 4 MiB. If the limit is exceeded, the server SHOULD return
-`RESOURCE_EXHAUSTED` gRPC status code. In this case the `RESOURCE_EXHAUSTED`
-code MUST NOT be accompanied by a `RetryInfo` status detail so that the client
-treats the error as not-retryable (see [Failures](#failures)).
+The server MUST enforce a message size limit when receiving the request to
+mitigate possible excessive memory usage caused by a misconfigured or malicious
+server. The server implementations typically enforce a default incoming message
+size limit of 4 MiB, which is acceptable to use. If the limit is exceeded, the
+client MUST treat the error as not-retryable. Note that in such a scenario,
+the gRPC server implementations return a `RESOURCE_EXHAUSTED` status code
+to the caller.
 
 The client SHOULD limit the size of the request message, including before
 compression, to avoid overwhelming the server. It is RECOMMENDED to limit the
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -125,8 +125,8 @@ the gRPC server implementations return a `RESOURCE_EXHAUSTED` status code
 to the caller.
 
 The client SHOULD limit the size of the request message, including before
-compression, to avoid overwhelming the server. It is RECOMMENDED to limit the
-request message to 4 MiB.
+compression, to avoid overwhelming the server. It is RECOMMENDED to use 4 MiB
+as the default limit. Implementations MAY allow this limit to be configured.
 
 #### OTLP/gRPC Concurrent Requests
 
@@ -490,14 +490,15 @@ sides.
 
 The server MUST limit the size of the request body when parsing it, including
 after decompression, to mitigate possible excessive memory usage caused by a
-misconfigured or malicious client. It is RECOMMENDED to limit the request body
-to 20 MiB. If the limit is exceeded, the server SHOULD respond with
-`HTTP 413 Content Too Large`. The client MUST NOT retry the request when it
-receives `HTTP 413 Content Too Large` response.
+misconfigured or malicious client. It is RECOMMENDED to use 20 MiB as the
+default limit. Implementations MAY allow this limit to be configured. If the
+limit is exceeded, the server SHOULD respond with `HTTP 413 Content Too Large`.
+The client MUST NOT retry the request when it receives
+`HTTP 413 Content Too Large` response.
 
 The client SHOULD limit the size of the request body, including before
-compression, to avoid overwhelming the server. It is RECOMMENDED to limit the
-request body to 20 MiB.
+compression, to avoid overwhelming the server. It is RECOMMENDED to use 20 MiB
+as the default limit. Implementations MAY allow this limit to be configured.
 
 #### OTLP/HTTP Response
 
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -120,7 +120,7 @@ The server MUST enforce a message size limit when receiving the request to
 mitigate possible excessive memory usage caused by a misconfigured or malicious
 client sending an oversized request. The server implementations typically
 enforce a default incoming message size limit of 4 MiB, which is acceptable to
-use. If the limit is exceeded, the client MUST treat the error as not-retryable.
+use. If the limit is exceeded, the client MUST treat the error as non-retryable.
 Note that in such a scenario, the gRPC server implementations return a
 `RESOURCE_EXHAUSTED` status code to the caller.
 
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -116,13 +116,13 @@ paths._
 
 #### OTLP/gRPC Request
 
-The server MUST enforce a message size limit when receiving the request to
-mitigate possible excessive memory usage caused by a misconfigured or malicious
-client sending an oversized request. The server implementations typically
-enforce a default incoming message size limit of 4 MiB, which is acceptable to
-use. If the limit is exceeded, the client MUST treat the error as non-retryable.
-Note that in such a scenario, the gRPC server implementations return a
-`RESOURCE_EXHAUSTED` status code to the caller.
+The server MUST enforce a message size limit when receiving the request,
+including after decompression, to mitigate possible excessive memory usage
+caused by a misconfigured or malicious client sending an oversized request.
+The server implementations typically enforce a default incoming message size
+limit of 4 MiB, which is acceptable to use. If the limit is exceeded, the client
+MUST treat the error as non-retryable. Note that in such a scenario, the gRPC
+server implementations return a `RESOURCE_EXHAUSTED` status code to the caller.
 
 The client SHOULD limit the size of the request message, including before
 compression, to avoid overwhelming the server. It is RECOMMENDED to use 4 MiB
@@ -490,11 +490,11 @@ sides.
 
 The server MUST limit the size of the request body when parsing it, including
 after decompression, to mitigate possible excessive memory usage caused by a
-misconfigured or malicious client. It is RECOMMENDED to use 20 MiB as the
-default limit. Implementations MAY allow this limit to be configured. If the
-limit is exceeded, the server SHOULD respond with `HTTP 413 Content Too Large`.
-The client MUST NOT retry the request when it receives
-`HTTP 413 Content Too Large` response.
+misconfigured or malicious client client sending an oversized request. It is
+RECOMMENDED to use 20 MiB as the default limit. Implementations MAY allow this
+limit to be configured. If the limit is exceeded, the server SHOULD respond with
+`HTTP 413 Content Too Large`. The client MUST NOT retry the request when it
+receives `HTTP 413 Content Too Large` response.
 
 The client SHOULD limit the size of the request body, including before
 compression, to avoid overwhelming the server. It is RECOMMENDED to use 20 MiB
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -32,6 +32,7 @@ nodes such as collectors and telemetry backends.
 
 - [Protocol Details](#protocol-details)
   * [OTLP/gRPC](#otlpgrpc)
+    + [OTLP/gRPC Request](#otlpgrpc-request)
     + [OTLP/gRPC Concurrent Requests](#otlpgrpc-concurrent-requests)
     + [OTLP/gRPC Response](#otlpgrpc-response)
       - [Full Success](#full-success)
@@ -112,6 +113,20 @@ delivery guarantees in such systems is outside of the scope of OTLP. The
 acknowledgements described in this protocol happen between a single
 client/server pair and do not span intermediary nodes in multi-hop delivery
 paths._
+
+#### OTLP/gRPC Request
+
+The server MUST limit the size of the request message when parsing it, including
+after decompression, to mitigate possible excessive memory usage caused by a
+misconfigured or malicious client. It is RECOMMENDED to limit the request
+message to 64 MiB. If the limit is exceeded, the server SHOULD return
+`RESOURCE_EXHAUSTED` gRPC status code. In this case the `RESOURCE_EXHAUSTED`
+code MUST NOT be accompanied by a `RetryInfo` status detail so that the client
+treats the error as not-retryable (see [Failures](#failures)).
+
+The client SHOULD limit the size of the request message, including before
+compression, to avoid overwhelming the server. It is RECOMMENDED to limit the
+request message to 64 MiB.
 
 #### OTLP/gRPC Concurrent Requests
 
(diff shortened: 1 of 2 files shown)


docs/specification.md
```diff
@@ -113,6 +114,20 @@ acknowledgements described in this protocol happen between a single
...
+mitigate possible excessive memory usage caused by a misconfigured or malicious
+server. The server implementations typically enforce a default incoming message
+size limit of 4 MiB, which is acceptable to use. If the limit is exceeded, the
+client MUST treat the error as not-retryable. Note that in such a scenario,
+the gRPC server implementations return a `RESOURCE_EXHAUSTED` status code
+to the caller.
+
+The client SHOULD limit the size of the request message, including before
+compression, to avoid overwhelming the server. It is RECOMMENDED to limit the
+request message to 4 MiB.
```
pellared: I am worried that the value is too small for profiles. However, I wanted to use the Collector defaults for initializing the conversation. Should we change it to 20 MiB? This would also follow the pattern from https://github.com/open-telemetry/opentelemetry-proto/pull/781 where we agreed on the same size for gRPC and HTTP.

CC @open-telemetry/collector-maintainers @open-telemetry/profiling-maintainers

docs/specification.md
```diff
@@ -113,6 +114,20 @@ acknowledgements described in this protocol happen between a single
...
+mitigate possible excessive memory usage caused by a misconfigured or malicious
+client sending an oversized request. The server implementations typically
+enforce a default incoming message size limit of 4 MiB, which is acceptable to
+use. If the limit is exceeded, the client MUST treat the error as non-retryable.
+Note that in such a scenario, the gRPC server implementations return a
+`RESOURCE_EXHAUSTED` status code to the caller.
+
+The client SHOULD limit the size of the request message, including before
+compression, to avoid overwhelming the server. It is RECOMMENDED to use 4 MiB
+as the default limit. Implementations MAY allow this limit to be configured.
```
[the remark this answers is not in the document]
pellared: https://github.com/open-telemetry/opentelemetry-proto/pull/782/commits/d450703d4b1ff59883f41bbedd44d927558e6805

docs/specification.md
felixge: Thanks for raising this. We discussed this in our [2026-04-02 profiling SIG meeting](https://docs.google.com/document/d/19UqPPPlGE83N37MhS93uRlxsP1_wGxQ33Qv6CDHaEp0/edit?tab=t.0#heading=h.tdl4orcd86r).

4 MiB is likely to be too low of a limit for us, and we'd be very interested in bumping the limits up to perhaps 32 MiB or more for both gRPC and HTTP.

I'll provide a more in-depth analysis on the data volumes we are seeing in our deployments here ASAP to help making an informed decision.

pellared: Going to change to 32MiB and then undraft this PR.

tigrannajaryan:
@open-telemetry/profiling-maintainers do you think 32MiB request size (uncompressed) is large enough that we practically never hit this limit? If it is not large enough is there such number or we think that legitimate profiles can be arbitrarily large? If there is no practical limit for profiles do we feel comfortable with the approach that there is a known default limitation for payloads (e.g. 32MiB) and all profile emitters are expected to split collected profiles to keep individual payloads under this default limit?

docs/specification.md
pellared: https://github.com/open-telemetry/opentelemetry-proto/pull/782/commits/94aba0b037b09a4d77bf674d09c7505b18ce4208

martincostello (approved):

pellared:
Changed backed to draft as I decided to first create a prototype PR at for the client-side (OTLP exporters).

florianl (approved):
> do you think 32MiB request size (uncompressed) is large enough that we practically never hit this limit? If it is not large enough is there such number or we think that legitimate profiles can be arbitrarily large? If there is no practical limit for profiles do we feel comfortable with the approach that there is a known default limitation for payloads (e.g. 32MiB) and all profile emitters are expected to split collected profiles to keep individual payloads under this default limit?

Since this limit is configurable, environments using larger profiles can still increase it. However, for most environments, 32 MiB appears to be a reasonable default.

pellared:
> However, for most environments, 32 MiB appears to be a reasonable default.

Have you done any testing? Do we have some numbers? Could you e.g.. setup the https://github.com/open-telemetry/opentelemetry-ebpf-profiler for all services (including e.g. Prometheus, Jaeger) on https://github.com/open-telemetry/opentelemetry-demo/?

Also waiting on @felixge with (from https://github.com/open-telemetry/opentelemetry-proto/pull/782#discussion_r3032163377): 

> I'll provide a more in-depth analysis on the data volumes we are seeing in our deployments here ASAP to help making an informed decision.

tigrannajaryan:
> Since this limit is configurable, environments using larger profiles can still increase it.

I think this only works as a strategy if users have control over receiver's configuration, which may not always be the case.

My preference would be one the following approaches, in the order of priority:
1. Establish a maximum request size that any reasonable production profile cannot possibly exceed. Anything larger than that is to be classified as malicious. If such a maximum number exists and it is not too big to be defined as default max size then we choose this approach. The limit can still be configurable, but it should practically be never required to change this configuration. We also likely don't need to implement sender payload splitting in this case.
2. If approach 1 is not possible (i.e. no realistic max size is possible to establish), then we choose a reasonable default max size for receivers (e.g. 32MiB). _Also_ require senders by default to split payloads if they exceed a configurable max size. Default max for senders is the same as default max for receivers (or slightly below if splitting cannot be done precisely).

It is important that interoperability works by default, without requiring configuration. Configuration can be used to fine tune performance (e.g. if larger payloads result in better overall compression and memory usage is acceptable).

felixge (commented):
Sorry for the delay, but I'm still planning to share some hard data here. Hope to be able to post it later this week.

jack-berg:
@open-telemetry/collector-maintainers given that collector is primary OTLP server in the ecosystem, can you review?

tigrannajaryan (changes requested):
Blocking temporarily until we get more data for profiles.

docs/specification.md
```diff
@@ -113,6 +114,24 @@ acknowledgements described in this protocol happen between a single
...
+Implementations MAY allow this limit to be configured. If the limit is
+exceeded, the gRPC server implementations MUST report a
+`RESOURCE_EXHAUSTED` code to the caller which the client MUST treat as a
+non-retryable error.
+
+The client SHOULD limit the size of the request message, including before
+compression, to avoid overwhelming the server. It is RECOMMENDED to use 32 MiB
+as the default limit. Implementations MAY allow this limit to be configured. If
+the limit is exceeded, the client MUST NOT make the request and SHOULD record
+the fact that the request was discarded.
```
pellared, line 132: There was a discussion during the SIG meeting regarding splitting the request instead of only dropping. I think nothing is against adding later adding a sentence like:

> In such scenario, the discarded request MAY be spitted into smaller requests that  are below this limit.

felixge:
In an earlier comment on this PR, I had suggested a limit of 32 MiB based on some quick napkin math. I finally found some time last week to do a more in-depth analysis:

* Notebook: [otlp-size-analysis.html](https://felixge.s3.us-east-1.amazonaws.com/2026-05-10/otlp-size-analysis.html)
* PR: https://github.com/open-telemetry/sig-profiling/pull/123

This analysis confirms the 32 MiB limit, but critically depends on a few assumptions. In particular, I'm proposing that we should try to support receiving CPU profiles from the ebpf profiler (ReportInterval: 60s) from fully loaded hosts with up to 64 vCPUs while dropping less than 1% of those profiles. When extrapolating from 10k real profiles, these requirements produce 32 MiB as a reasonable limit.

visualization

Please check out the notebook and LMK what you think 🙇

felixge (approved):

tigrannajaryan:
@felixge thanks for the analysis.

This explicitly targets that approximately 1% (or less) of profiles may exceed the limit and get dropped. We believe this is an acceptable behavior since profiles are inherently statistical data and losing a small portion of it by design is not catastrophic. Is that fair to say?

The reason I want this to be explicitly stated is because for other signals the same assertion is not true (e.g. it is not acceptable to drop 1% of audit logs by design).

felixge:
> This explicitly targets that approximately 1% (or less) of profiles may exceed the limit and get dropped. We believe this is an acceptable behavior since profiles are inherently statistical data and losing a small portion of it by design is not catastrophic. Is that fair to say?
> 
> The reason I want this to be explicitly stated is because for other signals the same assertion is not true (e.g. it is not acceptable to drop 1% of audit logs by design).

It would certainly be better to not drop any profiles. The 1% of profiles that would be dropped under the 32 MiB limit would be extremely interesting. 64 MiB would likely allow us to avoid dropping any profiles.

But all of this assumes 64 vCPU. In theory people can rent machines with up to [1920 vCPU](https://aws.amazon.com/ec2/instance-types/u7i/) these days. Not dropping any profiles from these machines would perhaps require a ~2 GiB size limit.

That being said, there is another way out: Profilers could split their payloads when they start to exceed the limit. There is a pending comment about that from @pellared.

The only downside with that approach is that it decreases efficiency (a lower limit implies a higher data rate, as dictionary data will be to be duplicated across requests more often):

2026-05-12 otlp-size-analysis at 17 27 11@2x

In other words, 32 MiB should okay. 64 MiB would be better. FWIW, we use a limit of ~100 MiB for our proprietary profiling intake at Datadog. The datadog-agent has no limit as it's typically streaming the profiling data without decoding it.

tigrannajaryan:
>It would certainly be better to not drop any profiles. The 1% of profiles that would be dropped under the 32 MiB limit would be extremely interesting.

> But all of this assumes 64 vCPU. In theory people can rent machines with up to [1920 vCPU](https://aws.amazon.com/ec2/instance-types/u7i/) these days. Not dropping any profiles from these machines would perhaps require a ~2 GiB size limit.

The way I am reading this is there is no specific number that we can choose that:
1. Ensures the data is not dropped in real production use cases and
2. Dropping the data is not as benign as I was suggesting.

> That being said, there is another way out: Profilers could split their payloads when they start to exceed the limit. There is a pending comment about that from @pellared.

I think we should look into this before making a final call on profiles. Can we consider the following:
- Confirm that the implementations can indeed perform splitting to fit the limit and do this splitting efficiently. We don't have to implement it now, but we need to know it is doable.
- If splitting can't be done efficiently and we can't rely on that then our other option is configurability of the limit.

I would like to understand what is the way forward for profiles: splitting or configurability.

pellared:
> I would like to understand what is the way forward for profiles: splitting or configurability.

@tigrannajaryan, I don’t think these are mutually exclusive options. In fact, this PR already introduces configurability. For example, OTel Go already added support for configuring this limit: https://github.com/open-telemetry/opentelemetry-go/pull/8157. I also think we should expose it via https://github.com/open-telemetry/opentelemetry-configuration/ (CC @open-telemetry/configuration-maintainers). 

Because of that, I would support strengthening the wording from MAY to SHOULD:

```
Implementations SHOULD allow this limit to be configured.
```

Regarding splitting: it definitely comes with additional complexity, while the “efficiency” argument is still fairly subjective without concrete goals or benchmarks. For sure it would cause additional overhead (some is mentioned in https://github.com/open-telemetry/opentelemetry-proto/pull/782#issuecomment-4432200649 by @felixge). It is also worth noting that splitting alone would not prevent DDoS scenarios, and it could potentially overlap or conflict with existing batch processor functionality. Because of this probably even the splitting should be an opt-in functionality?

Given those tradeoffs, I currently lean toward configurability as the primary direction. We could also consider increasing the default limit to 64 MiB.

felixge:
@pellared SGTM. 64 MiB as a default would be great.

Splitting will always be a compromise. On the client side it will require additional CPU and Memory (from splitting the dictionary itself) while on the collector side it will cause overall work (and costs) to increase in exchange for lowering memory peaks.

I think most users will be better off with configuring higher limits.

That being said, if I understand correctly, the spec currently doesn't suggest a default concurrency limit. It seems incomplete to recommend a maximum payload size without also defining a concurrency limit?

pellared:
> That being said, if I understand correctly, the spec currently doesn't suggest a default concurrency limit. It seems incomplete to recommend a maximum payload size without also defining a concurrency limit?

@felixge, do you mean https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlphttp-concurrent-requests and https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlpgrpc-concurrent-requests? If so then we should have a separate issue/PR for it.

PS. I am also not sure if this is actually implemented anywhere expect the OpenTelemetry Collector and configured via https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md#sending-queue

felixge:
> > That being said, if I understand correctly, the spec currently doesn't suggest a default concurrency limit. It seems incomplete to recommend a maximum payload size without also defining a concurrency limit?
> 
> @felixge, do you mean [`main`/docs/specification.md#otlphttp-concurrent-requests](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md?rgh-link-date=2026-05-19T11%3A00%3A16Z#otlphttp-concurrent-requests) and [`main`/docs/specification.md#otlpgrpc-concurrent-requests](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md?rgh-link-date=2026-05-19T11%3A00%3A16Z#otlpgrpc-concurrent-requests)? If so then we should have a separate issue/PR for it.

Yeah, that's the section I looked at. It seems to apply only to the concurrency of clients, and not receivers. I can open an issue if you also think it's a good idea.

> PS. I am also not sure if this is actually implemented anywhere expect the OpenTelemetry Collector and configured via [open-telemetry/opentelemetry-collector@`main`/exporter/exporterhelper/README.md#sending-queue](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md?rgh-link-date=2026-05-19T11%3A00%3A16Z#sending-queue)

@pellared yeah, looks like the OTLP receiver doesn't have a concurrency limit. I know there is a memory_limiter processor, but since that's downstream from the receiver, it doesn't seem like it would do a good job protecting against overload from big payloads being received concurrently.

docs/specification.md
```diff
@@ -113,6 +114,24 @@ acknowledgements described in this protocol happen between a single
...
+caused by a misconfigured or malicious client sending an oversized request.
+The server implementations typically enforce a default incoming message size
+limit of 4 MiB. However, it is RECOMMENDED to use 32 MiB as the default limit.
+Implementations MAY allow this limit to be configured. If the limit is
+exceeded, the gRPC server implementations MUST report a
+`RESOURCE_EXHAUSTED` code to the caller which the client MUST treat as a
+non-retryable error.
+
+The client SHOULD limit the size of the request message, including before
+compression, to avoid overwhelming the server. It is RECOMMENDED to use 32 MiB
```
aabmass: At least in python, I think it would be disruptive to start enforcing this without implementing splitting. Some of the linked issues have context (e.g. https://github.com/open-telemetry/opentelemetry-python/issues/4533), but it's not uncommon to have 1MB span/log attributes with GenAI Telemetry.

The current controls in BSP/BLRP are not size aware, and generally with GenAI observability you're likely to have really widely distributed span sizes (normal RPC spans and LLM prompt/response spans), so configuring there might not be very helpful.

pellared: The consensus during the SIG meeting was to add splitting separately after this one. Note that it seems only to be "disruptive" only for GenAI telemetry. Also I do not think that this would make the situation worse.

Moreover, note that in some scenarios users may prefer to have a hard limit and drop and not use splitting for the reasons described in https://github.com/open-telemetry/opentelemetry-proto/pull/782#issuecomment-4486631665

pellared:
@tigrannajaryan, I pushed https://github.com/open-telemetry/opentelemetry-proto/pull/782/commits/9d381d48600b6187c9e7888be204adf9ecfc5995 and https://github.com/open-telemetry/opentelemetry-proto/pull/782/commits/794438c8a73047fb40e1e46bba38040957ac176a. I think that this PR can be now unblocked. I will work on adding the configuration to https://github.com/open-telemetry/opentelemetry-configuration/ in the following days.

docs/specification.md
aabmass: Not disagreeing on this being a good practice or making it configurable. But do you not consider this a breaking change to OTLP exporters?

tigrannajaryan:
I do not consider this to be a breaking change. 

There is a always a limit, the limit was just not previously defined and the spec didn't define what happens when that limit is hit.

docs/specification.md
```diff
@@ -481,6 +500,20 @@ The client MAY gzip the content and in that case MUST include
 Non-default URL paths for requests MAY be configured on the client and server
 sides.
 
+The server MUST limit the size of the request body when parsing it, including
```
tigrannajaryan, line 502:
```suggestion
The server SHOULD limit the size of the request body when parsing it, including
```

I think this should be a SHOULD instead of MUST so that existing old implementations formally remain compliant with the spec. And of course we must highly encourage them to implement the limits.

This allows us to meet the formal requirement of not breaking backward compatibility and matches the reality that today many implementations don't have well-defined limits.

docs/specification.md
```diff
@@ -113,6 +114,24 @@ acknowledgements described in this protocol happen between a single
 client/server pair and do not span intermediary nodes in multi-hop delivery
 paths._
 
+#### OTLP/gRPC Request
+
+The server MUST enforce a message size limit when receiving the request,
```
pellared, line 119: @tigrannajaryan same as https://github.com/open-telemetry/opentelemetry-proto/pull/782/files#r3274810087 (given the consensus would be to change to SHOULD)

```suggestion
The server SHOULD enforce a message size limit when receiving the request,
```

Do we want to the same change for the OTLP/gRPC server part for consistency?

The thing is that I believe that everyone already uses  limit when having a gRPC server (as it usually defaults to 4 MiB). On the other hand, I like the spec to be consistent.

docs/specification.md
pellared: I did this intentionally. I think this is a security fix as having no limit is not safe (e.g. see https://github.com/open-telemetry/opentelemetry-go/security/advisories/GHSA-w8rr-5gcm-pp58).

Therefore, yes, old implementations that do not have a limit would no longer be compliant, but I think this is a good call.

docs/specification.md
pellared: > But do you not consider this a breaking change to OTLP exporters?

I see that this can break **some** users, but I see this more a reliability fix than a breaking change. It reduces the possibility of the client or server from running out of memory and degrading performance.


Also see https://github.com/open-telemetry/opentelemetry-proto/pull/782#issuecomment-4499445019

docs/specification.md
tigrannajaryan: OK, works either way, that's a minor nuance.

docs/specification.md
tigrannajaryan: Yes, as far as I know gRPC has this enforced out of the box, unlike http.

docs/specification.md
pellared: Then resolving.

docs/specification.md
pellared: Then there is nothing that we need to change.

felixge (approved):

aabmass (approved):

docs/specification.md
```diff
@@ -113,6 +114,24 @@ acknowledgements described in this protocol happen between a single
...
+including after decompression, to mitigate possible excessive memory usage
+caused by a misconfigured or malicious client sending an oversized request.
+The server implementations typically enforce a default incoming message size
+limit of 4 MiB. However, it is RECOMMENDED to use 64 MiB as the default limit.
+Implementations SHOULD allow this limit to be configured. If the limit is
+exceeded, the gRPC server implementations MUST report a
+`RESOURCE_EXHAUSTED` code to the caller which the client MUST treat as a
```
tigrannajaryan: We already have a section explaining how clients must react to `RESOURCE_EXHAUSTED` and it has more nuance than this. I suggest we delete the second part of this sentence since it lacks that important nuance.

tigrannajaryan (approved):

tigrannajaryan:
@open-telemetry/spec-sponsors please review.

docs/specification.md
pellared: Addressed https://github.com/open-telemetry/opentelemetry-proto/pull/782/commits/34ce0f4455544254ffabdbbd9a710097a62546bf

yurishkuro (approved):

docs/specification.md
```diff
@@ -113,6 +114,23 @@ acknowledgements described in this protocol happen between a single
...
+The server implementations typically enforce a default incoming message size
+limit of 4 MiB. However, it is RECOMMENDED to use 64 MiB as the default limit.
+Implementations SHOULD allow this limit to be configured. If the limit is
+exceeded, the gRPC server implementations MUST report a
+`RESOURCE_EXHAUSTED` code as a non-retryable error.
+
+The client SHOULD limit the size of the request message, including before
+compression, to avoid overwhelming the server. It is RECOMMENDED to use 64 MiB
+as the default limit. Implementations SHOULD allow this limit to be configured.
+If the limit is exceeded, the client MUST NOT make the request and SHOULD record
```
kamphaus, line 131:
```suggestion
If the limit is exceeded, the client MUST NOT retry the request and SHOULD record
```

pellared: The current phrasing is correct. The idea is that the exporter does not make any request at all.

kamphaus: Ok, I get it now.
I thought this was meant to handle the `RESOURCE_EXHAUSTED` non-retryable error code.

kamphaus (approved):

reyang (approved):

PR #786, opened 2026-04-07
[DO NOT MERGE] profiles: more efficient and flexible original_payload handling.

jhalliday:
Move original_payload data to a dictionary table to avoid duplication when a single payload encodes to multiple Profiles.

### DIFF
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -176,10 +176,10 @@ message ProfilesDictionary {
   repeated Stack stack_table = 7;
 
   // Where the OpenTelemetry data is a conversion of some other format (e.g. pprof, JFR)
-  // the original file(s) may be included for reference, particularly if the conversion may be lossy.
+  // the original source(s) may be included for reference, particularly if the conversion may be lossy.
   //
   // original_payload_table[0] must always be a zero-length byte[] i.e. an empty file.
-  repeated bytes original_payload_table = 8;
+  repeated OriginalDataSource original_data_table = 8;
 }
 
 // ProfilesData represents the profiles data that can be stored in persistent storage,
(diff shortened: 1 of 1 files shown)


aalexand:
A bit unusual to have such a large value in the dictionary but makes sense overall.

jhalliday:
Switched to using OriginalDataSource, a new message type that allows each payload content to be accompanied by attributes. These can be used via semconv to e.g. associate a file name with the data.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -340,10 +346,20 @@ message Profile {
...
+  reserved 9, 10; // do not reuse.
+
+  // The format(s) of each file referenced by the corresponding original_payload_indices.
+  // Indexes into ProfilesDictionary.string_table.
+  repeated int32 original_payload_format_stridx = 12;
+  // The original payload bytes. See also original_payload_format_stridx.
+  // Optional, but format and the bytes must be set or unset together.
+  // i.e. original_payload_format_stridx and original_payload_indices must be arrays of equal length.
+  // Indexes into ProfilesDictionary.original_payload_table.
+  repeated int32 original_payload_indices = 13;
```
florianl, line 362: As there is a SemConv attribute dedicated to file formats, [file.extension](https://opentelemetry.io/docs/specs/semconv/registry/attributes/file/#file-extension), I'm wondering if we can drop this field here.

```suggestion
  // The original payload bytes.
  // Indexes into ProfilesDictionary.original_payload_table.
  repeated int32 original_payload_indices = 12;
```

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -523,3 +539,15 @@ message KeyValueAndUnit {
...
+// Where the Profile content is created by translation from another format
+// the original data may also be transmitted, along with meta-data attributes
+// e.g. a JFR file with its file name.
+//
+// Status: [Alpha]
+message OriginalDataSource {
+  // The file or buffer content
+  bytes content = 1;
+  // References to attributes in ProfilesDictionary.attribute_table. [optional]
+  repeated int32 attribute_indices = 2;
```
florianl, line 552: Should we make [file.extension](https://opentelemetry.io/docs/specs/semconv/registry/attributes/file/#file-extension) and/or [file.name](https://opentelemetry.io/docs/specs/semconv/registry/attributes/file/#file-name) a required attribute?

felixge: Discussed in today's SIG, we should probably not make these required as the data could come from in-memory buffers rather than files (h/t Jonathan).

PR #791, opened 2026-04-29
profiles: Add better doc and samples for the period_type / period fields.

aalexand:


aalexand:
@open-telemetry/profiling-approvers

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -302,10 +302,23 @@ message Profile {
   // was gathered. This ensures Sample.timestamps_unix_nano values such as
   // allocation timestamp fall into the profile time range.
   uint64 duration_nano = 4;
-  // The kind of events between sampled occurrences.
-  // e.g [ "cpu","cycles" ] or [ "heap","bytes" ]
+  // The type and the unit of the events between sampled occurrences for
+  // periodic sampling profiles. It can be the same as sample_type or it can be
+  // different depending on the case, for example:
+  // - sample_type=(cpu, milliseconds), period_type=(cpu, milliseconds),
+  //   period=10 signals that we sample the program every 10 milliseconds and
```
florianl, line 314: As the field `period` reflects the distance **between** sampled occurrences, shouldn't `period` set to `1000/10 = 100` instead of `10`?

```suggestion
  //   period=100 signals that we sample the program every 10 milliseconds and
```

aalexand: > As the field `period` reflects the distance **between** sampled occurrences, shouldn't `period` set to `1000/10 = 100` instead of `10`?
> 
> ```suggestion
>   //   period=100 signals that we sample the program every 10 milliseconds and
> ```

Hmm, I don't think so? The period is the distance - e.g. the period of rotation of Earth around the Sun is 365 days.

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -302,10 +302,23 @@ message Profile {
...
+  //   capture samples that each represent that sampling distance.
+  // - sample_type=(off_cpu, nanoseconds), period_type=(context switch, count),
+  //   period=1000 describes a profile where sampling is done every so often in
+  //   terms of context switches, but the recorded metric is the time spent by
+  //   the thread off CPU.
+  // - sample_type=(inuse_space, bytes), period_type=(allocated bytes, bytes),
```
christos68k: Are these (`ValueType.type` strings) mostly arbitrary / free-form or is there a registry somewhere with possible values? 

If the former, should we establish an underscore convention for consistency? e.g.


```suggestion
  // - sample_type=(cpu, milliseconds), period_type=(cpu, milliseconds),
  //   period=10 signals that we sample the program every 10 milliseconds and
  //   capture samples that each represent that sampling distance.
  // - sample_type=(off_cpu, nanoseconds), period_type=(context_switch, count),
  //   period=1000 describes a profile where sampling is done every so often in
  //   terms of context switches, but the recorded metric is the time spent by
  //   the thread off CPU.
  // - sample_type=(inuse_space, bytes), period_type=(allocated_bytes, bytes),
```

christos68k (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
florianl: Sorry for the late responds and thanks for the clarification!
You are right, I was mixing sampling rate with the period value.

florianl (approved):

jsuereth (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
aalexand: Done.

aalexand:
@florianl @christos68k I addressed the comment. I think we can merge this as is and continue separately the discussion about the exact relationship between the sample type, sample shapes and period type.

Gandem (approved):

PR #792, opened 2026-05-01
profiles: improve Link message encoding documentation

jhalliday:
Clarify appropriate representation of invalid Links for better compatibility with existing codecs.

### DIFF
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -140,7 +140,12 @@ message ProfilesDictionary {
 
   // Links referenced by samples via Sample.link_index.
   //
-  // link_table[0] must always be zero value (Link{}) and present.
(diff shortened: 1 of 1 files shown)


jhalliday:
@open-telemetry/profiling-maintainers per [#8349 #issuecomment-4352927214 ](https://github.com/open-telemetry/opentelemetry-java/pull/8349/#issuecomment-4352927214) experience has shown that existing span id / trace id codecs can be strict about type length, i.e. don't play well with the profiles alpha spec definition of invalidity as 'minimum message bytes on the wire'
This doc change clarifies that the ProfilesDictionary.link_table[0] element should be encoded such as to maximise compatibility with the existing ecosystem at the cost of a small increase in message payload size.
As it still permits the previous encoding this is not considered a breaking change and could be merged immediately, unlike e.g. #786

florianl (approved):

felixge (approved):

PR #801, opened 2026-06-01
docs: add server response size limits

pellared:
## Changes

Adds server-side OTLP response size limit requirements:

- OTLP/gRPC servers MUST limit response message size before compression.
- OTLP/HTTP servers MUST limit response body size before compression.
- If a Partial Success response would otherwise exceed the limit, the server SHOULD reduce the response size without changing response semantics if it is possible and practical.
- To reduce the response size, the server MAY decrease the verbosity of optional diagnostic fields, such as `partial_success.error_message`, or omit optional diagnostic fields.
- If the response still cannot fit within the limit, the server MUST fail the request. After such a failure, whether any telemetry data from the request was accepted is unspecified.
- OTLP/gRPC servers MUST fail this request with the `RESOURCE_EXHAUSTED` code as a non-retryable error.
- OTLP/HTTP servers MUST fail this request with `HTTP 500 Internal Server Error`.
- Reuses the existing recommended default response limit of 4 MiB.

## Motivation

The specification already requires clients to enforce response size limits when receiving responses. This change completes the server-side behavior so servers avoid producing responses that can overwhelm clients.

This is inspired by the work done in [open-telemetry/opamp-spec#346](https://github.com/open-telemetry/opamp-spec/pull/346).

Follows [open-telemetry/opentelemetry-proto#781](https://github.com/open-telemetry/opentelemetry-proto/pull/781).

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -197,8 +197,8 @@ semantics if it is possible. To do so, the server MAY decrease the verbosity of
 optional diagnostic fields, such as `partial_success.error_message`, or omit
 optional diagnostic fields. If the response still cannot fit within the limit,
 the server MUST fail the request with the `RESOURCE_EXHAUSTED` code as a
-non-retryable error and MUST NOT partially accept telemetry data from the
-request.
+non-retryable error. After such a failure, whether any telemetry data from the
+request was accepted is unspecified.
 
 ##### Full Success
 
@@ -546,8 +546,9 @@ the limit, the server MUST reduce the response size without changing response
 semantics if it is possible. To do so, the server MAY decrease the verbosity of
 optional diagnostic fields, such as `partial_success.error_message`, or omit
 optional diagnostic fields. If the response still cannot fit within the limit,
-the server MUST fail the request with `HTTP 500 Internal Server Error` and MUST
-NOT partially accept telemetry data from the request.
+the server MUST fail the request with `HTTP 500 Internal Server Error`. After
+such a failure, whether any telemetry data from the request was accepted is
+unspecified.
 
 The server MUST set "Content-Type: application/x-protobuf" header if the
 response body is binary-encoded Protobuf payload. The server MUST set
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -187,6 +187,12 @@ limit to be configured. If the limit is exceeded, the client MUST treat the
 response as a non-retryable error. Note that in such scenario, the gRPC client
 implementations are reporting a `RESOURCE_EXHAUSTED` code to the caller.
 
+The server MUST limit the size of the response message, including before
+compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
+as the default limit. Implementations MAY allow this limit to be configured. If
+the limit is exceeded, the server MUST NOT send the oversized response message
+and SHOULD record the fact that the response was discarded.
+
 ##### Full Success
 
 The success response indicates telemetry data is successfully accepted by the
@@ -524,6 +530,12 @@ misconfigured or malicious server. It is RECOMMENDED to use 4 MiB
 as the default limit. Implementations MAY allow this limit to be configured. If
 the limit is exceeded, the client MUST treat the response as a non-retryable
 error and SHOULD record the fact that the response was discarded.
+
+The server MUST limit the size of the response body, including before
+compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
+as the default limit. Implementations MAY allow this limit to be configured. If
+the limit is exceeded, the server MUST NOT send the oversized response body and
+SHOULD record the fact that the response was discarded.
 
 The server MUST set "Content-Type: application/x-protobuf" header if the
 response body is binary-encoded Protobuf payload. The server MUST set
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -193,10 +193,11 @@ compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
 as the default limit. Implementations MAY allow this limit to be configured.
 For a [Partial Success](#partial-success) response that would otherwise exceed
 the limit, the server MUST reduce the response size without changing response
-semantics. To do so, the server MAY decrease the verbosity of optional
-diagnostic fields, such as `partial_success.error_message`, or omit optional
-diagnostic fields. If the response cannot fit within the limit, the server MUST
-fail the request with the `RESOURCE_EXHAUSTED` code as a non-retryable error.
+semantics if it is possible. To do so, the server MAY decrease the verbosity of
+optional diagnostic fields, such as `partial_success.error_message`, or omit
+optional diagnostic fields. If the response cannot fit within the limit, the
+server MUST fail the request with the `RESOURCE_EXHAUSTED` code as a
+non-retryable error.
 
 ##### Full Success
 
@@ -541,10 +542,10 @@ compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
 as the default limit. Implementations MAY allow this limit to be configured.
 For a [Partial Success](#partial-success-1) response that would otherwise exceed
 the limit, the server MUST reduce the response size without changing response
-semantics. To do so, the server MAY decrease the verbosity of optional
-diagnostic fields, such as `partial_success.error_message`, or omit optional
-diagnostic fields. If the response cannot fit within the limit, the server MUST
-fail the request with `HTTP 500 Internal Server Error`.
+semantics if it is possible. To do so, the server MAY decrease the verbosity of
+optional diagnostic fields, such as `partial_success.error_message`, or omit
+optional diagnostic fields. If the response cannot fit within the limit, the
+server MUST fail the request with `HTTP 500 Internal Server Error`.
 
 The server MUST set "Content-Type: application/x-protobuf" header if the
 response body is binary-encoded Protobuf payload. The server MUST set
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -190,13 +190,13 @@ reporting a `RESOURCE_EXHAUSTED` code to the caller.
 
 The server MUST limit the size of the response message, including before
 compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
-as the default limit. Implementations MAY allow this limit to be configured. For
-a [Partial Success](#partial-success) response, if necessary, the server MUST
-reduce the response size by shortening or omitting optional diagnostic fields,
-such as `partial_success.error_message`, if doing so preserves the response
-semantics so that the response message does not exceed the limit. If the
-response cannot fit within the limit, the server MUST fail the request with
-the `RESOURCE_EXHAUSTED` code as a non-retryable error.
+as the default limit. Implementations MAY allow this limit to be configured.
+For a [Partial Success](#partial-success) response that would otherwise exceed
+the limit, the server MUST reduce the response size without changing response
+semantics. To do so, the server MAY decrease the verbosity of optional
+diagnostic fields, such as `partial_success.error_message`, or omit optional
+diagnostic fields. If the response cannot fit within the limit, the server MUST
+fail the request with the `RESOURCE_EXHAUSTED` code as a non-retryable error.
 
 ##### Full Success
 
@@ -538,13 +538,13 @@ error and SHOULD record the fact that the response was discarded.
 
 The server MUST limit the size of the response body, including before
 compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
-as the default limit. Implementations MAY allow this limit to be configured. For
-a [Partial Success](#partial-success-1) response, if necessary, the server MUST
-reduce the response size by shortening or omitting optional diagnostic fields,
-such as `partial_success.error_message`, if doing so preserves the response
-semantics so that the response body does not exceed the limit. If the response
-cannot fit within the limit, the server MUST fail the request with
-`HTTP 500 Internal Server Error`.
+as the default limit. Implementations MAY allow this limit to be configured.
+For a [Partial Success](#partial-success-1) response that would otherwise exceed
+the limit, the server MUST reduce the response size without changing response
+semantics. To do so, the server MAY decrease the verbosity of optional
+diagnostic fields, such as `partial_success.error_message`, or omit optional
+diagnostic fields. If the response cannot fit within the limit, the server MUST
+fail the request with `HTTP 500 Internal Server Error`.
 
 The server MUST set "Content-Type: application/x-protobuf" header if the
 response body is binary-encoded Protobuf payload. The server MUST set
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -192,8 +192,8 @@ The server MUST limit the size of the response message, including before
 compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
 as the default limit. Implementations SHOULD allow this limit to be configured.
 For a [Partial Success](#partial-success) response that would otherwise exceed
-the limit, the server MUST reduce the response size without changing response
-semantics if it is possible. To do so, the server MAY decrease the verbosity of
+the limit, the server SHOULD reduce the response size without changing response
+semantics if it is possible and practical. To do so, the server MAY decrease the verbosity of
 optional diagnostic fields, such as `partial_success.error_message`, or omit
 optional diagnostic fields. If the response still cannot fit within the limit,
 the server MUST fail the request with the `RESOURCE_EXHAUSTED` code as a
@@ -542,8 +542,8 @@ The server MUST limit the size of the response body, including before
 compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
 as the default limit. Implementations SHOULD allow this limit to be configured.
 For a [Partial Success](#partial-success-1) response that would otherwise exceed
-the limit, the server MUST reduce the response size without changing response
-semantics if it is possible. To do so, the server MAY decrease the verbosity of
+the limit, the server SHOULD reduce the response size without changing response
+semantics if it is possible and practical. To do so, the server MAY decrease the verbosity of
 optional diagnostic fields, such as `partial_success.error_message`, or omit
 optional diagnostic fields. If the response still cannot fit within the limit,
 the server MUST fail the request with `HTTP 500 Internal Server Error`. After
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -193,12 +193,13 @@ compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
 as the default limit. Implementations SHOULD allow this limit to be configured.
 For a [Partial Success](#partial-success) response that would otherwise exceed
 the limit, the server SHOULD reduce the response size without changing response
-semantics if it is possible and practical. To do so, the server MAY decrease the verbosity of
-optional diagnostic fields, such as `partial_success.error_message`, or omit
-optional diagnostic fields. If the response still cannot fit within the limit,
-the server MUST fail the request with the `RESOURCE_EXHAUSTED` code as a
-non-retryable error. After such a failure, whether any telemetry data from the
-request was accepted is unspecified.
+semantics if it is possible and practical. To do so, the server MAY decrease the
+verbosity of optional diagnostic fields, such as
+`partial_success.error_message`, or omit optional diagnostic fields. If the
+response still cannot fit within the limit, the server MUST fail the request
+with the `RESOURCE_EXHAUSTED` code as a non-retryable error. After such a
+failure, whether any telemetry data from the request was accepted is
+unspecified.
 
 ##### Full Success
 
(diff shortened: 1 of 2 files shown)


docs/specification.md
```diff
@@ -187,6 +187,14 @@ limit to be configured. If the limit is exceeded, the client MUST treat the
 response as a non-retryable error. Note that in such scenario, the gRPC client
 implementations are reporting a `RESOURCE_EXHAUSTED` code to the caller.
 
+The server MUST limit the size of the response message, including before
+compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
+as the default limit. Implementations MAY allow this limit to be configured. If
+necessary, the server MUST reduce the response size by shortening
```
tigrannajaryan: Is this a useful thing to try? Supposedly error messages are important. How much shorten is OK before the meaning is lost?

We don't try this shortening for requests, why have this asymmetric requirement for responses?

pellared: > why have this asymmetric requirement for responses?

I think there is a difference between a "request" and "response". I do not think we want to avoid failing requests which would result in Partial Success. 

> Supposedly error messages are important. How much shorten is OK before the meaning is lost?

This is already defined in https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#partial-success

> Additionally, the server MUST initialize the `partial_success` field
> (`ExportTracePartialSuccess` message for traces, `ExportMetricsPartialSuccess`
> message for metrics, `ExportLogsPartialSuccess` message for logs and
> `ExportProfilesPartialSuccess` for profiles), and it MUST set the respective
> `rejected_spans`, `rejected_data_points`, `rejected_log_records` or `rejected_profiles` field with
> the number of spans/data points/log records it rejected.
> 
> The server SHOULD populate the `error_message` field with a human-readable
> error message in English. The message should explain why the
> server rejected parts of the data and might offer guidance on how users
> can address the issues.
> The protocol does not attempt to define the structure of the error message.
> 
> Servers MAY also use the `partial_success` field to convey
> warnings/suggestions to clients even when it fully accepts the request.
> In such cases, the `rejected_<signal>` field MUST have a value of `0`, and
> the `error_message` field MUST be non-empty.

I will try to refine the spec to make it clearer (e.g. refer to this section) and I think we should cover the edge case when 

>MUST initialize the `partial_success` field

exceeds the limit. 

Related conversation: https://github.com/open-telemetry/opentelemetry-proto/pull/781#discussion_r3010008338

pellared: PTAL https://github.com/open-telemetry/opentelemetry-proto/pull/801/commits/872592a9cd8235e80804e143a7645be1b747e413

pellared:
@open-telemetry/collector-maintainers, PTAL

docs/specification.md
```diff
@@ -187,6 +187,16 @@ limit to be configured. If the limit is exceeded, the client MUST treat the
 response as a non-retryable error. Note that in such scenario, the gRPC client
 implementations are reporting a `RESOURCE_EXHAUSTED` code to the caller.
 
+The server MUST limit the size of the response message, including before
+compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
+as the default limit. Implementations MAY allow this limit to be configured. For
+a [Partial Success](#partial-success) response, if necessary, the server MUST
+reduce the response size by shortening or omitting optional diagnostic fields,
+such as `partial_success.error_message`, if doing so preserves the response
+semantics so that the response message does not exceed the limit. If the
```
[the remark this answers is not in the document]
pellared: I prefer the current version

docs/specification.md
```diff
@@ -525,6 +535,16 @@ as the default limit. Implementations MAY allow this limit to be configured. If
 the limit is exceeded, the client MUST treat the response as a non-retryable
 error and SHOULD record the fact that the response was discarded.
 
+The server MUST limit the size of the response body, including before
+compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
+as the default limit. Implementations MAY allow this limit to be configured. For
+a [Partial Success](#partial-success-1) response, if necessary, the server MUST
+reduce the response size by shortening or omitting optional diagnostic fields,
+such as `partial_success.error_message`, if doing so preserves the response
+semantics so that the response body does not exceed the limit. If the response
```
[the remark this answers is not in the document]
pellared: I prefer the current version

martincostello (approved):

docs/specification.md
pellared: By @tigrannajaryan and myself during the SIG meeting:
Let's try to be more specific e.g. say that instead of shortening (this cannot be a naive trim) saying like decreasing the verbosity of `partial_success.error_message`.

pellared: https://github.com/open-telemetry/opentelemetry-proto/pull/801/commits/ae683ff044e8cf015502d5948543cb91499ad090

docs/specification.md
```diff
@@ -188,6 +188,16 @@ If the limit is exceeded, the client MUST treat the response as a non-retryable
...
 
+The server MUST limit the size of the response message, including before
+compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
+as the default limit. Implementations MAY allow this limit to be configured.
+For a [Partial Success](#partial-success) response that would otherwise exceed
+the limit, the server MUST reduce the response size without changing response
+semantics. To do so, the server MAY decrease the verbosity of optional
+diagnostic fields, such as `partial_success.error_message`, or omit optional
+diagnostic fields. If the response cannot fit within the limit, the server MUST
+fail the request with the `RESOURCE_EXHAUSTED` code as a non-retryable error.
```
jsuereth: Should we comment on whether the server should ALSO reject points - or do partially accepted points still get accepted?

docs/specification.md
```diff
@@ -526,6 +536,16 @@ as the default limit. Implementations SHOULD allow this limit to be configured.
...
 
+The server MUST limit the size of the response body, including before
+compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
+as the default limit. Implementations MAY allow this limit to be configured.
+For a [Partial Success](#partial-success-1) response that would otherwise exceed
+the limit, the server MUST reduce the response size without changing response
+semantics. To do so, the server MAY decrease the verbosity of optional
+diagnostic fields, such as `partial_success.error_message`, or omit optional
+diagnostic fields. If the response cannot fit within the limit, the server MUST
+fail the request with `HTTP 500 Internal Server Error`.
```
jsuereth: Same question as above - can we comment on whether points must be rejected or partial acceptance is ok here?

docs/specification.md
pellared: What do you think about specifying the behavior for this scenario explicitly as unpredictable/unspecified?

docs/specification.md
pellared: Same as before - What do you think about specifying the behavior for this scenario explicitly as unpredictable/unspecified?

pellared:
ping @open-telemetry/collector-maintainers

docs/specification.md
```diff
@@ -188,6 +188,16 @@ If the limit is exceeded, the client MUST treat the response as a non-retryable
 error. Note that in such scenario, the gRPC client implementations are
 reporting a `RESOURCE_EXHAUSTED` code to the caller.
 
+The server MUST limit the size of the response message, including before
```
mx-psi, line 191: Given OTLP is already stable, shouldn't this be a SHOULD?

docs/specification.md
```diff
@@ -188,6 +188,16 @@ If the limit is exceeded, the client MUST treat the response as a non-retryable
...
+compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
+as the default limit. Implementations MAY allow this limit to be configured.
+For a [Partial Success](#partial-success) response that would otherwise exceed
+the limit, the server MUST reduce the response size without changing response
+semantics. To do so, the server MAY decrease the verbosity of optional
```
mx-psi:
```suggestion
semantics if it is possible. To do so, the server MAY decrease the verbosity of optional
```

docs/specification.md
pellared: In https://github.com/open-telemetry/opamp-spec/pull/346 we decided to use a stronger MUST for the server given older clients may not handle big responses and causing https://cwe.mitre.org/data/definitions/789.html. We see this also as a security bugfix (and not only a new feature).

docs/specification.md
pellared: Addressed in https://github.com/open-telemetry/opentelemetry-proto/pull/801/commits/e00a7275932da5cc46d17ec9be550e9f413c0771

docs/specification.md
pellared: @mx-psi, thoughts?

docs/specification.md
mx-psi: I am unsure if I would consider this a security bugfix given our new [security model](https://github.com/open-telemetry/sig-security/blob/main/security-response.md#security-model) but if there is precedent I am fine with going forward with a MUST

mx-psi (approved):

docs/specification.md
mx-psi: > What do you think about specifying the behavior for this scenario explicitly as unpredictable/unspecified?

Do we have examples of both behaviors being desirable? Rejecting the points is the option that would make more sense to me intuitively/for the Collector, but maybe there are other use cases I am missing

pellared: I have similar opinion that rejecting seems safer and more intuitive.

pellared: PTAL 99b435b

docs/specification.md
pellared: PTAL fbb2017

pellared:
@open-telemetry/collector-approvers, can you make a prototype to validate this proposal?

docs/specification.md
```diff
@@ -188,6 +188,18 @@ If the limit is exceeded, the client MUST treat the response as a non-retryable
...
+compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
+as the default limit. Implementations SHOULD allow this limit to be configured.
+For a [Partial Success](#partial-success) response that would otherwise exceed
+the limit, the server MUST reduce the response size without changing response
+semantics if it is possible. To do so, the server MAY decrease the verbosity of
+optional diagnostic fields, such as `partial_success.error_message`, or omit
+optional diagnostic fields. If the response still cannot fit within the limit,
+the server MUST fail the request with the `RESOURCE_EXHAUSTED` code as a
+non-retryable error and MUST NOT partially accept telemetry data from the
+request.
```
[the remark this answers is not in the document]
pellared: repetition is not what we want

jade-guiton-dd: I'm a bit confused by this part. How can a server "NOT partially accept telemetry data from the request" _after_ it determined it needed to send a partial success response? Are we asking servers to determine whether the data will lead to a partial success _before_ starting processing on any of it?

pellared: Can you please reply in this thread to not create a new one https://github.com/open-telemetry/opentelemetry-proto/pull/801#discussion_r3458787408? 

PS. What you write is one of the reasons I thought about explicitly saying that in this scenario the server behavior is not predictable.

jade-guiton-dd: Will do

docs/specification.md
jade-guiton-dd: (From my comment below) I'm not sure how rejecting all points would be implemented in practice. How can a server _"NOT partially accept telemetry data from the request"_ after it determined it needed to send a partial success response? Are we asking servers to determine whether the data will lead to a partial success before starting processing on any of it?

pellared: @jade-guiton-dd, @mx-psi, @open-telemetry/collector-maintainers, I would appropriate if you make a PR with a prototype to validate what suits you better.

Personally, I think that in such scenario we can describe that that the behavior is unspecified (or that it can be unpredictable). PTAL fbb2017

jade-guiton-dd: Thinking about how to implement this, I'm also not sure in what context the server should deem itself fully "unable" to fit the response below the limit. Would "decreasing the verbosity" allow for simple measures like truncating the error message to whatever fits? By my count, this can only truly fail (as in, the result is no longer a valid partial success response) if the limit is 6 bytes or less, which we should probably forbid anyway.

pellared: >  this can only truly fail (as in, the result is no longer a valid partial success response) if the limit is 6 bytes or less, which we should probably forbid anyway.

This is "currently" but in the future it may change.

> which we should probably forbid anyway

I think this is fine for the collector to have some minimal acceptable value. Then the "unspecified behavior" would be unreachable. Right?

dmitryax (approved):

jade-guiton-dd:
> @open-telemetry/collector-approvers, can you make a prototype to validate this proposal?

After looking into it, the OTLP receiver never generates a `PartialResponse` at the moment. So I don't think we can make a useful prototype.

jade-guiton-dd:
It may be possible to limit the size of the `Status` value returned in case of error however. Maybe how to do that should be explicitly described in the spec similarly to the `PartialResponse` case? But it seems difficult to implement reliably in all cases, especially for gRPC, since the gRPC library usually controls the generation and serialization of the final `Status`, and some `Status` values may not come from application code at all.

I think the best we can easily do is assume large responses typically come from serializing application-controled error values, and do a best effort truncation of those, but it's difficult to translate that into a hard limit on payload size. It's also hard to predict how easily this can be implemented in languages that do error handling differently. So overall I think we may want to relax the specification-level requirements here.

(I also noticed that our gRPC library interprets `MaxRecvMsgSize` and `MaxSendMsgSize` as _compressed_ size, not uncompressed, so even the options available in the library to reject large messages, without trying to shorten them first, may not allow the OTLP receiver to be up-to-spec given the _"including before compression"_ wording...)

pellared:
@jade-guiton-dd, thanks a lot for your review!

> It may be possible to limit the size of the `Status` value returned in case of error however.

Thanks, this makes sense. I was mainly thinking about the regular OTLP response message and the Partial Success case, but I agree failure responses need separate treatment.

Do you mean the transport error status here: the gRPC status message/details for OTLP/gRPC and the `google.rpc.Status` response body for OTLP/HTTP? Right?

> I think the best we can easily do is assume large responses typically come from serializing application-controled error values, and do a best effort truncation of those, but it's difficult to translate that into a hard limit on payload size. It's also hard to predict how easily this can be implemented in languages that do error handling differently. So overall I think we may want to relax the specification-level requirements here.

For application-controlled error values, I think the spec can still recommend best-effort behavior: avoid attaching large diagnostic details, omit optional `Status.details` where possible, and use a short generic message instead of truncating structured values. I would prefer omitting optional diagnostic fields/details over trimming them, because trimming may change or corrupt their meaning.

> I also noticed that our gRPC library interprets MaxRecvMsgSize and MaxSendMsgSize as compressed size

I would still consider that compressed-size behavior a bug or at least a misleading interpretation of `MaxRecvMsgSize` / `MaxSendMsgSize`. From an OTLP safety perspective, the uncompressed size is the value that matters most for memory pressure.

jade-guiton-dd:
> Do you mean the transport error status here: the gRPC status message/details for OTLP/gRPC and the google.rpc.Status response body for OTLP/HTTP? Right?

Yes, the error information in [this struct](https://pkg.go.dev/google.golang.org/genproto/googleapis/rpc/status#Status), which I believe is used in both protocols in different forms. It would make sense to handle this case separately.

> I would prefer omitting optional diagnostic fields/details over trimming them, because trimming may change or corrupt their meaning.

I agree, but I'm unsure if it's common for the Details field to be filled in the first place, so I think very little trimming could be done in practice if we don't allow truncation of the error message string. The spec can recommend shorter error messages, but that's hard to enforce across an entire codebase.

> I would still consider that compressed-size behavior a bug or at least a misleading interpretation of MaxRecvMsgSize / MaxSendMsgSize. From an OTLP safety perspective, the uncompressed size is the value that matters most for memory pressure.

We could set up a Collector-specific gRPC interceptor to do this, although it would probably have a non-negligible runtime cost. So maybe the best option would be to file an issue with grpc-go? I think in general this shows how difficult it is to follow this part of the spec if you don't write your own gRPC implementation from scratch.

pellared:
> So maybe the best option would be to file an issue with grpc-go? I think in general this shows how difficult it is to follow this part of the spec if you don't write your own gRPC implementation from scratch.

You should be good when you are using grpc-go. You can also look at e.g. https://github.com/open-telemetry/opentelemetry-go/pull/8157/

From Gemini:

> In `google.golang.org/grpc`, the `MaxCallRecvMsgSize` limit is applied **after** decompression.
> 
> When you receive a compressed message, the gRPC-Go library first decompresses the payload and then checks the size of the resulting uncompressed bytes. If the uncompressed size exceeds the configured limit (which defaults to 4MB), it will reject the message and return an error that specifically notes the decompression phase:
> 
> `rpc error: code = ResourceExhausted desc = grpc: received message after decompression larger than max (...)`
> 
> **Why is it designed this way?**
> This behavior is a security measure to protect your application from **"zip bomb" (or decompression bomb) attacks**. A malicious client could send a highly compressed payload that is very small over the wire (e.g., under 4MB) but expands to gigabytes of data when uncompressed. By applying the limit *after* decompression, gRPC ensures that your application won't allocate more memory than the strict limit you've configured, effectively preventing memory exhaustion.

> In `google.golang.org/grpc`, the `MaxCallSendMsgSize` limit is applied **before** compression.
> 
> When you attempt to send a message, the gRPC-Go library processes it in the following order:
> 
> 1. It serializes your protobuf message into a raw, uncompressed byte slice.
> 2. It checks the length of those uncompressed bytes against the `MaxCallSendMsgSize` limit.
> 3. If the limit is exceeded, it immediately returns a `ResourceExhausted` error: `trying to send message larger than max...`
> 4. If it passes the check, it passes the bytes to the compressor (if compression is enabled) and sends the framed payload over the wire.
> 
> **Why is it designed this way?**
> 
> * **Resource Protection (CPU & Memory):** Compression is a CPU-intensive and memory-intensive process. If your application accidentally tries to send a massively oversized payload (e.g., 2 GB of data), trying to compress it first would waste a significant amount of compute resources and likely cause memory spikes. By failing the message *before* compression, gRPC saves those resources.
> * **Semantic Consistency:** The send limit is meant to govern the *logical* size of the data your application is allowed to emit, rather than how well a specific payload happens to compress on the wire.
> 
> So, while the **receive limit** (`MaxCallRecvMsgSize`) acts as a shield *after* decompression to protect against malicious zip-bombs from external sources, the **send limit** (`MaxCallSendMsgSize`) acts as a short-circuit *before* compression to protect your own application from wasting resources on oversized outbound payloads.

I remember that I was double-checking this :wink:

jade-guiton-dd:
These are the options for clients (`MaxCallRecvMsgSize` / `MaxCallSendMsgSize`), but we're discussing server behavior (`MaxRecvMsgSize` / `MaxSendMsgSize`). [This code](https://github.com/grpc/grpc-go/blob/1eb52d9b6ee11ba952f2d2127402c4e7b4ab379b/server.go#L1195-L1216) (and the TODO comment) suggests that `MaxSendMsgSize` is only checked _after_ compression.

(However, reading the code again, it looks like `MaxRecvMsgSize` is applied before _and after_ decompression, so I think the OTLP receiver should be okay on the receiving front, although it doesn't apply the RECOMMENDED default limit.)

Even if we assume that this is a bug and that all gRPC libraries should have a pre-compression send / post-decompression receive limit option, I don't think we can assume there is an easy way to reduce the response size when the limit is reached, so to avoid overly burdening implementors I suggest that it would be good to reduce the _"the server MUST reduce the response size [...] if it is possible"_ requirement to a "SHOULD" and "if it is practical to do so".

This would allow for a simple "hard limit" implementation to be compliant without needing to evaluate whether it would _technically_ be possible to automatically shorten the response by completely refactoring error handling and plugging deep into the gRPC library. I think we especially want to avoid this complexity given that if an application trusts an OTLP endpoint with its telemetry, there is little reason to believe it would be actively malicious and send back a zip bomb as an error message, so this is a bit of a niche concern in the first place.

docs/specification.md
```diff
@@ -188,6 +188,18 @@ If the limit is exceeded, the client MUST treat the response as a non-retryable
 error. Note that in such scenario, the gRPC client implementations are
 reporting a `RESOURCE_EXHAUSTED` code to the caller.
 
+The server MUST limit the size of the response message, including before
+compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
+as the default limit. Implementations SHOULD allow this limit to be configured.
+For a [Partial Success](#partial-success) response that would otherwise exceed
+the limit, the server MUST reduce the response size without changing response
+semantics if it is possible. To do so, the server MAY decrease the verbosity of
```
```suggestion
the limit, the server SHOULD reduce the response size without changing response
semantics if it is possible and practical. To do so, the server MAY decrease the verbosity of
```

docs/specification.md
```diff
@@ -526,6 +538,18 @@ as the default limit. Implementations SHOULD allow this limit to be configured.
 If the limit is exceeded, the client MUST treat the response as a non-retryable
 error and SHOULD record the fact that the response was discarded.
 
+The server MUST limit the size of the response body, including before
+compression, to avoid overwhelming the client. It is RECOMMENDED to use 4 MiB
+as the default limit. Implementations SHOULD allow this limit to be configured.
+For a [Partial Success](#partial-success-1) response that would otherwise exceed
+the limit, the server MUST reduce the response size without changing response
+semantics if it is possible. To do so, the server MAY decrease the verbosity of
```
```suggestion
the limit, the server SHOULD reduce the response size without changing response
semantics if it is possible and practical. To do so, the server MAY decrease the verbosity of
```

pellared:
@jade-guiton-dd

>  I suggest that it would be good to reduce the "the server MUST reduce the response size [...] if it is possible" requirement to a "SHOULD" and "if it is practical to do so".

Addressed in https://github.com/open-telemetry/opentelemetry-proto/pull/801/commits/fb69c457691c83a598958a90f228f5966562d8a1

Regarding the Status field e.g. for Failures I think we are good as then the specified behavior when the response would exceed the limit will be:

> If the response still cannot fit within the limit, the server MUST fail the request with the `RESOURCE_EXHAUSTED` code as a non-retryable error. After such a failure, whether any telemetry data from the request was accepted is unspecified.

(as far as I remember this is what grpc-go is doing)

and

> If the response still cannot fit within the limit, the server MUST fail the request with `HTTP 500 Internal Server Error`. After such a failure, whether any telemetry data from the request was accepted is unspecified.

This will also allow sharing the same code.

Let me know if the current revision makes sense for you.

florianl (approved):

jade-guiton-dd (approved):

mx-psi (approved):
I discussed this with Jade offline, the SHOULD change makes sense to me

jack-berg (approved):

docs/specification.md
jack-berg: @tigrannajaryan can you confirm this is reasonable from your POV?

jmacd (approved):

tigrannajaryan (approved):

PR #827, opened 2026-07-17
Fix BCR publishing and add tests

### LINKED ISSUE #796: Bazel9 support missing
Hi,

Although opentelemetry-proto could support bazel9, it (1) doesn't say so in BCR, (2) it in practice doesn't because critically relies on dependencies which do not. Dependencies such as rules_go are listed as so old (0.50.1) that they do not support the backwards-incompatible changes such as loss of implicit CcFile.

Would it be possible to update opentelemetry-proto to support bazel9? I tried to give it a try myself, but I'm fairly unfamiliar with the BCR ecosystem or what automation is in place.

My intent was to:

- Modify this file https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/opentelemetry-proto/1.8.0/MODULE.bazel to refer latest versions of each package, and also to replace dependency on rules_proto to instead use protobuf BCR package.
- Modify https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/opentelemetry-proto/1.8.0/presubmit.yml to include 9.x bazel compatibility.

I stopped short of doing so because I wasn't sure about the publish-to-bcr workflow and how it would interact with any manual changes.

Thanks,
 - Stepan

jsuereth:
- Add minimal MODULE template and BUILD file for BCR.
- Add `.bazelignore` so bazel ignores `.bcr/` tempaltes in local build.
- Create github action to do brief check of BUILD/MODULE
- Update all docs to denote the level of support and expectation *in this repo*.  BCR users should self-service fixes via OVERLAY mechanics of BCR - we will only fix the "notify BCR on release via a PR" publishing step here.
- Adds rennovate config to keep BCR module dependencies up to date.

Fixes #776
Fixes #796


.github/workflows/build-check.yaml
```diff
@@ -18,6 +18,23 @@ jobs:
...
+        uses: bazel-contrib/setup-bazel@0.8.5
+        with:
+          disk-cache: ${{ github.workflow }}
+          repository-cache: true
+      - name: Prepare minimal Bazel workspace from .bcr templates
+        run: |
+          cp .bcr/BUILD.bazel BUILD.bazel
+          sed 's/{VERSION}/0.0.0-test/' .bcr/MODULE.bazel.template > MODULE.bazel
+      - name: Build all minimal Bazel proto_library targets
+        run: bazel build //...
```
ps-mir, line 36: I think it would be good to pin the bazel version.

.github/workflows/build-check.yaml
```diff
@@ -18,6 +18,23 @@ jobs:
...
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - name: Setup Bazel
+        uses: bazel-contrib/setup-bazel@0.8.5
+        with:
+          disk-cache: ${{ github.workflow }}
+          repository-cache: true
+      - name: Prepare minimal Bazel workspace from .bcr templates
+        run: |
+          cp .bcr/BUILD.bazel BUILD.bazel
+          sed 's/{VERSION}/0.0.0-test/' .bcr/MODULE.bazel.template > MODULE.bazel
```
ps-mir, line 34: This works for build, but i think bcr release workflow could still fail as it won't find this file.

ps-mir (commented):
Removing dependency from `rules_go` solves bazel 9.x issues. 

On BCR release, unless bcr publish workflow allows generating MODULE.bazel similar to build-check.yaml, this might get stuck with same error.

`Keeping bazel out of core` and `Automated BCR` seems mutually exclusive unless there is way include bazel files in release archive. Happy to be proven wrong, and learn something new.

jsuereth:
> Keeping bazel out of core and Automated BCR seems mutually exclusive unless there is way to prepare tar file after generating bazel files. Happy to be proven wrong, and learn something new.

What we do *not* do in this proto repo is generate client libraries that are consumable by others, we expect that to be owned/maintained donwstream.  E.g. In many Otel SDKs / Collector there's bespoke / optimised OTLP proto generation that wouldn't use the protoc generated code.

If my understanding is correct - BCR allows creating overlays on top of our ZIP distributions.  The hope here was we could publish a "baseline" that would have the community update via overlays in the BCR github repo.

If BCR publishing requires a full BAZEL build for this repo - does the overlay mechanism still work if we provide a BUILD?

The current action I added already significantly delays the CI/CD pipeline - so I was debating moving it to something that doesn't run on every PR.  Is there I way we can be bazel friendly without having to depend directly on bazel or pay the cost of rebuilding protoc in our build?

ps-mir:
> What we do _not_ do in this proto repo is generate client libraries that are consumable by others, we expect that to be owned/maintained donwstream. E.g. In many Otel SDKs / Collector there's bespoke / optimised OTLP proto generation that wouldn't use the protoc generated code.

The approach makes more sense given this ask. I'm still doubtful about overlay as I haven't seen any working example or doc references. I'll keep digging.

ps-mir:
Overlay won't be useful as its a downstream mechanism/control for non bazel upstream deps. It can be done locally, and also in BCR by creating manual entry with Overlay (BCR release process so far).

But automated BCR needs the lib to be bazel module, and a bazel module cannot be overlayed (Immutable entry in BCR).

Assuming automated BCR is still preferred, the cleanest way is to make it a bazel module, and release proto only as baseline, let downstream generate the lib bindings.

If more time is needed for downstream community, we have to include proto+lang bindings (same as last BCR overlay) before eventually switching to proto only release.

jsuereth:
FYI discussed in specification call.  Unless we get a maintainer to stand up and maintain this we're likely to remove this from the proto repo and ask that this be maintained directly in BCR.

I'm not happy with build time increase from a few minutes to half hour or more with verification of these rules, plus no existing proto maintainer is comfortable with bazel

ps-mir:
> Hi @jsuereth — just a friendly reminder that this pull request has been waiting on you for a week.

I think this PR can be closed. BCR flow needs to be removed from the repo.

pellared (changes requested):
Blocking as it seems that there is no maintainer that wants to support BCR. More https://github.com/open-telemetry/opentelemetry-proto/pull/827#issuecomment-5037106658

jsuereth:
No need to officially block - sorry I went on vacation before I could close this.

Closing now - will follow up with PR to remove existing BCR publish workflow and close related bugs with rationale

PR #845, opened 2026-08-14
Add special note about UTF-8 handling to help impelmentors.

### LINKED ISSUE #814: Invalid UTF-8 error handling policy
## Summary

The OpenTelemetry project lacks a clear cross-project policy for invalid UTF-8 in telemetry string fields.

This came up in the Collector in https://github.com/open-telemetry/opentelemetry-collector/issues/11449: the OTLP receiver can accept non-UTF-8 data because the Collector's current protobuf path is permissive. That data can later fail when another component or backend uses a stricter protobuf implementation, potentially causing a whole batch of otherwise valid telemetry to be rejected.

There was already an OTEP for this: https://github.com/open-telemetry/oteps/pull/257. It was closed because OTEPs moved to this repository, with a request to reopen it against the new location (at least I was unable to find it).

## Related work

- Accepted/open spec issue: https://github.com/open-telemetry/opentelemetry-specification/issues/3421
- Related/open spec issue: https://github.com/open-telemetry/opentelemetry-specification/issues/3950
- Closed old OTEP PR: https://github.com/open-telemetry/oteps/pull/257
- Collector example: https://github.com/open-telemetry/opentelemetry-collector/issues/11449
- Downstream batch-loss example: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35723

## Current gap

The current spec already has useful pieces:

- `specification/common/attribute-type-mapping.md` says valid UTF-8 strings should map to `string_value`, while invalid Unicode strings should map to `bytes_value`.
- `specification/common/README.md` defines `AnyValue`, including byte arrays.
- `specification/error-handling.md` says APIs and SDKs should provide safe defaults for invalid arguments.

The missing part is a general policy for OTLP protobuf `string` fields that are not just `AnyValue` string attributes, such as span names, event names, instrumentation scope names, metric units, log event names, status descriptions, and attribute keys.

Without that policy, permissive components can pass invalid data forward until a strict component rejects it later. That makes failures hard to diagnose and can lose valid telemetry batched with the invalid item.

## Proposal

Create a UTF-8 validation policy, based on https://github.com/open-telemetry/oteps/pull/257.

The OTEP should stay at the interoperability-policy layer:

- SDK export behavior for invalid UTF-8 in telemetry string fields.
- Receiver-boundary behavior when external telemetry contains invalid UTF-8.
- Interaction with the existing `AnyValue` mapping rule for invalid string values.
- Compatibility tradeoffs between repair, partial rejection, and permissive pass-through.

## Suggested direction

The least harmful default appears to be repair at boundaries: replace invalid UTF-8 byte sequences with the Unicode replacement character before the data reaches stricter protobuf implementations.

Strict rejection should remain possible for users who want it, ideally rejecting individual telemetry items rather than whole batches where the protocol supports partial success.

Permissive pass-through may be needed for compatibility, but it should be documented as allowing invalid data to fail later in the pipeline.

jsuereth:
Closes #814 

Updates our policy to be clear:

- We are no stricter than Protobuf itself
- JSON already requires valid UTF-8 string for both sending/parsing so it is unaffected here.
- Offer recommendations to implementors for addressing this issue.

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -59,7 +59,7 @@ nodes such as collectors and telemetry backends.
 - [Implementation Recommendations](#implementation-recommendations)
   * [Multi-Destination Exporting](#multi-destination-exporting)
   * [Empty Telemetry Envelopes](#empty-telemetry-envelopes)
-    + [UTF-8 String Handling](#utf-8-string-handling)
+  * [UTF-8 String Handling](#utf-8-string-handling)
 - [Known Limitations](#known-limitations)
   * [Request Acknowledgements](#request-acknowledgements)
     + [Duplicate Data](#duplicate-data)
@@ -747,7 +747,7 @@ payloads that contain zero spans, zero metric points or zero log records),
 receivers MAY ignore empty envelopes, and implementations that receive and send
 (forward) OTLP payloads MAY drop empty envelopes.
 
-#### UTF-8 String Handling
+### UTF-8 String Handling
 
 While, by specification, all protocol buffer implementations are required to
 send UTF-8 or ASCII-7 strings, in practice many implementations allow native
@@ -762,19 +762,17 @@ benefit not worth the performance hit.
 
 Consequently:
 
+* Encoders SHOULD ensure `string` fields contain valid UTF-8. Producing
+  invalid UTF-8 is not specification compliant and is considered a bug within
+  OpenTelemetry.
+* Decoders SHOULD be prepared to receive non-compliant strings from permissive
+  runtimes and SHOULD replace invalid UTF-8 sequences (code units) with the
+  Unicode replacement character (`U+FFFD`).
 * Intermediary nodes (such as the OpenTelemetry Collector) are not required
   to perform in-line UTF-8 validation on `string` fields and MAY pass them
-  them through without inspection.
-* Implementations should be aware that many Protobuf runtimes are permissive and
-  allow non-compliant strings to be serialized, which may cause downstream nodes
-  using stricter Protobuf runtimes to fail deserialization. Implementations
-  which do NOT provide valid UTF-8 are not specification compliant, and this is
-  considered a bug within OpenTelemetry.
+  through without inspection.
 * Intermediary components MAY provide optional, configurable validation or
   sanitization.
-* Implementations SHOULD replace invalid UTF-8 sequences (code units) with
-  the Unicode replacement character (`U+FFFD`) when interacting with invalid
-  Unicode strings.
 
 ## Known Limitations
 
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -772,8 +772,9 @@ Consequently:
   considered a bug within OpenTelemetry.
 * Intermediary components MAY provide optional, configurable validation or
   sanitization.
-* We recommend replacing invalid UTF-8 with the unicode replacement character
-  (`U+FFFD`) when interacting with an invalid string.
+* Implementations SHOULD replace invalid UTF-8 sequences (code units) with
+  the Unicode replacement character (`U+FFFD`) when interacting with invalid
+  Unicode strings. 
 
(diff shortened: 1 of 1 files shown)


MrAlias (approved):

dashpole (approved):

florianl (approved):

docs/specification.md
```diff
@@ -746,6 +747,34 @@ payloads that contain zero spans, zero metric points or zero log records),
...
+  them through without inspection.
+* Implementations should be aware that many Protobuf runtimes are permissive and
+  allow non-compliant strings to be serialized, which may cause downstream nodes
+  using stricter Protobuf runtimes to fail deserialization. Implementations
+  which do NOT provide valid UTF-8 are not specification compliant, and this is
+  considered a bug within OpenTelemetry.
+* Intermediary components MAY provide optional, configurable validation or
+  sanitization.
+* We recommend replacing invalid UTF-8 with the unicode replacement character
+  (`U+FFFD`) when interacting with an invalid string.
```
pellared: Maybe normative language?
Also (nit) Unicode should be capitalized and a little more words to be more precise

```suggestion
* Implementations SHOULD replace invalid UTF-8 sequences (code units) with
  the Unicode replacement character (`U+FFFD`) when interacting with invalid
  Unicode strings. 
```

pellared (approved):
Just one suggestion. LGTM

iblancasa (approved):

jmacd (approved):

jsuereth:
Build failure is link-check on our markdown document linking to CNCF slack and getting permission denied.

Merging and investigating that failure out of band.

pellared:
@jsuereth, there are some problems with EasyCLA that prevent this PR from being merged.

docs/specification.md
```diff
@@ -746,6 +747,35 @@ payloads that contain zero spans, zero metric points or zero log records),
...
+  to perform in-line UTF-8 validation on `string` fields and MAY pass them
+  them through without inspection.
+* Implementations should be aware that many Protobuf runtimes are permissive and
+  allow non-compliant strings to be serialized, which may cause downstream nodes
+  using stricter Protobuf runtimes to fail deserialization. Implementations
+  which do NOT provide valid UTF-8 are not specification compliant, and this is
+  considered a bug within OpenTelemetry.
+* Intermediary components MAY provide optional, configurable validation or
+  sanitization.
+* Implementations SHOULD replace invalid UTF-8 sequences (code units) with
```
tigrannajaryan: We say "implementations" here, but we may want to distinguish encoder and decoder implementations.

For example we could say that encoders SHOULD (or even MUST) ensure UTF-8 validity, while decoders should be prepared to receive invalid UTF-8 sequences.

You already differentiate intermediaries (rightly so), so perhaps for clarity we should go a step further and list decoder and encoder requirements explicitly.

jsuereth: We could - but I think I would still consistently use SHOULD on both sides.

We know encoders cannot be MUST - because we allow using protobuf libraries and they are the ones failing to uphold that requirement in some cases / languages.  I.e. you can trigger the UTF-8 issue independently of any OTEL / OTLP if you wanted.

jsuereth: @tigrannajaryan do you consider this blocking? If so please propose verbage / wording and I'll take it.  Otherwise I'd like to merge this as an improvement over not stating anything about the issue in the spec.

tigrannajaryan: This particular phrase is unclear to me and I don't know if having it in unclear state is better than not having it. Does "replacement" requirement apply both to encoders and decoders?

jsuereth: I'll clarify - my focus here is on decoders being ready for invalid strings due to common practice.

jsuereth: Ok updated, PTAL

PR #848, opened 2026-08-21
profiles: Make non-zero duration being delta relationship explicit

brancz:


brancz:
Looks like all PRs are failing with this link check. Seems fine.

brancz:
@felixge @christos68k

jsuereth (approved):

jsuereth:
@brancz Yes - all PRs are failing - think we need to update the link or remove it.

brancz:
Sorry I should have added to the PR description that this was discussed in the last profiling working group call and we wanted to call this out explicitly.

aalexand (approved):

christos68k (approved):

opentelemetry/proto/profiles/v1development/profiles.proto
```diff
@@ -301,11 +301,13 @@ message Profile {
...
-  // to set time_unix_nano to the process start time and duration_nano to
-  // the relative time when the profile was gathered so that Sample.timestamps_unix_nano
-  // values fall within the profile time range.
+  // Duration of the profile in nanoseconds. For instant profiles like live
+  // heap snapshot, the duration can be zero but it may be preferable to set
+  // time_unix_nano to the process start time and duration_nano to the relative
+  // time when the profile was gathered so that Sample.timestamps_unix_nano
+  // values fall within the profile time range. If duration is larger than
+  // zero, then the samples describe a delta of occurrences in the
+  // [ time_unix_nano, time_unix_nano + duration_nano ) window.
```
florianl, line 310: What about rephrasing it to something like this?
```suggestion
  // Duration of the profile in nanoseconds.
  // If zero, the profile is an instant snapshot, like a live heap snapshot.
  // If non-zero, the profile has delta semantics, where sample values
  // reflect occurrences within [time_unix_nano, time_unix_nano + duration_nano),
  // not cumulative totals.
```

To me, the part `it may be preferable to set time_unix_nano to the process start time and duration_nano to the relative time when the profile was gathered so that Sample.timestamps_unix_nano values fall within the profile time range.` sounds to vage. What does `preferable` mean and how can users differentiate this `preferable` interpretation to other interpretations?

aalexand: The current comment is from [this discussion](https://docs.google.com/document/d/19UqPPPlGE83N37MhS93uRlxsP1_wGxQ33Qv6CDHaEp0/edit?tab=t.0#bookmark=id.an4px2jo7lgp) in the past.

The idea was to allow Sample timestamps (`Sample.timestamps_unix_nano`) in instantaneous profiles (like live heap snapshot) and also always maintain the Sample.timestamps_unix_nano ∈ [Profile.time_unix_nano, Profile.time_unix_nano+Profile.duration_nano) invariant. If we say that instantaneous profiles must have zero duration then technically it means that for such profiles the sample timestamps are either disallowed or they point outside the profile time range and we wanted to avoid this special casing.

aalexand: This is also how we check the timestamps in profcheck today - [here](https://github.com/open-telemetry/sig-profiling/blob/854821e3ec25f534fb9d814d71274af85fc398e8/profcheck/check.go#L144).

florianl: I'm open to further refinement, I suggest we follow our [previous discussion](https://docs.google.com/document/d/19UqPPPlGE83N37MhS93uRlxsP1_wGxQ33Qv6CDHaEp0/edit?tab=t.0#bookmark=id.an4px2jo7lgp) and use clear, relevant prefixes, such as `Profile.time_unix_nano`, `Profile.duration_nano`, and `Sample.timestamps_unix_nano` to prevent confusion. Additionally, to provide explicit guidance for implementers, we should replace vague phrasing like `[...] it may be preferable to set [...]` with more definitive language.

felixge: @florianl I like your version, especially if we add the protobuf message names in front of the fields as you suggested.

> The idea was to allow Sample timestamps (Sample.timestamps_unix_nano) in instantaneous profiles (like live heap snapshot) and also always maintain the Sample.timestamps_unix_nano ∈ [Profile.time_unix_nano, Profile.time_unix_nano+Profile.duration_nano) invariant. If we say that instantaneous profiles must have zero duration then technically it means that for such profiles the sample timestamps are either disallowed or they point outside the profile time range and we wanted to avoid this special casing.

@aalexand I don't see any use cases for populating `Sample.timestamps_unix_nano` for instantaneous profiles. I prefer to document the special case here rather than being more lax.

PR #849, opened 2026-08-21
docs: fix lychee link failure for slack channel

jsuereth:
Updates link to one that requires auth - also includes link to CNCF

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,7 +37,7 @@ OpenTelemetry follows the [Protobuf style guide](https://protobuf.dev/programmin
 
 ## Further Help
 
-If you have any questions or need assistance while contributing, feel free to reach out to the [`#otel-specification`](https://cloud-native.slack.com/archives/C01N7PP1THC) Slack channel.
+If you have any questions or need assistance while contributing, feel free to reach out to the [`#otel-specification`](https://app.slack.com/client/T08PSQ7BQ/C01N7PP1THC) channel on the [CNCF Slack](https://slack.cncf.io/).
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -37,7 +37,7 @@ OpenTelemetry follows the [Protobuf style guide](https://protobuf.dev/programmin
 
 ## Further Help
 
-If you have any questions or need assistance while contributing, feel free to reach out to the [`#otel-specification`](https://cloud-native.slack.com/archives/C01N7PP1THC) Slack channel.
+If you have any questions or need assistance while contributing, feel free to reach out to the [`#otel-specification`](https://app.slack.com/client/T08PSQ7BQ/C01N7PP1THC) channel on the [CNCF Slack](https://slack.cncf.io/).
```
jack-berg: Huh.. I recently had a link like this broken in otel-java and fixed it by deactivating lychee on slack links. The link still resolved for me in my browser, but was a 403 for lychee.

jack-berg (approved):

CONTRIBUTING.md
jsuereth: The link wouldn't resolve for me at all - but maybe it was an ephemeral thing?

jack-berg: The old link works for me 🤷 

Screenshot 2026-08-21 at 3 51 51 PM

jsuereth: :shrug: It now works for me - I'll update this.  It seems like perhaps slack / CNCF just generally had a problem where they were rejecting *some* traffic, leading to the build failures.

florianl (approved):
For me both links work, old and new. But I like the explicit use of `slack.cncf.io` 👍

MrAlias (approved):

CONTRIBUTING.md
jsuereth: Ok - reverted to previous url, exclude slack links now, but left new cncf slack link

PR #783, opened 2026-04-02
processcontext: add `ProcessContext` message from OTEP 4719

ivoanjo:
As https://github.com/open-telemetry/opentelemetry-specification/pull/4719 looks to be merged soon, it came up as we were implementing this spec in the OTel eBPF Profiler
(https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1181) that it'd be nice to stop copy-pasting the `process_context.proto` and it's time to add it to the proper place.

This is my first contribution to this repo so please do point out if I missed something! I didn't touch the collector parts since this message is not expected to be processed by the collector directly.

PR #854, opened 2026-08-27
Add `since <version>` requirement to new proto fields and backfill

### LINKED ISSUE #535: Specify version of the proto that introduced a field in the comments
The spec uses labels like `since 1.23.0` to indicate which version added the particular capability.

I wonder if we want to use a similar labeling in the form of comments when new capabilities are added to OTLP proto. This would make it easier for OTLP implementors to understand what new capabilities they need to implement. Implementations can also publish the version number they conform to and that in turn will allow users of OTLP implementations to understand which capabilities of OTLP the implementation supports.

An example of a field that would be labeled is Metric metadata [introduced here](https://github.com/open-telemetry/opentelemetry-proto/pull/514).

Important: this has nothing to do with having a runtime version number in the proto or elsewhere on the wire (which was a feature that we discussed in the past and decided not to add).

@dashpole @jsuereth and @open-telemetry/specs-approvers

jsuereth:
Fixes #535

- Updated CONTRIBUTING.md, RELEASING.md etc. to define this requirement
- Adds copilot instructions to do automatic code review based on CONTRIBUTING.md sections.
- Backfills `since`

Note: for *profiling* this is a little ugly since profiling never really released, but it was interesting seeing the history.  cc @florianl for thoughts

Also - `event_name` is an interesting one because I think we added it after dropping it.  I'm leaving the "since" for when it was restored.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,10 +19,12 @@ for general information about the project.
 ## Making Changes to the .proto Files
 
 When proposing schema changes or additions:
+
 1. Ensure all new messages, enums, enum values, and fields include a `// [Since next]` annotation in their leading doc comment (see [Version Annotations](#version-annotations-since-labels)).
 2. Follow the [Schema Evolution and Compatibility](#schema-evolution-and-compatibility) and [Style Guide](#style-guide) rules.
 3. Run `make gen-all` to regenerate implementation stubs across all supported target languages.
 4. Validate schema linting and backward compatibility before opening a pull request:
+
    ```bash
    make check
    ```
@@ -41,12 +43,14 @@ OpenTelemetry protocol definitions are the foundation of the OTLP ecosystem and 
 OpenTelemetry follows the [Protobuf style guide](https://protobuf.dev/programming-guides/style/) with the following clarifications:
 
 ### Naming Conventions
+
 - **Fields**: Use `snake_case` (e.g. `trace_id`, `dropped_attributes_count`).
 - **Messages & Enums**: Use `PascalCase` (e.g. `ResourceMetrics`, `SpanFlags`).
 - **Enum Values**: Use `UPPER_SNAKE_CASE` prefixed with the enum name (e.g. `SPAN_FLAGS_TRACE_FLAGS_MASK`).
 - **Service RPCs**: Use `PascalCase` for RPC methods (e.g. `Export`).
 
 ### Documentation Comments
+
 - All messages, enums, enum values, and fields MUST be documented via comments.
 - Field comments should document purpose or behavior with active verbs, or a simple definition noun phrase (similar to a dictionary entry).
   - valid: "Represents ..."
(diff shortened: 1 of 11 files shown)


opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -352,7 +358,10 @@ enum SpanFlags {
   // Bits 8 and 9 are used to indicate that the parent span or link span is remote.
   // Bit 8 (`HAS_IS_REMOTE`) indicates whether the value is known.
   // Bit 9 (`IS_REMOTE`) indicates whether the span or link is remote.
+  //
+  // [Since v1.5.0]
```
dashpole: These mask constants were introduced in PR #484 and released in `v1.2.0`. Consider updating the version label:
```suggestion
  // [Since v1.2.0]
```

opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -352,7 +358,10 @@ enum SpanFlags {
...
   SPAN_FLAGS_CONTEXT_HAS_IS_REMOTE_MASK = 0x00000100;
+  // [Since v1.5.0]
```
dashpole: Same here, this mask was released in `v1.2.0`:
```suggestion
  // [Since v1.2.0]
```

opentelemetry/proto/processcontext/v1development/process_context.proto
```diff
@@ -37,6 +37,8 @@ option go_package = "go.opentelemetry.io/proto/otlp/processcontext/v1development
 // for details of this mechanism.
 //
 // Status: [Development]
+//
+// [Since next]
```
dashpole: `ProcessContext` was introduced in PR #783 and released in `v1.11.0`. Should this be updated from `next` to `v1.11.0`?
```suggestion
// [Since v1.11.0]
```

opentelemetry/proto/trace/v1/trace.proto
jsuereth: Ah - I was using `git blame` in my script - You're right, needed to go deeper.

CONTRIBUTING.md
```diff
@@ -16,25 +16,65 @@ for general information about the project.
...
 
-## Style-Guide
+1. Ensure all new messages, enums, enum values, and fields include a `// [Since next]` annotation in their leading doc comment (see [Version Annotations](#version-annotations-since-labels)).
+2. Follow the [Schema Evolution and Compatibility](#schema-evolution-and-compatibility) and [Style Guide](#style-guide) rules.
+3. Run `make gen-all` to regenerate implementation stubs across all supported target languages.
+4. Validate schema linting and backward compatibility before opening a pull request:
 
-OpenTelemetry follows the [Protobuf style guide](https://protobuf.dev/programming-guides/style/) with the following clarifications:
+   ```bash
+   make check
```
dashpole: There is no `check` target in the `Makefile`.

jsuereth: You're right - I had a local addition of that, and then I realized it was just `make` due to default targets and removed it, cleaned it up.

dashpole (approved):

jmacd (approved):
👏

florianl (approved):
Sorry for the delayed response 🙏 
> [..]  for profiling this is a little ugly since profiling never really released

As the OTel Profiles signal is marked as `Alpha`, I think this is fine.

carlosalberto (approved):

PR #585, opened 2024-09-11
Rename profiles v1experimental to v1development

dmathieu:
As discussed in https://github.com/open-telemetry/opentelemetry-proto/pull/582#discussion_r1751850589, this renames the `v1experimental` version for profiles to `v1development`.

I've also taken this opportunity to update the versioning and stability link.

PR #857, opened 2026-09-01
profiles: specify OTLP Profiles development version metadata

Gandem:
## Changes

Defines a temporary development-version mechanism for OTLP Profiles `v1development` requests:

- Defines `otlp-profiles-development-version` request metadata for OTLP/gRPC and the corresponding `OTLP-Profiles-Development-Version` HTTP header.
- Uses positive integer versions, with missing metadata interpreted as version`1`.
- Documents the current version in `profiles.proto` and requires incrementing it once during preparation of a release containing one or more incompatible Profiles schema changes.
- Requires later-version clients to send exactly one version value and intermediaries forwarding an encoded payload to preserve it.
- Requires receivers to reject malformed, repeated, or unsupported versions before deserializing the request, using `INVALID_ARGUMENT` for OTLP/gRPC or `HTTP 400 Bad Request` for OTLP/HTTP.
- Requires removing the mechanism when Profiles moves to the stable `v1` service.

This PR defines the protocol behavior. Collector, SDK, and backend implementations will follow separately.

## Motivation

Profiles will continue using the same `v1development` package, service, and HTTP endpoint while incompatible Alpha and Beta schema changes remain possible. Because development formats share the same endpoint, a receiver cannot identify the expected schema before decoding the request. Protobuf decoding may still succeed while silently discarding or misinterpreting data, as could happen with changes such as https://github.com/open-telemetry/opentelemetry-proto/pull/786.

Carrying the development version outside the payload allows version-aware receivers to reject unsupported formats before decoding them.

This follows the direction discussed in https://github.com/open-telemetry/sig-profiling/issues/82. 

The detailed proposal and tradeoffs are documented in [Versioning OTLP Profiles](https://docs.google.com/document/d/1ZYf5CvxphaAXHzDxnC5PQr7Petn78-EImxjNiagwzF0/edit?tab=t.0#heading=h.xf6khaboen2a).

## Compatibility and rollout

Existing clients remain compatible because requests without the metadata are treated as version `1`.

Version-aware receivers should be deployed before clients begin sending a later version. Receivers that predate this mechanism may ignore the metadata and attempt to decode an incompatible payload. The linked proposal contains the detailed rollout sequence.

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -56,6 +56,7 @@ nodes such as collectors and telemetry backends.
     + [OTLP/HTTP Connection](#otlphttp-connection)
     + [OTLP/HTTP Concurrent Requests](#otlphttp-concurrent-requests)
     + [OTLP/HTTP Default Port](#otlphttp-default-port)
+- [Profiles Development Version](#profiles-development-version)
 - [Implementation Recommendations](#implementation-recommendations)
   * [Multi-Destination Exporting](#multi-destination-exporting)
   * [Empty Telemetry Envelopes](#empty-telemetry-envelopes)
@@ -714,6 +715,65 @@ connections SHOULD be configurable.
 #### OTLP/HTTP Default Port
 
 The default network port for OTLP/HTTP is 4318.
+
+## Profiles Development Version
+
+The Profiles signal uses the `v1development` package and service while
+incompatible changes to its Protobuf schema remain possible. To let a server
+identify the Profiles development format before deserializing a request, clients
+use temporary request metadata called the Profiles development version.
+
+This mechanism applies only to the
+`opentelemetry.proto.collector.profiles.v1development.ProfilesService` service
+and the `/v1development/profiles` HTTP endpoint. The version is carried as:
+
+* `otlp-profiles-development-version` request metadata for OTLP/gRPC.
+* The `OTLP-Profiles-Development-Version` request header for OTLP/HTTP.
+
+The version value MUST be a positive base-10 integer without leading zeros.
+Version `1` is the first Profiles development version. The current version is
+documented in
+[`profiles.proto`](../opentelemetry/proto/profiles/v1development/profiles.proto).
+
+The Profiles development version MUST be incremented by one for a schema change
+that could cause a server supporting the current version to fail to decode the
+request, lose information, or interpret the request incorrectly. Compatible
+changes MUST retain the current version.
+Changing the Profiles maturity level without such an incompatible change MUST
+NOT change the version.
+
+A client that serializes version `1` MAY omit the metadata; a client that
+serializes any later version MUST send it. When sent, the metadata MUST contain
+exactly one value assigned to that format. The value applies to the entire
+Export request.
+
+An intermediary that forwards an Export request without decoding and re-encoding
+its payload MUST preserve the Profiles development version metadata. If the
+metadata is removed, a request using a later version will be treated as version
+`1` and may be decoded incorrectly.
+
+A server MAY support one or more Profiles development versions. It MUST handle
+the metadata as follows:
+
+| Request metadata | Server behavior |
+| --- | --- |
+| Absent | Treat the request as version `1` |
+| Exactly one well-formed value | Use the value as the request version |
+| Malformed or repeated value | Reject the request before deserializing it |
+
+If the identified version is unsupported, the server MUST reject the request
+before deserializing it. A server MUST NOT substitute a different version or
+partially process an unsupported request. Rejections required by this section
+MUST use `INVALID_ARGUMENT` for OTLP/gRPC or `HTTP 400 Bad Request` for
+OTLP/HTTP. The client MUST NOT retry the same payload without the metadata or
+with a different version.
+
+Servers that predate this mechanism may ignore the metadata and attempt to
+process a request using an incompatible schema.
+
+The Profiles development version metadata MUST be removed when the Profiles
+signal becomes Stable and moves to the `v1` package, service, and HTTP endpoint.
+Clients and servers MUST NOT use this metadata with the stable Profiles service.
 
 ## Implementation Recommendations
 
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -735,15 +735,16 @@ Version `1` is the first Profiles development version. The current version is
 documented in
 [`profiles.proto`](../opentelemetry/proto/profiles/v1development/profiles.proto).
 
-Profiles development versions are assigned to `opentelemetry-proto` releases,
-not individual changes. During release preparation, the version MUST be
-incremented by one if the release contains one or more Profiles schema changes
-that could cause a server built from the preceding release to fail to decode the
-request, lose information, or interpret the request incorrectly. The version
-MUST be incremented at most once per release. A release containing only
-compatible Profiles schema changes MUST retain the current version. Changing the
-Profiles maturity level without an incompatible change MUST NOT change the
-version.
+The Profiles development version MUST be incremented by one for each
+incompatible Profiles schema change. A schema change is incompatible when it
+could cause a server supporting the current version to fail to decode the
+request, lose information, or interpret the request incorrectly. Each increment
+MUST update the version documented in
+[`profiles.proto`](../opentelemetry/proto/profiles/v1development/profiles.proto).
+Compatible changes MUST retain the current version. Changing the Profiles
+maturity level without an incompatible change MUST NOT change the version.
+Between `opentelemetry-proto` releases, the development version may increase
+by more than one, and gaps in the version sequence may occur.
 
 A client that serializes version `1` MAY omit the metadata; a client that
 serializes any later version MUST send it. When sent, the metadata MUST contain
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -735,12 +735,15 @@ Version `1` is the first Profiles development version. The current version is
 documented in
 [`profiles.proto`](../opentelemetry/proto/profiles/v1development/profiles.proto).
 
-The Profiles development version MUST be incremented by one for a schema change
-that could cause a server supporting the current version to fail to decode the
-request, lose information, or interpret the request incorrectly. Compatible
-changes MUST retain the current version.
-Changing the Profiles maturity level without such an incompatible change MUST
-NOT change the version.
+Profiles development versions are assigned to `opentelemetry-proto` releases,
+not individual changes. During release preparation, the version MUST be
+incremented by one if the release contains one or more Profiles schema changes
+that could cause a server built from the preceding release to fail to decode the
+request, lose information, or interpret the request incorrectly. The version
+MUST be incremented at most once per release. A release containing only
+compatible Profiles schema changes MUST retain the current version. Changing the
+Profiles maturity level without an incompatible change MUST NOT change the
+version.
 
 A client that serializes version `1` MAY omit the metadata; a client that
 serializes any later version MUST send it. When sent, the metadata MUST contain
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The full list of changes can be found in the compare view for the respective rel
 
 ### Added
 
+- profiles: add temporary OTLP Profiles development version metadata. [#857](https://github.com/open-telemetry/opentelemetry-proto/pull/857)
+
 ### Changed
 
 ### Fixed
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -43,8 +43,9 @@ option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
 
 // Current OTLP Profiles development version: 1.
 //
-// Increment this value by one when an incompatible Profiles schema change
-// requires a new development version. See
+// During release preparation, increment this value by one if the release
+// contains one or more incompatible Profiles schema changes since the preceding
+// release. Increment it at most once per release. See
 // docs/specification.md#profiles-development-version.
 
 //                Relationships Diagram
--- a/opentelemetry/proto/profiles/v1development/profiles.proto
+++ b/opentelemetry/proto/profiles/v1development/profiles.proto
@@ -41,6 +41,12 @@ option java_package = "io.opentelemetry.proto.profiles.v1development";
 option java_outer_classname = "ProfilesProto";
 option go_package = "go.opentelemetry.io/proto/otlp/profiles/v1development";
 
+// Current OTLP Profiles development version: 1.
+//
+// Increment this value by one when an incompatible Profiles schema change
+// requires a new development version. See
+// docs/specification.md#profiles-development-version.
+
 //                Relationships Diagram
 //
 // ┌──────────────────┐                      LEGEND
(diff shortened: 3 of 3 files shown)


docs/specification.md
```diff
@@ -715,6 +716,68 @@ connections SHOULD be configurable.
...
+
+Profiles development versions are assigned to `opentelemetry-proto` releases,
+not individual changes. During release preparation, the version MUST be
+incremented by one if the release contains one or more Profiles schema changes
+that could cause a server built from the preceding release to fail to decode the
+request, lose information, or interpret the request incorrectly. The version
+MUST be incremented at most once per release. A release containing only
+compatible Profiles schema changes MUST retain the current version. Changing the
+Profiles maturity level without an incompatible change MUST NOT change the
+version.
```
felixge: This seems brittle. What if somebody forgets to bump the profiles development version when cutting a release? I'm especially worried about this because not all people cutting proto releases may be part of the profiling SIG.

Why don't we simply require that each PR that makes a breaking change to profiles increments the version number? We only need to enforce this rule for members of the profiling SIG which is easy. And if somebody forgets to do it, we still have until the next release to fix it. Incrementing by more than 1 per proto release shouldn't cause any issues AFAICT.

felixge (changes requested):
LGTM for the most parts, but I found one issue that might be easy to fix.

docs/specification.md
Gandem: Thanks, I went back and forth on this one. I think that makes sense and bumping with each schema change will be easier to enforce. Updated in https://github.com/open-telemetry/opentelemetry-proto/pull/857/commits/db6508d38a0a6a4ad986e3f6833cf831c7a6a96c.

felixge (approved):
LGTM

docs/specification.md
```diff
@@ -715,6 +716,69 @@ connections SHOULD be configurable.
...
+
+The version value MUST be a positive base-10 integer without leading zeros.
+Version `1` is the first Profiles development version. The current version is
+documented in
+[`profiles.proto`](../opentelemetry/proto/profiles/v1development/profiles.proto).
+
+The Profiles development version MUST be incremented by one for each
+incompatible Profiles schema change. A schema change is incompatible when it
+could cause a server supporting the current version to fail to decode the
+request, lose information, or interpret the request incorrectly. Each increment
```
aalexand, line 741: Should "Each increment" be "Each incompatible Profiles schema change"?

Otherwise right now it reads as "Each increment must increment ..."

docs/specification.md
```diff
@@ -715,6 +716,69 @@ connections SHOULD be configurable.
...
+MUST update the version documented in
+[`profiles.proto`](../opentelemetry/proto/profiles/v1development/profiles.proto).
+Compatible changes MUST retain the current version. Changing the Profiles
+maturity level without an incompatible change MUST NOT change the version.
```
aalexand, line 745: Sorry, what is maturity level? Does it mean updating to Beta and so on?

I think this is also related to the "The Profiles development version metadata MUST be removed when the Profiles signal becomes Stable" part below, maybe combine?

docs/specification.md
```diff
@@ -715,6 +716,69 @@ connections SHOULD be configurable.
...
+
+A client that serializes version `1` MAY omit the metadata; a client that
+serializes any later version MUST send it. When sent, the metadata MUST contain
+exactly one value assigned to that format. The value applies to the entire
+Export request.
+
+An intermediary that forwards an Export request without decoding and re-encoding
+its payload MUST preserve the Profiles development version metadata. If the
+metadata is removed, a request using a later version will be treated as version
+`1` and may be decoded incorrectly.
```
aalexand, line 757: ... or rejected

?

aalexand (approved):

docs/specification.md
```diff
@@ -715,6 +716,69 @@ connections SHOULD be configurable.
 
 The default network port for OTLP/HTTP is 4318.
 
+## Profiles Development Version
+
+The Profiles signal uses the `v1development` package and service while
+incompatible changes to its Protobuf schema remain possible. To let a server
+identify the Profiles development format before deserializing a request, clients
+use temporary request metadata called the Profiles development version.
```
florianl, line 724: Should we phrase it in a way so that it is not mandatory for clients to set this version? At the moment it reads to me every client needs to set this metadata.

```suggestion
identify the Profiles development format before deserializing a request, clients can
use temporary request metadata called the Profiles development version.
```

docs/specification.md
```diff
@@ -715,6 +716,69 @@ connections SHOULD be configurable.
...
+
+* `otlp-profiles-development-version` request metadata for OTLP/gRPC.
+* The `OTLP-Profiles-Development-Version` request header for OTLP/HTTP.
+
+The version value MUST be a positive base-10 integer without leading zeros.
+Version `1` is the first Profiles development version. The current version is
+documented in
+[`profiles.proto`](../opentelemetry/proto/profiles/v1development/profiles.proto).
+
+The Profiles development version MUST be incremented by one for each
```
florianl, line 738: Who is responsible for the increment? The PR that introduces the incompatible schema change or is the person releasing proto responsible doing so?

docs/specification.md
Gandem: In the current proposal, lack of metadata means the request is sent with version 1 for backwards compatibility. Clients emitting version 1 may omit the metadata. Clients emitting a later version must set it otherwise the server may incorrectly parse the request or drop it.

IMHO the current wording covers both: the sentence doesn't clearly say it's mandatory or optional, and the specs makes it explicit later on.

docs/specification.md
Gandem: The PR that introduces the change, see https://github.com/open-telemetry/opentelemetry-proto/pull/857#pullrequestreview-5087193767.

docs/specification.md
```diff
@@ -715,6 +716,69 @@ connections SHOULD be configurable.
...
+| Absent | Treat the request as version `1` |
+| Exactly one well-formed value | Use the value as the request version |
+| Malformed or repeated value | Reject the request before deserializing it |
+
+If the identified version is unsupported, the server MUST reject the request
+before deserializing it. A server MUST NOT substitute a different version or
+partially process an unsupported request. Rejections required by this section
+MUST use `INVALID_ARGUMENT` for OTLP/gRPC or `HTTP 400 Bad Request` for
+OTLP/HTTP. The client MUST NOT retry the same payload without the metadata or
+with a different version.
```
christos68k, line 773: Not sure about the specificity of this. Maybe a little far-fetched but one failure can be the client sending v3 payload as v2 which a non-v2 supporting server will reject. Then the client can bump the version to v3 but keep the payload intact and complete a valid request.

```suggestion
OTLP/HTTP. The client MUST NOT retry the same payload without the metadata or the same payload with the same version.
```

christos68k (approved):
LGTM

florianl (approved):

rogercoll (approved):

jsuereth (approved):

docs/specification.md
```diff
@@ -715,6 +716,69 @@ connections SHOULD be configurable.
...
+maturity level without an incompatible change MUST NOT change the version.
+Between `opentelemetry-proto` releases, the development version may increase
+by more than one, and gaps in the version sequence may occur.
+
+A client that serializes version `1` MAY omit the metadata; a client that
+serializes any later version MUST send it. When sent, the metadata MUST contain
+exactly one value assigned to that format. The value applies to the entire
+Export request.
+
+An intermediary that forwards an Export request without decoding and re-encoding
```
tigrannajaryan, line 752: Do we consider Collector to be an intermediary? 

If yes specify how is this expected to work in the Collector? The Collector pipelines don't normally preserve payload headers as far as I know.

If not please call out how the Collector should deal with version numbers.

Do we expect Collector otlp receiver to support multiple versions of profile data at the same time or only one particular version will be supported?

Do we expect Collector otlp exporter to be able to encode to different versions of profile data (e.g. based on user config setting)?

tigrannajaryan:
> This PR defines the protocol behavior. Collector, SDK, and backend implementations will follow separately.

I have multiple questions about the Collector. Before this PR goes ahead I think we want to at least see answers and possibly also prototype implementations if the answers are not straightforward.

tigrannajaryan (changes requested):
Blocking temporarily until there is clarity around Collector.

docs/specification.md
tigrannajaryan: cc @open-telemetry/collector-maintainers

Gandem: > Do we consider Collector to be an intermediary?

The Collector is an intermediary, but it doesn't forward requests without decoding and re-encoding their payloads, which is what this requirement applies to.

Under this proposal, the Collector’s receiver would validate the incoming version before decoding, and its exporter would independently set the version corresponding to the schema it serializes. The version metadata wouldn’t need to be preserved through the pipeline.

I can make the distinction explicit in the spec.

Gandem: > Do we expect Collector otlp receiver to support multiple versions of profile data at the same time or only one particular version will be supported?
> Do we expect Collector otlp exporter to be able to encode to different versions of profile data (e.g. based on user config setting)?

Our intent is to have the collector support a single version of the protocol at a time (both for receiving the data, and exporting it) and reject every other version for simplicity. The version it supports would be tied to the version that is exported in the current `opentelemetry-proto` release used by the collector, and would not be configurable.

I'm happy to explicitly call this out in the spec if that helps.

tigrannajaryan:
@jsuereth in the call today I think you said the PR specifies how the Collector works, but I am not able to find it in the PR. Can you please point it to me?

docs/specification.md
pellared, line 724: Do we have some documented reasoning and agreement that rejects using e.g. `opentelemetry.proto.profiles.v2development` namespace instead? Would then the breaking changes be more explicit and easier to follow? Would it also not be easier in stabilization so that we accidentally do not keep fields that are not used in the latest version?

jmacd (approved):

docs/specification.md
Gandem: Agreed, simplified this in 4f6e452.

docs/specification.md
Gandem: Yes I meant transitions from Alpha to Beta. Removed the sentence in 6f66192: we already enforce that compatible changes must retain the current version, and the final paragraph separately covers removing the mechanism when Profiles becomes Stable.

docs/specification.md
Gandem: Good point, added “rejected or decoded incorrectly” in 90e7e14.

docs/specification.md
Gandem: Good point! Updated in 5e97d66 to prohibit retrying “with a version that does not match the payload’s format,” instead of any different version. This allows the example above, but still prevents relabelling an incompatible payload as a supported version (sending a v2 payload labeled as v3).

docs/specification.md
Gandem: @tigrannajaryan I clarified the Collector behavior as an intermediary in 9197a0f, PTAL 🙏 

I also drafted a [Collector prototype](https://github.com/open-telemetry/opentelemetry-collector/pull/16009) that supports a single non-configurable version for receiving and exporting payloads.

I left the single-version restriction out of the spec since it’s a Collector implementation choice rather than a protocol requirement, but I’m happy to include it if you think it belongs here.

Gandem:
@tigrannajaryan I’ve responded to your feedback, updated the spec and linked a Collector prototype in https://github.com/open-telemetry/opentelemetry-proto/pull/857#discussion_r4075745046 (link to the Collector prototype: https://github.com/open-telemetry/opentelemetry-collector/pull/16009). Could you take another look when you have a chance? Thanks! 🙏

docs/specification.md
tigrannajaryan: Sounds good to me. No need to add to the spec.

tigrannajaryan (approved):

tigrannajaryan:
Proposed changes LGTM, including proposed Collector behavior.

I would still want to see a @open-telemetry/collector-maintainers approval based on the [prototype](https://github.com/open-telemetry/opentelemetry-collector/pull/16009) implementation and then we can merge.

dmitryax (approved):

docs/specification.md
Gandem: Agreed, changing the namespace would make the breaking changes more explicit.

Since we expect only a few breaking changes before stable, the idea was to avoid making everyone currently using the signal change packages, imports, and HTTP paths/gRPC services each time (which can also mean updating server-side routing rules or allowlists). 

We already went through this with the `v1experimental` → `v1development` rename (#585), and it was extra work for existing implementations.  To give an idea of where another rename would have an impact:
- [References within OpenTelemetry.](https://github.com/search?q=org%3Aopen-telemetry+%28%22profiles.v1development%22+OR+%22v1development.profiles%22+OR+%22profiles%2Fv1development%22+OR+%22v1development%2Fprofiles%22%29&type=code)
- [References in the wider ecosystem.](https://github.com/search?q=%28%22profiles.v1development%22+OR+%22v1development.profiles%22+OR+%22profiles%2Fv1development%22+OR+%22v1development%2Fprofiles%22%29&type=code)
- A concrete example: [Fluent Bit still explicitly handles both the old and new service paths.](https://github.com/fluent/fluent-bit/blob/be271c25b9b06f3e840740e9b9c4014d454dea37/plugins/in_opentelemetry/opentelemetry_prot.c#L69-L70)

Clients and servers still need to update their schemas, but we wanted to avoid adding this extra migration work on top of that.

To give some context, this builds on the earlier SIG decision to keep `v1development` through Alpha/Beta, then move to `v1` at release candidate (https://github.com/open-telemetry/sig-profiling/issues/70, https://github.com/open-telemetry/opentelemetry-proto/pull/771). We captured the trade-off in [the original working document, under “Alternatives Considered”](https://docs.google.com/document/d/1ZYf5CvxphaAXHzDxnC5PQr7Petn78-EImxjNiagwzF0/edit?tab=t.0#heading=h.sykqsk33vncy).

On unused fields, I’m not sure I understand the difference you have in mind. Our intent is to remove them when needed and increment the development version. We’d still need to check the schema carefully before moving to v1 either way.

Ultimately, it’s not perfect, but we think it’s a reasonable trade-off for the few changes we expect before stable. It lets us avoid extra migrations while people keep building and testing Profiles, and goes away when we move to v1. Let me know if that makes sense!

florianl: By renaming `v1experimental` to `v1development`, the Profiling SIG aligned with the maturity level specification outlined in [OTEP 232](https://github.com/open-telemetry/oteps/blob/main/text/0232-maturity-of-otel.md). Although the [versioning and stability document](https://opentelemetry.io/docs/specs/otel/versioning-and-stability/) does not explicitly address the `vX` prefix used before maturity levels, my understanding is that this prefix denotes the target stable version (e.g., a stable `v1`).

Transitioning from `v1development` to `v2development` to accommodate breaking changes, which are expected during the development stage, would challenge this assumption and create a naming conflict once the overall project eventually advances to `v2`.

Gandem:
@tigrannajaryan @jsuereth it looks like we have approval from a Collector maintainer here and multiple approvals on the [Collector PR](https://github.com/open-telemetry/opentelemetry-collector/pull/16009). Is there anything else we need to address before merging? Thanks! 🙏

tigrannajaryan:
> Is there anything else we need to address before merging?

LGTM, but before merging all open conversations must be addressed and resolved.

docs/specification.md
Gandem: @pellared does the explanation above address your concerns, or is there anything else you’d like us to clarify before resolving this thread? Thanks! 🙏

PR #865, opened 2026-09-15
feat: init policy spec

jaronoff97:


experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+
+```yaml
+id: keep-spans-with-exceptions
+name: Keep spans with exception events
+trace_sampling:
+  matches:
+    - target: { record_field: SPAN_RECORD_FIELD_EVENT_NAME }
+      predicate: { equals: exception }
+  sampling:
+    percentage: 100.0
```
dashpole: Since unmatched spans pass through by default and overlapping `TraceSamplingPolicy` rules take the lowest percentage (`min`), a policy with `percentage: 100.0` is a mathematical no-op that loses to any sampling policy with `< 100%` (meaning exception spans on sampled routes will still be dropped). To support span exemptions like GCP's `TraceFilterPolicy` (`ACTION_KEEP`), either deterministic span filtering is needed in the filter stage before sampling, or `TraceSamplingPolicy` needs an explicit exemption mechanism rather than relying on `percentage: 100.0`.

opentelemetry/proto/policies/v1development/traces.proto
```diff
@@ -0,0 +1,298 @@
...
+//
+// [Since next]
+message SamplingConfig {
+  // The share of matching spans to keep, from 0 to 100.
+  //
+  // A value of 0 drops every matching span. A value of 100 or more keeps every
+  // matching span. A value outside the range 0 to 100 is invalid.
+  //
+  // [Since next]
+  float percentage = 1;
```
dashpole: Saying a value of `100 or more` keeps every span contradicts the next sentence (and the compilation error table in `policy.md`) which marks any value outside `0` to `100` as invalid and inert. Also, `float` only has 24 bits of mantissa precision (~6 hex digits), which cannot represent 56-bit thresholds when `sampling_precision` is set up to `14` hex digits, so `double` should be used here instead.

opentelemetry/proto/policies/v1development/policies.proto
```diff
@@ -0,0 +1,111 @@
...
+option java_package = "io.opentelemetry.proto.policies.v1development";
+option java_outer_classname = "PoliciesProto";
+option go_package = "go.opentelemetry.io/proto/otlp/policies/v1development";
+
+// Policy carries one policy definition together with the metadata that a
+// control plane and an operator need.
+//
+// The envelope is optional. A transport that already supplies an identity and
+// a lifecycle, such as an xDS resource or an OpAMP custom message, MAY carry a
+// bare policy message instead.
```
dashpole: If a transport sends bare policy messages (`LogFilterPolicy`, `LogTransformPolicy`, etc.) without the `Policy` envelope, the unmarshaled protobuf messages have no `id` field. That breaks both `policy.evaluated.items` match tracking (which requires the `policy.id` attribute) and deterministic tie-breaking in `LogTransformPolicy` unless `id` is included on the bare policy messages as in GCP's `v1alpha1` schema.

opentelemetry/proto/policies/v1development/logs.proto
```diff
@@ -0,0 +1,299 @@
...
+message LogRename {
+  // The field to rename. A field that is absent yields no change.
+  //
+  // [Since next]
+  LogFieldSelector from = 1;
+
+  // The new key. An empty key is invalid.
+  //
+  // [Since next]
+  string to = 2;
```
dashpole: Reusing `LogFieldSelector` for `from` allows selecting fixed OTLP struct fields like `record_field` or `scope_field` (`SEVERITY_NUMBER`, `TRACE_ID`, `SCOPE_FIELD_NAME`) that cannot be renamed to an arbitrary string key, and having `to` as a flat `string` rather than an `AttributePath` makes it ambiguous whether renaming a nested attribute (`["http", "request", "method"]`) renames the leaf key in place or moves it to the root. Similarly, `LogRedact` and `LogAdd` should specify behavior or validation rules when targeting non-string first-class fields like `SEVERITY_NUMBER` or `TRACE_ID`.

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+apply any one of them.
+
+A span therefore has no keep-overrides-drop rule. A `percentage` of 100 states
+an intent to keep, but a policy with a lower percentage still wins, because a
+higher percentage has no sound claim over a lower one.
+
+**Transform stage.** Every matching transform policy contributes its operations.
+The operations run in the order `remove`, `redact`, `rename`, `add` across the
+union of all matching policies. When two policies target one field within one
+operation type, the result is implementation-defined but MUST be deterministic.
```
dashpole: Leaving transform conflicts implementation-defined means an SDK and Collector given the same `PolicySet` can produce different outputs (violating OTEP 4738's cross-implementation determinism requirement), and winner-takes-all tie-breaking on `redact` would skip one of the redactions and leak sensitive data. Instead, operations within each stage (`remove`, `redact`, `rename`, `add`) should execute sequentially across all matching policies in ascending lexicographical order of `(policy.id, operation_index)` so multiple redactions compose predictably.

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+- `miss` when another policy overrides that intent.
+
+Under the precedence rules above:
+
+| Situation                                              | `outcome = hit`                 | `outcome = miss`                      |
+| ------------------------------------------------------ | ------------------------------- | ------------------------------------- |
+| A keep policy and a drop policy match. Record is kept. | The keep policy.                | The drop policy.                      |
+| Only drop policies match. Record is dropped.           | Every drop policy.              | None.                                 |
+| A sampling decision keeps the span.                    | Every matching sampling policy. | None.                                 |
+| A sampling decision drops the span.                    | The most restrictive policy.    | Every other matching sampling policy. |
```
dashpole: This table and section omit `LogTransformPolicy` and the case where only `ACTION_KEEP` filter policies match a record. Please specify how `outcome` (`hit` vs `miss`) is recorded for transform policies (for example, whether a transform policy matching a record dropped in the filter stage is skipped or counted as `miss`, and whether `hit` requires modifying at least one field).

opentelemetry/proto/policies/v1development/common.proto
```diff
@@ -0,0 +1,246 @@
...
+    // OpenTelemetry data model, the target is present when the field holds a
+    // non-default value. Pair exists with negate to test for absence.
+    //
+    // [Since next]
+    google.protobuf.Empty exists = 1;
+
+    // Matches when the target value equals this typed scalar.
+    //
+    // An integer and a double are compared in one numeric domain, so an
+    // int_value MAY match a double target with an equal numeric value, and the
```
dashpole: Since this spec adopts RFC 2119 keywords, using `MAY` here makes cross-type numeric equality between `int_value` and `double_value` optional across implementations. Please change `MAY` to `MUST` (and note that comparisons should preserve full 64-bit integer precision rather than naively casting `int64` to `double`).

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+```
+
+### Granularity
+
+The matchers set the granularity:
+
+- **Instrument level.** When every matcher selects `descriptor_field`,
+  `resource_attribute`, `scope_attribute`, or `scope_field`, the action applies
+  to the whole metric stream.
+- **Data point level.** When any matcher selects `datapoint_attribute`, the
```
dashpole: It would be helpful to clarify how `ACTION_KEEP` overrides `ACTION_DROP` when an instrument-level policy conflicts with a datapoint-level policy on the same metric (for example, an instrument-level `ACTION_DROP` vs a datapoint-level `ACTION_KEEP`). Defining precedence per data point (where an instrument-level policy applies its action to every data point in the stream, and the metric container is dropped if all its data points are dropped) makes this interaction unambiguous and aligns with match tracking.

opentelemetry/proto/policies/v1development/logs.proto
```diff
@@ -0,0 +1,299 @@
...
+  // [Since next]
+  LOG_RECORD_FIELD_SEVERITY_NUMBER = 3;
+
+  // The trace id of the log record.
+  //
+  // A predicate reads the canonical 32-character lowercase hexadecimal text,
+  // as the W3C Trace Context specification defines it.
+  //
+  // [Since next]
+  LOG_RECORD_FIELD_TRACE_ID = 4;
```
dashpole: In OTLP and W3C Trace Context (as well as fixed-size byte array implementations like Go Collector's `pcommon.TraceID`), absent trace and span IDs are often represented as all-zero byte sequences (`16` or `8` zero bytes) rather than empty slices. Please specify here and in `policy.md` that all-zero trace and span IDs evaluate to `exists: false` and read as the empty string `""` (matching `SPAN_RECORD_FIELD_PARENT_SPAN_ID` on root spans and GCP's `trace_filter_policy.proto`), rather than matching `"00000000000000000000000000000000"`.

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+When a consumer reads YAML or JSON it MUST accept scalar shorthand. The type of
+the literal selects the variant:
+
+```yaml
+- log_attribute: ["http.response.status_code"]
+  gte: 500 # int_value
+- log_attribute: ["sampling.ratio"]
+  lt: 0.5 # double_value
+- log_attribute: ["cache.hit"]
+  equals: true # bool_value
```
Logiraptor, line 345: In JSON encoding, I think we'll have an ambiguity here. It seems we can use the non-shorthand form just fine, but it may be worth specifying what happens with `"equals": "true"` read from JSON

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+| `$$`             | A literal dollar sign.    |
+
+A reference to an absent capture group MUST expand to the empty string.
+
+### Rename
+
+```
+LogRename {
+  from:   LogFieldSelector  // REQUIRED
+  to:     string            // REQUIRED, non-empty
```
Logiraptor, line 487: Any reason to keep `to` here as a `string` instead of `LogFieldSelector`? I'm assuming it's because some moves would be ambiguous, like moving from a record to a scope attribute (which scope, etc). 

Again assuming it's intentional, can we specify how to is to be interpreted? Relative to from, I presume

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+
+**Transform stage.** Every matching transform policy contributes its operations.
+The operations run in the order `remove`, `redact`, `rename`, `add` across the
+union of all matching policies. When two policies target one field within one
+operation type, the result is implementation-defined but MUST be deterministic.
+A consumer MUST break the tie in a stable order, for example by policy
+identifier.
+
+This rule set is commutative: the outcome does not depend on the order of the
+policy set.
```
Logiraptor, line 705: This is the part that I'm most unsure about. If the goal of this spec is to establish a _portable_ policy system, it seems we'd want to have deterministic behavior when porting policy application between consumers.

With the rules here, it's also undefined what should happen given the following policies:

```
P1:
  rename: a -> b

P2:
  rename: b -> c
```

Given an input of `{a: "A"}`, we could end up with either `{c: "A"}` or `{b: "A"}` depending on application order. These policies do not target one field within one operation type, so there is no defined output.

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+A span therefore has no keep-overrides-drop rule. A `percentage` of 100 states
+an intent to keep, but a policy with a lower percentage still wins, because a
+higher percentage has no sound claim over a lower one.
+
+**Transform stage.** Every matching transform policy contributes its operations.
+The operations run in the order `remove`, `redact`, `rename`, `add` across the
+union of all matching policies. When two policies target one field within one
+operation type, the result is implementation-defined but MUST be deterministic.
+A consumer MUST break the tie in a stable order, for example by policy
+identifier.
```
Logiraptor, line 702: Similar to my other comment, there is another unspecified case:

```
P1:
  rename: a -> x
  upsert: false

P2:
  rename: b -> x
  upsert: false
```

Depending on execution order, this may result in either a or b being renamed. Does "one field within one operation type" include the case where `to` is the same?

Personally I'd be in favor of defining a canonical ordering somehow (or else requiring policies to be shipped as a user-ordered set in all cases)

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+
+### Match Tracking
+
+A consumer SHOULD report two counters, so that an operator can see how much
+telemetry a policy set saw and whether each policy has the effect it intends.
+
+| Counter                  | Counts                                     |
+| ------------------------ | ------------------------------------------ |
+| `policy.consumed.items`  | Every datum that enters policy evaluation. |
+| `policy.evaluated.items` | Every datum that a policy matched.         |
```
Logiraptor, line 715: Should these be `evaluated` and `matched` instead, to more directly describe what they track?

Logiraptor (commented):
Overall very happy to see this progressing - left a few comments on things that seem unclear

experimental-docs/specifications/policy.md
jaronoff97: yeah this was designed to move an attribute so maybe we should filter it to just to other attributes

experimental-docs/specifications/policy.md
jaronoff97: yeah i was doing this deterministically by using sorted alphanumeric priorities, but open to other ideas.

experimental-docs/specifications/policy.md
dashpole, line 926: Under the evaluation rules in this spec, unmatched spans pass through by default (100% kept) and overlapping TraceSamplingPolicy rules apply the lowest percentage (min). That means a policy with percentage: 100.0 is a no-op (min(100, x) = x) that can never cause a span to be kept that wouldn't otherwise be kept. In this example, if a database span with an exception also matches sample-database-spans (percentage: 5.0) below, min(100.0, 5.0) evaluates to 5.0% and 95% of exception spans are dropped, so this example does not actually keep spans with exceptions.

opentelemetry/proto/policies/v1development/policies.proto
dashpole, line 35: Could we make the Policy envelope mandatory across all transports so Policy is the single canonical wire unit? Consistency seems nice... Is there a downside?

opentelemetry/proto/policies/v1development/traces.proto
dashpole, line 74: Nit: saying a value of 100 or more keeps every matching span contradicts the next sentence (and the compilation error table in policy.md) which marks any value outside 0 to 100 (such as 150) as invalid and inert. Also, consider using double instead of float here for better precision.

opentelemetry/proto/policies/v1development/common.proto
dashpole, line 140: Using MAY here makes cross-type numeric equality between int_value and double_value optional across implementations. Could we change MAY to MUST (and note that comparisons should preserve full 64-bit integer precision rather than naively casting int64 to double)?

opentelemetry/proto/policies/v1development/logs.proto
dashpole, line 285: In OTLP and W3C Trace Context, absent trace and span IDs are frequently represented as all-zero byte sequences (16 zero bytes for trace_id, 8 zero bytes for span_id) rather than empty byte slices—for example, Go Collector pdata represents TraceID and SpanID as fixed-size [16]byte and [8]byte arrays in memory where unset IDs are all zeroes. Could we explicitly specify here and in policy.md that all-zero byte sequences MUST be treated as absent (evaluating exists to false and reading as empty text "" rather than "00000000000000000000000000000000")?

experimental-docs/specifications/policy.md
dashpole, line 758: This lookup table for outcome = hit | miss covers filter drop/keep conflicts and sampling, but omits two cases: (1) when only ACTION_KEEP filter policies match a record (presumably outcome = hit for all matching keep policies), and (2) how outcome = hit | miss is defined for LogTransformPolicy (for example, whether a transform policy matching a record dropped in the Filter stage is skipped or recorded as miss, and whether hit requires modifying at least one field).

experimental-docs/specifications/policy.md
dashpole, line 528: Since this spec adopts ACTION_KEEP overrides ACTION_DROP and defines both instrument-level and datapoint-level granularities for MetricFilterPolicy, what happens when an instrument-level ACTION_DROP matches a metric stream while a datapoint-level ACTION_KEEP matches specific data points within that stream? We should clarify that filter precedence resolves per data point (where an instrument-level policy applies its action to every data point in the stream, and the metric container is dropped if all its data points are dropped) to make this interaction unambiguous and align with match tracking.

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+A policy MUST carry at least one operation.
+
+### Execution Order
+
+A consumer MUST apply the operations in this order and MUST NOT reorder them:
+
+1. **Remove.** Delete the field.
+2. **Redact.** Mask the field value.
+3. **Rename.** Change the field key.
+4. **Add.** Insert the field.
```
dashpole, line 435: Running remove and redact before rename and add creates two ordering pitfalls: (1) any attributes inserted by add bypass remove and redact entirely, so a broad PII sanitization policy cannot redact or strip sensitive values introduced by an add rule, and (2) redact runs before rename, so if a redaction policy targets a canonical attribute name (like http.request.header.authorization), it will miss legacy attribute names that haven't been normalized by rename yet. Should we change the execution order to rename -> add -> remove -> redact so normalization happens first and sanitization/deletion is guaranteed to run last?

experimental-docs/specifications/policy.md
dashpole: Leaving the tie-breaking order implementation-defined means a Go Collector and a Java SDK evaluating the exact same policy set on the same telemetry can produce different output. Since policy portability across implementations is a core goal of OTEP 4738, could we mandate a specific canonical ordering here, such as lexicographical UTF-8 order by policy.id (and by operation index within the same policy)?

experimental-docs/specifications/policy.md
dashpole: Nit on phrasing: saying a consumer MUST break the tie suggests winner-takes-all selection where only one policy's operation runs. For operations like redact on body, if one policy masks email addresses and another masks credit card numbers, both operations should execute sequentially in deterministic order so both patterns are redacted rather than one overriding the other. Could we clarify that all matching operations within a stage execute sequentially in deterministic order?

opentelemetry/proto/policies/v1development/logs.proto
```diff
@@ -0,0 +1,299 @@
...
+    // An instrumentation scope attribute.
+    //
+    // [Since next]
+    AttributePath scope_attribute = 4;
+
+    // A first-class field of the instrumentation scope.
+    //
+    // [Since next]
+    ScopeField scope_field = 5;
+  }
```
dashpole, line 150: In OTLP, LogRecord.body is an AnyValue and frequently holds a structured map (kvlist_value) for JSON logs rather than a flat string. Right now, selecting LOG_RECORD_FIELD_BODY only targets the entire root body, so a policy cannot remove, redact, or rename an individual nested field inside a structured body (like body.user.ssn) without wiping or regex-replacing the entire body. This could definitely be added as a follow-up, but it would be great to support AttributePath traversal on body (for example, adding AttributePath body_field = 6 or similar to LogFieldSelector).

opentelemetry/proto/policies/v1development/logs.proto
dashpole: Using LogFieldSelector for from allows targeting fixed OTLP struct fields like LOG_RECORD_FIELD_SEVERITY_NUMBER or SCOPE_FIELD_NAME, which cannot have their keys renamed. In addition, having from accept a nested AttributePath while to is a flat string makes nested renames ambiguous: it is unclear whether renaming ["http", "request", "method"] to "http.request.method" renames only the leaf key inside the nested map or flattens it to the root attribute map. Could we introduce a dedicated LogAttributeSelector (containing only log_attribute, resource_attribute, and scope_attribute) and use it (or AttributePath) for both from and to? Since AttributePath already supports string shorthand in YAML, flat renames would keep the exact same YAML syntax while supporting both nested-to-flat and nested-to-nested renames unambiguously.

experimental-docs/specifications/policy.md
dashpole: +1 to filtering this to just attributes.  Also, having `from` accept a nested `AttributePath`, while `to` is a flat string makes it ambiguous whether you are trying to rename a leaf or flatten it.

Could we introduce a `LogAttributeSelector` and use it for both `from` and `to`?

experimental-docs/specifications/policy.md
dashpole: I would also like to see the spec define a canonical ordering. Alphanumeric sorting on policy.id works for me.

experimental-docs/specifications/policy.md
dashpole: I agree that canonical ordering would help here.

dashpole (commented):
Thanks for starting this! Looking good, but lots of details to work through :)

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+| `contains` | `Value`        | The target value holds the scalar. See below.         |
+
+#### **`exists`**
+
+For an attribute, the target is present when the key exists. For a first-class
+field that carries no presence information in the OpenTelemetry data model, the
+target is present when the field holds a non-default value. To test for absence,
+pair `exists` with `negate: true`.
+
+#### **`regex`**
```
anuraaga, line 271: I would recommend against regex, for cross-language SDKs, using an engine that supports re2 syntax can be difficult. Declarative configuration uses glob syntax for this reason and I think fits better for an SDK concern. It's unfortunate but the reality.

Incidentally I wonder about the overlap between declarative configuration and the policy documents, should they be the same schema where possible? /cc @jack-berg

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+  matches:  TraceMatcher[]  // REQUIRED, at least one
+  sampling: SamplingConfig  // REQUIRED
+}
+
+SamplingConfig {
+  percentage:         float  // REQUIRED, 0 to 100
+  mode:               SamplingMode  // OPTIONAL, defaults to hash_seed
+  sampling_precision: uint32 // OPTIONAL, defaults to 4, range 1 to 14
+  hash_seed:          uint32 // OPTIONAL, defaults to 0
+  fail_closed:        bool   // OPTIONAL, defaults to true
```
anuraaga, line 573: I think an `action` enum with KEEP and DROP similar to log policy, or maybe this should follow SamplingDecision in the SDK. Either way, there isn't really a failure happening here

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+
+`negate` inverts the whole matcher, not the quantifier. A matcher with
+`event_name`, `equals: "exception"`, and `negate: true` therefore reports true
+for a span that carries no exception event.
+
+### Sampling Modes
+
+| Mode                         | Description                                                                                                                                              |
+| ---------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `SAMPLING_MODE_HASH_SEED`    | Hashes the trace id with `hash_seed` for a deterministic decision. Every consumer in one tier that shares a seed reaches the same decision. The default. |
```
anuraaga, line 633: Need to specify the hashing mechanism. And I guess if hash_seed is set to `0` the trace ID should be used as is? For me it's a bit confusing how this interacts with the normal probability sampling

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+### Runtime Errors
+
+A runtime error occurs while a consumer applies a compiled, valid policy to one
+record. The record MUST pass through unmodified by that policy. A runtime error
+MUST NOT cause telemetry loss.
+
+A type mismatch is not a runtime error. It is a non-match. See
+[Predicates](#predicates).
+
+## YAML Representation
```
anuraaga, line 833: While I think policy servers will generally need to support YAML, does it actually need to be supported in the SDK protocol? SDKs will be much simpler if they only need to use the proto binary / JSON formats just like OTLP, notably support for YAML can be spotty i.e. in Go.

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+place different demands on that engine.
+
+| Field             | Needs                                                                                                    |
+| ----------------- | -------------------------------------------------------------------------------------------------------- |
+| `Predicate.regex` | One boolean answer: does the pattern match anywhere in the value.                                        |
+| `LogRedact.regex` | The range of every non-overlapping match, plus each capture group that the `replacement` template reads. |
+
+A consumer with a large policy set will scan one value against many patterns. A
+multiple-pattern engine, such as
+[Hyperscan](https://intel.github.io/hyperscan/dev-reference/), answers every
```
anuraaga, line 986: I really don't see SDKs taking on a dependency on hyperscan

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+`LogTransformPolicy`, `MetricFilterPolicy`, and `TraceSamplingPolicy`. A policy
+carries one matcher list and one action.
+
+**Tradeoff.** Four wire types instead of one per signal. A single record that
+needs both a filter rule and a transform rule now needs two policies, and the
+matcher list is duplicated between them. In exchange, each policy stays small,
+its validity is decidable on its own, and a new intent lands as a new type
+rather than as a new mode inside an existing one.
+
+### The envelope is optional
```
anuraaga, line 1040: There seems to be no reason for this and it can result in bugs are misunderstandings in SDK implementation. May as well define the envelope proto so SDKs just use it - like the YAML point above, services can choose to provide configuration UI with single documents, but that doesn't need to be reflected into the protocol which needs to be implemented across all SDKs

experimental-docs/specifications/policy.md
```diff
@@ -0,0 +1,1292 @@
...
+
+### The transport is out of scope
+
+**Decision.** This document defines the policy schema and its evaluation
+semantics. It does not define how a consumer receives a policy set or how it
+reports the results.
+
+**Tradeoff.** A conformant consumer still needs a transport binding, and
+[Match Tracking](#match-tracking) defines counters with no channel to report
+them on. Binding the schema to one transport, whether xDS, OpAMP, or a dedicated
```
anuraaga, line 1259: I feel it can be agnostic while still speccing out the "standard transport" which is presumably OpAMP since this is OTel. Otherwise, there's nothing to guide SDK implementations

anuraaga (commented):
Most feedback is from https://github.com/open-telemetry/opentelemetry-specification/issues/5287 but in a general sense the spec seems strong in specifying schema details, but in the end the spec is a document to read while implementing in SDKs, that's basically it. I think in general there can be some reworking for the spec to be "SDK-first" rather than "data model-first".

@jack-berg with your context from declarative configuration, I suspect you are in the best position to provide the technical vision for fleshing this doc out as an OTel spec (if you have the bandwidth of course). Will be great to get your thoughts.

opentelemetry/proto/policies/v1development/common.proto
```diff
@@ -0,0 +1,246 @@
...
+//
+// [Since next]
+message AttributePath {
+  // Path segments, applied left to right.
+  //
+  // [Since next]
+  repeated string path = 1;
+}
+
+// Value carries one typed scalar for the equals and contains predicates.
```
jsuereth, line 44: Should this exactly reference OTLP AnyValue?

I believe we should mirror some choices done for ProcessContext - and if this is intended to be the same value you'd get in OTLP - we should directly use the value from OTLP (another point about being in the same repository).

opentelemetry/proto/policies/v1development/common.proto
jsuereth: +1 - I feel like we should take some time to fully specify this out.

opentelemetry/proto/policies/v1development/policies.proto
jsuereth: I actually think there's value in the envelope too.

Think self-reporting metrics on policy application, e.g.

experimental-docs/specifications/policy.md
jsuereth: I'm also a bit concerned around the "half-typed" nature of the selector, e.g. switching between lookup-by-string and lookup-by-index and policy owner needs to know.

That's not so bad for read-only, that's a lot worse for write-paths.

experimental-docs/specifications/policy.md
jsuereth: Rename is the classic thing that will force ordering of some fashion.

As long as it's an *easy to understand* rule - I think we're ok.  Policy id seems reasonable.  That does assume all policies come from one source and we're not trying to blend policies between *two* systems - in which case there could be an arms-race to see which one wants to be first.

experimental-docs/specifications/policy.md
dashpole: SDKs aren't expected to use things like hyperscan, but Policies aren't just for SDKs. For example, we are also working on a collector component for enforcing policies. But more generally we definitely want it to be possible to efficiently enforce large numbers of policies.

experimental-docs/specifications/policy.md
dashpole: We plan to have separate SDK specifications to cover this in the specification repo.

anuraaga: Since sdks have more restrictions, how about starting with them? It's hard to see the general story and direction behind this spec, especially how it eventually lands in the sdks - a huge question mark is the relationship to declarative configuration, which has some features for telemetry policy, which almost assuredly should exist in some form.

Collector components tend to be a superset of SDK specs and I think it can help drive both if focusing on SDK.

opentelemetry/proto/policies/v1development/logs.proto
Isha-BH: Since AttributePath already works for the other attribute-like selectors, could this be as simple as adding:

`AttributePath body_field = 6;`

to LogFieldSelector, with the rule that if body isn't a kvlist_value (or the path doesn't resolve inside it), the target is treated as absent , same fallback behavior LOG_RECORD_FIELD_BODY already has for non-string bodies elsewhere in the spec? That would keep LogRemove/LogRedact/LogRename working on nested body fields without inventing a new type.

